PR #13, opened 2021-01-26
Add SharedLibrary class to use when dynamically loading driver API libraries.

jonlugoatx:
# Justification
In order for our gRPC service implementations to dynamically load the API libraries without linking against them, we need utilities those services can call into to invoke the API methods by name in an OS-independent way.

# Implementation
Defined a `SharedLibrary` class that allows for loading a library by name and obtaining function pointers from those libraries by name. This class should handle the OS-specific details so that the services do not need to handle it.

# Testing
- Created a new test API library that can be dynamically loaded by unit tests
- Wrote unit tests that dynamically load that test dll.


kreimnitz (approved):

reckenro (approved):

epetersoni (approved):

.gitignore
```diff
@@ -1,5 +1,6 @@
 /bin/
 /build/
+/third_party/
```
christag-ni, line 3: @jonlugoatx What's the reason for this additional "/third_party/" folder in .gitignore? 

For context, I'm working on pulling in the latest grpc submodule from third_party. Git complains since I'm attempting to add a change from within third_party to my commit.

CC @maxxboehme

bkeryan: This sounds vaguely familiar but I can't find any discussion of it in previous submodule update PRs. I don't think our other repos that use submodules list them in .gitignore.

PR #154, opened 2021-03-26
[Owners] Update example error handling

kreimnitz:

Updates the example error handling and standardizes some configurations across examples.  Also update example file names to use "betterproto" instead of "betterpython"

GetError has inconsistent behavior on digitizers and sometimes doesn't correctly report past errors.  GetErrorMessage is a reliable way to get the error message for a negative status.

Ran all examples locally against latest server.  Tested errors by modifying vertical range to 1000 and viewing the resulting error

### DIFF
--- a/examples/niscope/fetch-continuous.py
+++ b/examples/niscope/fetch-continuous.py
@@ -83,7 +83,7 @@ try :
     CheckForError(vi, init_with_options_response.status)
 
     # Configure vertical.
-    voltage = 1.0
+    voltage = 10.0
     CheckForError(vi, (client.ConfigureVertical(niscope_types.ConfigureVerticalRequest(
         vi = vi,
         channel_list = channels,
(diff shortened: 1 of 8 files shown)


examples/niscope/fetch-continuous.py
```diff
@@ -138,23 +138,24 @@ def ThrowOnError (vi, error_code):
 
         # We fetch each channel at a time so we don't have to de-interleave afterwards.
         # We do not keep the wfm_info returned from fetch.
-        for channel, waveform in zip(channel_list, waveforms):
+        for channel_name, waveform in zip(channel_list, waveforms):
```
kreimnitz, line 141: This is to not conflict with the "channel" variable declare at the start

examples/niscope/fetch-continuous.py
```diff
@@ -83,7 +83,7 @@ def ThrowOnError (vi, error_code):
     CheckForError(vi, init_with_options_response.status)
 
     # Configure vertical.
-    voltage = 1.0
+    voltage = 10.0
```
kreimnitz, line 86: Standardize vertical range to 10 in examples

examples/niscope/graph-measurement-betterproto.py
```diff
@@ -189,4 +189,6 @@ async def main():
     await CloseGrpcScope(*grpc_scope)
 
 ## Run the main
-asyncio.run(main())
+loop = asyncio.get_event_loop()
```
kreimnitz, line 192: Use this method of starting an async loop so that it doesn't throw when we exit inside the await

jonlugoatx (approved):

kylesull30 (approved):

PR #155, opened 2021-03-31
[Owners] Refactor mako templates and python helpers.

reckenro:

It re-organizes and cleans up a handful of the code generation mako and python files.
This is a pure refactoring so the generated code shouldn't be updated in this PR.

Major updates:

- Moved all mako template files into `source/codegen/templates` folder.
- Added a `TemplateLookup` to the `templates` folder for each template file so each can import defs and blocks from other templates if desired.
- Separated defs out of `service.cpp.mako` and `proto.mako` into `service_helpers.mako` and `proto_helpers.mako`, respectively.
- Added documentation to all of the mako defs
- Single sourced metadata manipulation to `metadata_mutation.py` and call the methods once from `generate_service.py`.
- Renamed `handler_helpers.py` to `service_helpers.py`
- Cleaned up python helper files to add documentation where appropriate, remove unused methods, re-use methods over re-implementing, etc.
- Removed unused file `test_metadata_parsing.py`

This should be merged so that when we add support for more driver APIs the code generation will be easier to parse and update if necessary.

Built locally and confirmed that generated folder has not changed.

All tests still pass (as expected since generated files haven't changed).


CMakeLists.txt
```diff
@@ -67,15 +67,18 @@ set(nidrivers "nifake" "niscope" "niswitch")
 set(codegen_scripts
   "${codegen_dir}/common_helpers.py"
   "${codegen_dir}/generate_service.py"
-  "${codegen_dir}/handler_helpers.py"
-  "${codegen_dir}/library_interface.h.mako"
-  "${codegen_dir}/library.cpp.mako"
-  "${codegen_dir}/library.h.mako"
-  "${codegen_dir}/mock_library.h.mako"
-  "${codegen_dir}/proto.mako"
+  "${codegen_dir}/service_helpers.py"
```
reckenro, line 70: Note: Moved all mako scripts to `templates` folder and included new mako and python helper files here.

source/codegen/common_helpers.py
```diff
@@ -1,43 +1,14 @@
-def driver_name_to_pascal(driver_name):
```
reckenro, line 1: Note: `driver_name_to_pascal` and `driver_name_add_underscore` weren't used anywhere so I went ahead and removed them.

source/codegen/generate_service.py
```diff
@@ -5,32 +5,42 @@
...
 
   os.makedirs(output_dir, exist_ok=True)
-  template = mako.template.Template(filename=template_file_path)
+  template_lookup = TemplateLookup(directories = template_directory + "/")
+  template = mako.template.Template(filename=template_file_path, lookup=template_lookup)
   f=open(output_file_path, "w+", newline="")
   f.write(template.render(data=metadata))
   f.close()
 
+def fix_up_metadata(metadata):
```
reckenro, line 27: Note: This used to be spread out in different spots in the code. Consolidated it here to apply all metadata changes our codegen relies on.

source/codegen/generate_service.py
```diff
@@ -5,32 +5,42 @@
...
+  template_file_path = os.path.join(current_dir, "templates", template_file_name)
+  template_directory = os.path.dirname(template_file_path)
   module_name = metadata["config"]["module_name"]
   output_dir = os.path.join(gen_dir, module_name)
   file_name = module_name + generated_file_suffix
   output_file_path = os.path.join(output_dir, file_name)
 
   os.makedirs(output_dir, exist_ok=True)
-  template = mako.template.Template(filename=template_file_path)
+  template_lookup = TemplateLookup(directories = template_directory + "/")
```
reckenro, line 21: Note: This makes it so any template we render can reference `def`s in other templates in the `templates` directory.

source/codegen/service_helpers.py
```diff
@@ -1,33 +1,4 @@
...
-    'params', 'private', 'protected', 'public',
-    'readonly', 'ref', 'return',
-    'sbyte', 'sealed', 'short', 'sizeof', 'stackalloc', 'static', 'status', 'string', 'struct', 'switch',
-    'this', 'throw', 'true', 'try', 'typeof',
-    'uint', 'ulong', 'unchecked', 'unsafe', 'ushort', 'using',
-    'virtual', 'void', 'volatile',
-    'while'
-]
-
-def sanitize_names(parameters):
```
reckenro, line 25: Note: `sanitize_names` and the `RESERVED_WORDS` array were moved to `metadata_mutation.py`.

source/codegen/generate_service.py
```diff
@@ -5,32 +5,43 @@
...
+  template = mako.template.Template(filename=template_file_path, lookup=template_lookup)
   f=open(output_file_path, "w+", newline="")
   f.write(template.render(data=metadata))
   f.close()
 
+def fix_up_metadata(metadata):
```
SparkingSpork, line 27: Nitpick: Maybe call this `preprocess_metadata`, `mutate_metadata`, or something else similar? I'm not sure calling it `fix_up` is descriptive.

source/codegen/templates/proto_helpers.mako
```diff
@@ -51,45 +19,60 @@ enum ${service_class_prefix}Attributes {
...
-%if has_custom_template:
+</%def>
+
+## If there is a custom proto mako template in the metadata, insert it.
+<%def name="insert_custom_template_if_found()">\
+<%
+  config = data["config"]
+  lookup = data["lookup"]
+  has_custom_template = False
+  if len(config["custom_types"]) > 0:
```
SparkingSpork, line 62:
```suggestion
  if config["custom_types"]:
```
In python, objects with a __len__() method have implicit boolean conversions and I think you should be able to use that here. See https://docs.python.org/3/library/stdtypes.html#truth-value-testing

SparkingSpork (approved):

source/codegen/templates/proto_helpers.mako
reckenro: Oh neat. Committed the suggested change.

source/codegen/generate_service.py
reckenro: You're right. I went with `mutate_metadata`.

source/codegen/generate_service.py
```diff
@@ -5,32 +5,43 @@
...
   spec = importlib.util.spec_from_file_location("metadata", init_file)
   module = importlib.util.module_from_spec(spec)
   spec.loader.exec_module(module)
 
-  metadata = module.metadata;
-  lookup = TemplateLookup(directories=metadata_dir)
+  metadata = module.metadata
+  lookup = TemplateLookup(directories = metadata_dir)
   metadata["lookup"] = lookup
+  mutate_metadata(metadata)
```
jonlugoatx, line 44: This is a much better place to do the mutation. At some point, if we end up creating a hapigen plugin to generate this, we should consider moving this mutation into that plugin.

source/codegen/metadata_mutation.py
```diff
@@ -0,0 +1,51 @@
...
+
+def mark_size_params(parameters):
+    """Marks the size parameters in the metadata."""
+    for param in parameters:
+        mechanism = common_helpers.get_size_mechanism(param)
+        if mechanism in {'len', 'ivi-dance', 'passed-in'}:
+            size_param = get_size_param(param, parameters)
+            size_param['is_size_param'] = True
+
+def mark_non_grpc_params(parameters):
```
jonlugoatx, line 39: I think `non_proto_params` would be better since we're removing it from the .proto file but still referring to it in the service handler.

source/codegen/templates/proto_helpers.mako
```diff
@@ -51,45 +19,60 @@ enum ${service_class_prefix}Attributes {
...
+% if enum.get("generate-mappings", False):
   ${enum_value_prefix}_${value["name"]} = ${nonint_index};
 <%
-  nonint_index = nonint_index+1
+  nonint_index = nonint_index + 1
 %>\
 % else:
   ${enum_value_prefix}_${value["name"]} = ${value["value"]};
-%endif
+% endif
```
jonlugoatx, line 49: Since mako allows for spaces between the `%` and the directive, it could sometimes help to use that spacing to indicate indentation levels. I previously did it somewhere in service.cpp where there were lots of nested if/else ladders.

source/codegen/templates/service_helpers.mako
```diff
@@ -349,16 +248,16 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
         response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(${parameter_name}));
 %  endif
         response->set_${parameter_name}_raw(${parameter_name});
-%elif common_helpers.is_array(parameter['type']):
-%  if handler_helpers.is_string_arg(parameter):
+% elif common_helpers.is_array(parameter['type']):
+%  if service_helpers.is_string_arg(parameter):
```
jonlugoatx, line 252: Add a little more space between the `%` and `if`.

jonlugoatx (approved):

kreimnitz (approved):

source/codegen/metadata_mutation.py
reckenro: Agreed, that's more accurate. Updated the name and description taking the `non_proto_params` terminology into account.

source/codegen/generate_service.py
reckenro: That's what I was thinking as well, will be easier to find this spot when looking into the manual and auto updates we did per metadata if we get to creating our own hapigen plugin.

source/codegen/templates/proto_helpers.mako
reckenro: I did another pass through all the mako templates and helpers and standardized how the spacing is done. Now it's one space at first indent level and two additional spaces per level after that. This seems to read better and is consistent now at least.

source/codegen/templates/service_helpers.mako
reckenro: Along with other comment, adjusted the spaces for these indentation levels to be 1 space for first level and two additional spaces for each subsequent indentation level.

PR #161, opened 2021-04-08
[Owners]Add DCPower metadata and generated files and register DCPower service 

ni-pkhare:
### Justification

Adds the DCPower metadata and the generated proto file and service handlers.
Link to task : https://ni.visualstudio.com/DevCentral/_workitems/edit/1367880
Register DCPower service.
Link to task : https://ni.visualstudio.com/DevCentral/_workitems/edit/1367885

### Implementation

- Added nidcpower.h , nidcpowerObsolete.h , nidcpower.lib
- Made some changes to config.py
- Cleaned up function.py
- Registered DCPower service in core_server.cpp
- Added DCPower in CMakeLists.txt
- Added service.custom file for DCPower

### Testing

Manually inspected the generated files


ganeshnie (approved):

imports/include/nidcpowerObsolete.h
```diff
@@ -0,0 +1,56 @@
+/****************************************************************************
+ *              National Instruments DC Power Supplies and SMUs
+ *--------------------------------------------------------------------------*
+ *      Copyright (c) National Instruments 2016.  All Rights Reserved.      *
+ *--------------------------------------------------------------------------*
+ *
+ * Title:    nidcpowerObsolete.h
```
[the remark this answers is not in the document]
ni-pkhare: @ganeshnie , **nidcpower.h** uses the line **#include "nidcpowerObsolete.h"** , so we have to include this file also else we would get build errors.

source/codegen/metadata/nidcpower/config.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+# This file is generated from NI-DCPower API metadata version 20.7.0d8
+config = {
+    'api_version': '20.7.0d8',
+    'c_header': 'nidcpower.h',
+    'c_function_prefix': 'niDCPower_',
+    'service_class_prefix': 'NiDCPower',
+    'java_package': 'com.ni.grpc.nidcpower',
+    'csharp_namespace': 'NationalInstruments.Grpc.DCPower',
```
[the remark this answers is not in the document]
ni-pkhare: @ganeshnie , I will change it to the convention used in scope. Scope has the line **'java_package': 'com.ni.grpc.scope'**, so accordingly for dcpower it would become **'java_package': 'com.ni.grpc.dcpower'**

source/codegen/metadata/nidcpower/config.py
```diff
@@ -0,0 +1,54 @@
...
+# This file is generated from NI-DCPower API metadata version 20.7.0d8
+config = {
+    'api_version': '20.7.0d8',
+    'c_header': 'nidcpower.h',
+    'c_function_prefix': 'niDCPower_',
+    'service_class_prefix': 'NiDCPower',
+    'java_package': 'com.ni.grpc.dcpower',
+    'csharp_namespace': 'NationalInstruments.Grpc.DCPower',
+    'namespace_component': 'nidcpower',
+    'close_function': 'close',
```
reckenro, line 11:
```suggestion
    'close_function': 'Close',
```
Match the capitalization used in `functions.py` for the close method so it can find it through the interface. I believe that's why the builds are failing currently.

source/codegen/metadata/nidcpower/functions.py
```diff
@@ -0,0 +1,2985 @@
...
+      },
+      {
+        'name': 'filePath',
+        'direction': 'in',
+        'type': 'ViConstString'
+      }
+    ],
+    'returns': 'ViStatus'
+  },
+  'InitExtCal': {
```
reckenro, line 2368: Should this be marked as an `'init_method' : True,`? It seems like it initiates a ViSession.

source/codegen/metadata/nidcpower/functions_addon.py
```diff
@@ -0,0 +1,639 @@
+# These dictionaries are merged with the extracted function metadata at build time.
+# Changes to the metadata should be made here, because functions.py is generated thus any changes get overwritten.
+
+functions_override_metadata = {
+}
+
+functions_additional_create_advanced_sequence = {
```
reckenro, line 7: I think you can just delete this, I don't believe it's used or intended to be used in through the grpc server.

reckenro (changes requested):
It looks good for the most part. Mainly just want to see it again after it's successfully building to see how it all looks.

imports/include/nidcpowerObsolete.h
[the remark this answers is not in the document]
jonlugoatx: We should add the obsolete headers to the `include` directory to avoid needing to modify `nidcpower.h`, but we shouldn't add service handlers fo rthem.

source/codegen/metadata/nidcpower/attributes.py
```diff
@@ -0,0 +1,1759 @@
+# -*- coding: utf-8 -*-
+# This file is generated from NI-DCPower API metadata version 20.7.0d8
+attributes = {
+    1050003: {
+        'access': 'read-write',
+        'channel_based': False,
+        'documentation': {
+            'description': '\nSpecifies whether NI-DCPower queries the device status after each operation.\nQuerying the device status is useful for debugging. After you validate your program, you can set this  attribute to VI_FALSE to disable status checking and maximize performance.\nNI-DCPower ignores status checking for particular attributes regardless of the setting of this attribute.\nUse the niDCPower_InitializeWithChannels function to override this value.\nDefault Value: VI_TRUE\n'
```
jonlugoatx, line 8: We don't use the `documentation` and `lv_property` values. I realize we forgot to remove them from NI-SCOPE and NI-SWITCH, but consider removing them from here.

source/codegen/metadata/nidcpower/functions.py
```diff
@@ -0,0 +1,2985 @@
...
+        'name': 'channelName',
+        'direction': 'in',
+        'type': 'ViConstString'
+      },
+      {
+        'name': 'voltageMeasurements',
+        'direction': 'out',
+        'type': 'ViReal64[]',
+        'size': {
+          'mechanism': 'source-code'
```
jonlugoatx, line 2487: Usually when the mechanism is `source-code`, there's a `value` entry with the source code to run. Try to include what the source code would look like in C++. See the niscope functions as an example.

jonlugoatx (changes requested):

imports/include/nidcpowerObsolete.h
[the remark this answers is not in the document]
ni-pkhare: @jonlugoatx , There are only 3 Obsolete APIs, I checked they are not present as part of metadata and hence they are also generated in the proto and service handlers.

source/codegen/metadata/nidcpower/config.py
ni-pkhare: @reckenro, Thanks for putting this to notice, I changed them and now the build is also successful.

source/codegen/metadata/nidcpower/functions.py
ni-pkhare: @reckenro, InitExtCal also initializes a session, hence I included 'init_method':True in the metadata. Thanks for putting this to notice.

source/codegen/metadata/nidcpower/functions_addon.py
ni-pkhare: @reckenro, Done deleting.

source/codegen/metadata/nidcpower/attributes.py
ni-pkhare: @jonlugoatx, Done removing these two values from attributes.py file.

ni-pkhare:
> It looks good for the most part. Mainly just want to see it again after it's successfully building to see how it all looks.

@reckenro, Thanks for suggesting the changes, Now the build is successful.

source/codegen/metadata/nidcpower/functions.py
ni-pkhare: @jonlugoatx, This API needs to be custom implemented, so we have planned to make changes in metadata and do the custom implementation as part of subsequent PR's. I would probably comment-out this function in the metadata for this PR.

jonlugoatx: Don't comment out the metadata. Instead, do this to be consistent with [NI-SCOPE metadata](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/niscope/functions.py#L1193):
```python
'size': {
  'mechanism': 'custom-code',
  'value': 'number_of_channels'
}
```
Explanation:
The custom code you need to implement is a call to `niDCPower_ParseChannelCount` (see the [nimi-python metadata](https://github.com/ni/nimi-python/blob/master/src/nidcpower/metadata/functions.py#L1983) & [helper implementation](https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/session.py#L3531)). Then your custom code would pass the output parameter of that call as the `voltageMeasurements` argument when calling `niDCPower_MeasureMultiple`. The output variable of that function call should go in the `value` field of the metadata.

This is very similar to the NI-SCOPE plugin where scope needs to call `niScope_ActualNumWfms`. We considered adding a gRPC codegen feature to specify custom functions to call but cut that codegen feature. If we implement that feature, then we would need the value in the size metadata to be the C++ code to write for the argument in the function call.

When you make this metadata change, you don't need to provide a custom implementation of the function until you add `'codegen_method': 'CustomCode'` before the `'parameters'` array.

jonlugoatx (approved):

SparkingSpork (approved):

reckenro (approved):

ni-pkhare:
I will not be merging this PR anywhere in ni/grpc-device because after approval from owners I have created a replicated PR in doshirohan/grpc-device.
So I am closing this PR as it has been already approved.
Replicated PR : [PR](https://github.com/doshirohan/grpc-device/pull/9)

PR #163, opened 2021-04-09
Added an example for NI DMM grpc APIs

ShantanuShinde:

A C example. ContinuousAcquireMultiplePoints, has been ported to python using the NIDMM gRPC APIs

This example can be used to test and understand the generated DMM APIs

The example has been locally tested and is working as expected


examples/nidmm/continuous-acquire-multiple-points.py
```diff
@@ -0,0 +1,186 @@
...
+import sys
+import time
+import nidmm_pb2 as nidmm_types
+import nidmm_pb2_grpc as grpc_nidmm
+import session_pb2 as session_types
+import session_pb2_grpc as grpc_session
+
+server_address = "localhost"
+server_port = "31763"
+session_name = "NI-DMM-Session-1"
```
ganeshnie, line 35: nit: you could remove the suffix "-1" in the session name.

examples/nidmm/continuous-acquire-multiple-points.py
```diff
@@ -0,0 +1,186 @@
...
+        CheckForError(vi, read_status_response.status)
+        pts_available = read_status_response.acquisition_backlog
+
+        # if there are more than MAXPTSTOREAD measurements
+        # available, set pts_available to MAXPTSTOREAD in order to
+        # avoid reallocating the array for measurements
+        pts_available = min(pts_available, MAXPTSTOREAD)
+
+        if pts_available > 0:
+
```
ganeshnie, line 143: Is this empty line consistent with the styling conventions?

ganeshnie (approved):

examples/nidmm/continuous-acquire-multiple-points.py
```diff
@@ -0,0 +1,186 @@
...
+    error_message_request = nidmm_types.ErrorMessageRequest(
+        vi = vi,
+        error_code = error_code
+    )
+    error_message_response = nidmm_client.ErrorMessage(error_message_request)
+    raise Exception (error_message_response.error_message)
+
+try:
+    # Open session to NI-DMM with options
+    print('\nInitializing session...')
```
harsha-bhushan, line 79: no need to print messages twice - Initializing, Initialized, configuring, configured etc. 
it will be almost immediate & would not add any value. In fact in the DMM example we may not need to print these messages at all. (In switch example we print those messages to indicate that the example is indeed running as there is no output as such.)

examples/nidmm/continuous-acquire-multiple-points.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # if there are more than MAXPTSTOREAD measurements
+        # available, set pts_available to MAXPTSTOREAD in order to
+        # avoid reallocating the array for measurements
+        pts_available = min(pts_available, MAXPTSTOREAD)
+
+        if pts_available > 0:
+
+            # Fetch data
+            fetch_multipoints_response = nidmm_client.FetchMultiPoint(nidmm_types.FetchMultiPointRequest(
```
harsha-bhushan, line 145: Can you continuously plot the acquired points on a graph using matplot lib? It would be easy as this is python example. refer to scope example (https://github.com/ni/grpc-device/blob/main/examples/niscope/graph-measurement.py) 

This might slightly deviate from the c example, but would be evident for the user if they can visualize the acquired measurements. 
Then you can remove printing all the messages to the console too, as the plot will stop & return to console to print the number of measurements & avg value.

harsha-bhushan (changes requested):

examples/nidmm/continuous-acquire-multiple-points.py
ShantanuShinde: removed

examples/nidmm/continuous-acquire-multiple-points.py
ShantanuShinde: removed the line

examples/nidmm/continuous-acquire-multiple-points.py
```diff
@@ -0,0 +1,186 @@
+# 
+# This example performs continuous multipoint acquisition and prints the number of points read and average value of measurements.
```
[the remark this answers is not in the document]
ShantanuShinde: changed the explanation

examples/nidmm/continuous-acquire-multiple-points.py
ShantanuShinde: removed prints

examples/nidmm/continuous-acquire-multiple-points.py
ShantanuShinde: added graph plotting and changed the file name

harsha-bhushan (approved):

PR #166, opened 2021-04-21
[Owners] Fix status checks in two custom niscope service methods.

### LINKED ISSUE #165: A couple methods in niscope_service.custom.cpp aren't checking status correctly.
A few of the custom scope methods are setting the response fields only if the status is not zero.
This should be reversed, as a zero status represents a successful API call.

For instance, `GetNormalizationCoefficients` currently has this:

```
if (status != 0) {
      response->set_number_of_coefficient_sets(number_of_coefficient_sets);
      Copy(coefficient_info, response->mutable_coefficient_info());
    }
```

When it should be keying off of `status == 0` like the other methods.

reckenro:

Closes #165 

It fixes the status check on `GetNormalizationCoefficients` and `GetScalingCoefficients` so that it sets the output fields on the response message upon success instead of upon failure.

Added two system level tests for the broken NI Scope service implementations. Prior to the changes they would fail and now pass.

### DIFF
--- a/source/custom/niscope_service.custom.cpp
+++ b/source/custom/niscope_service.custom.cpp
@@ -455,7 +455,7 @@ void CheckStatus(int status)
     std::vector<niScope_coefficientInfo> coefficient_info(number_of_coefficient_sets, niScope_coefficientInfo());
     auto status = library_->GetNormalizationCoefficients(vi, channel_list, coefficient_info.size(), coefficient_info.data(), &number_of_coefficient_sets);
     response->set_status(status);
-    if (status != 0) {
+    if (status == 0) {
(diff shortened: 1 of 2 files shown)


reckenro:
The one open question I have about this is how it will be incorporated into a release. Should it stay in the code base until something else comes along to bump a release or should we proactively make a new release to incorporate this fix?

jonlugoatx (approved):

jonlugoatx (commented):
The change looks good, but we should probably write an integration test for this.

reckenro:
> The change looks good, but we should probably write an integration test for this.

Good point. I added a test for each of the methods that both fail without the custom service updates. Reset you to get another eye on them before submitting.

jonlugoatx (approved):

PR #173, opened 2021-05-03
Add MeasureEx

ppease:

Add the method `niSync_MeasureFrequencyEx` for the measure test panel for NI-Sync.

Adding methods need for the measure test panel for NI-Sync.

[ RUN      ] NiSyncDriverApiTest.MeasureFrequencyExOnTerminalWithNoFrequency_ReturnsNoFrequency
[       OK ] NiSyncDriverApiTest.MeasureFrequencyExOnTerminalWithNoFrequency_ReturnsNoFrequency (106 ms)
[ RUN      ] NiSyncDriverApiTest.MeasureFrequencyExOnOscillatorWithFrequency_ReturnsFrequency
[       OK ] NiSyncDriverApiTest.MeasureFrequencyExOnOscillatorWithFrequency_ReturnsFrequency (105 ms)
[----------] 18 tests from NiSyncDriverApiTest (750 ms total)

### DIFF
--- a/generated/nisync/nisync.proto
+++ b/generated/nisync/nisync.proto
@@ -25,6 +25,7 @@ service NiSync {
   rpc DisconnectSWTrigFromTerminal(DisconnectSWTrigFromTerminalRequest) returns (DisconnectSWTrigFromTerminalResponse);
   rpc ConnectTrigTerminals(ConnectTrigTerminalsRequest) returns (ConnectTrigTerminalsResponse);
   rpc DisconnectTrigTerminals(DisconnectTrigTerminalsRequest) returns (DisconnectTrigTerminalsResponse);
+  rpc MeasureFrequencyEx(MeasureFrequencyExRequest) returns (MeasureFrequencyExResponse);
   rpc GetAttributeViInt32(GetAttributeViInt32Request) returns (GetAttributeViInt32Response);
   rpc SetAttributeViInt32(SetAttributeViInt32Request) returns (SetAttributeViInt32Response);
   rpc GetAttributeViString(GetAttributeViStringRequest) returns (GetAttributeViStringResponse);
@@ -236,6 +237,20 @@ message DisconnectTrigTerminalsRequest {
 
 message DisconnectTrigTerminalsResponse {
   int32 status = 1;
+}
+
+message MeasureFrequencyExRequest {
+  nidevice_grpc.Session vi = 1;
+  string src_terminal = 2;
+  double duration = 3;
+  uint32 decimation_count = 4;
+}
+
+message MeasureFrequencyExResponse {
+  int32 status = 1;
+  double actual_duration = 2;
+  double frequency = 3;
+  double frequency_error = 4;
 }
 
 message GetAttributeViInt32Request {
--- a/source/tests/system/nisync_driver_api_tests.cpp
+++ b/source/tests/system/nisync_driver_api_tests.cpp
@@ -557,17 +557,34 @@ TEST_F(NiSyncDriverApiTest, AttributeSet_GetAttributeViString_ReturnsValue)
   EXPECT_STREQ(expectedValue, value.c_str());
 }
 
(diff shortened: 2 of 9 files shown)


source/tests/system/nisync_driver_api_tests.cpp
```diff
@@ -540,6 +557,36 @@ TEST_F(NiSyncDriverApiTest, AttributeSet_GetAttributeViString_ReturnsValue)
...
+  ViStatus viStatus;
+  ViConstString srcTerminal = NISYNC_VAL_OSCILLATOR;
+  ViReal64 duration = 0.1; // duration is in seconds
+  ViUInt32 decimationCount = 0; // ignored for 6674
+  ViReal64 actualDuration, frequency, frequencyError;
+  auto grpcStatus = call_MeasureFrequencyEx(srcTerminal, duration, decimationCount, &actualDuration, &frequency, &frequencyError, &viStatus);
+
+  EXPECT_TRUE(grpcStatus.ok());
+  EXPECT_EQ(VI_SUCCESS, viStatus);
+  EXPECT_GT(actualDuration, 0);
```
ppease, line 586: For both of these values I'm just checking that they are greater than zero.  In reality it should be really close to 0.1 for the duration and really close to 10 for the frequency.

source/tests/system/nisync_driver_api_tests.cpp
```diff
@@ -249,6 +249,23 @@ class NiSyncDriverApiTest : public ::testing::Test {
     return grpcStatus;
   }
 
+  ::grpc::Status call_MeasureFrequencyEx(ViConstString srcTerminal, ViReal64 duration, ViUInt32 decimationCount, ViReal64* actualDurationOut, ViReal64* frequencyOut, ViReal64* frequencyErrorOut, ViStatus* viStatusOut)
```
yixinguo2, line 252: This line is a bit long, put parameters on separate lines?

yixinguo2 (approved):

source/tests/system/nisync_driver_api_tests.cpp
ppease: Fixed in commit 5 and 6

generated/nisync/nisync.proto
```diff
@@ -25,6 +25,7 @@ service NiSync {
   rpc DisconnectSWTrigFromTerminal(DisconnectSWTrigFromTerminalRequest) returns (DisconnectSWTrigFromTerminalResponse);
```
ppease, line 25: I'm guessing my changes will conflict with Zach's recent PR to add some additional get/set methods.  I'll wait for his to go in and resolve conflicts.

tkrebes (approved):

source/tests/system/nisync_driver_api_tests.cpp
reckenro: This seems fine to me. More of an FYI, but Google Test has some assertions for floats / doubles that assert the values are "close" to each other and one that takes in an error bound (see https://github.com/google/googletest/blob/master/docs/advanced.md#floating-point-comparison).

I expect just checking greater than zero is fine here but those are some alternative assertions you can use.

reckenro (approved):

source/tests/system/nisync_driver_api_tests.cpp
jonlugoatx: For cases where you would like to test specific values or corner cases, you may want to consider adding an integration test that uses a mocked version of your service. Also the benefit in doing so is that the integration tests run on the CI while the system tests do not. This test looks good as is though.

jonlugoatx (approved):

source/tests/system/nisync_driver_api_tests.cpp
ppease: I really just wanted to do sanity testing here to make sure things look reasonable.  Thanks for the suggestions about floating point and the integration tests.  I'll keep that in mind for future changes.  I'm going to leave this as is for now.

PR #174, opened 2021-05-03
[Peer Review] Add Windows artifact for merges to main

SparkingSpork:
To facilitate releases, this PR adds Windows artifacts containing the server binary any time a merge to main occurs.

To allow usage of the Windows artifacts for incremental usage and creating releases without requiring a local build.

Ran a private workflow to ensure the artifact contains the right files. See [here](https://github.com/ni/grpc-device/actions/runs/808000693).

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -126,7 +126,7 @@ jobs:
 
     - name: Upload Windows Server Binaries Artifact
       uses: actions/upload-artifact@v2
-      if: "(runner.os == 'Windows')"
+      if: "(github.ref == 'refs/heads/main') && (runner.os == 'Windows')"
       with:
         name: ni-grpc-device-server-windows
         path: |
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -123,3 +123,13 @@ jobs:
         name: ni-grpc-device-server-linux-glibc2_23
         path: ni-grpc-device-server-linux-glibc2_23.tar
         retention-days: 5
+
+    - name: Upload Windows Server Binaries Artifact
+      uses: actions/upload-artifact@v2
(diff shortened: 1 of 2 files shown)


kreimnitz (approved):

reckenro (approved):
This looks good for the main branch. The one thing I'm not sure about is how to have these artifacts for the [releases/1.0 branch](https://github.com/ni/grpc-device/tree/releases/1.0). The main branch has some ni-sync stuff in it now that we don't want in a 1.0.1 release so we made the releases/1.0 branch to cherry pick bug fixes we want in the 1.0.x releases.

This goes for both the linux and windows artifacts but maybe in that branch it makes sense to submit a change so the if statement refers to that branch instead of main. @jonlugoatx , thoughts on this?

SparkingSpork:
@reckenro, @jonlugoatx  - We would simply need to add pushes to the `release/**` branches and a similar check for the specific actions to cover release branches. I can make that change in this PR if desired.

.github/workflows/build_cmake.yml
```diff
@@ -123,3 +123,13 @@ jobs:
         name: ni-grpc-device-server-linux-glibc2_23
         path: ni-grpc-device-server-linux-glibc2_23.tar
         retention-days: 5
+
+    - name: Upload Windows Server Binaries Artifact
+      uses: actions/upload-artifact@v2
+      if: "(github.ref == 'refs/heads/main') && (runner.os == 'Windows')"
```
jonlugoatx, line 129: This should also run on `refs/heads/releases/*`

jonlugoatx (approved):

.github/workflows/build_cmake.yml
SparkingSpork: I used `startsWith` as it wasn't clear from the documentation if operators accepted wildcard filters.

reckenro (approved):

PR #185, opened 2021-05-06
[Owners] Adding licenses to build artifacts

SparkingSpork:
Adds LICENSE and ThirdPartyNotices.txt to build artifacts

Makes packaging releases easier and ensures people who download our artifacts get a copy of the licenses.

Testing will be a run on main once this PR is approved.

### DIFF
--- a/.github/workflows/build_nilrt.yml
+++ b/.github/workflows/build_nilrt.yml
@@ -134,8 +134,8 @@ jobs:
           message(FATAL_ERROR "Build failed")
         endif()
 
-    - name: Tar Build Files
-      if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/releases') }}
+    - name: Tar Development Build Files
+      if: ${{ github.ref == 'refs/heads/main' }}
       run: >-
         tar -cvf NILinuxRealTime-x86_64.tar
         -C ${GITHUB_WORKSPACE}/build
@@ -147,6 +147,20 @@ jobs:
         SystemTestsRunner
         test_mutual_tls_config.json
         UnitTestsRunner
+        -C ${GITHUB_WORKSPACE}
+        LICENSE
+        ThirdPartyNotices.txt
+
+    - name: Tar Release Build Files
+      if: ${{ startsWith(github.ref, 'refs/heads/releases') }}
+      run: >-
+        tar -cvf NILinuxRealTime-x86_64.tar
+        -C ${GITHUB_WORKSPACE}/build
+        ni_grpc_device_server
+        server_config.json
+        -C ${GITHUB_WORKSPACE}
+        LICENSE
+        ThirdPartyNotices.txt
 
     - name: Upload Artifact
       uses: actions/upload-artifact@v2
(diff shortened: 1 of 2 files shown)


amstewart (approved):

.github/workflows/build_nilrt.yml
```diff
@@ -134,8 +134,8 @@ jobs:
           message(FATAL_ERROR "Build failed")
         endif()
 
-    - name: Tar Build Files
-      if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/releases') }}
+    - name: Tar Development Build Files
+      if: ${{ github.ref == 'refs/heads/main' }}
       run: >-
         tar -cvf NILinuxRealTime-x86_64.tar
```
kreimnitz, line 140: If we make these -cvzf will Github still automatically zip them?

.github/workflows/build_nilrt.yml
```diff
@@ -147,6 +147,20 @@ jobs:
...
         test_mutual_tls_config.json
         UnitTestsRunner
+        -C ${GITHUB_WORKSPACE}
+        LICENSE
+        ThirdPartyNotices.txt
+
+    - name: Tar Release Build Files
+      if: ${{ startsWith(github.ref, 'refs/heads/releases') }}
+      run: >-
+        tar -cvf NILinuxRealTime-x86_64.tar
```
kreimnitz, line 157: Might be nice to rename the artifacts to exactly match the release names:

Linux Desktop: 
ni-grpc-device-server-linux-glibc2_23-x64.tar.gz  (created with "tar -cvzf")

LinuxRT: 
ni-grpc-device-server-ni-linux-rt-x64.tar.gz   (created with "tar -cvzf")

Windows: 
ni-grpc-device-server-windows-x64.zip

.github/workflows/build_nilrt.yml
SparkingSpork: Yes. All artifact downloads are zipped, even if you've already zipped them once. Just an annoying thing about how GitHub hosts them.

kreimnitz (commented):
Just added some minor naming comments, thanks for taking care of this!

.github/workflows/build_nilrt.yml
kreimnitz: Probably still worth using -cvzf and ending files with .tar.gz I guess. We'll still have to unzip it to get down to the .tar.gz, but that's not too bad

.github/workflows/build_nilrt.yml
SparkingSpork: Done

.github/workflows/build_nilrt.yml
SparkingSpork: I've gone back and forth on this a couple of times. I don't really like having the extra layer of compression (Linux can handle zip files without issue I believe) but after thinking about it more, the extra compression to save space on our artifact storage is probably worthwhile even if we currently have unlimited space for the public repo.

Fixed.

kreimnitz (approved):

.github/workflows/build_nilrt.yml
jonlugoatx: The more important aspect of this is the ease of creating the release. I don't want to have to unzip the files then tar.gz them up. That adds an extra source of possible user error. If we zip up the compressed .tar.gz, then unzipping will give you the right thing to upload to the release.

jonlugoatx (approved):

PR #195, opened 2021-05-12
CMakeLists: enable OpenEmbedded compilation

amstewart:

This patchset makes several changes to the CMakeLists file which enable cross-compilation in the context of an OpenEmbedded build environment. In the OE-case, it is not reasonable to satisfy build dependencies using git dependencies.

This patchset is needed to enable the `ni-grpc-device` recipe in the [meta-nilrt:nilrt/master/dunfell](https://github.com/ni/meta-nilrt/tree/nilrt/master%2Fdunfell) repo.

OE builds this CMakeList successfully on my dev machine. I expect the GH workflows can verify that it builds correctly in the mundane windows/linux build context.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,6 +198,8 @@ target_link_libraries(ni_grpc_device_server
    nlohmann_json::nlohmann_json
    )
 
+set_target_properties(ni_grpc_device_server PROPERTIES BUILD_WITH_INSTALL_RPATH TRUE)
+
 if(WIN32)
   target_link_libraries(ni_grpc_device_server Delayimp niScope niswitch nisync nidcpower)
   set_target_properties(ni_grpc_device_server PROPERTIES LINK_FLAGS "/DELAYLOAD:niScope.dll /DELAYLOAD:niswitch.dll /DELAYLOAD:nisync.dll /DELAYLOAD:nidcpower.dll")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,9 @@ include_directories(
   "./imports/include"
   "./source"
 )
-link_directories("./imports/lib/win64")
+if(WIN32)
+  link_directories("./imports/lib/win64")
+endif()
 
 #----------------------------------------------------------------------
 # Get list of NI Driver APIs and directories to generate from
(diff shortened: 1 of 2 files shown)


jonlugoatx (approved):

CMakeLists.txt
```diff
@@ -208,12 +218,17 @@ add_custom_command(
...
+  find_package(nlohmann_json REQUIRED)
+  find_library(gtest REQUIRED)
+else()
+  add_subdirectory(third_party/json ${CMAKE_CURRENT_BINARY_DIR}/json EXCLUDE_FROM_ALL)
+  add_subdirectory(third_party/gtest ${CMAKE_CURRENT_BINARY_DIR}/gtest EXCLUDE_FROM_ALL)
+endif()
 
 # Add googletest tests
 enable_testing()
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
```
SparkingSpork, line 231: Per the Teams conversation, move these so that they're called prior to adding gtest. I think it should be OK for them to be outside of the if block.

CMakeLists.txt
```diff
@@ -208,12 +218,17 @@ add_custom_command(
 #----------------------------------------------------------------------
 # Add JSON parser
 #----------------------------------------------------------------------
```
SparkingSpork, line 220: We should probably update this comment to indicate gtest is added here now too.

SparkingSpork (approved):

CMakeLists.txt
amstewart: Will fix.

CMakeLists.txt
amstewart: I think I like the idea that testing won't be enabled at all in the NILRT-OE case. I'll see if I can get it working within the branch structure.

amstewart:
## Update 2
* Added a new cmake variable `_GRPC_DEVICE_NILRT_LEGACY_TOOLCHAIN`. When `TRUE`, it asserts that CMakeLists should setup the toolchain dependencies to satisfy the published NIRLT toolchain archive. This change fixes the "NILRT" GH workflow.
* Fixed the Windows compilation workflow by moving the `set(gtest_force_shared_crt...` line before `add_subdirectory(third_party/gtest...`. Thanks; @SparkingSpork.

All workflows should pass now.

Ready for final review and merge.

SparkingSpork (approved):

PR #202, opened 2021-05-17
[RFC] Support for per-type enums for attribute values

doshirohan:

Earlier we used to define separate enum, per enum name defined for attributes. This has usability issue where users would face difficulty figuring out list of possible values for given enum. This PR addresses that by creating single enum per-type of attribute which contains all possible constant values of that type. So we have enums for ViInt32, ViReal64 and so on.

## What is done
- Enums are created per type for group of attributes
- Support for generate_map didn't need any change.
- Build for ni_grpc_device_server.exe works fine

## Pending work
- I haven't yet updated request messages for Set*Attribute APIs to specify enum type as `oneof` data types
- Detailed testing to make sure generated-mappings work fine.
- Examples need to fixed to use new enum values
- Testing

TODO: Detail what testing has been done to ensure this submission meets requirements.


generated/nidcpower/nidcpower.proto
```diff
@@ -511,19 +431,92 @@ enum SourceMode {
...
-}
 
-enum TriggerType {
-  TRIGGER_TYPE_UNSPECIFIED = 0;
-  TRIGGER_TYPE_NIDCPOWER_VAL_NONE = 1012;
-  TRIGGER_TYPE_NIDCPOWER_VAL_DIGITAL_EDGE = 1014;
-  TRIGGER_TYPE_NIDCPOWER_VAL_SOFTWARE_EDGE = 1015;
+enum NiDCPowerViInt32Enum{
+  option allow_alias = true;
+  NI_D_C_POWER_VI_INT32_ENUM_AUTORANGE_BEHAVIOR_UNSPECIFIED = 0;
```
doshirohan, line 437: Should we add only one UNSPECIFIED value for entire `NiDCPowerViInt32Enum` instead of adding it for each enum that goes into it?

jonlugoatx: Yes, we should have a single `UNSPECIFIED` value. It's okay to have multiple values that are `0` though.

jonlugoatx, line 437: We need to find a way to make the prefix look better: `NI_DCPOWER_` or `NIDCPOWER_

source/codegen/templates/proto_helpers.mako
```diff
@@ -53,6 +53,49 @@ enum ${enum_name} {
...
+<%def name="define_per_type_attribute_enums(attribute_enums)">\
+<%
+  config = data["config"]
+  service_class_prefix = config["service_class_prefix"]
+  enums = data["enums"]
+%>\
+% for type_name in attribute_enums:
+<%
+  type_enums = attribute_enums[type_name]
+  type_enum_name = service_class_prefix + type_name + "Enum"
```
jonlugoatx, line 66: I think a better name would be `service_class_prefix + type_name + "AttributeValues"`

source/codegen/templates/proto_helpers.mako
```diff
@@ -53,6 +53,49 @@ enum ${enum_name} {
...
+  # allow alias if we have more than one enums (more than one UNSPECIFIED enum values) or
+  # the only enum we have has duplicate values
+  allos_alias_option = len(type_enums) > 1 or proto_helpers.should_allow_alias(enums[list(type_enums)[0]])
+%>\
+
+enum ${type_enum_name}{
```
jonlugoatx, line 72: Add a space before the `{`

generated/nidcpower/nidcpower.proto
```diff
@@ -511,19 +431,92 @@ enum SourceMode {
...
+  NI_D_C_POWER_VI_INT32_ENUM_AUTORANGE_BEHAVIOR_NIDCPOWER_VAL_RANGE_UP_TO_LIMIT_THEN_DOWN = 1107;
```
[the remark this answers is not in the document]
jonlugoatx: I agree. There's also now a bit of duplication in the name.

jonlugoatx, line 438: We shouldn't be including the individual enum name, e.g. `AUTORANGE_BEHAVIOR`, in this combined enum. The 2nd part also already has `NIDCPOWER_VAL` in it, so the `NIDCPOWER_VIINT32_ENUM` prefix isn't helping. IMO, this one would ideally be named `NIDCPOWER_ATTRIBUTE_VAL_RANGE_UP_TO_LIMIT_THEN_DOWN`

doshirohan: Removing the `type_enum_name` prefix `(NI_D_C_POWER_VI_INT32_ENUM)` will cause conflicts if there are two attributes of type int32 and real64, which share same enum values. In that case we duplicate same set of values in both enums. If I duplicate `NIDCPOWER_VAL_INTERNAL` value across two enums, this is the error I receive.
image


Removing the enum name from value `(AUTORANGE_BEHAVIOR)` also has potential for conflict if we have more than two enums defined in our metadata for same set of values. For e..g we have `PowerSource` and `PowerSourceInUse` enums in dcpower which have `NIDCPOWER_VAL_INTERNAL/AUXILIARY/AUTOMATIC` values. Last value doesn't exist in `PowerSourceInUse` because it is not applicable for corresponding attribute.

That said, we could probably do something to remove NIDCPOWER from middle of enum value to remove duplication. That comes from enum values defined in enums.py though, so some sort of string manipulation would be required for this. And if users copy potential attribute value from help (wihch contains NIDCPOWER_ prefix), and search for it with intellisense they might not get a hit. Is that something we should do?

jonlugoatx: Is this a realistic scenario? The C headers would probably not define a single constant to be used by both a `ViReal64` attribute and a `ViInt32` attribute, would it? If so, then maybe the prefix should be `NIDCPOWER_INT32_VAL` and we take only the part of the enum value that comes after the `NIDCPOWER_VAL`, resulting in this:
```proto
enum NiDCPowerInt32AttributeValue {
  option allow_alias = true;
  NIDCPOWER_INT32_VAL_UNSPECIFIED = 0;
  NIDCPOWER_INT32_VAL_RANGE_UP_TO_LIMIT_THEN_DOWN = 1107;
  ...
}
```

generated/nidcpower/nidcpower.proto
doshirohan: Updated.

source/codegen/templates/proto_helpers.mako
doshirohan: Updated enum name

generated/nidcpower/nidcpower.proto
doshirohan: I have updated PR to use `NIDCPOWER_INT32_VAL` prefix and also added name of the enum (from enums.py file) compared to example you posted above. Name of the enum is required because same enum values could be present in multiple enums. For e.g. `NIDCPOWER_VAL_INTERNAL` is present in `PowerSource` and `PowerSourceInUse` enums in [dcpower](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nidcpower/enums.py#L521).

I was able to remove `NIDCPOWER_VAL` prefix that comes with all enum values, so the names looks shorter now. Can you take another look and let me know if we are good to go for peer review in separate PR (after some cleanup).

PR #204, opened 2021-05-17
Response strings are 1 nul-character too big

zhindes:

Fixes a bug in the service generator that is resulting in all out-strings being 1 nul-character too large. For example, if a string "foo" is returned from the library, we were returning "foo\0" (raw buffer is then "foo\0\0") over gRPC. This resulted in empty strings that were actually length 1 with a single character '\0'. This greatly confused LabVIEW.

Fix a major string handling bug.

Modified a sync test to validate the strlen in addition to str value. This was failing before my fix. Validated a string property node in NI-Sync is now correctly returning empty strings.

### DIFF
--- a/source/codegen/templates/service_helpers.mako
+++ b/source/codegen/templates/service_helpers.mako
@@ -218,12 +218,12 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
 %>\
 %     if common_helpers.is_struct(parameter) or underlying_param_type == 'ViBoolean':
       std::vector<${underlying_param_type}> ${parameter_name}(${size}, ${underlying_param_type}());
-## byte arrays are leveraging a string as a simple buffer wrapper
+## Byte arrays are leveraging a string as a buffer, so we don't need to take special consideration of the null terminator.
 %     elif service_helpers.is_string_arg(parameter) and parameter['type'] == 'ViInt8[]':
       std::string ${parameter_name}(${size}, '\0');
-## real strings leverages the underlying string buffer's null terminator as part of the size
+## Driver string APIs require room in the buffer for the null terminator. We need to account for that when sizing the string.
 %     elif service_helpers.is_string_arg(parameter) and common_helpers.get_size_mechanism(parameter) == 'fixed':
-      std::string ${parameter_name}(${size}-1, '\0');
+      std::string ${parameter_name}(${size} - 1, '\0');
 %     elif service_helpers.is_string_arg(parameter):
       std::string ${parameter_name};
       if (${size} > 0) {
--- a/source/codegen/templates/service_helpers.mako
+++ b/source/codegen/templates/service_helpers.mako
@@ -218,8 +218,13 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
 %>\
 %     if common_helpers.is_struct(parameter) or underlying_param_type == 'ViBoolean':
       std::vector<${underlying_param_type}> ${parameter_name}(${size}, ${underlying_param_type}());
+%     elif service_helpers.is_string_arg(parameter) and common_helpers.get_size_mechanism(parameter) == 'fixed':
+      std::string ${parameter_name}(${size}-1, '\0');
 %     elif service_helpers.is_string_arg(parameter):
-      std::string ${parameter_name}(${size}, '\0');
+      std::string ${parameter_name};
+      if (${size} > 0) {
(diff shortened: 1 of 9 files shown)


reckenro (approved):

source/codegen/templates/service_helpers.mako
```diff
@@ -218,8 +218,17 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
 %>\
 %     if common_helpers.is_struct(parameter) or underlying_param_type == 'ViBoolean':
       std::vector<${underlying_param_type}> ${parameter_name}(${size}, ${underlying_param_type}());
-%     elif service_helpers.is_string_arg(parameter):
+## byte arrays are leveraging a string as a simple buffer wrapper
+%     elif service_helpers.is_string_arg(parameter) and parameter['type'] == 'ViInt8[]':
       std::string ${parameter_name}(${size}, '\0');
+## real strings leverages the underlying string buffer's null terminator as part of the size
```
jonlugoatx, line 224: Are we confident this is the case for all drivers?

source/codegen/templates/service_helpers.mako
```diff
@@ -218,8 +218,17 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
...
+%     elif service_helpers.is_string_arg(parameter) and common_helpers.get_size_mechanism(parameter) == 'fixed':
+      std::string ${parameter_name}(${size}-1, '\0');
```
jonlugoatx, line 226: Add spaces around the `-` operator. Same below.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -980,7 +980,8 @@ TEST(NiFakeServiceTests, NiFakeService_ReturnMultipleTypes_CallsReturnMultipleTy
   EXPECT_EQ(nifake_grpc::FloatEnum::FLOAT_ENUM_SIX_POINT_FIVE, response.a_float_enum());
   EXPECT_EQ(a_float_enum, response.a_float_enum_raw());
   EXPECT_THAT(response.an_array(), ElementsAreArray(an_array, array_size));
-  EXPECT_THAT(response.a_string(), ElementsAreArray(a_string, string_size));
+  EXPECT_STREQ(response.a_string().c_str(), a_string);
+  EXPECT_EQ(response.a_string().length(), string_size-1);
```
jonlugoatx, line 984: Space around `-`.

jonlugoatx (approved):

source/codegen/templates/service_helpers.mako
zhindes: This (poorly worded - I'll fix!) comment is referring to the code we're generating, not the behavior of the driver. The drivers definitely require that the buffer be big enough to hold the null-terminator. You can search ni-code for `sizeof('\0')` to see lots of examples!

tkrebes (approved):

source/codegen/templates/service_helpers.mako
```diff
@@ -218,8 +218,17 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
 %>\
 %     if common_helpers.is_struct(parameter) or underlying_param_type == 'ViBoolean':
       std::vector<${underlying_param_type}> ${parameter_name}(${size}, ${underlying_param_type}());
-%     elif service_helpers.is_string_arg(parameter):
+## Byte arrays are leveraging a string as a buffer, so we don't need to take special consideration of the null terminator.
+%     elif service_helpers.is_string_arg(parameter) and parameter['type'] == 'ViInt8[]':
```
jonlugoatx, line 222: @zhindes, I just caught this, but `service_helpers.is_string_arg(parameter)` is redundant with `parameter['type'] == `ViInt8[]`, and we should have also included `ViUInt8[]` in the list. Not sure it's worth spinning up a PR to fix it though.

zhindes: Good catches, I'll make a new PR today

zhindes: FYI - We don't seem to use ViUInt8 anywhere, yet. The proto helper only has this case:
```
  if 'ViInt8' in type:
    if is_array == True:
      type = "bytes"
      add_repeated = False
    else:
      type = 'uint32'
```

zhindes: https://github.com/ni/grpc-device/pull/206

PR #205, opened 2021-05-19
[Owners] [Tiny] Fix enum name for few dmm attributes

doshirohan:

There was typo in name of enums for some of the attributes in metadata file. Somehow our current codegen didn't catch that theses enums are not present in `enums.py` and fail. Changes I am making for per-type enum for attribute values fails build because they can't find matching enums.

This changes fixes `attributes.py` metadata for DMM. There are no generated file changes since there are other attributes/functions using the correct enum name, which added them to the `.proto` file.

No changes in generated files, so no testing required.

### DIFF
--- a/source/codegen/metadata/nidmm/attributes.py
+++ b/source/codegen/metadata/nidmm/attributes.py
@@ -388,7 +388,7 @@ attributes = {
     1150120: {
         'access': 'read-write',
         'channel_based': False,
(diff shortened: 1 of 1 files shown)


reckenro (approved):

reckenro (approved):
@jonlugoatx , simple change so I removed you. GitHub seems to let me so we could try doing this on simple changes.

PR #200, opened 2021-05-14
Updated nisync functions.py with a code generated version

tkrebes:
### What does this Pull Request accomplish?

This adds an implementation for the whole NI-Sync C API (expect niSync_ReadMultipleTriggerTimeStamp)
- Checked in codegen'ed version of functions.py
- Renamed `terminalName` parameters to `terminal` to match names found in nisync.h.

### Why should this Pull Request be merged?

This provides a version of functions.py with the whole NI-Sync C API that follows the black style.

### What testing has been done?

Ran `.\SystemTestsRunner.exe --gtest_filter=NiSyncSessionTest*`
```
Note: Google Test filter = NiSyncSessionTest*
[==========] Running 5 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 5 tests from NiSyncSessionTest
[ RUN      ] NiSyncSessionTest.InitializeSessionWithDeviceAndSessionName_CreatesDriverSession
[       OK ] NiSyncSessionTest.InitializeSessionWithDeviceAndSessionName_CreatesDriverSession (1807 ms)
[ RUN      ] NiSyncSessionTest.InitializeSessionWithDeviceAndNoSessionName_CreatesDriverSession
[       OK ] NiSyncSessionTest.InitializeSessionWithDeviceAndNoSessionName_CreatesDriverSession (8 ms)
[ RUN      ] NiSyncSessionTest.InitializeSessionWithoutDevice_ReturnsDriverError
[       OK ] NiSyncSessionTest.InitializeSessionWithoutDevice_ReturnsDriverError (16 ms)
[ RUN      ] NiSyncSessionTest.InitializedSession_CloseSession_ClosesDriverSession
[       OK ] NiSyncSessionTest.InitializedSession_CloseSession_ClosesDriverSession (9 ms)
[ RUN      ] NiSyncSessionTest.InvalidSession_CloseSession_NoErrorReported
[       OK ] NiSyncSessionTest.InvalidSession_CloseSession_NoErrorReported (2 ms)
[----------] 5 tests from NiSyncSessionTest (1849 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 1 test suite ran. (1858 ms total)
[  PASSED  ] 5 tests.
```

PR #207, opened 2021-05-19
Changed nisync lower_case_with_underscores function names to CamelCase

tkrebes:

- Changed lower_case_with_underscores function names to CamelCase
  This is a follow up to PR #200 
- Updated nisync CHANGES.md since functions.py is codegen'ed and checked in without modification

The NiSyncLibraryInterface function names follow a consistent naming convention.

Built locally on Windows machine.

### DIFF
--- a/generated/nisync/nisync_library.cpp
+++ b/generated/nisync/nisync_library.cpp
@@ -21,13 +21,13 @@ NiSyncLibrary::NiSyncLibrary() : shared_library_(kLibraryName)
   if (!loaded) {
     return;
   }
-  function_pointers_.init = reinterpret_cast<initPtr>(shared_library_.get_function_pointer("niSync_init"));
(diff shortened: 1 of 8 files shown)


zhindes (approved):

reckenro (approved):
I think you need to update `"close_function"` in config.py to also be "Close" now (see [here](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nisync/config.py#L10)). That's what's used in all of the Init methods' close lambdas.

jonlugoatx (approved):
I'm not opposed to these changes, but since they only affect the library wrapper, it might be simpler to leave them the way they were so that the metadata more closely matches the C header case. The code generator already converts snake_case C header names to CamelCase for the service handler methods.

PR #210, opened 2021-05-21
[Owners] Create consolidated attribute values enum based on attribute data types

doshirohan:

Earlier we used to define separate enum per enum name defined in `enums.py` for attribute values. This has usability issue where users would face difficulty figuring out list of possible values for given attribute. This PR addresses that by creating single consolidated enum per-type of attribute which contains all possible constant values of that data type. With this approach we would have enums for ViInt32, ViInt64, ViReal64, ViString type attributes.

With this, whenever user wants to call `SetAttribute*` or `CheckAttribute*` methods for specific attribute they can figure out supported values for the attribute by looking at enum for corresponding type and intellisense would show all possible values. Users can now filter the values by typing part of attribute name or by copying values from C API help file and let intellisense do partial match to show matching constants defined in the enum.

image

This solves the discoverability issue for attribute values to great extent. As long as users know data type of the attribute, they can easily lookup pre-defined values for the attribute under corresponding enum and further filter using attribute name or constants specified in C documentation.

### Implementation
- Added new `*AttributeValues` enums per data type consolidating all enums that are applicable for attribute values
- Updated signature of `SetAttribute*` and `CheckAttribute*` APIs to accept enums of corresponding type alongside raw values.
- Both of above is achieved in codegen by using `metadata_mutation` to add metadata for new enums that need to be created and tagging function metadata with appropriate enum names.
- Any enums tagged with `generate-mappings` are ignored from consolidated enums as of now, since we do not have enums in our released drivers which need this.
- If any of the `AttributeValues` enum is empty, in case there are no enums/constants for attributes of that type, we do not generate the consolidated enum and also not update signature of the method to accept the enum
- Introduced new `proto_values` that can be added on `function` in metadata to specify value to be assigned for specific parameter in the proto file. This is to maintain backward compatibility of messages whose signature is changing because of this change (scope and switch drivers as of now).
 
### Further improvements
- We need to identify mechanism to uniquely number enum values in case of `generate-mappings`. Current approach of using incremental numbers in our codegen has shortcoming of breaking backward compatibility when new enums are added. Further PR (if needed) will implement solution for this case.

- Builds pass
- Existing auto tests pass on windows
- Executed all shipping examples manually


source/codegen/common_helpers.py
```diff
@@ -77,17 +79,24 @@ def filter_proto_rpc_functions(functions):
   functions_for_proto = {'public', 'CustomCode'}
   return [name for name, function in functions.items() if function.get('codegen_method', 'public') in functions_for_proto]
 
-def get_used_enums(functions, attributes):
-  '''Returns a set of enums used with functions or attributes.'''
-  used_enums = set()
+def get_attribute_enums(attributes):
```
doshirohan, line 82: note: Returns something like...
```
{
  'ViInt32': {'VideoPolarity', 'NotificationType', 'TriggerType'},
  'ViReal64': {'CableSenseMode'},
  'ViInt64': {'AddressType'}
}
```

source/codegen/templates/proto_helpers.mako
```diff
@@ -53,6 +53,48 @@ enum ${enum_name} {
...
+<%
+  config = data["config"]
+  service_class_prefix = config["service_class_prefix"]
+  enums = data["enums"]
+%>\
+% for type_name in attribute_enums:
+<%
+  type_enum_name = service_class_prefix + type_name[2:] + "AttributeValues"
+  enum_value_type_specific_prefix = (f"{service_class_prefix.upper()}_{type_name[2:]}_VAL").upper()
+  values_to_create = { enum_value_type_specific_prefix + "_UNSPECIFIED": 0 }
```
doshirohan, line 67: note: Unlike function enums, we need to create dictionary of all enum names/values before writing to output file so as to check if to set `allow_alias` by checking duplicate values across values from all the enums that we are combining.

doshirohan:
@jonlugoatx  can you look at this review before I add other owners?

harsha-bhushan:
Could you also add a bullet point under point 3 here on using these enums. You could ask Jeff to do a quick review of the write up. - https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client#converting-c-api-calls-to-grpc

source/tests/system/nidcpower_driver_api_tests.cpp
```diff
@@ -357,7 +357,7 @@ TEST_F(NiDCPowerDriverApiTest, SetAttributeViReal64_GetAttributeViReal64ReturnsS
   set_int32_attribute(
       channel_name,
       dcpower::NiDCPowerAttributes::NIDCPOWER_ATTRIBUTE_MEASURE_WHEN,
-      dcpower::MeasureWhen::MEASURE_WHEN_NIDCPOWER_VAL_AUTOMATICALLY_AFTER_SOURCE_COMPLETE);
+      dcpower::NiDCPowerInt32AttributeValues::NIDCPOWER_INT32_VAL_MEASURE_WHEN_AUTOMATICALLY_AFTER_SOURCE_COMPLETE);
```
kruthi-NI, line 360: I feel it would be nice if the value was same as what is in C header which is "NIDCPOWER_VAL_AUTOMATICALLY_AFTER_SOURCE_COMPLETE" . But currently it had INT32 as well as attribute name. Was there any reason to have both of those?



NIDCPOWER_VAL_AUTOMATICALLY_AFTER_SOURCE_COMPLETE
--

kruthi-NI (approved):

jonlugoatx:
> @jonlugoatx can you look at this review before I add other owners?

Sorry, I missed this comment.

source/codegen/templates/proto_helpers.mako
```diff
@@ -53,6 +53,48 @@ enum ${enum_name} {
...
+  for enum_name in sorted(attribute_enums[type_name]):
+    enum_value_prefix = enum_value_type_specific_prefix + "_" + common_helpers.pascal_to_snake(enum_name)
+    enum = enums[enum_name]
+    nonint_index = 1
+    for value in enum["values"]:
+      enum_value_suffix = value['name'].replace(f"{service_class_prefix.upper()}_VAL_", "")
+      value_name = f"{enum_value_prefix}_{enum_value_suffix}".upper()
+      if enum.get("generate-mappings", False):
```
jonlugoatx, line 75: If one of the values needs to generate a mapping, then all of them need to be mapped. You can't do mapping on a case by case basis.

source/codegen/templates/proto_helpers.mako
```diff
@@ -53,6 +53,48 @@ enum ${enum_name} {
...
+  type_enum_name = service_class_prefix + type_name[2:] + "AttributeValues"
+  enum_value_type_specific_prefix = (f"{service_class_prefix.upper()}_{type_name[2:]}_VAL").upper()
+  values_to_create = { enum_value_type_specific_prefix + "_UNSPECIFIED": 0 }
+  for enum_name in sorted(attribute_enums[type_name]):
+    enum_value_prefix = enum_value_type_specific_prefix + "_" + common_helpers.pascal_to_snake(enum_name)
+    enum = enums[enum_name]
+    nonint_index = 1
+    for value in enum["values"]:
+      enum_value_suffix = value['name'].replace(f"{service_class_prefix.upper()}_VAL_", "")
+      value_name = f"{enum_value_prefix}_{enum_value_suffix}".upper()
```
jonlugoatx, line 74:
```python
enum_value_prefix_to_replace = service_class_prefix.upper() + "_VAL_"
...
for value in enum["values"]:
  value_name = value['name'].replace(enum_val_prefix_to_replace, enum_value_prefix)
```

source/codegen/common_helpers.py
```diff
@@ -77,17 +79,24 @@ def filter_proto_rpc_functions(functions):
...
+  '''Returns a dictionary of different attribute data types that use enum alongwith set of enums used'''
+  attribute_enums = defaultdict(set)
+  for attribute in attributes:
+    if("enum" in attributes[attribute]):
```
jonlugoatx, line 86: Parens not needed.

generated/nisync/nisync.proto
```diff
@@ -200,6 +200,18 @@ enum NiSyncAttributes {
   NISYNC_ATTRIBUTE_8021AS_AS_CAPABLE = 1151115;
 }
```
doshirohan, line 202: There are absolutely on enums in this file. Does nisync not have constants in header file? Or did they not port those constants as enums.

source/codegen/templates/proto_helpers.mako
```diff
@@ -53,6 +53,48 @@ enum ${enum_name} {
...
+  service_class_prefix = config["service_class_prefix"]
+  enums = data["enums"]
+%>\
+% for type_name in attribute_enums:
+<%
+  type_enum_name = service_class_prefix + type_name[2:] + "AttributeValues"
+  enum_value_type_specific_prefix = (f"{service_class_prefix.upper()}_{type_name[2:]}_VAL").upper()
+  values_to_create = { enum_value_type_specific_prefix + "_UNSPECIFIED": 0 }
+  for enum_name in sorted(attribute_enums[type_name]):
+    enum_value_prefix = enum_value_type_specific_prefix + "_" + common_helpers.pascal_to_snake(enum_name)
```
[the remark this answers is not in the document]
doshirohan: Enum name needs to be part of value. In case same enum value is defined in multiple enums in enums.py, not prefixing enum name here would cause conflict. 

Example from dcpower: Two of the values from [PowerSource](https://github.com/doshirohan/grpc-device/blob/main/source/codegen/metadata/nidcpower/enums.py#L521) are also present in `PowerSourceInUse` enum.

I have updated enum names to use your suggestion though and moved `_VAL_` after all the prefixes to keep names close to what C APIs have.

source/codegen/templates/proto_helpers.mako
doshirohan: updated.

source/codegen/templates/proto_helpers.mako
doshirohan: Yes, I have updated code to do `generate-mappings` by looking for duplicate values over entire consolidated enum, instead of looking at individual split enums.

source/codegen/common_helpers.py
doshirohan: Remove redundant parenthesis.

source/codegen/metadata_mutation.py
```diff
@@ -50,3 +50,51 @@ def mark_non_proto_params(parameters):
...
+        enums.update({type_enum_name: new_enum})
+    
+def mark_attr_value_param_if_required(function, attribute_enums, service_class_prefix):
+    """For SetAttribute* and CheckAttribute* APIs, update function metadata to mark attribute_value parameter as enum."""
+    attribute_value_param = next((param for param in function["parameters"] if param["name"] in {"value", "attributeValue"}), None)
+    if attribute_value_param["type"] == "ViConstString":
+        enum_type = "ViString"
+    else:
+        enum_type = attribute_value_param["type"]
+    if(attribute_value_param != None and enum_type in attribute_enums):
```
jonlugoatx, line 96: Space out the `if` and `(`. Same elsewhere.

source/codegen/metadata_mutation.py
```diff
@@ -50,3 +50,51 @@ def mark_non_proto_params(parameters):
...
+        values_to_create = {}
+        for enum_name in sorted(attribute_enums[type_name]):
+            enum = enums[enum_name]
+            for value in enum["values"]:
+                value_name = value['name'].replace(f"{service_class_prefix.upper()}_", f"{common_helpers.pascal_to_snake(enum_name).upper()}_")
+                value_value = value["value"]
+                values_to_create[value_name] = value_value
+        type_enum_name = get_attribute_values_enum_name(service_class_prefix, type_name)
+        enum_value_prefix = (f"{service_class_prefix}_{type_name[2:]}").upper()
+        generate_mappings = type_name in mapping_types
```
jonlugoatx, line 71: `type_name in mapping_types and values_to_create`

jonlugoatx (approved):

generated/niscope/niscope_service.h
```diff
@@ -125,6 +125,12 @@ class NiScopeService final : public NiScope::Service {
   void Copy(const std::vector<NIComplexNumber_struct>& input, google::protobuf::RepeatedPtrField<niscope_grpc::NIComplexNumber>* output);
   void Copy(const NIComplexI16_struct& input, niscope_grpc::NIComplexInt32* output);
   void Copy(const std::vector<NIComplexI16_struct>& input, google::protobuf::RepeatedPtrField<niscope_grpc::NIComplexInt32>* output);
+  std::map<std::int32_t, std::int64_t> niscopeint64attributevalues_input_map_ { {0, 0},{1, 0},{2, 1}, };
+  std::map<std::int64_t, std::int32_t> niscopeint64attributevalues_output_map_ { {0, 0},{0, 1},{1, 2}, };
+  std::map<std::int32_t, double> niscopereal64attributevalues_input_map_ { {0, 0},{1, 0},{2, 1}, };
+  std::map<double, std::int32_t> niscopereal64attributevalues_output_map_ { {0, 0},{0, 1},{1, 2}, };
+  std::map<std::int32_t, std::string> niscopestringattributevalues_input_map_ {  };
+  std::map<std::string, std::int32_t> niscopestringattributevalues_output_map_ {  };
```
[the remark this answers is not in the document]
doshirohan: We do need to do this to prevent backward compatibility issues in future. We need to generate empty enum and add that in `SetAttribute*` method signature, so that in case new enums are introduced in future - we don't break backward compatibility for the API by updating API signature.

source/codegen/metadata_mutation.py
```diff
@@ -50,3 +50,51 @@ def mark_non_proto_params(parameters):
...
+            'allow_alias': allow_alias,
+            'values': values
+        }
+        if generate_mappings:
+            new_enum['value-type'] = mapping_types[type_name]
+        enums.update({type_enum_name: new_enum})
+    
+def mark_attr_value_param_if_required(function, attribute_enums, service_class_prefix):
+    """For SetAttribute* and CheckAttribute* APIs, update function metadata to mark attribute_value parameter as enum."""
+    attribute_value_param = next((param for param in function["parameters"] if param["name"] in {"value", "attributeValue"}), None)
```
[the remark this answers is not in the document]
doshirohan: dmm, dcpower and switch use `attributeValue` in metadata file. tClk, nidigital and scope use `value`.

generated/nidcpower/nidcpower.proto
```diff
@@ -1984,7 +1995,10 @@ message SetAttributeViInt64Request {
   nidevice_grpc.Session vi = 1;
   string channel_name = 2;
   NiDCPowerAttributes attribute_id = 3;
-  int64 attribute_value = 4;
+  oneof attribute_value_enum {
+    NiDCPowerInt64AttributeValues attribute_value = 4;
```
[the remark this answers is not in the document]
doshirohan: Removed auto-generation of mappings for Int64 types.

source/codegen/metadata_mutation.py
```diff
@@ -50,3 +50,51 @@ def mark_non_proto_params(parameters):
...
+        type_enum_name = get_attribute_values_enum_name(service_class_prefix, type_name)
+        enum_value_prefix = (f"{service_class_prefix}_{type_name[2:]}").upper()
+        generate_mappings = type_name in mapping_types
+        allow_alias = False
+        if not generate_mappings:
+            enum_values = values_to_create.values()
+            allow_alias = len(enum_values) != len(set(enum_values))
+        values = []
+        for name in values_to_create:
+            values.append({"name": name, "value": values_to_create[name]})
```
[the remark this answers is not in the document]
doshirohan: Updated to use list comprehension.

generated/niscope/niscope_service.h
```diff
@@ -125,6 +125,12 @@ class NiScopeService final : public NiScope::Service {
   void Copy(const std::vector<NIComplexNumber_struct>& input, google::protobuf::RepeatedPtrField<niscope_grpc::NIComplexNumber>* output);
   void Copy(const NIComplexI16_struct& input, niscope_grpc::NIComplexInt32* output);
   void Copy(const std::vector<NIComplexI16_struct>& input, google::protobuf::RepeatedPtrField<niscope_grpc::NIComplexInt32>* output);
+  std::map<std::int32_t, std::int64_t> niscopeint64attributevalues_input_map_ { {0, 0},{1, 0},{2, 1}, };
+  std::map<std::int64_t, std::int32_t> niscopeint64attributevalues_output_map_ { {0, 0},{0, 1},{1, 2}, };
+  std::map<std::int32_t, double> niscopereal64attributevalues_input_map_ { {0, 0},{1, 0},{2, 1}, };
+  std::map<double, std::int32_t> niscopereal64attributevalues_output_map_ { {0, 0},{0, 1},{1, 2}, };
```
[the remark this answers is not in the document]
doshirohan: In subsequent draft PR I will be trying out mapping using different mechanism which should solve confusing mapping of 0 to 1 and 1 to 2. If that doesn't work out, I will add something to prevent this enum from generating. Keeping the comment open till then.

source/codegen/metadata_mutation.py
doshirohan: Removed `(` around `if` conditions everywhere. Old C# habits.... :)

source/codegen/metadata_mutation.py
doshirohan: Even if there are no `values_to_create`, we do need to create mappings dictionary. Otherwise we will have weird typecasting on server. for e.g. if there are no values to create for ViStringAttributeValues enum, typecasting on server would try to typecast `request->attribute_value()`, which is an enum to ViString and throw compile time error.

source/codegen/service_helpers.py
```diff
@@ -55,6 +55,8 @@ def create_param(parameter):
         return f'{type} {name}'
 
 def python_to_c(enum):
```
doshirohan, line 57: note: If the enum is empty, return default value. This could happen now since we could have empty `AttributeValues` enums for specific types if there are no attributes of that type. Same check is done for some of the helpers below as well.

doshirohan (commented):
@jonlugoatx I haven't incorporated some of your suggestions. Resetting you so that you can look at replies to those unresolved comments.

generated/nisync/nisync.proto
reckenro: I'm guessing they just haven't ported any constants over as enums. There's a handful of constants in the header file in the includes folder at least: [include/nisync.h](https://github.com/ni/grpc-device/blob/main/imports/include/nisync.h)

reckenro (approved):

generated/niscope/niscope_service.h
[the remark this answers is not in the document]
jonlugoatx: The maps are implementation details. If we add them later that wouldn't break compatibility over the wire. The main source of compatibility breakage would be if we change something in the .proto file or reinterpret the meaning or behavior of an attribute value. If it's easier to generate the empty maps to keep the .proto file the same then I can understand that.

generated/niscope/niscope_service.h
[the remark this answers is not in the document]
jonlugoatx: Like I said in the other PR, it would be good to find a way to not map 1->0 and 2->1.

generated/niscope/niscope_service.h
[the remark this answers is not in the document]
doshirohan: Ah, that's correct. I could just not add the `generate-mappings` tag to enums if they are empty during metadata mutation in such case. I was thinking it would be difficult to get the information about if enum is empty or not in so many different places where map is referenced, but not adding the tag would do the trick. I will update to remove these maps.

kylesull30 (approved):

generated/nidcpower/nidcpower.proto
```diff
@@ -2013,7 +1987,7 @@ message SetAttributeViInt64Request {
   nidevice_grpc.Session vi = 1;
   string channel_name = 2;
   NiDCPowerAttributes attribute_id = 3;
-  int64 attribute_value = 4;
+  int64 attribute_value_raw = 4;
```
doshirohan, line 1990: Note: This rename is to keep all SetAttribute APIs consistent. Even though this particular instance of SetAttribute doesn't have enum yet, adding `_raw` in the name makes it clear that users are supposed to pass in raw value.

generated/niscope/niscope.proto
```diff
@@ -927,7 +909,10 @@ message CheckAttributeViInt32Request {
   nidevice_grpc.Session vi = 1;
   string channel_list = 2;
   NiScopeAttributes attribute_id = 3;
-  sint32 value = 4;
+  oneof value_enum {
```
doshirohan, line 912: Note: This function signature is kept consistent using override from metadata file. Even though this particular override was not required, since the raw value and enum's constant values are still same (no mappings used yet).

source/codegen/templates/proto_helpers.mako
```diff
@@ -69,66 +69,79 @@ ${lookup.get_template(custom_template).render()}
 </%def>
 
 ## Define a proto request message for a given API function.
-<%def name="define_request_message(function, input_parameters)">\
+<%def name="define_request_message(function, input_parameters, proto_values)">\
```
doshirohan, line 72: Note: Changes here and in `define_response_message` below are for consuming proto-value overrides provided through metadata.

source/codegen/metadata_mutation.py
doshirohan: Updated to early return if there are no `values_to_create`.

generated/niscope/niscope_service.h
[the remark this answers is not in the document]
doshirohan: This weird mapping of 1 to 0 and 2 to 1 is now resolved, since we use unmapped values everywhere.

generated/niscope/niscope_service.h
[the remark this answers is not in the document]
doshirohan: Removed the addition of `generate-mappings` completely, since it is not used by any of the drivers yet. Addition of `generate-mappings` will be handled in separate PR later.

jonlugoatx (approved):
I'm fine with you submitting once you address my comments about the field number metadata.

source/codegen/metadata/niscope/functions.py
```diff
@@ -277,7 +277,10 @@
                 'type':'ViInt32'
             }
         ],
-        'returns':'ViStatus'
+        'returns':'ViStatus',
+        'proto_values': {
```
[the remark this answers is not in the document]
doshirohan: Actually, the reason I had added separate `proto_values` dictionary at function level was for easy lookup of "field names" in our codegen to check if the field should be assigned custom field number. The key for entries in dictionary was name of the proto field which should be assigned custom field number, so we would have entries like `value`, `value_raw`, `value_mapped` etc (`value` being name of the field in this specific case).

That said, I think keeping that information with parameter is fine too. I will add `grpc_field_number` attribute on all of these functions I had modified earlier. `grpc_raw_field_number` need not be explicitly mentioned in these cases, since they will automatically get next incremental number.

source/codegen/templates/proto_helpers.mako
```diff
@@ -69,66 +69,79 @@ ${lookup.get_template(custom_template).render()}
...
 %>\
 message ${common_helpers.snake_to_pascal(function)}Request {
 % for parameter in input_parameters:
 <%
-  index  = index + 1
+  parameter_name = common_helpers.camel_to_snake(parameter["name"])
+  parameter_index = proto_values.get(parameter_name, None)
+  if not parameter_index:
+    parameter_index = index
+    index = index + 1
```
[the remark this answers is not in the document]
doshirohan: Updated to use helper function and lookup using range instead of incrementing the index.

source/codegen/templates/proto_helpers.mako
```diff
@@ -69,66 +69,79 @@ ${lookup.get_template(custom_template).render()}
...
   else:
     parameter_type = proto_helpers.determine_function_parameter_type(parameter, service_class_prefix)
 %>\
 %   if common_helpers.is_enum(parameter):
 <%
-  index = index + 1
   is_array = common_helpers.is_array(parameter["type"])
   non_enum_type = proto_helpers.get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
-  parameter_name = common_helpers.camel_to_snake(parameter["name"])
+  parameter_raw_index = proto_values.get(f"{parameter_name}_raw", None)
```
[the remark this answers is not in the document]
doshirohan: Updated to move field number info to `parameter`

PR #211, opened 2021-05-21
Add syslog support to ni_grpc_device_server

zhindes:

This adds syslog support to ni_grpc_device_server. It does this by:
* Adding a simple logging interface to ni_grpc_device_server.
   * This should be used instead of direct stdout/stderr access.
   * The implementation uses a simple va_list-function that can be implemented for other logging.
   * There is a default stdout/stderr implementation
* Implemented a syslog logging function available on linux
* Added _slightly_ more robust command line option parsing with support for `--use-syslog`. I expect this may be replaced with a more generic daemon option in the future.

This is early work towards making ni_grpc_device_server natively-daemonizable for on-boot configuration.

Normal invocation and sad/happy paths:
```
admin@NI-PXIe-8880-03095FC1:~/grpc# ./ni_grpc_device_server
Server listening on port 31763
Security is configured with insecure credentials.
^C
admin@NI-PXIe-8880-03095FC1:~/grpc# ./ni_grpc_device_server foo
Failed to parse config: The server configuration file was not found at: foo
```

syslog invocation with sad/happy paths:
```
admin@NI-PXIe-8880-03095FC1:~/grpc# ./ni_grpc_device_server --use-syslog
^C
admin@NI-PXIe-8880-03095FC1:~/grpc# ./ni_grpc_device_server --use-syslog foo
admin@NI-PXIe-8880-03095FC1:~/grpc# tail /var/log/messages
...
2021-05-21T19:16:47.096+00:00 NI-PXIe-8880-03095FC1 ni_grpc_server[31221]: Server listening on port 31763
2021-05-21T19:16:47.096+00:00 NI-PXIe-8880-03095FC1 ni_grpc_server[31221]: Security is configured with insecure credentials.
2021-05-21T19:16:54.260+00:00 NI-PXIe-8880-03095FC1 ni_grpc_server[31255]: Failed to parse config: The server configuration file was not found at: foo
````

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,15 @@ foreach(api ${nidrivers})
   endif()
 endforeach()
 
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  set(SOURCE_OS
+    "source/server/linux/syslog_logging.cpp")
+else()
+  set(SOURCE_OS "")
+endif()
+
 add_executable(ni_grpc_device_server
+   ${SOURCE_OS}
    "source/server/core_server.cpp"
    "source/server/device_enumerator.cpp"
    "source/server/logging.cpp"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,13 +176,6 @@ foreach(api ${nidrivers})
   endif()
 endforeach()
 
-if(CMAKE_SYSTEM_NAME STREQUAL Linux)
-  set(OSDEP_SOURCE
-    "source/server/linux/syslog_logging.cpp")
-else()
-  set(OSDEP_SOURCE "")
-endif()
-
 add_executable(ni_grpc_device_server
    ${OSDEP_SOURCE}
    "source/server/core_server.cpp"
@@ -198,6 +191,11 @@ add_executable(ni_grpc_device_server
    ${session_proto_srcs}
    ${session_grpc_srcs}
    ${nidriver_service_srcs})
+
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  target_sources(ni_grpc_device_server
+    PRIVATE "source/server/linux/syslog_logging.cpp")
+endif()
 
 target_link_libraries(ni_grpc_device_server
    ${_REFLECTION}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,7 @@ endforeach()
 add_executable(ni_grpc_device_server
    "source/server/core_server.cpp"
    "source/server/device_enumerator.cpp"
+   "source/server/logging.cpp"
    "source/server/semaphore.cpp"
    "source/server/server_configuration_parser.cpp"
    "source/server/server_security_configuration.cpp"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,14 +177,14 @@ foreach(api ${nidrivers})
 endforeach()
 
 if(CMAKE_SYSTEM_NAME STREQUAL Linux)
-  set(SOURCE_OS
+  set(OSDEP_SOURCE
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: source/server/linux/syslog_logging.cpp [hunk 1]
```diff
@@ -0,0 +1,44 @@
+#include "syslog_logging.h"
+
+#include <syslog.h>
+
+namespace nidevice_grpc {
+namespace Logging {
+
+static const char* syslog_identity = "ni_grpc_server";
```
zhindes, line 8: that name is wrong, it should be "ni_grpc_device_server"

CMakeLists.txt
```diff
@@ -176,9 +176,18 @@ foreach(api ${nidrivers})
   endif()
 endforeach()
 
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  set(SOURCE_OS
```
jonlugoatx, line 180: How about `OSDEP_SOURCE`? Without looking at where this is used and what it's set to, I would expect a variable named `SOURCE_OS` to contain the name of an operating system.

source/server/core_server.cpp
```diff
@@ -87,25 +96,74 @@ static void RunServer(const std::string& config_file_path)
...
+{
+  Options options;
+  for (int i = 1; i < argc; ++i) {
+ #if defined(__GNUC__)
+    if (strcmp("--use-syslog", argv[i]) == 0) {
+      options.use_syslog = true;
+    }
+    else
+ #endif
+    if (strcmp("--help", argv[i]) == 0) {
```
jonlugoatx, line 141: Add `-h` too.

jonlugoatx (approved):

CMakeLists.txt
```diff
@@ -176,9 +176,18 @@ foreach(api ${nidrivers})
   endif()
 endforeach()
 
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+  set(OSDEP_SOURCE
+    "source/server/linux/syslog_logging.cpp")
+else()
+  set(OSDEP_SOURCE "")
+endif()
```
SparkingSpork, line 184: I think an alternative would be to add the source conditionally with[ target_sources](https://cmake.org/cmake/help/latest/command/target_sources.html), but I've never tried that so feel free to leave as-is for now.
```
if(CMAKE_SYSTEM_NAME STREQUAL Linux)
   target_sources(ni_grpc_device_server
      PRIVATE "source/server/linux/syslog_logging.cpp")
endif()
```

SparkingSpork (approved):

source/server/linux/syslog_logging.cpp
```diff
@@ -3,7 +3,8 @@
 #include <syslog.h>
 
 namespace nidevice_grpc {
-namespace Logging {
+
```
jonlugoatx, line 6: We tend not to have blank lines between our namespaces. Same with the `}`s at the end.

jonlugoatx (approved):

tkrebes (approved):

CMakeLists.txt
zhindes: That worked great, thanks! I'm new at CMake, so it's been interesting to try it out.

CMakeLists.txt
```diff
@@ -177,8 +177,10 @@ foreach(api ${nidrivers})
 endforeach()
 
 add_executable(ni_grpc_device_server
+   ${OSDEP_SOURCE}
```
SparkingSpork, line 180: You shouldn't need this anymore.

CMakeLists.txt
SparkingSpork: Glad it worked! I'm definitely still learning about it as we've refined things here.

PR #212, opened 2021-05-21
Fix Windows build

zhindes:

Fixes the Windows build. It currently fails because `static_cast`ing from floating-point to an enum isn't allowed. You gotta stop by the underlying enum type on the way.

Fix the Windows build.

It builds on both Linux and Windows.

### DIFF
--- a/source/codegen/templates/service_helpers.mako
+++ b/source/codegen/templates/service_helpers.mako
@@ -263,6 +263,8 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
         if(${iterator_name} != ${map_name}.end()) {
           response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(${iterator_name}->second));
         }
+%     elif parameter['type'] == 'ViReal64':
+        response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(static_cast<int>(${parameter_name})));
 %     else:
         response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(${parameter_name}));
 %     endif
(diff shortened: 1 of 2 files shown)


reckenro (approved):

source/codegen/templates/service_helpers.mako
```diff
@@ -263,6 +263,8 @@ one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{PascalFieldNam
         if(${iterator_name} != ${map_name}.end()) {
           response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(${iterator_name}->second));
         }
+%     elif parameter['type'] == 'ViReal64':
+        response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(static_cast<int>(${parameter_name})));
```
jonlugoatx, line 267: The problem is that we're defining `int` enums for `ViReal64` args. Our code generator supports defining the enum value to a floating point and generating maps from sequential `int` values to actual values. I think we should mark the aperture time enum with [`generate-mappings`](https://github.com/ni/grpc-device/blob/main/source/codegen/templates/service_helpers.mako#L94) and assign floating-point values to them.

But then this ties into an issue that @doshirohan is looking into about how to generate those mappings in a backwards-compatible way.

jonlugoatx (changes requested):

source/codegen/templates/service_helpers.mako
zhindes: I think using enums for this is pretty gross, so I agree with your recommendation to use mappings instead. If you're fine with the Windows build being broken until it can be fixed by @doshirohan, then I can just drop this. These builds don't seem super critical since I think it has been broken for ~2 weeks since https://github.com/ni/grpc-device/commit/cb3cf4f07ffd3d1df0b25bdca3679e783a208b19.

jonlugoatx: We'll need stable builds by the time we release NI-DMM and NI-DCPower, but we need to work out the enum issue before then anyway.

PR #216, opened 2021-06-14
[Owners] Refactor codegen for enum input paramters

doshirohan:

Enum input parameter handling was done in two places earlier, one was `initialize_enum_input_param` that was handling only `generate-mappings` enum and for other enums codegen was happening during `initialize_standard_input_param`. This change combines both pieces of code under `initialize_enum_input_param`. This allows consistent generation of enum handling code for both above cases and would also make it easier to add support for mapped/unmapped enums subsequently.

Most of the changes in generated files are
1. Using `static_cast` instead of C-style cast for typecasting from enum to actual type. This is more restrictive and hence better than C-style cast. We were using `static_cast` for mapped enums, but not for unmapped enums earlier.
2. Wrapping of first case in enum switch block with `{ }`. This is required only for mapped enums to mark scope for dictionary iterator variable we define. Since the codegen is merged, the brackets get generated for both mapped and unmapped enums. Without the bracket, compiler fails with error "initialization of variable is skipped by case label".
image

This refactoring brings in consistency in the way code is generated for both mapped and unmapped enums. This will also make it easier go write codegen when we introduce separate enums and separate field `field_name_mapped` in parameters for mapped enums.

- Builds pass
- Manually verified that there are no other changes in generated files


Akshata440 (approved):

source/codegen/templates/service_helpers.mako
```diff
@@ -101,41 +101,48 @@ ${initialize_standard_input_param(function_name, parameter)}\
...
-          ${parameter_name} = const_cast<${parameter['type']}>((${iterator_name}->second).c_str());
+          ${parameter_name} = const_cast<${parameter['type']}>((${enum_request_snippet}).c_str());
 % else:
-          ${parameter_name} = static_cast<${parameter['type']}>(${iterator_name}->second);
+          ${parameter_name} = static_cast<${parameter['type']}>(${enum_request_snippet});
 % endif
           break;
         }
-        case ${enum_type_prefix}k${pascal_parameter_name}Raw: {
+        case ${one_of_case_prefix}::k${pascal_field_name}Raw:
```
jonlugoatx, line 136: Why keep the `{` on the first switch case but remove it from this one? I recommend putting the `{` and `}` back in.

source/codegen/templates/service_helpers.mako
```diff
@@ -101,41 +101,48 @@ ${initialize_standard_input_param(function_name, parameter)}\
 </%def>
 
 ## Initialize an enum input parameter for an API call.
-<%def name="initialize_enum_input_param(function_name, parameter)">\
+<%def name="initialize_enum_input_param(function_name, parameter, enums)">\
 <%
+  config = data['config']
```
jonlugoatx, line 106: You pull `config` out of `data` here. Why not do the same with `enums` instead of passing it in as a param?

jonlugoatx (approved):

source/codegen/templates/service_helpers.mako
doshirohan: Updated to add `{ }` for all cases in switch.

source/codegen/templates/service_helpers.mako
```diff
@@ -129,36 +126,46 @@ ${initialize_standard_input_param(function_name, parameter)}\
...
-          auto ${iterator_name} = ${map_name}.find(request->${parameter_name}());
+      switch (request->${field_name}_enum_case()) {
+        case ${one_of_case_prefix}::k${pascal_field_name}: {
+% if enums[parameter["enum"]].get("generate-mappings", False):
+          auto ${iterator_name} = ${map_name}.find(${enum_request_snippet});
           if (${iterator_name} == ${map_name}.end()) {
             return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The value for ${parameter_name} was not specified or out of range.");
           }
+<%
+enum_request_snippet = f'{iterator_name}->second'
```
astarche, line 150: Consider:
`enum_request_snippet`  is being reassigned in a style that's usually OK:
```c++
auto str = "one thing";
if (generate_mappings)
  str = "some other thing";
```

But it's a lot harder to read in mako. 

Can you give `iterator_name` a better name? `mapped_enum_request_iterator_name`?

Or Consider:
Would it work to pull it up into the python block?

```python
generate_mappings = enums[parameter["enum"]].get("generate-mappings", False)
direct_enum_request_snippet = f'request->{field_name}()'
mapped_enum_request_snippet =  f'{iterator_name}->second'
enum_request_snippet = mapped_enum_request_snippet if generate_mappings else direct_enum_request_snippet 
````

astarche (approved):
Looks good!

source/codegen/templates/service_helpers.mako
doshirohan: > Consider:
enum_request_snippet is being reassigned in a style that's usually OK:

I have a follow up request where we will have 3 variants of this `enum_request_snippet`, `mapped_enum_request_snippet` and `raw_request_snippet`, where this reassignment is not required. I am leaving it as is for now.

> Can you give iterator_name a better name? mapped_enum_request_iterator_name?

Updated to `mapped_enum_iterator_name` in both `initialize_enum_input_param` and `set_response_values` functions.

>  Would it work to pull it up into the python block?

I couldn't update `enum_request_snippet` directly based on `generate_mappings`, because I still need original snippet to use in `map_name.find(enum_request_snippet)` before reassigning the variable to use `mapped_enum_request_snippet`. Irrespective, this code is further refactored in subsequent PR to consider 3 cases, I can try to figure out best way to pull out pieces of code into python block.

PR #217, opened 2021-06-14
[Owners] [Tiny] Fix windows build error while typecasting float to enum

doshirohan:

On some dev machines build fails on windows when we try to typecast float values to enum. Since we know we use enums for floats without maps only when underlying type is `int`, we could safely typecast them to `int` and then typecast the int to enum.

This issue is observed only during `set_response_values`. While reading the input parameters, we typecast from enum to float, and that doesn't show any issues.

This changes fixes build issues observed on certain dev machines.

- Build passes on affected machine
- No specific testing done on execution, since there are no tests for the one API that got modified because of the change.

### DIFF
--- a/generated/nidmm/nidmm_service.cpp
+++ b/generated/nidmm/nidmm_service.cpp
@@ -1449,7 +1449,9 @@ namespace nidmm_grpc {
       auto status = library_->GetApertureTimeInfo(vi, &aperture_time, &aperture_time_units);
       response->set_status(status);
       if (status == 0) {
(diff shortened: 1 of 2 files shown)


kruthi-NI (approved):
Build is passing for this branch on my machine

doshirohan:
FYI @zhindes @tkrebes, since you have tried to fix this build issue in past. After finalizing on how we want mapped enums to be represented in gRPC, we decided best option to fix the build is by typecasting raw float values to int before typecasting it to enums. This shouldn't cause any precision lost since we would typecast only when underlying value is `int`. In other cases, enum response value will remain unspecified and clients will have to use `_raw` API to read the actual value.

jonlugoatx (approved):

PR #221, opened 2021-06-16
[Owner Review] Removed Calibration APIs and Enums from NI-DMM

ShantanuShinde:

Removed calibration APIs and enums from metadata as well as generated files as it is decided not to support external calibration in gRPC
https://nio365-my.sharepoint.com/:x:/g/personal/akshata_k_ni_com/EcTGOgHrnPhDrJCOJWMXe5sBMWJhiq-dItyT6nlvL5Xayg

Removed calibration APIs and enums

Build successful

### DIFF
--- a/source/codegen/metadata/nidmm/CHANGES.md
+++ b/source/codegen/metadata/nidmm/CHANGES.md
@@ -61,7 +61,6 @@ Enums with more than 2 consecutive uppercase letters in their name (as abbreviat
 
 Metadata for following enums added:
 - `'BufferSize'`
-- `'CalibrationCloseAction'`
 - `'CalibrationType'`
 - `'CompensatedOhms'`
 - `'ControlCommit'`
@@ -69,7 +68,6 @@ Metadata for following enums added:
 - `'Configuration4022'`
 - `'DcBias'`
 - `'Latency'`
-- `'MiscCalibration'`
 - `'PowerLineFrequencies'`
 - `'Range'`
 - `'SettleTime'`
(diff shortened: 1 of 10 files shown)


source/codegen/metadata/nidmm/CHANGES.md
```diff
@@ -133,3 +137,19 @@ with the session_repository.
 The following function was tagged with `'custom_close_method': True,` to ensure that the generated service handler of this function removes
```
doshirohan, line 137: Since `CloseExtCal` is removed, we don't need this line.

source/codegen/metadata/nidmm/functions.py
```diff
@@ -26,155 +26,6 @@
         ],
         'returns': 'ViStatus'
     },
-    "CalAdjustGain": {
```
doshirohan, line 29: @jonlugoatx would you prefer removing these metadata or marking them with `codegen: no`? I see we have few APIs marked with `codegen: no` elsewhere.

I guess advantage of retaining metadata would be 
1. Less changes from base metadata. So it would be easier to merge if we ever want to sync from base metadata or regenerate the metadata tile.
2. In case someone wants to use these APIs, it would be easier for them to regenerate. Not sure if we want this.

Removing the metadata on the other hand, would just keep it cleaner.

This PR doesn't have any attributes, but any external calibration related attributes will need to be removed from metadata since we don't have support for `codegen: no` for attributes.

FYI @Akshata440 for [this PR](https://github.com/doshirohan/grpc-device/pull/123).

Akshata440 (approved):

source/codegen/metadata/nidmm/functions.py
jonlugoatx: Removing them is okay. When we take on the work to integrate with hapigen hiding the work to remove vs hide them would be the same.

jonlugoatx (approved):

PR #224, opened 2021-06-17
[Owners] Add support for generate-mappings enums

doshirohan:

For any enums that have non-int32 values for constants, we create maps for these enums in header file of service handler and use mapped integer values in proto files so that clients can pass these values as enum. In this change, we rename parameters with enum type and generate-mappings to `<field-name>_mapped` to make it obvious to client users that these fields and corresponding enum types don't represent actual values.

Typically a function parameter will either use mapped enum or unmapped enum. Special case is for `SetAttribute` and `CheckAttribute` APIs which can have both mapped and unmapped enums if some of the attribute values can be represented in 32 bit and some can't. To accommodate this we tag function parameter metadata with either `enum` or `mapped-enum` during metadata mutation based on if the enum values should be mapped. Rest of the codegen for proto file generation, header file generation and service cpp file generation uses this information to figure out naming of parameters, dictionary lookup etc.

With this change, we can add enums of string and float types for attributes and function parameters that we have left out so far.

- Manually verified generated files for input/output mapped enum parameter by modifying metadata


Akshata440 (approved):

source/codegen/templates/proto_helpers.mako
```diff
@@ -30,16 +30,22 @@ enum ${service_class_prefix}Attributes {
 <%
   enum = enums[enum_name]
   enum_value_prefix = enum.get("enum-value-prefix", common_helpers.pascal_to_snake(enum_name).upper())
+  is_mapped_enum = enum.get("generate-mappings", False)
+  enum_name = f"{enum_name}Mapped" if (is_mapped_enum and not enum_name.endswith("Mapped")) else enum_name
   allow_alias = enum.get("allow-alias", proto_helpers.should_allow_alias(enum))
```
ganeshnie, line 35: I feel the suffix string for mapped name `{enum_name}Mapped`  cold be replaced with `{enum_name}+MAPPED_ENUM_SUFFIX`  with MAPPED_ENUM_SUFFIX = "Mapped".

source/codegen/templates/proto_helpers.mako
```diff
@@ -30,16 +30,22 @@ enum ${service_class_prefix}Attributes {
...
   nonint_index = 1
 %>\
 enum ${enum_name} {
 %   if allow_alias == True:
   option allow_alias = true;
 %   endif
+%   if is_mapped_enum:
+  ${enum_value_prefix}_MAPPED_UNSPECIFIED = 0;
```
ganeshnie, line 43: Is there a reason we added Mapped suffix explicitly for UNSPECIFIED enums?

source/codegen/templates/proto_helpers.mako
```diff
@@ -79,23 +85,38 @@ message ${common_helpers.snake_to_pascal(function)}Request {
 % for parameter in input_parameters:
 <%
   parameter_name = common_helpers.camel_to_snake(parameter["name"])
-  grpc_field_number = proto_helpers.generate_parameter_field_number(parameter, used_indexes)
+  is_array = common_helpers.is_array(parameter["type"])
   if 'grpc_type' in parameter:
     parameter_type = parameter['grpc_type']
   else:
-    parameter_type = proto_helpers.determine_function_parameter_type(parameter, service_class_prefix)
+    parameter_type = proto_helpers.get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
```
ganeshnie, line 92: Is this replacement of function `determine_function_parameter_type` with `get_grpc_type_from_ivi` expected?

ganeshnie (commented):
Could not complete the review. Will complete the review tomorrow.

ganeshnie (approved):

source/codegen/templates/proto_helpers.mako
doshirohan: Yes, I have removed `determine_function_parameter_type` function. That function was used for determining parameter type for enum parameter as well as other parameters. For other parameters, it used to call `get_grpc_type_from_ivi`. That function doesn't make sense anymore, as even with enum we could have two potential types, one for Mapped enum and other for Unampped enum. It is difficult to customize `determine_function_parameter_type` to handle 2 different enum types, so I have inlined type generation for enums.

source/codegen/templates/proto_helpers.mako
doshirohan: Yes, `MAPPED` suffix is added intentionally. If you take case of `NiDmmInt32AttributeValues` and `NiDmmInt32AttributeValuesMapped` enums, both of them will have `NIDMM_INT32_UNSPECIFIED` constant if we don't add `MAPPED` substring. Multiple enum constants with same name is not allowed in proto.

generated/nifake/nifake.proto
```diff
@@ -94,17 +94,17 @@ enum Color {
   COLOR_BLACK = 42;
 }
 
-enum FloatEnum {
-  FLOAT_ENUM_UNSPECIFIED = 0;
+enum FloatEnumMapped {
+  FLOAT_ENUM_MAPPED_UNSPECIFIED = 0;
```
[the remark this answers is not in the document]
doshirohan: Yes, MAPPED substring in UNSPECIFIED constant is intentional because of special case of `*AttributeValues` enums. I have replied to your comment in `proto_helpers.mako` with more details.

generated/nifake/nifake.proto
```diff
@@ -602,7 +611,7 @@ message ReturnMultipleTypesResponse {
   Turtle an_int_enum = 5;
   sint32 an_int_enum_raw = 6;
   double a_float = 7;
-  FloatEnum a_float_enum = 8;
+    FloatEnumMapped a_float_enum_mapped = 8;
```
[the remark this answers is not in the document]
doshirohan: Fixed spacing in codegen.

source/codegen/templates/proto_helpers.mako
doshirohan: We have used `Mapped` keyword in lot of places. I think using formatted string makes it concise and more readable than using a variable.

generated/nifake/nifake_service.h
```diff
@@ -87,6 +87,8 @@ class NiFakeService final : public NiFake::Service {
   void Copy(const std::vector<CustomStruct>& input, google::protobuf::RepeatedPtrField<nifake_grpc::FakeCustomStruct>* output);
   std::map<std::int32_t, float> floatenum_input_map_ { {1, 3.5f},{2, 4.5f},{3, 5.5f},{4, 6.5f},{5, 7.5f}, };
   std::map<float, std::int32_t> floatenum_output_map_ { {3.5f, 1},{4.5f, 2},{5.5f, 3},{6.5f, 4},{7.5f, 5}, };
+  std::map<std::int32_t, float> nifakereal64attributevalues_input_map_ { {1, 3.5f},{2, 4.5f},{3, 5.5f},{4, 6.5f},{5, 7.5f}, };
+  std::map<float, std::int32_t> nifakereal64attributevalues_output_map_ { {3.5f, 1},{4.5f, 2},{5.5f, 3},{6.5f, 4},{7.5f, 5}, };
```
astarche, line 91: `map` lookup on `float` is OK because we assume that they're constants in the driver API? So there's not going to be some roundoff issue?

You could also use `std::find_if` to search with a `tolerance` delegate.

source/codegen/templates/service_helpers.mako
```diff
@@ -282,23 +296,31 @@ ${initialize_standard_input_param(function_name, parameter)}\
...
+    mapped_enum_name = parameter["mapped-enum"]
+    mapped_enum_type = mapped_enum_name[:-len("Mapped")] if mapped_enum_name.endswith("Mapped") else mapped_enum_name
+    map_name = mapped_enum_type.lower() + "_output_map_"
+    mapped_enum_iterator_name = parameter_name + "_omap_it"
 %>\
+%     if has_mapped_enum:
         auto ${mapped_enum_iterator_name} = ${map_name}.find(${parameter_name});
         if(${mapped_enum_iterator_name} != ${map_name}.end()) {
-          response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(${mapped_enum_iterator_name}->second));
+          response->set_${parameter_name}_mapped(static_cast<${namespace_prefix}${mapped_enum_type}Mapped>(${mapped_enum_iterator_name}->second));
```
astarche, line 311: Can this be the type in the map instead of a `static_cast`?

source/codegen/templates/proto_helpers.mako
```diff
@@ -79,23 +85,38 @@ message ${common_helpers.snake_to_pascal(function)}Request {
...
+  if parameter.get('mapped-enum', None):
+    mapped_type = f"repeated {parameter['mapped-enum']}" if is_array else parameter['mapped-enum']
+    mapped_type = mapped_type if mapped_type.endswith('Mapped') else f'{mapped_type}Mapped'
+    grpc_mapped_field_number = proto_helpers.generate_parameter_field_number(parameter, used_indexes, "_mapped")
   grpc_raw_field_number = proto_helpers.generate_parameter_field_number(parameter, used_indexes, "_raw")
 %>\
   oneof ${parameter_name}_enum {
-    ${parameter_type} ${parameter_name} = ${grpc_field_number};
-    ${non_enum_type} ${parameter_name}_raw = ${grpc_raw_field_number};
+%   if enum_parameter_type:
```
astarche, line 108: Is there any way this conditional can collapse to:

`${parameter_type} ${parameter_name} = ${field_number};`

And the conditional code above (or python helper code) figures out the correct values for each of those?

source/codegen/templates/proto_helpers.mako
```diff
@@ -79,23 +85,38 @@ message ${common_helpers.snake_to_pascal(function)}Request {
...
-    parameter_type = proto_helpers.determine_function_parameter_type(parameter, service_class_prefix)
+    parameter_type = proto_helpers.get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
 %>\
 %   if common_helpers.is_enum(parameter):
 <%
-  is_array = common_helpers.is_array(parameter["type"])
-  non_enum_type = proto_helpers.get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
+  enum_parameter_type = None
+  mapped_type = None
+  if parameter.get('enum', None):
```
astarche, line 98: These conditionals are really exploding and have some duplicated code.

What about a python helper that returns a list of (type, name, number) as namedtuples? And then loop over them in the mako?

source/codegen/templates/proto_helpers.mako
```diff
@@ -114,21 +135,36 @@ message ${common_helpers.snake_to_pascal(function)}Response {
...
     parameter_type = parameter["grpc_type"]
   else:
-    parameter_type = proto_helpers.determine_function_parameter_type(parameter, service_class_prefix)
+    parameter_type = proto_helpers.get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
 %>\
 %   if common_helpers.is_enum(parameter):
 <%
-  is_array = common_helpers.is_array(parameter["type"])
-  non_enum_type = proto_helpers.get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
+  enum_parameter_type = None
```
astarche, line 146: This really should be shared somehow with the input case. Or at least have a bunch of shared helpers.

source/codegen/metadata_mutation.py
```diff
@@ -51,30 +51,32 @@ def get_size_param(param, parameters):
     named_params = { p['name'] : p for p in parameters }
     return named_params.get(param['size']['value'], None)
 
+def mark_mapped_enum_params(parameters, enums):
+    for param in (p for p in parameters if 'enum' in p):
+        enum_name = param['enum']
+        enum = enums[enum_name]
+        if 'generate-mappings' in enum:
+            del param['enum']
+            param['mapped-enum'] = enum_name
```
astarche, line 60: Some of the downstream logic looks better to me if `mapped-enum` is a `bool` and `enum` is left alone. For example, there are a few blocks that do the same thing to `mapped-enum` or `enum` with a little extra for `mapped-enum`. 

Why is it better to swap the fields? Is this to support the case where both exist? Is that implemented with special metadata?

I think you cover some of this in the PR description. But this is very non-obvious. More of these conditionals/methods need comments on why they exist.

source/codegen/templates/proto_helpers.mako
```diff
@@ -30,16 +30,22 @@ enum ${service_class_prefix}Attributes {
 <%
   enum = enums[enum_name]
   enum_value_prefix = enum.get("enum-value-prefix", common_helpers.pascal_to_snake(enum_name).upper())
+  is_mapped_enum = enum.get("generate-mappings", False)
+  enum_name = f"{enum_name}Mapped" if (is_mapped_enum and not enum_name.endswith("Mapped")) else enum_name
```
astarche, line 34: Why is this conditional needed? Can you comment it? And maybe the corresponding logical case in metadata mutations? The `generate-mappings` vs `{enum_name}Mapped` is already a little tricky to follow.

source/codegen/templates/service.h.mako
```diff
@@ -66,9 +67,10 @@ private:
 % for enum in enums_to_map:
 <%
   enum_value = service_helpers.python_to_c(enums[enum])
+  map_prefix = enum[:-len("Mapped")] if enum.endswith("Mapped") else enum
```
astarche, line 70: This condition is everywhere!!

1. Do we really have to check ends with `Mapped`? (can we just use the name as-is?)
2. Do we really have to strip `Mapped` off the name?
3. Can we put code with logic based on `Mapped` in a python helpers file so that it's shared and close together?

source/codegen/templates/service_helpers.mako
```diff
@@ -282,23 +296,31 @@ ${initialize_standard_input_param(function_name, parameter)}\
   parameter_name = common_helpers.camel_to_snake(parameter['cppName'])
 %>\
 %   if common_helpers.is_enum(parameter) == True:
-%     if enums[parameter["enum"]].get("generate-mappings", False):
 <%
-  map_name = parameter["enum"].lower() + "_output_map_"
-  mapped_enum_iterator_name = parameter_name + "_omap_it"
+  has_mapped_enum = 'mapped-enum' in parameter
+  has_unmapped_enum = 'enum' in parameter
+  if has_mapped_enum:
```
astarche, line 302: Consider: there's less duplication (and waaay less conditional logic) in the server than in the proto. But shared helpers could still make sense for input vs. output.

source/codegen/metadata_mutation.py
```diff
@@ -85,12 +87,30 @@ def mark_attr_value_param_if_required(function, enums, attribute_enums_by_type,
         enum_type = attribute_value_param["type"]
     if attribute_value_param != None and enum_type in attribute_enums_by_type:
         enum_name = get_attribute_values_enum_name(service_class_prefix, enum_type)
-        if(enum_name in enums):
-            attribute_value_param['enum'] = enum_name
+        if enum_name in enums or f'{enum_name}Mapped' in enums:
```
astarche, line 90: I was going to say that this is a duplicate nested if. But is it? Can you add both mapped and unmapped in the same pass here?

astarche: Is that covered in existing test cases?

astarche (changes requested):
I'd like to see the responses to my comments.

jonlugoatx (commented):
@astarche, here's some background on the problem we were trying to solve with these attribute enums
1. Since gRPC doesn't support constants and only supports integer enums, we need to map string enums to ints and vice versa. Double and int64 values are in an interesting in-between where in most cases the constants fit nicely in integers, but sometimes they don't. For an enum with values `{ -1, 0, 1 }` it would be a shame to define maps like this: `{ 0: -1, 1: 0, 2: 1}`. For those, it would be better to just cast the raw enum value into a `double` and vice versa.
1. Whether the enum is mapped or not mapped is determined by the `generate-mappings: true` in the enum metadata. We would always put that attribute on string enums and selectively tag `double` and `int64` enums with it where appropriate.
1. For Get/SetAttribute functions, we want to combine all possible enum values into a single enum for that type. This allows us to provide good intellisense to the user without needing to define separate Get/SetAttribute methods for each enum or add separate parameters for each enum in a big `oneof` block.
1. It can be problematic if when we merge all `ViReal64` enum values referenced by `Get/SetViReal64Attribute` together if some of the enums have `generate-mappings` and some of them don't. Our solution was to merge like types into separate enums keeping mapped enums with mapped enums and unmapped enums with unmapped enums. That means defining `SetViReal64Attribute` with 3 parameters wrapped in a `oneof`: raw, enum, & mapped enum.

source/codegen/templates/proto_helpers.mako
jonlugoatx: I agree, we need to find a way of factoring this out.

source/codegen/templates/service.h.mako
jonlugoatx: I think the problem is that we're adding that suffix to anything with `generate-mappings`. The code generator shouldn't be doing that.

source/codegen/templates/service_helpers.mako
```diff
@@ -132,31 +132,45 @@ ${initialize_standard_input_param(function_name, parameter)}\
   parameter_name = common_helpers.camel_to_snake(parameter['cppName'])
   field_name = common_helpers.camel_to_snake(parameter["name"])
   pascal_field_name = common_helpers.snake_to_pascal(field_name)
-  map_name = parameter["enum"].lower() + "_input_map_"
-  mapped_enum_iterator_name = field_name + "_imap_it"
   one_of_case_prefix = f'{namespace_prefix}{function_name}Request::{pascal_field_name}EnumCase'
-  enum_request_snippet = f'request->{field_name}()'
+  has_mapped_enum = 'mapped-enum' in parameter
```
jonlugoatx, line 136: I reached review fatigue here. I'll try to pick up from here later.

jonlugoatx (changes requested):

source/codegen/metadata_mutation.py
```diff
@@ -85,12 +87,30 @@ def mark_attr_value_param_if_required(function, enums, attribute_enums_by_type,
...
+            if enum_name in enums:
+                attribute_value_param['enum'] = enum_name
+            if f'{enum_name}Mapped' in enums:
+                attribute_value_param['mapped-enum'] = f'{enum_name}Mapped'
         else:
             # Ideally there must be an enum associated with this parameter for this SetAttribute* API.
             # Since the enum is empty, users will be passing in raw values. Make it clear via naming.
             attribute_value_param['name'] = attribute_value_param['name'] + "_raw"
```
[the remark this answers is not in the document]
doshirohan: This place adds `_raw` suffix to "value" params of `SetAttribute` APIs where there is no associated enum. As the comment above mentions, it keeps all `SetAttribute` APIs consistent even if some of them don't have enum for the `value` parameter.  For e.g. [this API](https://github.com/ni/grpc-device/blob/main/generated/nidcpower/nidcpower.proto#L1844) in dcpower

If we don't rename the parameter here, since there is no enum associated with it, codegen will generate `request` message with `value` as parameter name - and this parameter actually accepts raw value unlike similar APIs from other drivers where `value` parameter accepts enum. This would be confusing for users.

generated/nifake/nifake_service.h
doshirohan: Actually, we do have few float values in DMM we didn't add because it was not supported yet.

> #define NIDMM_VAL_1_MICROAMP 0.000001
#define NIDMM_VAL_10_MICROAMP 0.00001
#define NIDMM_VAL_100_MICROAMP 0.0001
#define NIDMM_VAL_1_MILLIAMP 0.001

Should we add `tolerance` support for these roundoff issues? How would that look like? I see quite a few different ways of implementing [tolerance](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/).

source/codegen/templates/proto_helpers.mako
```diff
@@ -30,16 +30,22 @@ enum ${service_class_prefix}Attributes {
...
   allow_alias = enum.get("allow-alias", proto_helpers.should_allow_alias(enum))
   nonint_index = 1
 %>\
 enum ${enum_name} {
 %   if allow_alias == True:
   option allow_alias = true;
 %   endif
+%   if is_mapped_enum:
```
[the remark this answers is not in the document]
doshirohan: Are you talking about not adding `_MAPPED_` for `UNSPECIFIED` value of non-SetAttribute API enums? That would be difficult to figure out here, since we just have list of `enums` here and don't know which APIs use these enums. We can figure out that information by iterating over all functions, but that will just make this codegen more unreadable.

source/codegen/metadata_mutation.py
```diff
@@ -51,30 +51,32 @@ def get_size_param(param, parameters):
...
-            'enum-value-prefix': enum_value_prefix,
-            'allow-alias': allow_alias,
-            'values': values
-        }
         type_enum_name = get_attribute_values_enum_name(service_class_prefix, type_name)
-        enums.update({type_enum_name: new_enum})
+        add_enum(enums, enum_value_prefix, type_enum_name, umapped_values_to_create, is_mapped=False)
+        add_enum(enums, enum_value_prefix, type_enum_name, mapped_values_to_create, is_mapped=True)
     
 def mark_attr_value_param_if_required(function, enums, attribute_enums_by_type, service_class_prefix):
```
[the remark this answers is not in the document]
doshirohan: Yeah, these named helper functions might help. I will try to refactor to improve readability.

source/codegen/templates/service_helpers.mako
doshirohan: We probably could add it to map. We might need to write some mutation or a lookup helper so that this type information is available where we declare maps. Let me see what it takes to get the value there.

Although, even though map's type would have an enum name, we might still have to put integer constants there instead of using enum constants names for better readability of map. Otherwise the map declaration will have to be multi-line with (enum_name, raw_value) pair on each line.

astarche: That's fine. I'd like less `cast`s overall. But if it's not an easy fix here, you can leave it as is.

source/codegen/metadata_mutation.py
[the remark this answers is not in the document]
jonlugoatx: How is that confusing to customers? If there's a field that looks like this, it's clear that it's raw and how to use it.
```
double value = 5;
```

source/codegen/templates/proto_helpers.mako
jonlugoatx: I don't want the code generator to modify the names of enums. If the name of the enum needs to have "mapped" in it, we should do that in the `enums.py`.

source/codegen/templates/proto_helpers.mako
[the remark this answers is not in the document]
jonlugoatx: Since I'm saying I don't want "mapped" to be automatically added to enum names, I wouldn't want `_MAPPED_` automatically added to the value names.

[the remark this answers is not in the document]
doshirohan: Even if we don't add "Mapped" to other enums from `enums.py` as per your other comment, `_MAPPED_` needs to be added here for `SetAttribute` APIs, since they might have both mapped and unmapped enums. And both of them will end up having `NIDMM_INT32_UNSPECIFIED` if I don't add `_MAPPED_` explicitly. And even to do this specifically for `SetAttribute` APIs, I will need to know if the enum corresponds to `SetAttribute` API which will further complicate the logic here (I can probably just do `enum_name.endswith("AttributeValues")`).

Other option would be adding `_MAPPED_` to `enum_value_prefix` attribute on consolidated mapped enum when we generate it during mutation. But that means all other values would get `_MAPPED_` prefix too.

source/codegen/metadata_mutation.py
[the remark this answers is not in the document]
doshirohan: Yes, it is clear when you look at proto file. But when you are using it as API, it would be confusing if you have to use `.value` for setting raw value in SetAttributeViInt64Request  of [nidigitalpattern](https://github.com/ni/grpc-device/blob/main/generated/nidigitalpattern/nidigitalpattern.proto#L1687), but have to use `.value_raw` for setting raw value in same API of [niscope](https://github.com/ni/grpc-device/blob/main/generated/niscope/niscope.proto#L1821). The linked files are with this change where I rename `value` to `value_raw`, but imagine the digital pattern file with param name as `.value` above.

source/codegen/templates/proto_helpers.mako
doshirohan: Problem is, if we are asking developers to add "Mapped" suffix in `enums.py` some of our consolidated enums might have "Mapped" keyword twice in the constant name. 

If equivalent of below from enums.py is enum associated with attribute...
```
enum OsNames {
  VAL_ANDROID: "ANDROID"
  VAL_IOS: "IOS"
}
```

...consolidated string enum for above in `.proto` file would currently look like
```
enum NiDmmStringAttributeValuesMapped {
    NIDMM_MAPPED_UNSPECIFIED: 0,
    NIDMM_OS_NAMES_ANDROID: 1,
    NIDMM_OS_NAMES_IOS: 2
}
```
But if developers add `Mapped` to `OsNames` enum, it would look like...
```
enum NiDmmStringAttributeValuesMapped {
    NIDMM_MAPPED_UNSPECIFIED: 0,
    NIDMM_OS_NAMES_MAPPED_ANDROID: 1,
    NIDMM_OS_NAMES_MAPPED_IOS: 2
}
```
or worse, based on discussion about how to add `_MAPPED_` for `_UNSPECIFIED` value [here](https://github.com/ni/grpc-device/pull/224/files/5f3df308d10ecd15d240eafe7cd01ef3cc9e8fee#r655863859)
```
enum NiDmmStringAttributeValuesMapped {
    NIDMM_MAPPED_UNSPECIFIED: 0,
    NIDMM_MAPPED_OS_NAMES_MAPPED_ANDROID: 1,
    NIDMM_MAPPED_OS_NAMES_MAPPED_IOS: 2
}
```
Adding `Mapped` suffix to all enums with `generate-mapping` tag during mutation/codegen has advantage that developers just have to tag enum with above attribute to say that mappings should be generated and gives logical name (that comes from base metadata for MI enums) for the enum and our code generator will consistently add same suffix to all such enums (even for mapped enums in DAQmx and so on).

source/codegen/templates/service.h.mako
doshirohan: > When you say "add the suffix to all enum names" you are referring to the automatically-generated merged attribute enums, right? 

I was actually talking about all enums tagged with `generate-mappings`. But based on our discussion on other thread about whether or not we should add `Mapped` suffix automatically, I can do it just for automatically-generated merged attribute values enums only.

generated/nifake/nifake_service.h
jonlugoatx: Just for the record, we should not be fixing the tolerance as part of this PR, which is already quite large. I do agree that with a value like `0.000001`, we will need to address it though.

source/codegen/templates/proto_helpers.mako
jonlugoatx: I still don't see much benefit in putting "Mapped" at the end of the type name unless we're talking about the attribute value enums.

If a normal function refers to a mapped enum, we should see this:
```
oneof {
  double foo_raw = 2,
  MyEnum foo_enum_mapped = 3
}
```
When we merge enums for attributes, we'll end up with this:
_(note: I'm not trying to be precise with names here, I just don't have the real naming scheme in front of me while writing the example)_
```
enum NiDmmReal64AttributeValues {
  ...
}
enum NiDmmReal64AttributeMappedValues {
  ...
}
message NiDmmSetViReal64AttributeRequest {
  ...
  oneof {
    double value_raw = 5,
    NiDmmReal64AttributeValues value_enum = 6,
    NiDmmReal64AttributeMappedValues value_enum_mapped = 7
  }
}
```

reckenro (changes requested):
I'm a little late to this review but agree with the suggestions laid out in the comments. Like the others I want to take another pass after the comments are addressed to see how everything looks.

source/codegen/metadata_mutation.py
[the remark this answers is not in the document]
doshirohan: I have added helper for `get_attribute_function_value_param`, but other one helper you suggested is probably not possible if we want to keep renaming of `value` parameter to `value_raw` when there are no enum values which is being discussed in other comment thread here.

source/codegen/templates/service_helpers.mako
doshirohan: Turns out getting enum type in the map was quite easy, but following code doesn't compile since integer values in the map don't match with enum type specified in the map type. Looks like we can't make that change unless we add enum values in the map itself, which will make it unreadable.

>   std::map<nidevice_grpc::MobileOSNames, std::string> mobileosnames_input_map_ { {1, "Android"},{2, "iOS"},{3, "None"}, };

I am leaving the casts as is.

astarche (approved):

reckenro (approved):

source/codegen/metadata_mutation.py
```diff
@@ -51,46 +51,71 @@ def get_size_param(param, parameters):
...
+        param_type = value_param["type"]
+    if value_param != None and param_type in attribute_enums_by_type:
+        enum_name = get_attribute_values_enum_name(service_class_prefix, param_type)
+        mapped_enum_name = get_attribute_values_enum_name(service_class_prefix, param_type, is_mapped=True)
+        if enum_name in enums:
+            value_param['enum'] = enum_name
+        if mapped_enum_name in enums:
+            value_param['mapped-enum'] = mapped_enum_name
+        has_enum = enum_name in enums or mapped_enum_name in enums
+        if not has_enum:
```
jonlugoatx, line 97:
```suggestion
        enum_exists = enum_name in enums
        mapped_enum_exists = mapped_enum_name in enums
        if enum_exists:
            value_param['enum'] = enum_name
        if mapped_enum_exists:
            value_param['mapped-enum'] = mapped_enum_name
        if not enum_exists and not mapped_enum_exists:
```

generated/nifake/nifake.proto
```diff
@@ -129,6 +129,15 @@ enum NiFakeInt32AttributeValues {
   NIFAKE_INT32_BLACK = 42;
 }
 
+enum NiFakeReal64AttributeValuesMapped {
```
jonlugoatx, line 132: It would be nice to add some nifake metadata that exercises the various scenarios this new mapping logic is supposed to support. I don't think that needs to hold up this PR though.

jonlugoatx (approved):

generated/nifake/nifake.proto
doshirohan: Makes sense. I have created [tech debt](https://dev.azure.com/ni/DevCentral/_workitems/edit/1517663) to add nifake metadata for this.

generated/nifake/nifake_service.h
doshirohan: I have created [tech debt](https://dev.azure.com/ni/DevCentral/_workitems/edit/1517667/) to address float lookup scenarios. These affects only returning of enum value for floats and not returning raw value or getting enum/raw value from request objects.

PR #226, opened 2021-06-18
Trigger AzDO CI testing after build

zhindes:

This PR causes the build to trigger AzDO pipelines for system testing. Those AzDO pipelines are still in work. For this to work, we need to add a Secret to the GitHub project (see https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/github-actions?view=azure-devops) named `NI_AZURE_DEVOPS_TOKEN` with an AzDO PAT. I can provide one.

To enable system testing!

Here is a successful desktop run: https://github.com/zhindes/grpc-device/actions/runs/947869919 that queued an AzDO CI.
Here is a successful RT run: https://github.com/zhindes/grpc-device/actions/runs/950454402 that queued an AzDO CI.


yixinguo2 (approved):
We will need to add test runners to the desktop build artifacts but we can do that separately.

zhindes:
> 
> 
> We will need to add test runners to the desktop build artifacts but we can do that separately.

Yep, testing that now. I'll make a new PR this afternoon 🤞

.github/workflows/build_cmake.yml
```diff
@@ -139,3 +139,16 @@ jobs:
           LICENSE
           ThirdPartyNotices.txt
         retention-days: 5
+
+  trigger_ci:
+    name: Trigger CI
+    needs: build
+    runs-on: windows-latest
+    if: ${{ github.ref == 'refs/heads/main' }}
```
SparkingSpork, line 147: Should this be run on release branches (such as [here](https://github.com/ni/grpc-device/blob/5a0b132ffc41e51a3ff2f926d7f1fbb6ed75b36d/.github/workflows/build_cmake.yml#L133)) as well? Same for below.

SparkingSpork (approved):

.github/workflows/build_cmake.yml
zhindes: @yixinguo2 - we only wanted main, right? Why is that...?

yixinguo2: I think we would like to run it on release branches but we would need to trigger a different pipeline because we can't pass variables, yet.

yixinguo2: Maybe a releases branch only change to the yml

zhindes: Ah, right, we need to know where to get artifacts from. If we don't have plans to create CI for releases (assuming that everything goes through main), then that seems OK for now...?

yixinguo2: I think we would like to trigger the CI for releases/1.1. maybe only make a change to that branch?

yixinguo2: Sorry I just realized we could update the condition if it is possible to conditionally set the pipeline name. Or we could duplicate the job with a startswith("refs/heads/releases") condition

SparkingSpork: I don't think this is worth cherry-picking to older release branches and I agree that having it run on main only is probably OK for now. It might just be a "nice-to-have" in the future.

yixinguo2: Ok. Since I'm already working on downloading and exporting release artifacts, let's just check in the yml change into main so when the next release branch is created, it will start kicking off CI builds.

zhindes: @yixinguo2 - added both release/main in the actions. Can you double-check it looks right? main is [running on my fork](https://github.com/zhindes/grpc-device/actions) and should trigger the two main pipelines in a bit.

PR #227, opened 2021-06-18
Add SystemTestsRunner to artifacts

zhindes:

This refactors the artifacts for all platforms that each has 2: one for the server and one for the tests.

This will allow the CI testing (that is being triggered in https://github.com/ni/grpc-device/pull/226) to run the System Tests.

Desktop run here: https://github.com/zhindes/grpc-device/actions/runs/958271879.
Linux RT run here: https://github.com/zhindes/grpc-device/actions/runs/958272207.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -116,7 +116,6 @@ jobs:
         -C ${GITHUB_WORKSPACE}/build
         ni_grpc_device_server
         server_config.json
-        SystemTestsRunner
         -C ${GITHUB_WORKSPACE}
         LICENSE
         ThirdPartyNotices.txt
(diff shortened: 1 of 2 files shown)


yixinguo2 (approved):

.github/workflows/build_cmake.yml
```diff
@@ -116,6 +116,7 @@ jobs:
         -C ${GITHUB_WORKSPACE}/build
         ni_grpc_device_server
         server_config.json
+        SystemTestsRunner
```
SparkingSpork, line 119: @kreimnitz and @jonlugoatx - If I recall correctly, we specifically decided to exclude these items from artifacts for main and release as it made it simpler to upload our "releases" since no manual changes were needed from the artifact to the shipping compressed files. 
- Do you all have any concerns with changing that decision here so that NI-Sync can run Systems tests as part of their pipeline? 

Personally, I'm OK with this but a simple alternative would be adding steps to upload the test runners as a separate artifact. It might be better to upload all of the test runners and support files as a separate artifact since, as I recall, those test runners have no dependencies on the actual server binaries.

kreimnitz: I'm fine with either including these in the existing artifact or making a separate one.  There is already some small manual file updates with the releases, so this doesn't significantly impact that.

SparkingSpork (approved):

.github/workflows/build_cmake.yml
SparkingSpork: If Kevin is fine with it, I'm fine with it.

jonlugoatx: I maintain that we should be able to take the `.tar.gz` from the artifact and publish it in a release without needing to unpack/repack on a Linux machine. That's a manual step that could introduce human error into the release process, which is already manual enough.

@SparkingSpork, is there a way for this workflow to produce two artifacts? One with and one without tests?

jonlugoatx (changes requested):

.github/workflows/build_cmake.yml
SparkingSpork: Yes. You would need to:
- Create the new `.tar.gz` either by adding a new tar command to the current step or creating a new step
- Add a new `upload-artifact` step for the new item

astarche (approved):

jonlugoatx (approved):

PR #230, opened 2021-06-23
[Owners Review] Display Port Number in addition to server_address in exception handling

ni-pkhare:

port number is also displayed in addition to the server_address as part of the message in the case when client fails to connect to the server.

If the client fails to connect to the server when either server is not running or client puts wrong server_address or port_number, the client gets a proper message indicating server_adress and port_number both.

For eg.
image

Manual

### DIFF
--- a/examples/nidcpower/measure-record.py
+++ b/examples/nidcpower/measure-record.py
@@ -196,7 +196,7 @@ try :
 except grpc.RpcError as rpc_error:
     error_message = rpc_error.details()
     if rpc_error.code() == grpc.StatusCode.UNAVAILABLE :
-        error_message = f"Failed to connect to server on {server_address}"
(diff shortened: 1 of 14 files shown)


ni-pkhare:
@jonlugoatx @kreimnitz As per discussion with PO, for the case when user inputs wrong port_number, it would be helpful if we display the message like "Failed to connect to the server on <server_address>:<server_port>" instead of "Failed to connect to the server on <server_address>". This was proposed in some previous example but was removed. Is there any specific reason as to why port_number should not be displayed in the message?

kruthi-NI (approved):

doshirohan (approved):

harsha-bhushan (approved):

jonlugoatx (approved):

kylesull30 (approved):

kreimnitz (approved):

PR #233, opened 2021-06-23
Add basic feature toggles to server configuration

astarche:

Add support for feature toggles to the server configuration json file format so that features can be explicitly enabled and disabled.

DAQmx feature development will not be ready for release until significant work is completed and multiple internal features are complete.

Adding the concept of feature toggles allows us to use an NI-standard mainline model for development.

Ran and passed new tests.
Manual testing with a server with valid and invalid configuration.

### DIFF
--- a/source/server/feature_toggles.cpp
+++ b/source/server/feature_toggles.cpp
@@ -1,12 +1,14 @@
 #include "feature_toggles.h"
 
 namespace nidevice_grpc {
-bool FeatureToggles::is_enabled(const std::string& feature_name) const
+FeatureToggles::FeatureState FeatureToggles::get_feature_state(const std::string& feature_name) const
 {
   auto found = map_.find(feature_name);
   if (found != map_.end()) {
-    return found->second;
+    return found->second 
+    ? FeatureState::kEnabled 
+    : FeatureState::kDisabled;
   }
-  return false;
+  return FeatureState::kUnspecified;
 }
 }  // namespace nidevice_grpc
--- a/source/server/feature_toggles.cpp
+++ b/source/server/feature_toggles.cpp
@@ -1,0 +1,12 @@
+#include "feature_toggles.h"
+
+namespace nidevice_grpc {
+bool FeatureToggles::is_enabled(const std::string& feature_name) const
+{
+  auto found = map_.find(feature_name);
+  if (found != map_.end()) {
+    return found->second;
+  }
+  return false;
+}
+}  // namespace nidevice_grpc
--- a/source/server/feature_toggles.h
+++ b/source/server/feature_toggles.h
@@ -1,0 +1,17 @@
+#pragma once
+
+#include <string>
+#include <unordered_map>
+namespace nidevice_grpc {
+
+using FeatureToggleConfigurationMap = std::unordered_map<std::string, bool>;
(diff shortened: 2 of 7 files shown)


source/server/feature_toggles.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <string>
+#include <unordered_map>
+namespace nidevice_grpc {
+
+using FeatureToggleConfigurationMap = std::unordered_map<std::string, bool>;
+class FeatureToggles {
+ public:
+  FeatureToggles() {}
+  FeatureToggles(FeatureToggleConfigurationMap&& map) : map_(map) {}
```
reckenro, line 11: I don't know if I've used this before. Is the rvlaue reference here an optimization to avoid unneccessary copies?

astarche: Yeah. It goes with the `std::move` in the parser and then passes through to the move constructor of `std::map`. Basically the logic is that `FeatureToggles` will always "own" the map even though it doesn't construct it. I think there's a pretty decent chance that it would get optimized to the same thing if it was passed by value.

source/tests/unit/server_configuration_parser_tests.cpp
```diff
@@ -357,6 +357,114 @@ TEST(ServerConfigurationParserTests, JsonConfigWithMissingRootCertFile_ParseRoot
     EXPECT_THAT(ex.what(), testing::HasSubstr(nidevice_grpc::kFileNotFoundMessage));
   }
 }
+TEST(ServerConfigurationParserTests, JsonConfigWithEnabledFeature_ParseFeatureToggles_FeatureIsEnabled)
```
reckenro, line 360: [Minor] Can you add a blank line between previous test and this new one?

source/server/feature_toggles.h
reckenro: Cool, I wanted to make sure I was understanding it right. That sounds good to me, matches the standard use case I found mentioned on stackoverflow and such.

reckenro (approved):

SparkingSpork (approved):

source/server/feature_toggles.cpp
```diff
@@ -0,0 +1,12 @@
+#include "feature_toggles.h"
+
+namespace nidevice_grpc {
+bool FeatureToggles::is_enabled(const std::string& feature_name) const
+{
+  auto found = map_.find(feature_name);
+  if (found != map_.end()) {
+    return found->second;
+  }
+  return false;
```
jonlugoatx, line 10: Returning a simple bool makes it difficult to distinguish from a feature that is explicitly toggled to false and one that isn't  specified. In the latter case, you'd want in-progress features to default to false and released features to true.

jonlugoatx (approved):

source/server/feature_toggles.cpp
astarche: I was thinking that adding the concept of declaring features and readiness would be a future enhancement. And then this would still boil down to a boolean that takes all that into account. But, yeah, in the current setup a tri-state is better. So I'll go with that. I hope that the readiness thing goes in, but it's not a current priority. So better be ready for it to be stuck as is.

PR #234, opened 2021-06-24
[Owners] [Tiny] Updated fgen system tests to use device_server

doshirohan:

Updated system tests for ni-fgen to use device_server for session management.

Session management is handled by common device_server framework and also makes fgen system tests consistent with other driver system tests.

- Executed fgen system tests manually

### DIFF
--- a/source/tests/system/nifgen_session_tests.cpp
+++ b/source/tests/system/nifgen_session_tests.cpp
@@ -1,6 +1,6 @@
 #include <gtest/gtest.h>
 
-#include "nifgen/nifgen_library.h"
+#include "device_server.h"
 #include "nifgen/nifgen_service.h"
 
 namespace ni {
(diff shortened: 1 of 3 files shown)


source/tests/system/nifgen_session_tests.cpp
```diff
@@ -21,15 +21,10 @@ const char* kTestInvalidFgenRsrc = "";
 class NiFgenSessionTest : public ::testing::Test {
  protected:
   NiFgenSessionTest()
+    : device_server_(DeviceServerInterface::Singleton()),
```
kruthi-NI, line 24: @doshirohan , we need to update fgen driver api tests also know?

doshirohan: Somehow I couldn't find then in VS, I thought they didn't exist. I will update them in a bit.

doshirohan: Updated api tests as well.

kruthi-NI (approved):

ganeshnie (approved):

reckenro (approved):

jonlugoatx (approved):

PR #242, opened 2021-06-29
[Owners Review] Making both betterproto examples consistent with other examples

ni-pkhare:
### Justification

This PR makes both the betterproto examples consistent with other examples by adding Exception handling and finally block to close the session.

### Implementation

The cases explicitly handled are:
1. Displaying proper message when server is not running and client runs the example.
2. Displaying appropriate message when client calls an API which is not yet supported by the server.

Adding finally block was necessary because without it the example stops whenever any API gives error or any sort of exception occurs and the Close API is not called at all. This is resolved by adding Close API inside finally block. 

### Testing
Manual

### DIFF
--- a/examples/niscope/graph-measurement-betterproto.py
+++ b/examples/niscope/graph-measurement-betterproto.py
@@ -69,10 +69,13 @@ def on_close(event):
 # Error Reporting
 async def CheckStatus(scope_service, vi, result):
     if (result.status != 0):
-        error_result2 = await scope_service.get_error(vi = vi)
-        raise Exception (error_result2.description)
+        error_result = await scope_service.get_error_message(vi = vi, error_code = result.status)
+        raise Exception (error_result.error_message)
+
 # Entry Points
+channel = ()
 async def OpenGrpcScope(resource_name: str, server_address: str, server_port):
+    global channel
     channel = Channel(host=server_address, port=server_port)
     scope_service = niscope_grpc.NiScopeStub(channel)
 
@@ -90,7 +93,8 @@ async def OpenGrpcScope(resource_name: str, server_address: str, server_port):
 
 
 async def CloseGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, vi):
-    await scope_service.close(vi=vi)
+    if vi.id != 0:
+        await scope_service.close(vi=vi)
     channel.close()
     return 0
 
@@ -194,10 +198,16 @@ async def main():
         await ConfigureGrpcScope(*grpc_scope)
         await MeasureGrpcScope(*grpc_scope)
     except Exception as e:
-        print(e)
+        error_message = e.args[1]
+        if e.args[0] == 22:
+            error_message = f"Failed to connect to server on {server_address}:{server_port}"
+        elif e.status.name == "UNIMPLEMENTED":
+            error_message = "The operation is not implemented or is not supported/enabled in this service"
+        print(error_message)
     finally:
         if grpc_scope:
             await CloseGrpcScope(*grpc_scope)
+        channel.close()
 
 ## Run the main
 loop = asyncio.get_event_loop()
--- a/examples/niscope/graph-measurement-betterproto.py
+++ b/examples/niscope/graph-measurement-betterproto.py
@@ -60,13 +60,17 @@ if len(sys.argv) == 4:
 #set to false to disable graph and increase speed
 show_plot = True 
 
+# Handle closing of plot window.
+closed = False
+def on_close(event):
+    global closed
+    closed = True
+
 # Error Reporting
 async def CheckStatus(scope_service, vi, result):
     if (result.status != 0):
-        error_result = await scope_service.get_error_message(vi = vi, error_code = result.status)
-        print(error_result.error_message)
-        exit()
-
+        error_result2 = await scope_service.get_error(vi = vi)
+        raise Exception (error_result2.description)
 # Entry Points
 async def OpenGrpcScope(resource_name: str, server_address: str, server_port):
     channel = Channel(host=server_address, port=server_port)
@@ -119,7 +123,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
     set_trigger_result = await scope_service.set_attribute_vi_int32(
         vi = vi,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_TRIGGER_TYPE,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
+        value_raw = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
     )
     await CheckStatus(scope_service, vi, set_trigger_result)
 
@@ -137,7 +141,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
         vi = vi,
         channel_list = channels,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_MEAS_REF_LEVEL_UNITS,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_REF_LEVEL_UNITS_VAL_PERCENTAGE
+        value_raw = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_REF_LEVEL_UNITS_VAL_PERCENTAGE
     )
     await CheckStatus(scope_service, vi, set_units_result)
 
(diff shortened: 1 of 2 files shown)


examples/niscope/graph-measurement-betterproto.py
```diff
@@ -60,13 +60,17 @@
...
+    closed = True
+
 # Error Reporting
 async def CheckStatus(scope_service, vi, result):
     if (result.status != 0):
-        error_result = await scope_service.get_error_message(vi = vi, error_code = result.status)
-        print(error_result.error_message)
-        exit()
-
+        error_result2 = await scope_service.get_error(vi = vi)
```
kruthi-NI, line 72: @ni-pkhare , why get error is being used instead of get error message which was there earlier?

ni-pkhare: GetError returns additional information about the error which is much more descriptive. Thats why I used it.

doshirohan: Rename from `error_result2` to `error_result`

examples/niscope/graph-measurement-betterproto.py
```diff
@@ -119,7 +123,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
     set_trigger_result = await scope_service.set_attribute_vi_int32(
         vi = vi,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_TRIGGER_TYPE,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
+        value_raw = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
```
doshirohan, line 126: @harsha-bhushan  we have to set `value_raw` instead of `value` for any of the `oneof` fields in `.proto` file because of following bug in `betterproto`
> Note: The Better Protobuf compiler has a bug generating helpers for gRPC messages with oneof types. The problem is that zero-value messages in a oneof group take priority based on order and overwrite other set values. Therefore, only the last field, the field that has '_raw` appended to the name, can be properly set without wrapper modification.

We have already documented this in our [Creating-a-gRPC-Client](https://github.com/ni/grpc-device/wiki/) wiki. Should we add a comment here to similar effect so that users referring to betterproto example are made aware of it.

examples/niscope/graph-measurement-betterproto.py
kruthi-NI: But get error does not seem to take error code as input.  I am guessing it might return most recent error which happened. This might be dangerous in multi client case. Can you confirm that behaviour once again? And if that's the case, please change it any other error api which takes error code also as input

examples/niscope/graph-measurement-betterproto.py
doshirohan: Also, do you think this bug should be cherry-picked to releases\1.1 branch? Should we hold off release for this?

examples/niscope/graph-measurement-betterproto.py
```diff
@@ -181,12 +186,18 @@ async def MeasureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, vi)
...
-    grpc_scope = await OpenGrpcScope(resource, server_address, server_port)
-    await ConfigureGrpcScope(*grpc_scope)
-    await MeasureGrpcScope(*grpc_scope)
-    await CloseGrpcScope(*grpc_scope)
+    try:
+        global grpc_scope
+        grpc_scope = await OpenGrpcScope(resource, server_address, server_port)
+        await ConfigureGrpcScope(*grpc_scope)
+        await MeasureGrpcScope(*grpc_scope)
+    except Exception as e:
```
kruthi-NI, line 196: @ni-pkhare , can exception handling be similar to other examples?

doshirohan (approved):
Diff in `plot-read-waveform-betterproto.py` is confusing. Add a comment on that file that there is no functionality change except addition of try-except-finally.

examples/niscope/graph-measurement-betterproto.py
kruthi-NI: @doshirohan , other examples handle a couple of cases in exception handling know. Would it not be possible to do that in better proto?

examples/niscope/graph-measurement-betterproto.py
kruthi-NI: @harsha-bhushan , what would you prefer? The older error api or the new api which does not take error code?

examples/niscope/graph-measurement-betterproto.py
ni-pkhare: For the case when server is not running and we run the client, a WinError exception is thrown instead of gRPC exception. That's why I am catching any type of exception in general.

kruthi-NI: Ok

kruthi-NI: @ni-pkhare   For the case of not found and Unimplemented what is the behaviour?

examples/niscope/graph-measurement-betterproto.py
kruthi-NI: If we just cherry pick to the release branch, the released client will not have the fix. Since example was working without error in previous release, and our changes have caused the bug, I would prefer that atleast just the bug fix part to get submitted by tomorrow before creating release. @ni-pkhare , would it be possible to separate just bug fix part to another tiny PR and we can ask owners to review it today since we need that for release? This PR of handling exception is fine to be just cherry picked to release branch

doshirohan: I don't think size of change matters for cherry picking, we need to get change in main and create cherrypick PR after that to submit it in releases branch. Question is can we delay release by another day for that.

examples/niscope/graph-measurement-betterproto.py
doshirohan: Can we catch WinError and gRPC error separately instead of generic Exception? To be frank, I don't have problem with catching generic exception either as long as we have good error messages reported for UNIMPLEMENTED and NOT_FOUND cases.

examples/niscope/graph-measurement-betterproto.py
harsha-bhushan: 1. A one-liner educating the users to use raw value should be good. We should avoid having a big paragraph.
2. Since we are anyway waiting on the NI-Sync team for updating release notes, I don't have a concern waiting for another day. My recommendation - fix the functionality bug as part of this PR & get it into the release.
The consistency related changes from other comments is not that critical. We could also do the clean up in future releases when we provide more better proto examples & discover better coding patterns.

examples/niscope/graph-measurement-betterproto.py
```diff
@@ -181,12 +190,24 @@ async def MeasureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, vi)
...
+        error_message = e.args[1]
+        if e.args[0] == 22:
+            error_message = f"Failed to connect to server on {server_address}:{server_port}"
+        elif e.status.name == "UNIMPLEMENTED":
+            error_message = "The operation is not implemented or is not supported/enabled in this service"
+        print(error_message)
+    finally:
+        if grpc_scope:
+            await CloseGrpcScope(*grpc_scope)
+        channel.close()
```
ni-pkhare, line 210: Here, in the case when server is not running, Channel is still being created but not the session, so `OpenGrpcScope` does not return anything `grpc_scope` tuple. Hence Channel needs to be closed seperately and not as a part of `CloseGrpcScope`

examples/niscope/graph-measurement-betterproto.py
ni-pkhare: @doshirohan , @kruthi-NI I have made changes now regarding UNIMPLEMENTED case and when server is not running. Could you pl have a look at it?

examples/niscope/plot-read-waveform-betterproto.py
```diff
@@ -60,82 +60,90 @@
...
+        error_message = e.args[1]
+        if e.args[0] == 22:
+            error_message = f"Failed to connect to server on {server_address}:{server_port}"
+        elif e.status.name == "UNIMPLEMENTED":
+            error_message = "The operation is not implemented or is not supported/enabled in this service"
+        print(error_message)
+    finally:
+        if('vi' in vars() and vi.id != 0):
+            await scope_service.close(vi = vi)
+        channel.close()
```
ni-pkhare, line 146: Here everything is just copied and pasted into try-except-finally block. Thats it.

examples/niscope/graph-measurement-betterproto.py
ni-pkhare: I have changed it back to original getErrorMessage API.

examples/niscope/graph-measurement-betterproto.py
ni-pkhare: I think, this conversation can be resolved as it has been addressed seperately.

kruthi-NI (approved):

doshirohan (approved):

astarche (approved):

PR #238, opened 2021-06-25
Add nifake GetAttributeViSesion test

astarche:
### What does this Pull Request accomplish?

Add a test covering for an `ViSession` out parameter from a non-open method.

### Why should this Pull Request be merged?

A similar method exists on many of the MI drivers.

I'm working on changes to the `SessionRepository`. A version of my change broke multiple `TClk` tests but no unit or integration tests.

### What testing has been done?

Ran and passed included test.
Confirmed how it was broken and can be fixed with the same fixes that worked for `TClk`.

PR #243, opened 2021-06-29
Add SessionResourceRepository to support non-IVI sessions

astarche:

Create a `SessionResourceRepository` class that wraps `SessionRepository` and allows storing driver-specific resource handles.

DAQmx will be the first non-IVI driver. Up to now, the `SessionRepository` has stored the 32-bit `ViSession` directly.

For DAQmx, we need to support a pointer-sized `TaskHandle`. We also need to be able to store incompatible (and potentially conflicting) resource types independently.

This change supports the DAQmx requirement while still preserving the session management features of the `SessionRepository`.

Note that all MI drivers will share a single `SessionResourceRepository`. This is required for `TClk` timing and sync, which adds functionality to existing sessions.

Ran and passed new auto-tests.
Ran and passed all system tests.
Used a prototype version of this in a prototype DAQ server.

### DIFF
--- a/generated/nidcpower/nidcpower_service.cpp
+++ b/generated/nidcpower/nidcpower_service.cpp
@@ -14,7 +14,7 @@
 
 namespace nidcpower_grpc {
 
-  NiDCPowerService::NiDCPowerService(NiDCPowerLibraryInterface* library, nidevice_grpc::SessionRepository* session_repository)
+  NiDCPowerService::NiDCPowerService(NiDCPowerLibraryInterface* library, ResourceRepositorySharedPtr session_repository)
       : library_(library), session_repository_(session_repository)
   {
   }
@@ -754,14 +754,14 @@ namespace nidcpower_grpc {
       ViBoolean reset = request->reset();
       ViConstString option_string = request->option_string().c_str();
 
-      auto init_lambda = [&] () -> std::tuple<int, uint32_t> {
+      auto init_lambda = [&] () {
         ViSession vi;
         int status = library_->InitializeWithIndependentChannels(resource_name, reset, option_string, &vi);
         return std::make_tuple(status, vi);
       };
       uint32_t session_id = 0;
       const std::string& session_name = request->session_name();
-      auto cleanup_lambda = [&] (uint32_t id) { library_->Close(id); };
+      auto cleanup_lambda = [&] (ViSession id) { library_->Close(id); };
       int status = session_repository_->add_session(session_name, init_lambda, cleanup_lambda, session_id);
       response->set_status(status);
       if (status == 0) {
@@ -2385,7 +2385,8 @@ namespace nidcpower_grpc {
       auto status = library_->GetAttributeViSession(vi, channel_name, attribute_id, &attribute_value);
       response->set_status(status);
       if (status == 0) {
-        response->mutable_attribute_value()->set_id(attribute_value);
+        auto session_id = session_repository_->resolve_session_id(attribute_value);
+        response->mutable_attribute_value()->set_id(session_id);
       }
       return ::grpc::Status::OK;
     }
@@ -2789,14 +2790,14 @@ namespace nidcpower_grpc {
       ViBoolean reset = request->reset();
       ViConstString option_string = request->option_string().c_str();
 
-      auto init_lambda = [&] () -> std::tuple<int, uint32_t> {
+      auto init_lambda = [&] () {
         ViSession vi;
         int status = library_->InitializeWithChannels(resource_name, channels, reset, option_string, &vi);
         return std::make_tuple(status, vi);
       };
       uint32_t session_id = 0;
       const std::string& session_name = request->session_name();
-      auto cleanup_lambda = [&] (uint32_t id) { library_->Close(id); };
+      auto cleanup_lambda = [&] (ViSession id) { library_->Close(id); };
       int status = session_repository_->add_session(session_name, init_lambda, cleanup_lambda, session_id);
       response->set_status(status);
       if (status == 0) {
--- a/source/codegen/templates/service.h.mako
+++ b/source/codegen/templates/service.h.mako
@@ -15,6 +15,9 @@ namespace_prefix = config["namespace_component"] + "_grpc::"
 if len(config["custom_types"]) > 0:
   custom_types = config["custom_types"]
 (input_custom_types, output_custom_types) = common_helpers.get_input_and_output_custom_types(functions)
+resource_handle_type = "ViSession"
+resource_repository_type = f"nidevice_grpc::SessionResourceRepository<{resource_handle_type}>"
+resource_repository_ptr = f"std::shared_ptr<{resource_repository_type}>"
 %>\
 
 //---------------------------------------------------------------------
@@ -31,7 +34,7 @@ if len(config["custom_types"]) > 0:
 #include <grpcpp/health_check_service_interface.h>
 #include <grpcpp/ext/proto_server_reflection_plugin.h>
 #include <map>
-#include <server/session_repository.h>
+#include <server/session_resource_repository.h>
 #include <server/shared_library.h>
 
 #include "${config["module_name"]}_library_interface.h"
@@ -40,8 +43,11 @@ namespace ${config["namespace_component"]}_grpc {
 
 class ${service_class_prefix}Service final : public ${service_class_prefix}::Service {
 public:
-  ${service_class_prefix}Service(${service_class_prefix}LibraryInterface* library, nidevice_grpc::SessionRepository* session_repository);
+  using ResourceRepositorySharedPtr = ${resource_repository_ptr};
+
+  ${service_class_prefix}Service(${service_class_prefix}LibraryInterface* library, ResourceRepositorySharedPtr session_repository);
   virtual ~${service_class_prefix}Service();
+  
 % for function in common_helpers.filter_proto_rpc_functions(functions):
 <%
   f = functions[function]
@@ -51,7 +57,7 @@ public:
 % endfor
 private:
   ${service_class_prefix}LibraryInterface* library_;
-  nidevice_grpc::SessionRepository* session_repository_;
+  ResourceRepositorySharedPtr session_repository_;
 % if common_helpers.has_viboolean_array_param(functions):
   void Copy(const std::vector<ViBoolean>& input, google::protobuf::RepeatedField<bool>* output);
 % endif
--- a/source/server/session_repository.cpp
+++ b/source/server/session_repository.cpp
@@ -27,12 +27,11 @@ int SessionRepository::add_session(
     return 0;
   }
   auto info = std::make_shared<SessionInfo>();
-  auto init_result = init_func();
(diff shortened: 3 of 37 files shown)


generated/nidcpower/nidcpower_service.cpp
```diff
@@ -2385,7 +2385,8 @@ namespace nidcpower_grpc {
       auto status = library_->GetAttributeViSession(vi, channel_name, attribute_id, &attribute_value);
       response->set_status(status);
       if (status == 0) {
-        response->mutable_attribute_value()->set_id(attribute_value);
+        auto session_id = session_repository_->resolve_session_id(attribute_value);
```
astarche, line 2388: This change is what fixes the tests covered in #238 in this implementation. The client can't do anything with a `ViSession` in this implementation. They need to resolve to the server allocated session id.

source/codegen/templates/service.h.mako
```diff
@@ -15,6 +15,9 @@ namespace_prefix = config["namespace_component"] + "_grpc::"
 if len(config["custom_types"]) > 0:
   custom_types = config["custom_types"]
 (input_custom_types, output_custom_types) = common_helpers.get_input_and_output_custom_types(functions)
+resource_handle_type = "ViSession"
```
astarche, line 18: Note: this will be replaced by some conditional lookup for DAQ.

source/server/session_repository.cpp
```diff
@@ -27,12 +27,11 @@ int SessionRepository::add_session(
...
   auto info = std::make_shared<SessionInfo>();
-  auto init_result = init_func();
-  int status = std::get<0>(init_result);
-  if (status != 0) {
+  auto status = init_func();
+  if (status) {
     return status;
   }
-  session_id = std::get<1>(init_result);
+  session_id = next_id();
```
astarche, line 34: `SessionRepository`s `InitFunc` no longer returns a tuple because the repository allocates the unique id.

source/server/session_resource_repository.h
```diff
@@ -0,0 +1,221 @@
+#pragma once
```
astarche, line 1: TODO remove pragma once :(. Google's style guide calls for include guards so I'll try to stick with that unless/until I can update everything.

source/server/session_resource_repository.h
```diff
@@ -0,0 +1,221 @@
...
+    SessionToResourceMap map_;
+    ResourceToSessionMap reverse_map_;
+    mutable std::recursive_mutex map_mutex_;
+  };
+
+  SessionRepository* session_repository_;
+  ResourceMap resource_map_;
+};
+
+template <typename TResourceHandle>
```
astarche, line 77: Note: [Here's ](https://google.github.io/styleguide/cppguide.html#Self_contained_Headers) the style guide section that covers this.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -31,37 +31,55 @@ using ::testing::SetArgPointee;
...
-  };
-  uint32_t session_id;
-  session_repo.add_session("", init_lambda, NULL, session_id);
-  return session_id;
+  auto resource_repository = std::make_shared<FakeResourceRepository>(session_repository);
+  return nifake_grpc::NiFakeService(library, resource_repository);
 }
 
-std::uint32_t create_session(std::string session_name, nidevice_grpc::SessionRepository& session_repo, ViSession sessionToStore)
+std::int32_t create_session(
```
astarche, line 47: Note: In an earlier version of this change, I had the services each own their own `SessionResourceRepository`. Then I discovered that this broke `TClk` and reverted to taking in the `SessionResourceRepository` as a `shared_ptr` (technically this could be conditional on whether-or-not the resources are share-able, but the codegen is simpler if it's always shared).

In that earlier version of the change, I removed the "seam" that made it possible to inject sessions directly. So I updated this test helper to add sessions through the init method. When I switch to a `shared_ptr` I reintroduced a similar seam. But, I like the tests like this. I think it's a more consistent "level" for this test class. And I think it's an advantage that it "works" with different implementations.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -82,15 +100,15 @@ TEST(NiFakeServiceTests, NiFakeService_InitWithOptionsSucceeds_CreatesAndStoresS
   EXPECT_TRUE(status.ok());
   EXPECT_EQ(kDriverSuccess, response.status());
   nidevice_grpc::Session session = response.vi();
-  EXPECT_EQ(kTestViSession, session_repository.access_session(session.id(), ""));
```
astarche, line 85: I weakened all these asserts from "I expect the session_repository to give me the driver-specific session id" to "I expect the session_repository to give *some valid server allocated session id*". 

This is also a case where I could add back more specific asserts on the contests of the session_**resource**_repository. But I think it's better to leave that out and focus on higher level testing in this class.

source/server/session_resource_repository.h
```diff
@@ -0,0 +1,226 @@
...
+  if (id) {
+    session_repository_->remove_session(id);
+  }
+}
+
+template <typename TResourceHandle>
+uint32_t SessionResourceRepository<TResourceHandle>::SessionResourceCreator::init_resource_handle()
+{
+  auto init_result = init_func_();
+  auto status = std::get<0>(init_result);
```
reckenro, line 147: Do you need this anymore? The `init_func_()` will just return an `int32_t` now, right?

reckenro: Nevermind, marking this resolved. I was getting confused between the `InitFunc` typedef in this file and the one in the `session_repository`. This makes sense now.

astarche: There are 2 different `InitFunc`s now.

On `SessionRepository`:  `typedef std::function<int32_t()> InitFunc;`
On `SessionResourceRepository`: `typedef std::function<std::tuple<int, TResourceHandle>()> InitFunc;`

This is the second one. The `InitFunc` passed in from the service still needs 2 return values because it creates the TResourceHandle (either a TaskHandle or a ViSession) and also returns a status. The `SessionResourceRepository` holds on to that and only returns the status to the `SessionRepository`.

Maybe just some renames would help with clarity?
Maybe `SessionResourceRepository::InitFunc` -> `SessionResourceRepository::CreateResourceFunc`?

Technically they both *do* the same thing, but they have different return values because of where they're invoked.

source/tests/unit/ni_fake_service_tests.cpp
reckenro: This weakening of the asserts makes sense to me.

I agree, this high level of testing feels right here.

source/tests/unit/session_resource_repository_tests.cpp
```diff
@@ -0,0 +1,259 @@
...
+      [](TResourceHandle handle) {},
+      session_id);
+  EXPECT_EQ(kNoError, result);
+  return session_id;
+}
+
+TEST(SessionResourceRepositoryTests, AddSessionResource_ResourceIsAdded)
+{
+  SessionRepository repository;
+  ;
```
reckenro, line 33: [Minor] Floating semi-colon here and throughout the test methods.

source/tests/unit/session_resource_repository_tests.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  EXPECT_EQ(kNoError, result);
+  EXPECT_NE(kNoSession, session_id);
+  EXPECT_EQ(
+      kResourceHandle,
+      resource_repository.access_session(session_id, ""));
+  EXPECT_EQ(
+      kResourceHandle,
+      resource_repository.access_session(0, kSessionName));
+  EXPECT_EQ(
```
reckenro, line 54: Do you think we should also have an assert like
```c++
EXPECT_EQ(
    session_id,
    repository.access_session(session_id, ""));
```
or is that redundant?

source/tests/unit/session_resource_repository_tests.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  EXPECT_EQ(
+      kNoSession,
+      resource_repository.access_session(session_id, ""));
+  EXPECT_EQ(
+      kNoSession,
+      repository.access_session(session_id, ""));
+}
+
+TEST(SessionResourceRepositoryTests, SessionResource_RemoveFromResourceRepository_ResourceIsRemoved)
```
reckenro, line 78: I might be missing something but is this test any different than `SessionResource_RemoveSessionResource_ResourceIsRemoved` besides the extra assert on `resolve_session_id`.

source/tests/unit/session_resource_repository_tests.cpp
```diff
@@ -0,0 +1,259 @@
...
+  SessionRepository repository;
+  ;
+  SessionResourceRepository<int32_t> resource_repository(
+      &repository);
+  SessionResourceRepository<int32_t> other_resource_repository(
+      &repository);
+  const int32_t kResourceHandle1 = -123456;
+  uint32_t session_id;
+  auto result = resource_repository.add_session(
+      "test_resource",
```
reckenro, line 240: [Nit] Can you pull "test_resource" out to a constant?

source/tests/unit/session_resource_repository_tests.cpp
```diff
@@ -0,0 +1,259 @@
...
+  using MockInitDelegate = ::testing::MockFunction<std::tuple<int32_t, int32_t>(void)>;
+  MockInitDelegate mock_init;
+  uint32_t second_session_id;
+  result = other_resource_repository.add_session(
+      "test_resource",
+      mock_init.AsStdFunction(),
+      [](int32_t handle) { FAIL() << "Unexpected Cleanup"; },
+      second_session_id);
+
+  EXPECT_EQ(kNoSession, second_session_id);
```
reckenro, line 255: Should we assert that `result` is 1 or possibly just non-zero here as well?

reckenro (approved):
Mostly just test cleanup suggestions and  commentary. Overall looks good though.

source/server/session_resource_repository.h
reckenro: Yeah, it was clear after I noticed there were two of these `InitFunc` functions defined. I had missed the one in defined in this file the first passthrough which led to the confusion.

If you want to do a rename I'd consider updating the cleanup function name as well so they're symmetrical like:
1. `InitFunc` and `CleanupFunc`
2. `CreateResourceFunc` and `CleanupResourceFunc`

source/tests/unit/session_resource_repository_tests.cpp
astarche: Added. I think that makes sense. Maybe should be covered by SessionRepository tests, but it's important to make sure all the links are working.

source/tests/unit/session_resource_repository_tests.cpp
astarche: I mistyped the test. It should be deleting the session_id from the session_repository. So we test remove both ways.

SparkingSpork (approved):

PR #244, opened 2021-06-29
[Owners Review] Bug fix in graph_measurement betterproto example

ni-pkhare:
### Justification
This PR fixes the [bug](https://ni.visualstudio.com/DevCentral/_workitems/edit/1512638) in graph_measurement betterproto example. 

### Implementation
1. Why did the bug arise?
Recently SetAttribute APIs were modified to include `raw_values` which are included inside `oneof` in the proto file. The Better Protobuf compiler has a bug generating helpers for gRPC messages with oneof types. The problem is that zero-value messages in a oneof group take priority based on order and overwrite other set values. Therefore, only the last field, the field that has '_raw` appended to the name, can be properly set without wrapper modification.
**The original example did not use _raw field at the places where SetAttribute APIs were used which was causing this issue.**
On changing those fields to raw type resolves the issue.

Note: The wiki already contains a highlight about the betterproto bug related to oneof gRPC types. https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client#creating-a-python-grpc-client

### Testing
Verified that the bug was resolved.

### DIFF
--- a/examples/niscope/graph-measurement-betterproto.py
+++ b/examples/niscope/graph-measurement-betterproto.py
@@ -119,7 +119,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
     set_trigger_result = await scope_service.set_attribute_vi_int32(
         vi = vi,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_TRIGGER_TYPE,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
+        value_raw = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
     )
     await CheckStatus(scope_service, vi, set_trigger_result)
 
@@ -137,7 +137,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
         vi = vi,
         channel_list = channels,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_MEAS_REF_LEVEL_UNITS,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_REF_LEVEL_UNITS_VAL_PERCENTAGE
+        value_raw = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_REF_LEVEL_UNITS_VAL_PERCENTAGE
     )
     await CheckStatus(scope_service, vi, set_units_result)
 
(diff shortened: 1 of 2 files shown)


examples/niscope/graph-measurement-betterproto.py
```diff
@@ -119,7 +119,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
     set_trigger_result = await scope_service.set_attribute_vi_int32(
         vi = vi,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_TRIGGER_TYPE,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
+        value_raw = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
```
doshirohan, line 122: Can you add this comment here so that customers are aware issue with betterproto in gRPC and why we are setting `value_raw` here.

> We have to set `value_raw` instead of `value` here because of `betterproto` bug related to `oneof` types in gRPC messages. More information here: https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client

doshirohan (approved):

examples/niscope/graph-measurement-betterproto.py
```diff
@@ -119,7 +119,7 @@ async def ConfigureGrpcScope(scope_service: niscope_grpc.NiScopeStub, channel, v
     set_trigger_result = await scope_service.set_attribute_vi_int32(
         vi = vi,
         attribute_id = niscope_grpc.NiScopeAttributes.NISCOPE_ATTRIBUTE_TRIGGER_TYPE,
-        value = niscope_grpc.NiScopeInt32AttributeValues.NISCOPE_INT32_TRIGGER_TYPE_VAL_EDGE_TRIGGER
```
astarche, line 122: > The problem is that zero-value messages in a oneof group take priority based on order and overwrite other set values. 

I think I understand this issue, but the way it's worded is confusing to me.

The spec'd implementation of `oneof` in protobuf is that only one can be set. If you set multiple fields of the `oneof` the "last one wins". It sounds like `betterproto` will set all fields on the `oneof` even if you didn't pass anything in for an argument. So, before your change, `value_raw` would be assigned of default zero-value after `value` is set and that would clear `value`.

Is that right?

doshirohan: Yes, that seems to be the behavior. This behavior (bug) is also documented at top of this example and in our wiki under 2nd bullet here: [creating-a-python-grpc-client](https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client#creating-a-python-grpc-client).

 I agree that the wording is confusing.

astarche (approved):

PR #245, opened 2021-06-30
Add daqmx service

astarche:

Creates an `nidaqmx` service in `grpc-device` with `CreateTask`\`ClearTask`.

## Details
* Use type_to_grpc_type and resource_handle_type to support non-IVI types.
* Support for using sessionName as a parameter to the driver init method.
* Add fake non-IVI service to unit test above.
* Add create/clear test for DAQ.
* Available in server behind toggle "nidaqmx".

Starts DAQmx feature.

Ran and passed new system test and spot-checked other system tests.
Manually tested server with toggle on and off.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -0,0 +1,77 @@
...
+    try {
+      const char* session_name = request->session_name().c_str();
+
+      auto init_lambda = [&] () {
+        TaskHandle task;
+        int status = library_->CreateTask(session_name, &task);
+        return std::make_tuple(status, task);
+      };
+      uint32_t session_id = 0;
+      const std::string& grpc_device_session_name = request->session_name();
```
astarche, line 62: Note: this method is an example of why the rename is needed in codegen. DAQmx (and the fake non-ivi service) are using `session_name` as a param to their init method. It's already in a `const char*` stack variable above ^^^. Renaming our internal management variable to a pseudo-namespaced name avoids the collision.

imports/include/niFakeNonIvi.h
```diff
@@ -0,0 +1,8 @@
+#ifndef NIFAKE_NON_IVI_H
+#define NIFAKE_NON_IVI_H
+#include <cstdint>
+
+using FakeHandle = uint64_t;
+using int32 = int32_t;
```
astarche, line 6: All drivers have some named handle type, so this needs one too. 

Both grpc and daqmx have a type named `int32`, so we rely on that existing in both. This gives fake non-ivi the same type.

source/codegen/metadata_mutation.py
```diff
@@ -59,6 +59,85 @@ def mark_mapped_enum_params(parameters, enums):
...
+        type = 'fixed64'
+    if 'int' == type:
+        type = 'sint32'
+    if "[]" in type:
+        type = type.replace("[]", "")
+
+    return "repeated " + type if add_repeated else type
+
+
+def populate_grpc_types(parameters, config):
```
astarche, line 125: Note: this method moves the pre-existing `get_parameter_type` to metadata mutation (setting the result to grpc_type) and augments it with support for a `type_to_grpc_type`.

Right now `type_to_grpc_type` is all-or-nothing (or really all-or-ivi). We can change that as it develops (or even use something more sophisticated to calculate mappings). But, for now, that makes it a little easier to catch "oops I forgot to add a new type to DAQ".

source/codegen/proto_helpers.py
```diff
@@ -119,7 +58,7 @@ def get_message_parameter_definitions(parameters, service_class_prefix, is_reque
   for parameter in parameters:
     is_array = common_helpers.is_array(parameter["type"])
     parameter_name = common_helpers.camel_to_snake(parameter["name"])
-    parameter_type = get_parameter_type(parameter, is_array, service_class_prefix)
```
astarche, line 122: Params no longer needed because it's computed as metadata mutation.

source/codegen/proto_helpers.py
```diff
@@ -168,10 +107,20 @@ def get_enum_parameters(parameter, parameter_name, parameter_type, is_array, use
...
-    parameter_type = parameter['grpc_type']
-  else:
-    parameter_type = get_grpc_type_from_ivi(parameter["type"], is_array, service_class_prefix)
-  return parameter_type
 
+def get_parameter_type(parameter):
+  return parameter['grpc_type']
+
+
+def get_parameters(function):
```
astarche, line 115: Pulled out of proto.mako with new support for `has_session_input` where any init method with a `sessionName` param will use that instead of adding a new one.

This allows DAQ to use session_name as task_name.

source/codegen/templates/service_helpers.mako
```diff
@@ -201,10 +201,11 @@ ${initialize_standard_input_param(function_name, parameter)}\
   field_name = common_helpers.camel_to_snake(parameter["name"])
   request_snippet = f'request->{field_name}()'
   c_type = parameter['type']
+  grpc_type = parameter.get('grpc_type', None)
   c_type_pointer = c_type.replace('[]','*')
   c_type_underlying_type = common_helpers.get_underlying_type_name(c_type)
 %>\
-% if c_type == 'ViConstString':
+% if c_type in ['ViConstString', 'const char*']:
```
astarche, line 208: Eventually I want to handle this a different way (at least for DAQ). But I think it's better to wait until we have more types to work with (maybe as part of the acquisition features).

source/server/core_server.cpp
```diff
@@ -123,6 +127,16 @@ static void RunServer(const ServerConfiguration& config)
   nifgen_grpc::NiFgenService nifgen_service(&nifgen_library, mi_shared_resource_repository);
   builder.RegisterService(&nifgen_service);
 
+  nidaqmx_grpc::NiDAQmxLibrary nidaqmx_library;
+  using DAQmxResourceRepository = nidevice_grpc::SessionResourceRepository<TaskHandle>;
+  auto daq_resource_repository = std::make_shared<DAQmxResourceRepository>(&session_repository);
+  nidaqmx_grpc::NiDAQmxService nidaqmx_service(&nidaqmx_library, daq_resource_repository);
```
astarche, line 133: Stack variables need to be outside the toggle if block.

This code is probably due for a refactor into unique_ptrs, but I think this change is OK without it.

source/codegen/metadata/nifake_non_ivi/config.py
```diff
@@ -0,0 +1,41 @@
...
+config = {
+    'api_version': '1.2.0d9',
+    'c_header': 'niFakeNonIvi.h',
+    'c_function_prefix': 'niFakeNonIvi_',
+    'service_class_prefix': 'NiFakeNonIvi',
+    'java_package': 'com.ni.grpc.fakenonivi',
+    'csharp_namespace': 'NationalInstruments.Grpc.FakeNonIvi',
+    'namespace_component': 'nifake_non_ivi',
+    'close_function': 'Close',
+    'driver_name': 'NI-FAKE-EXTENSION',
```
reckenro, line 12: Can you update the 'driver_name' to 'NI-FAKE-NON-IVI'?

CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ set(metadata_dir ${CMAKE_SOURCE_DIR}/source/codegen/metadata)
 set(service_output_dir ${CMAKE_SOURCE_DIR}/generated)
 set(codegen_dir ${CMAKE_SOURCE_DIR}/source/codegen)
 set(custom_dir ${CMAKE_SOURCE_DIR}/source/custom)
-set(nidrivers "nidcpower" "nidigitalpattern" "nidmm" "nifake" "nifake_extension" "nifgen" "niscope" "niswitch" "nisync" "nitclk")
+set(nidrivers "nidcpower" "nidigitalpattern" "nidmm" "nifake" "nifake_extension" "nifake_non_ivi" "nifgen" "niscope" "niswitch" "nisync" "nitclk" "nidaqmx")
```
SparkingSpork, line 70: I generally try to keep these alphabetical for easy reading. Can you move nidaqmx to be the first item in the list?

SparkingSpork: Same for below.

CMakeLists.txt
```diff
@@ -302,7 +307,8 @@ endif()
 
 target_include_directories(UnitTestsRunner
     PRIVATE "${service_output_dir}/nifake"
-    PRIVATE "${service_output_dir}/nifake_exension")
+    PRIVATE "${service_output_dir}/nifake_exension"
```
SparkingSpork, line 310: I somehow missed this previously, but looks like a typo here (nifake_extension)

source/codegen/metadata/nidaqmx/config.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+config = {
+    'api_version': '20.0.0',
+    'c_header': 'NIDAQmx.h',
+    'c_function_prefix': 'DAQmx',
+    'service_class_prefix': 'NiDAQmx',
```
SparkingSpork, line 6: @reckenro - You're more familiar with the naming we used than I am. Should this be NiDaqmx instead of NiDAQmx?

SparkingSpork (commented):
Should we consider splitting the addition of the nifake_non_ivi and the DAQmx service components into separate PRs? There's a lot to review here. 

Otherwise, the tests and build look good. I'll defer to Ryan on the actual codegen.

SparkingSpork (approved):
Chatted with Ryan offline, and since he's currently OK with this being a single PR for now I'm going to approve.

source/codegen/metadata_mutation.py
```diff
@@ -59,6 +59,85 @@ def mark_mapped_enum_params(parameters, enums):
...
+    if 'ViInt32' in type:
+        type = 'sint32'
+    if 'ViConstString' in type:
+        type = 'string'
+    if 'ViString' in type:
+        type = 'string'
+    if 'ViRsrc' in type:
+        type = 'string'
+    if 'ViChar' in type:
+        if is_array == True:
```
reckenro, line 80: [Nit] While you're here can you remove the `== True` from the `if is_array` evaluations?

source/codegen/proto_helpers.py
```diff
@@ -168,10 +107,20 @@ def get_enum_parameters(parameter, parameter_name, parameter_type, is_array, use
...
+  parameter_array = filter_parameters_for_grpc_fields(function['parameters'])
+  input_parameters = [p for p in parameter_array if common_helpers.is_input_parameter(p)]
+  if common_helpers.is_init_method(function):
+    has_session_input = any(param['name'] == 'sessionName' for param in input_parameters)
```
reckenro, line 119: An alternative to this which might give a little more flexibility is to have the parameter in functions.py have an extra field for `'is_session_name': True`. Then we could check that here if an input param has `is_session_name` set. If not, then create the `session_name_param` and also add that `'is_session_name': True` bit to it. Then in the codegen for init methods we could rely on finding the first (and only) param that has `is_session_name` set. This also would allow you to have the actual param in functions.py be named 'TaskName' if you want.

source/codegen/templates/service_helpers.mako
reckenro: Yeah, I agree. I think we can wait to refine this as we get further into the DAQ features.

source/tests/system/nidaqmx_session_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+  {
+    return nidaqmx_stub_;
+  }
+  DeviceServerInterface* device_server_;
+  std::unique_ptr<::nidevice_grpc::Session> driver_session_;
+  std::unique_ptr<NiDAQmx::Stub> nidaqmx_stub_;
+};
+
+TEST_F(NiDAQmxSessionTests, CreateTask_CloseTask_Succeeds)
+{
```
reckenro, line 51: Rename test name to `ClearTask` instead of `CloseTask` to match function names.

reckenro (approved):
Mostly small suggestions, looks good overall.

I did have one alternative for how to denote a param as the "sessionName" param without requiring it to be named "sessionName" that should give us some flexibility to name these session name parameters whatever we want.

source/codegen/metadata/nidaqmx/config.py
reckenro: I believe we've been using the same capitalization the `c_header` uses except "NI" => "Ni". So this seems fine. For example, that's what we've done with the tclk driver: its `service_class_prefix` capitilaztion follows its c_header: https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nitclk/config.py#L7

astarche: Interesting. Good question. The "little mx" always trips me up when I try to recapitalize DAQ and I fall back to the brand-name capitalization (or all-lower). `NiDCPower` also matches my choice on the allcaps "DC". So, I'll go with Ryan's rationale and leave it as-is.

generated/nifake/nifake_service.h
```diff
@@ -93,12 +93,12 @@ class NiFakeService final : public NiFake::Service {
   void Copy(const std::vector<CustomStruct>& input, google::protobuf::RepeatedPtrField<nifake_grpc::FakeCustomStruct>* output);
   CustomStruct ConvertMessage(const nifake_grpc::FakeCustomStruct& input);
   void Copy(const google::protobuf::RepeatedPtrField<nifake_grpc::FakeCustomStruct>& input, std::vector<CustomStruct>* output);
-  std::map<std::int32_t, float> nifakereal64attributevaluesmapped_input_map_ { {1, 3.5f},{2, 4.5f},{3, 5.5f},{4, 6.5f},{5, 7.5f}, };
```
astarche, line 96: I'm going to start a PR to fix this. It's iterating over a set so the ordering changes, it needs to be sorted or at least ordered.

source/codegen/proto_helpers.py
astarche: This had slightly more impact and I added a new test for it. Here's the [update(https://github.com/ni/grpc-device/pull/245/files/a0b6385a04928ddd7523c9f03a2a97c5259ff783..ed1991295e9aa081735f484dfe48af843cd20c10#diff-ae080b50783921b45d76b2b536fb4340b7e5fe6d3adb990956b5c8abe2bc493fL96).

Good idea though. We definitely could end up changing the name.

reckenro (approved):

PR #249, opened 2021-07-08
Add Create channel and Start/Stop Task to DAQ service

astarche:

Adds support for creating AI voltage channels, creating digital input/output channel, and starting and stopping a task.

`grpc-device-daq-tests.nce` is added to set up a simulated PXIe-6341 at `Dev1` for the system tests.

Incremental progress on DAQ grpc-device support.

Ran and passed all DAQ system tests, including new ones.


generated/nidaqmx/nidaqmx.proto
```diff
@@ -17,13 +17,40 @@ import "session.proto";
...
+}
+
+enum LineGrouping {
+  option allow_alias = true;
+  LINE_GROUPING_UNSPECIFIED = 0;
+  LINE_GROUPING_ChanForAllLines = 1;
+  LINE_GROUPING_ChanPerLine = 0;
+}
+
+enum VoltageUnits2 {
```
astarche, line 48: Note: this name is used in both the DAQ header and the CVI metadata files. There is a VoltageUnits1 but we're not using it yet.

generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -51,7 +181,7 @@ namespace nidaqmx_grpc {
       return ::grpc::Status::CANCELLED;
     }
     try {
-      const char* session_name = request->session_name().c_str();
+      auto session_name = const_cast<const char*>(request->session_name().data());
```
astarche, line 184: TODO: data "works" but it's not guranteed to be null-terminated.

source/codegen/templates/service_helpers.mako
```diff
@@ -208,8 +208,8 @@ ${initialize_standard_input_param(function_name, parameter)}\
   c_type_pointer = c_type.replace('[]','*')
   c_type_underlying_type = common_helpers.get_underlying_type_name(c_type)
 %>\
-% if c_type in ['ViConstString', 'const char*']:
-      ${c_type} ${parameter_name} = ${request_snippet}.c_str();\
+% if c_type in ['ViConstString', 'const char[]']:
+      auto ${parameter_name} = ${request_snippet}.c_str();\
```
astarche, line 212: Note: A lot of the diffs come from this change so that we don't have to conditionally encode the destination type on the let-hand side (instead the type conversion is happening on passing the parameter).

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -0,0 +1,180 @@
+#include <gtest/gtest.h>
```
astarche, line 1: Note: these are using a simulated DAQ device created in MAX. On windows, you can use the `.nce` file above to import the configuration. At some point we'll add instructions/scripts to set this up on linux.

astarche:
Added @doshirohan to check out the enums.

source/codegen/metadata/nidaqmx/enums.py
```diff
@@ -1,2 +1,60 @@
...
+                'name': 'ChanPerLine',
+                'value': 0
+            }
+        ]
+    },
+    'VoltageUnits2': {
+        'name': 'VoltageUnits2',
+        'values': [
+            {
+                'documentation': {
```
astarche, line 45: Note: The enums with documentation I was able to generate from CVI enum metadata. The ones *without* documentation showed up in the *function* metadata (as a list of valid values for the parameter, essentially) but weren't separately documented as an enum.

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -8,23 +7,142 @@
...
     'CreateTask': {
         'init_method': True,
         'parameters': [
             {
                 'direction': 'in',
+                'is_session_name': True,
                 'name': 'sessionName',
-                'type': 'const char*',
-                'is_session_name': True
+                'type': 'const char[]'
```
astarche, line 118: I changed these to `char[]` because it matches what the daq header uses and because I didn't see `*` used in any types in the rest of the grpc-metadata (though there are some typedefs of pointer types) and they're typically associated with outputs in the codegen.

source/tests/system/nidaqmx_driver_api_tests.cpp
SparkingSpork: It might be worth adding a GTEST_SKIP() to the setup if the device isn't found. Some of the [other tests](https://github.com/ni/grpc-device/blob/07850ae31a0c7086928e8ff5a1c984c1517b03c4/source/tests/system/nisync_driver_api_tests.cpp#L44) do this already.

SparkingSpork (approved):

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+TEST_F(NiDAQmxDriverApiTests, CreateAIVoltageChannel_Succeeds)
+{
+  CreateAIVoltageChanResponse response;
+  auto status = create_ai_voltage_chan(response);
+
+  EXPECT_SUCCESS(status, response);
+}
+
+TEST_F(NiDAQmxDriverApiTests, CreateDIChan_Succeeds)
```
reckenro, line 151: [Nit] Should the test names have "Channel" instead of "Chan" like "CreateAIVoltageChannel"? Minor but the test names seem like they should all use "Channel" or match the actual method names and use "Chan".

reckenro (approved):

source/tests/system/nidaqmx_driver_api_tests.cpp
astarche: I was on the fence on the actual behavior of skipping, but I like the idea of checking requirements up front and using that for documentation. So. I'm sold.

doshirohan (approved):

PR #251, opened 2021-07-13
Add ReadAnalogF64 to DAQ service

astarche:

Adds ReadAnalogF64 to DAQ service.

Enables very basic analog acquisition in DAQ.

Ran and passed all DAQ tests.

### DIFF
--- a/generated/nidaqmx/nidaqmx_library.cpp
+++ b/generated/nidaqmx/nidaqmx_library.cpp
@@ -26,6 +26,7 @@ NiDAQmxLibrary::NiDAQmxLibrary() : shared_library_(kLibraryName)
   function_pointers_.CreateDIChan = reinterpret_cast<CreateDIChanPtr>(shared_library_.get_function_pointer("DAQmxCreateDIChan"));
   function_pointers_.CreateDOChan = reinterpret_cast<CreateDOChanPtr>(shared_library_.get_function_pointer("DAQmxCreateDOChan"));
   function_pointers_.CreateTask = reinterpret_cast<CreateTaskPtr>(shared_library_.get_function_pointer("DAQmxCreateTask"));
+  function_pointers_.ReadAnalogF64 = reinterpret_cast<ReadAnalogF64Ptr>(shared_library_.get_function_pointer("DAQmxReadAnalogF64"));
   function_pointers_.StartTask = reinterpret_cast<StartTaskPtr>(shared_library_.get_function_pointer("DAQmxStartTask"));
   function_pointers_.StopTask = reinterpret_cast<StopTaskPtr>(shared_library_.get_function_pointer("DAQmxStopTask"));
 }
@@ -53,7 +54,7 @@ int32 NiDAQmxLibrary::ClearTask(TaskHandle task)
 #endif
 }
 
-int32 NiDAQmxLibrary::CreateAIVoltageChan(TaskHandle task, const char physicalChannel[], const char nameToAssignToChannel[], int terminalConfig, float64 minVal, float64 maxVal, int units, const char customScaleName[])
+int32 NiDAQmxLibrary::CreateAIVoltageChan(TaskHandle task, const char physicalChannel[], const char nameToAssignToChannel[], int32 terminalConfig, float64 minVal, float64 maxVal, int32 units, const char customScaleName[])
 {
   if (!function_pointers_.CreateAIVoltageChan) {
     throw nidevice_grpc::LibraryLoadException("Could not find DAQmxCreateAIVoltageChan.");
@@ -65,7 +66,7 @@ int32 NiDAQmxLibrary::CreateAIVoltageChan(TaskHandle task, const char physicalCh
 #endif
 }
 
-int32 NiDAQmxLibrary::CreateDIChan(TaskHandle task, const char lines[], const char nameToAssignToLines[], int lineGrouping)
+int32 NiDAQmxLibrary::CreateDIChan(TaskHandle task, const char lines[], const char nameToAssignToLines[], int32 lineGrouping)
 {
   if (!function_pointers_.CreateDIChan) {
     throw nidevice_grpc::LibraryLoadException("Could not find DAQmxCreateDIChan.");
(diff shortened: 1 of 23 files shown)


generated/nidaqmx/nidaqmx_library.cpp
```diff
@@ -53,7 +54,7 @@ int32 NiDAQmxLibrary::ClearTask(TaskHandle task)
 #endif
 }
 
-int32 NiDAQmxLibrary::CreateAIVoltageChan(TaskHandle task, const char physicalChannel[], const char nameToAssignToChannel[], int terminalConfig, float64 minVal, float64 maxVal, int units, const char customScaleName[])
+int32 NiDAQmxLibrary::CreateAIVoltageChan(TaskHandle task, const char physicalChannel[], const char nameToAssignToChannel[], int32 terminalConfig, float64 minVal, float64 maxVal, int32 units, const char customScaleName[])
```
astarche, line 57: Note: the CppHeaderParser tool that I'm using has a quirk where the `int32` typedef reports it's type as the underlying type `int` (but other similar ones do not). This started to cause a problem with out params. I fixed the metadata generation to correct this across the board, so it's in all the diffs.

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -125,6 +125,56 @@
...
+                'name': 'arraySizeInSamps',
+                'type': 'uInt32'
+            },
+            {
+                'direction': 'out',
+                'name': 'sampsPerChanRead',
+                'type': 'int32'
+            },
+            {
+                'direction': 'pass_null',
```
astarche, line 171: Note: this is documented in the [C API](https://zone.ni.com/reference/en-XX/help/370471AM-01/daqmxcfunc/daqmxreadanalogf64/). The desired behavior for grpc-device is to omit the parameter from the proto and hardcode a `nullptr` input.

I used `direction` for this for a few reasons. (1) the actual "direction" on the API is ambiguous, especially since ptrs are typically outputs. (2) it codegens easier this way.

astarche: > it codegens easier this way

It would be *almost* as easy to implement it as an `in` parameter with `pass_null: True`. There'd be a few more places where it needs to be removed from parameter lists (i.e., with `include_in_proto: False`)

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -141,6 +143,20 @@ class NiDAQmxDriverApiTests : public ::testing::Test {
...
 
+  ::grpc::Status read_analog_f64(
+      int32 samps_per_chan,
+      uint32 array_size_in_samps,
+      ReadAnalogF64Response& response)
+  {
+    ::grpc::ClientContext context;
+    ReadAnalogF64Request request;
+    set_request_session_id(request);
+    request.set_num_samps_per_chan(100);
```
reckenro, line 154: Did you want to use the `samps_per_chan` and `array_size_in_samps` args here instead of 100 for the request fields?

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -188,17 +204,25 @@ TEST_F(NiDAQmxDriverApiTests, CreateDOChannel_Succeeds)
...
 
   StartTaskResponse start_response;
   auto start_status = start_task(start_response);
+  ReadAnalogF64Response read_response;
+  auto read_status = read_analog_f64(100, 100, read_response);
   StopTaskResponse stop_response;
   auto stop_status = stop_task(stop_response);
 
+  EXPECT_EQ(read_response.read_array_size(), 100);
+  EXPECT_THAT(read_response.read_array(), Each(Not(Lt(AI_MIN))));
```
reckenro, line 222: [FYI] This is just in case you didn't see these matchers, but you could use >= and <= instead if you wanted for `Each(Ge(AI_MIN))` and `Each(Le(AI_MAX))`. But I prefer the way you have it, it reads better.

source/codegen/metadata/nidaqmx/functions.py
reckenro: Yeah, in the C API linked it's under the _Input_ parameters so I'm leaning towards keeping it that way. I think I'd prefer that as well because the direction being "pass_null" just sounds a little awkward to me. And I agree, `pass_null: True` sounds good for the indicator. I guess we could be more general with a `value: nullptr` but the pattern we probably'd rather capture is the pass_null part like you suggest.

reckenro (changes requested):
I want to avoid the `direction: 'pass_null'` and go with the alternative you suggested and see how that looks.

source/codegen/metadata/nidaqmx/functions.py
astarche: Good call, thanks for the feedback.

Addressed in [this change](https://github.com/ni/grpc-device/pull/251/commits/21ac3765c6e9023115cd41ed27b7ab7558df3539)

I like the `value: nullptr` idea too. The advantage of `pass_null` is that it captures both the `input_by_ptr` and `hardcoded_value` concept in one field. Of course we could also handle that with 2 fields, but I'd question whether that really generalizes to anything useful (at least for `input_by_ptr`; `hardcoded_value` seems pretty reusable).

source/codegen/templates/service_helpers.mako
```diff
@@ -197,6 +199,11 @@ ${initialize_standard_input_param(function_name, parameter)}\
       ${parameter['type']} ${parameter_name} = request->${field_name}().size();\
 </%def>
 
+## Initialize a 'pass_null' param.
+<%def name="initialize_pass_null_param(parameter)">\
+      auto ${parameter["name"]} = nullptr;\
```
reckenro, line 204: I think you'll have to do the same `common_helpers.camel_to_snake(parameter['cppName'])` other initialize methods do to come up with the variable name. This will make sure it aligns with what's expected later when we call `CreateArgs`. I think eventually this could be cleaned up and possibly refactored into the metadata_mutation stuff but for now just doing it in this method should be fine.

reckenro (approved):
Just a small update needed in the new initialize helper. But the `'pass_null': True` looks good in the metadata files.

PR #252, opened 2021-07-14
Add support for narrow integer types

gregstoll:

Adds support for arrays of narrow integer types that protobuf doesn't support, specifically: `int8`, `uInt8`, `int16`, `uInt16`. Per the spec, `int8[]` is represented as the protobuf type `bytes`, and all the others are `repeated int32`. If a user passes in an `int32` that is out of range for the real type, the wrapper code will throw an exception.

This is needed to support DAQmx methods such as `WriteDigitalU16`.

Wrote comprehensive unit tests for `nifake_non_ivi` and system tests that make real DAQmx calls. All automated tests pass, and the project builds in debug and release configurations.


source/codegen/common_helpers.py
```diff
@@ -139,6 +139,16 @@ def get_size_mechanism(parameter):
   size = parameter.get('size', {})
   return size.get('mechanism', None)
 
+def get_size_expression(parameter):
```
gregstoll, line 142: note: this code was moved from `service_helpers.mako` since I needed to call it from multiple places.

source/codegen/metadata_mutation.py
```diff
@@ -126,16 +126,17 @@ def populate_grpc_types(parameters, config):
...
-          parameter['grpc_type'] = type_map[parameter['type']]
-        else:
-          is_array = common_helpers.is_array(parameter["type"])
-          parameter['grpc_type'] = get_grpc_type_from_ivi(
-              parameter['type'],
-              is_array,
-              service_class_prefix)
+          if parameter['type'] in type_map:
+            parameter['grpc_type'] = type_map[parameter['type']]
+            continue
```
gregstoll, line 134: note: I refactored this code a little to try to do the lookup from `type_to_grpc_type`, but if that fails fall back on the other method.

source/codegen/common_helpers.py
```diff
@@ -27,7 +27,7 @@ def get_input_and_output_custom_types(functions):
   return (input_custom_types, output_custom_types)
 
 def is_string_arg(parameter):
-  return parameter['type'] in ['ViChar[]', 'ViInt8[]', 'ViUInt8[]']
+  return parameter['type'] in ['ViChar[]', 'ViInt8[]', 'ViUInt8[]', 'char[]', 'uInt8[]']
```
astarche, line 30: Could this be based on the `grpc_type`? I'd like to use hardcoded driver-specific types less if possible. Keying on `grpc_type` can be a cheap way to do that, though we may need to build up something else (not in this change).

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -125,6 +125,108 @@
...
+            },
+            {
+                'direction': 'in',
+                'name': 'dataLayout',
+                'type': 'int32',
+            },
+            {
+                'direction': 'in',
+                'name': 'writeArray',
+                'type': 'const uInt16*',
```
astarche, line 212: Can these be `const uInt16[]` instead of pointer just to be more consistent (both with the rest of grpc-device and with the daq header)?

source/codegen/templates/service.cpp.mako
```diff
@@ -129,6 +129,9 @@ ${mako_helper.define_simple_method_body(function_name=function_name, function_da
     catch (nidevice_grpc::LibraryLoadException& ex) {
       return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
     }
+    catch (nidevice_grpc::ValueOutOfRangeException& ex) {
```
astarche, line 132: Can this be conditional? Maybe on

```python
service_helpers.requires_checked_conversion(function)
```

which would currently be implemented as something like:
```python
return bool([p for p in function["parameters"] if is_input_array_that_needs_coercion(p)])
```

source/codegen/templates/service_helpers.mako
```diff
@@ -249,8 +250,30 @@ ${initialize_standard_input_param(function_name, parameter)}\
...
+                  message.append(std::to_string(x));
+                  message.append(" doesn't fit in datatype ");
+                  message.append("${c_element_type_that_needs_coercion}");
+                  throw nidevice_grpc::ValueOutOfRangeException(message);
+              }
+              return static_cast<${c_element_type_that_needs_coercion}>(x);
+        });
 % elif common_helpers.is_array(c_type):
       auto ${parameter_name} = const_cast<${c_type_pointer}>(${request_snippet}.data());\
+%elif parameter_name == 'reserved' and c_type == 'bool32*':
```
astarche, line 275: I have an implementation for `reserved` params in #251 . We can merge that in.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -11,9 +11,12 @@ using namespace nifake_non_ivi_grpc;
 
 using ::testing::_;
 using ::testing::DoAll;
+using ::testing::HasSubstr;
```
astarche, line 14: Consider: I think `using namespace ::testing;` would be a good alternative to all of these usings.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -105,6 +385,80 @@ TEST_F(NiFakeNonIviServiceTests, InitWithHandleNameAsSessionName_CloseSession_Cl
 
   EXPECT_EQ(kDriverSuccess, status);
 }
+
+TEST_F(NiFakeNonIviServiceTests, InputArraysWithNarrowIntegerTypes_U16DataGetsCoerced)
+{
+  auto status = input_arrays_with_narrow_integer_types_u16();
```
astarche, line 391: The structure of these test methods is a little confusing.

I think you can just inline all of the helper methods into the test methods if you're not sharing them. The test methods will have access to the protected members.

If you are sharing them you can have the response as an out param. Although it's harder to pull out the EXPECT cleanly if that's doing something significant.

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -125,6 +125,108 @@
         ],
         'returns': 'int32'
     },
+    'ReadDigitalU16': {
```
astarche, line 128: Try to keep these alphabetical so that it will diff better when I generate the metadata (whcih is always alphabetical). Although maybe we'll be able to switch over to generated before submitting.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -22,6 +25,58 @@ namespace ni {
 namespace tests {
 namespace unit {
 
+MATCHER(CustomU16Data, "")
```
astarche, line 28: What about using these matchers: http://google.github.io/googletest/gmock_cook_book.html#matching-containers ?

source/codegen/templates/service_helpers.mako
```diff
@@ -262,21 +285,16 @@ ${initialize_standard_input_param(function_name, parameter)}\
...
-    size = common_helpers.camel_to_snake(parameter['size']['value_twist'])
-  else:
-    size = common_helpers.camel_to_snake(parameter['size']['value'])
+  size = common_helpers.get_size_expression(parameter)
 %>\
 %     if common_helpers.is_struct(parameter) or underlying_param_type == 'ViBoolean':
       std::vector<${underlying_param_type}> ${parameter_name}(${size}, ${underlying_param_type}());
 ## Byte arrays are leveraging a string as a buffer, so we don't need to take special consideration of the null terminator.
-%     elif parameter['type'] in ['ViInt8[]', 'ViUInt8[]']:
+%     elif parameter['type'] in ['ViInt8[]', 'ViUInt8[]', 'uInt8[]']:
```
astarche, line 297: Use `grpc_type` here too if possible

source/codegen/templates/service_helpers.mako
```diff
@@ -286,7 +304,10 @@ ${initialize_standard_input_param(function_name, parameter)}\
       if (${size} > 0) {
           ${parameter_name}.resize(${size}-1);
       }
-%     elif underlying_param_type in ['ViAddr', 'ViInt32', 'ViUInt32', 'ViUInt16']:
+%     elif service_helpers.is_output_array_that_needs_coercion(parameter):
+      std::vector<${underlying_param_type}> ${parameter_name}(${size});
+## uInt32 requires cast because of int vs long in that typedef vs uint32_t
+%     elif underlying_param_type in ['ViAddr', 'ViInt32', 'ViUInt32', 'ViUInt16', 'uInt32']:
```
astarche, line 310: Ugh. I think you should leave this one the way you have it. I think we'd want to address this case in a generalized conversion implementation, but I don't think it makes sense to try to come up with a one-off for it.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
gregstoll: After playing around with this and carefully reading the documentation, those matchers only work on STL-like containers. (the `ElementsAreArray()` one accepts an array as input, but still acts on an STL-like container)

source/codegen/metadata/nidaqmx/config.py
```diff
@@ -14,11 +14,14 @@
         'TaskHandle': 'nidevice_grpc.Session',
         'const char[]': 'string',
         'float64': 'double',
-        'float64[]': 'repeated double',
-        'const float64[]': 'repeated double',
+        'bool32': 'bool',
+        'int': 'int32',
```
astarche, line 18: I don't think this is necessary. All `int` in daq should be `int32`s.

imports/include/niFakeNonIvi.h
```diff
@@ -1,8 +1,9 @@
 #ifndef NIFAKE_NON_IVI_H
 #define NIFAKE_NON_IVI_H
 #include <cstdint>
+// we need the typedefs from here
+#include <NIDAQmx.h>
```
astarche, line 5: We talked about removing this and having fake non-ivi use stdint types instead. If you want, make a follow on task for this.

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -232,13 +339,110 @@
...
+                'type': 'int32',
+            },
+            {
+                'direction': 'in',
+                'name': 'timeout',
+                'type': 'double',
+            },
+            {
+                'direction': 'in',
+                'name': 'dataLayout',
```
astarche, line 421: Add `'enum': 'GroupBy',`

(Although I'm going to come through and regenerate this, so we don't need to be too fussy)

source/codegen/service_helpers.py
```diff
@@ -1,75 +1,110 @@
...
+
+
+def is_output_array_that_needs_coercion(parameter):
+    return common_helpers.is_output_parameter(parameter) and get_c_element_type_for_array_that_needs_coercion(parameter) is not None
+
+
 def create_args(parameters):
     result = ''
-    is_twist_mechanism = common_helpers.has_ivi_dance_with_a_twist_param(parameters)
+    is_twist_mechanism = common_helpers.has_ivi_dance_with_a_twist_param(
```
astarche, line 34: Is this getting auto-formatted by clang format?

I'd leave this the way you have it for this change. Can you try out "Format On Save Mode: modifications" in the future?

astarche (approved):

source/codegen/service_helpers.py
gregstoll: Yeah, this is getting auto-formatted. I do have my Format On Save Mode set to modifications; I think in this case the indentation got messed up in a merge somehow, and I was fixing that, so maybe that's what did it?

imports/include/niFakeNonIvi.h
gregstoll: Sorry, I thought we ended up at a different place on this. But this is good, and I made this change.

source/codegen/templates/service_helpers.mako
```diff
@@ -269,21 +290,16 @@ ${initialize_standard_input_param(function_name, parameter)}\
 <%
   parameter_name = common_helpers.camel_to_snake(parameter['cppName'])
   underlying_param_type = common_helpers.get_underlying_type_name(parameter["type"])
+  c_element_type_that_needs_coercion = service_helpers.get_c_element_type_for_array_that_needs_coercion(parameter)
```
reckenro, line 293: I don't think `c_element_type_that_needs_coercion` is used in this `def`. I would've expected it to be used after `is_output_array_that_needs_coercion` was checked to mirror what's done for initializing the input params but I might be missing something.

reckenro (approved):

PR #253, opened 2021-07-14
Adding a workflow to create AzDO bugs from 'Issues' created on the repo

harsha-bhushan:

This PR enables creation of bugs on AzDO (in the appropriate area path) when an issue on the grpc-device github repo is created.
A workflow was created using the steps outlined in this video: https://www.youtube.com/watch?v=I9ItftRT9dU
Using the yml script provided here: https://github.com/marketplace/actions/github-issues-to-azure-devops

(note: the other way - syncing any changes on AzDO back to github will not be done.)

This avoids the below manual efforts:
1. Tracking issues opened on the grpc-device repo
2. Creating corresponding AzDO bugs so that they show up on domain dashboard & can be discussed during standups
3. Closing bugs after fixing the issue

Unable to test this on the branch. Once merged, need to manually test the below scenarios:
Creating an  issue on github creates a bug in AzDO in the right project area path
Closing an issue on github closes the bug in AzDO

### DIFF
--- a/.github/workflows/sync_github_issues_to_azdo.yml
+++ b/.github/workflows/sync_github_issues_to_azdo.yml
@@ -1,0 +1,24 @@
+name: Sync issue to Azure DevOps work item
+
+on:
+  issues:
+    types:
+      [opened, edited, deleted, closed, reopened, labeled, unlabeled, assigned]
+
+jobs:
+  alert:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: danhellem/github-actions-issue-to-work-item@master
+        env:
+          ado_token: "${{ secrets.ADO_PERSONAL_ACCESS_TOKEN }}"
+          github_token: "${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}"
+          ado_organization: "ni"
+          ado_project: "DevCentral"
+          ado_area_path: "DevCentral\Product RnD\SWP\CBS\HW Config"
+          ado_wit: "Bug"
+          ado_new_state: "New"
+          ado_active_state: "Active"
+          ado_close_state: "Closed"
+          ado_bypassrules: true
+          log_level: 100
--- a/.github/workflows/sync_github_issues_to_azdo.yml
+++ b/.github/workflows/sync_github_issues_to_azdo.yml
@@ -15,7 +15,7 @@ jobs:
           github_token: "${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}"
           ado_organization: "ni"
           ado_project: "DevCentral"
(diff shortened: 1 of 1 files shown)


.github/workflows/sync_github_issues_to_azdo.yml
```diff
@@ -0,0 +1,24 @@
...
+    types:
+      [opened, edited, deleted, closed, reopened, labeled, unlabeled, assigned]
+
+jobs:
+  alert:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: danhellem/github-actions-issue-to-work-item@master
+        env:
+          ado_token: "${{ secrets.ADO_PERSONAL_ACCESS_TOKEN }}"
```
harsha-bhushan, line 14: note: I have created a personal access token & added it under secrets

SparkingSpork: I think it's preferable to get tokens created for this by the organization/project owners in GitHub and AzDO respectively. If we use a personal token you created, we have to rely on you in the future for maintaining it and keeping it from expiring or swap to a new token owned by someone else occasionally.

I know for some other workflows @zhindes requested an organizational level secret (such as [here](https://github.com/ni/grpc-device/blob/main/.github/workflows/build_cmake.yml#L228)). We should at least entertain that approach instead, which will result in an organizational secret available on all NI repositories: 
image

SparkingSpork: If you'd like to get this in to test it before getting the new access tokens created, I'm OK with that as well.

SparkingSpork (approved):

.github/workflows/sync_github_issues_to_azdo.yml
zhindes: Yeah, I would recommend an org-wide secret that isn't a personal token. Ping one of the NI AzDO/GitHub admins to get that created and setup. I usually talk directly to David Corrigan for stuff like this, but he may direct you to a better method of reaching out.

astarche (approved):

doshirohan (approved):

.github/workflows/sync_github_issues_to_azdo.yml
davidcorrigan714: Sorry on the delay on this, had a busy week last week and want to dig into the security features of GitHub a bit before setting up a token with R/W access to AzDO work items. I'll try and do that in the next few days.

davidcorrigan714: Added the tokens. The AzDO one is scoped pretty narrowly to just that area path so if that changes we'll need to update it. That area path should really be an ID and not a string too but that's not something you all could fix - it's just not as nice if area paths get rearranged in DevCentral.

PR #255, opened 2021-07-15
Add DAQ Task Control and Configuration methods

astarche:

Add methods from the Task Control and Configuration API to the DAQ service.

Does not include methods with callbacks or timestamps. Support for those will be added separately.

Incremental progress towards DAQ grpc-device support.

Ran and passed all DAQ auto-tests.

### DIFF
--- a/source/codegen/templates/service_helpers.mako
+++ b/source/codegen/templates/service_helpers.mako
@@ -218,8 +218,6 @@ ${initialize_standard_input_param(function_name, parameter)}\
 %>\
 % if c_type in ['ViConstString', 'const char[]']:
       auto ${parameter_name} = ${request_snippet}.c_str();\
-% elif c_type == 'ViString' or c_type == 'ViRsrc':
-      ${c_type} ${parameter_name} = (${c_type})${request_snippet}.c_str();\
 % elif common_helpers.is_string_arg(parameter):
       ${c_type_pointer} ${parameter_name} = (${c_type_pointer})${request_snippet}.c_str();\
 % elif c_type == 'ViSession[]':
--- a/source/tests/system/nidaqmx_driver_api_tests.cpp
+++ b/source/tests/system/nidaqmx_driver_api_tests.cpp
@@ -29,7 +29,7 @@ struct ThrowawayResponse {
 
 class NiDAQmxDriverApiTests : public Test {
  protected:
-  const std::string DEVICE_NAME{"Dev1"};
+  const std::string DEVICE_NAME{"gRPCSystemTestDAQ"};
 
   NiDAQmxDriverApiTests()
       : device_server_(DeviceServerInterface::Singleton()),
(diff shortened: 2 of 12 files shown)


source/codegen/templates/service_helpers.mako
```diff
@@ -217,8 +217,6 @@ ${initialize_standard_input_param(function_name, parameter)}\
 %>\
 % if c_type in ['ViConstString', 'const char[]']:
       auto ${parameter_name} = ${request_snippet}.c_str();\
-% elif c_type == 'ViString' or c_type == 'ViRsrc':
```
astarche, line 220: Note: this case is handled by the `common_helpers.is_string_arg` case below. That case is more general because it can deal with the case where `c_type` is an array (in which case we need to cast to pointer type)

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -29,6 +29,8 @@ struct ThrowawayResponse {
 
 class NiDAQmxDriverApiTests : public Test {
  protected:
+  const std::string DEVICE_NAME{"Dev1"};
```
astarche, line 32: Note: potential merge issue with #252

source/codegen/common_helpers.py
```diff
@@ -33,7 +33,7 @@ def get_input_and_output_custom_types(functions):
   return (input_custom_types, output_custom_types)
 
 def is_string_arg(parameter):
-  return parameter['type'] in ['ViChar[]', 'ViInt8[]', 'ViUInt8[]']
+  return parameter['type'] in ['ViChar[]', 'ViInt8[]', 'ViUInt8[]'] or parameter['grpc_type'] == 'string'
```
reckenro, line 36: I'm guessing this will go away from the diff when you take in Greg's changes so it becomes `return parameter['grpc_type'] in ['string', 'bytes']`, right?

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -317,6 +349,34 @@ TEST_F(NiDAQmxDriverApiTests, AOVoltageChannel_WriteAODataWithOutOfRangeValue_Re
 
   EXPECT_EQ(DAQmxErrorInvalidAODataWrite, write_response.status());
 }
+
+TEST_F(NiDAQmxDriverApiTests, TaskWithAOChannel_GetNthTaskDevice_ReturnsDeviceForChannel)
+{
+  const double AO_MIN = 1.0;
+  const double AO_MAX = 10.0;
+  CreateAOVoltageChanResponse create_channel_response;
+  create_ao_voltage_chan(AO_MIN, AO_MAX, create_channel_response);
```
reckenro, line 358: You can simplify this to just be `create_ao_voltage_chan(AO_MIN, AO_MAX)` since you introduced the default throwaway response option for it.

reckenro (approved):

PR #264, opened 2021-07-19
Synchronous server streaming implementation of RegisterDoneEvent

astarche:

Implement `RegisterDoneEvent` as a thread-consuming synchronous server method.

Note: the final implementation will use an asynchronous server implementation to avoid blocking a thread for each registered callback.

Partially implements callbacks-as-response streams.
Allows staging some of the plumbing (`CallbackRouter`) and codegen (`stream out` and `callback_params`).
Provides a working server to test against.

Ran and passed all DAQ tests.
Some manual testing with python client.


source/codegen/metadata/nifake_non_ivi/functions.py
```diff
@@ -146,4 +146,33 @@
         ],
         'returns': 'int32'
     },
+    'RegisterCallback': {
```
astarche, line 149: Note: I decided to only test this with a synchronous callback instead of  streaming callback, The streaming implementation would require faking a grpc ServerWriter. Since that is not part of the final implementation, that would be wasted work.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,4 +1,46 @@
...
+
+    auto status = library_->RegisterDoneEvent(task, options, DoneEventCallbackRouter::handle_callback, registration.token());
+
+    if (status) {
+      RegisterDoneEventResponse failed_to_register_response;
+      failed_to_register_response.set_status(status);
+      writer->Write(failed_to_register_response);
+    }
+
+    while (!context->IsCancelled()) {
```
astarche, line 37: Note: The waiting and boundary conditions in this change are half-baked. I expect those to be different in the async implementation

source/server/callback_router.h
```diff
@@ -0,0 +1,112 @@
...
+  using Token = void*;
+
+  // Move-only RAII type exposed through register_handler, to auto-unregister_handler when leaving scope.
+  class CallbackRegistration {
+   public:
+    CallbackRegistration() : token_(0) {}
+    CallbackRegistration(Token token) : token_(token) {}
+    CallbackRegistration(const CallbackRegistration&) = delete;
+    CallbackRegistration& operator=(const CallbackRegistration&) = delete;
+    CallbackRegistration(CallbackRegistration&& rhs) : token_(rhs.token_) { rhs.token_ = reinterpret_cast<Token>(0); }
```
astarche, line 28: This needs to be move-able to allow returning from `register_handler` because the version of gcc we're using on linux RT doesn't implement "required move elision" (I think).

The return from that method will likely change to a `unique_ptr` to support the async implementation. In that case, this wouldn't be needed anyway.

source/tests/unit/callback_router_tests.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+TEST(CallbackRouterTests, IntCallbackHandlerRegistered_HandleCallbackMultipleTimes_CallsHandlerMultipleTimes)
+{
+  MockIntHandler handler;
+  auto registration = IntCallbackRouter::register_handler(handler.AsStdFunction());
+
+  const int32_t FIRST_CALLBACK_VAL = 0x1234;
+  EXPECT_CALLBACK(handler, FIRST_CALLBACK_VAL);
+  auto first_result = IntCallbackRouter::handle_callback(FIRST_CALLBACK_VAL, registration.token());
+  const int32_t SECOND_CALLBACK_VAL = 0x1234;
```
gregstoll, line 64: Would it make sense to have this value be different than `FIRST_CALLBACK_VAL`?

gregstoll (approved):

source/codegen/proto_helpers.py
```diff
@@ -44,22 +44,16 @@ def get_enum_definitions(enums_to_define, enums):
...
-      "type": "int32",
-      "grpc_field_number": "1"
-    })
-    used_indexes = [1]
   for parameter in parameters:
     is_array = common_helpers.is_array(parameter["type"])
     parameter_name = common_helpers.camel_to_snake(parameter["name"])
     parameter_type = get_parameter_type(parameter)
     if common_helpers.is_enum(parameter):
+      is_request_message = parameter['direction'] == 'in'
```
reckenro, line 56: You can use `common_helpers.is_input_parameter(parameter)` here.

source/codegen/proto_helpers.py
```diff
@@ -44,22 +44,16 @@ def get_enum_definitions(enums_to_define, enums):
     enum_definitions.update({enum_name: enum_definition})
   return enum_definitions
 
-def get_message_parameter_definitions(parameters, is_request_message):
+def get_message_parameter_definitions(parameters):
   """Get simplified list of all parameters that can be used for definiing request/respones messages in proto file."""
   parameter_definitions = []
   used_indexes = []
-  if not is_request_message:
```
reckenro, line 51: I appreciate the refactoring of this into `get_parameters`.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -519,6 +538,25 @@ TEST_F(NiDAQmxDriverApiTests, AIVoltageChannel_CfgSampClkTimingAndAcquireData_Su
   EXPECT_SUCCESS(read_status, read_response);
   EXPECT_EQ(NUM_SAMPS, read_response.samps_per_chan_read());
 }
+
+TEST_F(NiDAQmxDriverApiTests, ChannelWIthDoneEventRegistered_RunCompleteFiniteAcquisition_DoneEventResponseIsReceived)
```
reckenro, line 542: Typo:
```suggestion
TEST_F(NiDAQmxDriverApiTests, ChannelWithDoneEventRegistered_RunCompleteFiniteAcquisition_DoneEventResponseIsReceived)
```

reckenro (approved):
You can bypass Charlie since he's out this week.

PR #265, opened 2021-07-20
Add missing functions from DAQ basic API

astarche:

Add remaining functions from the DAQ basic API that don't require new codegen.

These were previously left off because either (a) they had missing enums or (b) they use datatypes that were not handled correctly in metadata generation scripts.

Enable `/bigobj` globally. DAQ tests, the service, and the generated protobuf files now exceed this limit.

Incremental progress on DAQ grpc support.

Ran and passed all DAQ tests.


reckenro (approved):
Charlie's out this week so feel free to submit without waiting. I think the CMakeLists change is fine but I'll let him know when he gets back about it.

SparkingSpork:
> Charlie's out this week so feel free to submit without waiting. I think the CMakeLists change is fine but I'll let him know when he gets back about it.

Just want to document that the changes look fine to me. We should probably move the location of this logic in a future PR since they're now intended to be applied globally instead of only to the target they're near.

PR #266, opened 2021-07-20
Fix handling of all caps acronyms

astarche:

Fix `pascal_to_camel` to handle acronyms at the beginning of word.
Fix `camel_to_snake` to handle acronyms in the middle of a word.

This prevents spelling out acronyms with `_` separators. I.e., `A_C_EXCIT_WIRE_MODE_6_WIRE`.

Ran and passed all DAQ, Scope, and DCPower tests.


generated/nidcpower/nidcpower.proto
```diff
@@ -424,8 +424,8 @@ enum NiDCPowerInt32AttributeValues {
   NIDCPOWER_INT32_COMPLIANCE_LIMIT_SYMMETRY_VAL_ASYMMETRIC = 1;
   NIDCPOWER_INT32_CURRENT_LIMIT_BEHAVIOR_VAL_CURRENT_REGULATE = 0;
   NIDCPOWER_INT32_CURRENT_LIMIT_BEHAVIOR_VAL_CURRENT_TRIP = 1;
```
astarche, line 426: @harsha-bhushan @doshirohan 

This is all of the changes to existing enums if we were to just change the behavior.

Any existing code will continue to work **until** you re-run protoc. At that point it will be "source-broken" and the user needs to patch the enums.

ALTERNATIVE:
1. Add an `introduced_version` to the configuration for each driver as needed. Use the version to switch to legacy behavior on codepaths where it's known to break APIs. (Default to the new behavior and use on any codepaths that don't currently change the output).
2. Reject this change and leave it as is.

reckenro (approved):

source/codegen/common_helpers.py
```diff
@@ -57,17 +59,15 @@ def is_unsupported_enum_array(parameter):
...
-  for x in camelString :
-      if x.isupper():
-          camelString[index] = camelString[index].lower()
-          camelString.insert(index, "_")
-
-      index = index + 1
-  return ("".join(camelString))
+  # Add _ before Words: 
+  # someDeviceIPAddress -> some_DeviceIP_Address
+  s1 = re.sub(r"([^_])([A-Z][a-z]+)", r"\1_\2", camel_string)
```
doshirohan, line 66: Should this regex also handle numerics in input string (e.g. `voltageLevel2`)

source/codegen/common_helpers.py
```diff
@@ -83,9 +83,19 @@ def snake_to_pascal(snake_string):
...
-  pascal_string = list(pascal_string)
-  pascal_string[0] = pascal_string[0].lower()
-  return ("".join(pascal_string))
+  # Full string all-caps: IEPE -> iepe
+  match = re.fullmatch(r"([A-Z]+)", pascal_string)
+  if match:
+    return match[1].lower()
+  
+  # Leading all-caps: IPAddress -> ipAddress
+  match = re.fullmatch(r"([A-Z]+)([A-Z].*)", pascal_string)
```
doshirohan, line 92: `.*` handles numerics here, but it won't handle if it comes after uppercase acronym in the beginning.

e.g.
`IP2` - probably should be handled by first case above
`IP2Host`

I am not sure if we need to handle all of the cases now or handle them if we notice any such names during reviews in future (could be missed during review though when you are migrating big chunk of APIs).

generated/nidcpower/nidcpower.proto
doshirohan: I like the new names better, we should have thought of this before. I think source-broken code is acceptable in this case since this is something that would result in obvious error when you upgrade to newer proto files and won't be something hard to debug/fix.

doshirohan (approved):

source/codegen/common_helpers.py
astarche: Good catch.

source/codegen/common_helpers.py
astarche: Currently: `voltageLevel2` -> `voltage_level2`.
Are you suggesting `voltageLevel2` -> `voltage_level_2`?

I like that change for the numbered enums. But that's also a breaking behavior change, and some of those changes would not be as good. For example: 
`NIDCPOWER_INT32_CURRENT_LIMIT_BEHAVIOR_VAL_CURRENT_TRIP` -> `NIDCPOWER_INT_32_CURRENT_LIMIT_BEHAVIOR_VAL_CURRENT_TRIP`

We could look at that separately potentially.

I think that this method is susceptible to the same issue you caught below:

`establishHTTP2Connection` -> `establish_htt_p2_connection` -> BAD
`establishHTTP2Connection` -> `establish_http2_connection` -> GOOD/PROPOSED
`establishHTTP2Connection` -> `establish_http_2_connection` -> Also OK

reckenro:
I'd recommend merging from main and regenerating the generated folder before submitting so the updates for new enums that have been added like `ForceIEPEUnits` get updated with this change.

generated/nidcpower/nidcpower.proto
harsha-bhushan: I'll check Cifra's thoughts once & get back. I'm not too concerned about the breaking the source code in this case, since it affects only a few enums & in languages other than python, intellisence would help the user remove & select new name.

source/codegen/common_helpers.py
doshirohan: Ah, I didn't think so far ahead. I couldn't figure out where numbers are being handled in you regex, so my comment was mostly regarding that. But looks like it does take care of numbers as well and create your PROPOSED string above. That handling looks good to me compared to adding `_` before the number as well.

harsha-bhushan (approved):

PR #268, opened 2021-07-20
Add CalculateReversePolyCoeff to DAQ server

astarche:

Add CalculateReversePolyCoeff to DAQ server
 * Use `custom-code` to calculate the size based on multiple inputs. See [C API documentation](https://zone.ni.com/reference/en-XX/help/370471AM-01/daqmxcfunc/daqmxcalculatereversepolycoeff/).

Add Fake non ivi tests for `custom-code` size param,

Incremental progress on grpc DAQ feature.

Ran and passed all DAQ tests.

### DIFF
--- a/generated/nidaqmx/nidaqmx_service.cpp
+++ b/generated/nidaqmx/nidaqmx_service.cpp
@@ -86,7 +86,7 @@ namespace nidaqmx_grpc {
       float64 max_val_x = request->max_val_x();
       int32 num_points_to_compute = request->num_points_to_compute();
       int32 reverse_poly_order = request->reverse_poly_order();
-      response->mutable_reverse_coeffs()->Resize((reverse_poly_order == -1) ? num_forward_coeffs_in : reverse_poly_order + 1, 0);
+      response->mutable_reverse_coeffs()->Resize((reverse_poly_order < 0) ? num_forward_coeffs_in : reverse_poly_order + 1, 0);
       float64* reverse_coeffs = response->mutable_reverse_coeffs()->mutable_data();
       auto status = library_->CalculateReversePolyCoeff(forward_coeffs, num_forward_coeffs_in, min_val_x, max_val_x, num_points_to_compute, reverse_poly_order, reverse_coeffs);
       response->set_status(status);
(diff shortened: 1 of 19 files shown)


source/codegen/common_helpers.py
```diff
@@ -47,7 +47,7 @@ def is_unsupported_parameter(parameter):
       or is_unsupported_scalar_array(parameter)
 
 def is_unsupported_size_mechanism(parameter):
-  return not get_size_mechanism(parameter) in {'fixed', 'len', 'ivi-dance', 'passed-in', 'ivi-dance-with-a-twist', None}
+  return not get_size_mechanism(parameter) in {'fixed', 'len', 'ivi-dance', 'passed-in', 'ivi-dance-with-a-twist', 'custom-code', None}
```
astarche, line 50: The default handling of `mechanism` just drops `value` on the right-hand side of the init expression. So it works without any other change.

generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -72,6 +72,33 @@ namespace nidaqmx_grpc {
...
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      auto forward_coeffs = const_cast<const float64*>(request->forward_coeffs().data());
+      uInt32 num_forward_coeffs_in = request->num_forward_coeffs_in();
+      float64 min_val_x = request->min_val_x();
+      float64 max_val_x = request->max_val_x();
+      int32 num_points_to_compute = request->num_points_to_compute();
+      int32 reverse_poly_order = request->reverse_poly_order();
+      response->mutable_reverse_coeffs()->Resize((reverse_poly_order == -1) ? num_forward_coeffs_in : reverse_poly_order + 1, 0);
```
reckenro, line 89: Two thoughts:

1. Are we concerned with the client sending -1 for `reverse_poly_order`? I know this can happen with some other sizes passed in but this one _might_ be more likely due to -1 being a sentinel value.
2. Is it fine if the client doesn't set the `reverse_poly_order` field and by default gets 0 so it's sized as one here?

reckenro (approved):
I left two open questions but am marking as approved. I don't think they are of immediate concern and may be more general than just the API you're adding in this change.

generated/nidaqmx/nidaqmx_service.cpp
astarche: Good question. I'm going to change the condition to `< 0` and let the driver report whatever errors it wants to report. I'm OK with the default. There are a lot of cases where we get less-than-awesome defaults.

PR #269, opened 2021-07-21
Add support for timestamps

gregstoll:

Adds support for timestamps in the API, specifically conversions between the protobuf `Timestamp` class and `CVIAbsoluteTime`. I introduced a templated method that should make writing future conversions in C++ code easier (as opposed to in mako).

Some DAQ entry points return timestamps, so some sort of conversion is necessary.

Added unit tests, they all pass.


imports/include/niFakeNonIvi.h
```diff
@@ -1,9 +1,10 @@
 #ifndef NIFAKE_NON_IVI_H
 #define NIFAKE_NON_IVI_H
 #include <cstdint>
+// For CVIAbsoluteTime
```
gregstoll, line 4: I'm not thrilled about this since we were trying to avoid it, but we need the definition of CVIAbsoluteTime, and copy/pasting it in this header file meant I also had to pull in some other things like uInt64 and such. If you feel strongly about it I can go down that route again and see how bad it looks...

source/codegen/common_helpers.py
```diff
@@ -19,6 +19,13 @@ def is_enum(parameter):
 def is_struct(parameter):
   return parameter["type"].startswith("struct")
 
+
+def any_function_uses_timestamp(functions):
```
astarche, line 23: What do you think about replacing this with some stuff in config.py for each driver. So the codegen doesn't have the types coded in?

additional_headers: ["custom/nidaqmx_conversions.h"],
types_require_conversion: ["CVIAbsoluteTime"]

I like keeping `convert_to_grpc` as a standard convention for naming converters.

imports/include/niFakeNonIvi.h
astarche: I think either way is ok. I guess you could also have a variant of CVITime that uses uint64_t/etc.

source/codegen/common_helpers.py
astarche: Oh, actually. I like this logic for the `import "google/protobuf/timestamp.proto";` but not as much for the `#include "custom/nidaqmx_conversions.h"`

astarche (changes requested):
I'd like to see the change to move the specification to `config.py` or a counter-argument.

generated/nifake_non_ivi/nifake_non_ivi_service.cpp
```diff
@@ -292,5 +293,44 @@ namespace nifake_non_ivi_grpc {
...
+  {
+    if (context->IsCancelled()) {
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      CVIAbsoluteTime when {};
+      auto status = library_->OutputTimestamp(&when);
+      response->set_status(status);
+      if (status == 0) {
+        convert_to_grpc(when, response->mutable_when());
```
astarche, line 326: Great line of code!

astarche (approved):

reckenro (approved):

PR #270, opened 2021-07-21
Silence some warnings in gRPC code

gregstoll:

Silences some warnings about iterator base type in gRPC code. These warnings seem innocuous, aren't in our code, and are pretty noisy.

Unfortunately the warnings I really wanted to fix are covered in protocolbuffers/protobuf#7103.

Reduce noise when compiling.

Built, verified warnings are gone.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,7 +302,7 @@ add_executable(UnitTestsRunner
 # ni_fake_service_tests.cpp and several DAQ cpp files exceed the MSVC limit for the number of sections 
 # in an obj file defined by PE-COFF. This line disables the limit.
 if(MSVC)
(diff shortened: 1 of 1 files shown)


astarche (approved):
One of the comments on the issue that you linked suggested putting all of the pb generated files in a subdirectory and ignoring warnings at that scope. Do you think we should try that for squashing the other warnings?

gregstoll:
> 
> 
> One of the comments on the issue that you linked suggested putting all of the pb generated files in a subdirectory and ignoring warnings at that scope. Do you think we should try that for squashing the other warnings?

I think the generated files we want to ignore those warnings on are already in a subdirectory, but to ignore those warnings I think we'd have to split them out into a separate build (like a library or something). As much as the warnings annoy me I'm not sure it's worth that effort. But hopefully the upstream issue gets fixed soon...

PR #275, opened 2021-07-23
Async implementation of RegisterDoneEvent

astarche:

Replace the synchronous, thread-blocking implementation of `RegisterDoneEvent` with an async implementation.

This is implemented as a "hybrid" grpc server, which starts with a synchronous server and opts out specific methods using `WithAsyncMethod_`.

Allows grpc-device to implement callbacks without blocking a grpc thread for each call.

Ran and passed all tests. Some extra testing with python client.


source/server/async_operation.h
```diff
@@ -0,0 +1,68 @@
...
+#include <memory>
+
+#include "async_method_context.h"
+#include "async_processing_loop.h"
+
+using nidevice_grpc::SharedMethodContextPtr;
+
+namespace nidevice_grpc {
+// Calls writer.Finish and then releases the method_context_.
+template <typename TRequest, typename TResponse>
```
astarche, line 15: Note: these are templated because they own a templated `SharedMethodContextPtr<TRequest, TResponse>`. I can fix that with some type-erasing/base-class, which would allow moving these implementations out of the header. I (probably) plan to do that, but want to make sure I'm on the right track first.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,48 +1,97 @@
...
 #include <server/callback_router.h>
 
+#include <memory>
 #include <thread>
 
 namespace nidaqmx_grpc {
 
 using DoneEventCallbackRouter = nidevice_grpc::CallbackRouter<int32, TaskHandle, int32>;
 
+class DoneEventCallbackOperation : public nidevice_grpc::CompletionQueueElement {
```
astarche, line 13: This "self-owning" pattern is pretty common in demos of async grpc:

1. `new` up the class but don't keep the pointer
2. Register with the CompletionQueue using `this` as the tag.
3. The CompletionQueue can trivially get back to the element by casting it and calling `process`.
4. In `process` the element either calls `delete this` or goes back to step 2.

Most implementations use a state machine and repeat steps 2-4. I'm always deleting and potentially allocating a new queue element. The big downside there is that I'm hitting the heap more, although that could be optimized in a couple different ways (using a pool; allocating the related steps as a block).

The "rationale" for my version:
1. Makes it easier to separate the common `FinishOperation` logic in totally separate classes.
2. The "state machine" works best when you continually reuse the same tag. But `AsyncNotifyWhenDone` always needs to be a separate tag. So we'd need multiple elements even if we consolidated.
3. Though the `shared_ptr` ownership is a little overkill for the strictly-async-grpc aspect of this implementation. It's really helpful for allowing the `callback` to participate in ownership. Since that can be called literally at any time, we need an ownership model that supports that.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,48 +1,97 @@
...
 //---------------------------------------------------------------------
 //---------------------------------------------------------------------
 ::grpc::Status
-NiDAQmxService::RegisterDoneEvent(::grpc::ServerContext* context, const RegisterDoneEventRequest* request, ::grpc::ServerWriter<RegisterDoneEventResponse>* writer)
+NiDAQmxService::process_RegisterDoneEvent(const RegisterDoneEventMethodContextPtr& async_method_context)
 {
-  if (context->IsCancelled()) {
-    return ::grpc::Status::CANCELLED;
-  }
+  //if (context->IsCancelled()) {
```
astarche, line 56: TODO: might still need this.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,48 +1,97 @@
...
-      std::this_thread::sleep_for(std::chrono::milliseconds(10));
-    }
+    //while (!context->IsCancelled()) {
+    //  std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    //}
     return ::grpc::Status::OK;
   }
   catch (nidevice_grpc::LibraryLoadException& ex) {
-    return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
+    writer.Finish(::grpc::Status(::grpc::NOT_FOUND, ex.what()), nidevice_grpc::NO_TAG);
```
astarche, line 93: TODO: I think this is the right idea, but I'm not sure it works with the rest of my code.

source/server/async_operation.h
```diff
@@ -0,0 +1,68 @@
...
+  static void register_completion_queue_element(const SharedMethodContextPtr<TRequest, TResponse>& method_context)
+  {
+    new FinishCallbackOperation(method_context);
+  }
+
+  void process()
+  {
+    // Allow any driver code to start cleaning up and release it's references back to the
+    // method_context_ (i.e., to access writer.Write).
+    method_context_->registration.reset();
```
astarche, line 54: This logic around the "registration" is maybe the hardest to follow. It gets registered in the process_RegisterDoneEvent method. And then released here. I don't know if there's a way to make that clearer/more-symmetric other than spamming comments.

If this is a big problem for understandability, that would be an argument to have more of the logic in a single "call" object. Rather than having separate operation/elements operating on the same state.

source/server/async_processing_loop.h
```diff
@@ -0,0 +1,28 @@
...
+  virtual ~CompletionQueueElement() {}
+  virtual void process() = 0;
+};
+
+template <typename TService>
+void run_async_processing_loop(::grpc::ServerCompletionQueue* completion_queue, TService& service)
+{
+  void* tag;
+  bool ok;
+  service.register_async_functions(completion_queue);
```
astarche, line 20: This register call should go in the `DAQService` constructor. Then this wouldn't need to be a template. I plan to do this for this change , just putting off that kind of housekeeping until the direction is set.

source/server/callback_router.h
```diff
@@ -16,44 +18,36 @@ template <typename TReturn, typename... TArgs>
...
       return token_;
     }
 
    private:
-    Token token_;
+    AsyncOperationToken token_;
   };
 
-  static CallbackRegistration register_handler(const Handler& handler)
+  static std::unique_ptr<AsyncOperationRegistration> register_handler(const Handler& handler)
```
astarche, line 44: Note: Returning `unique_ptr` makes all the move stuff above unnecessary ^^^. Just delete `move` instead and allow `unique_ptr` to implement it.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -727,6 +727,31 @@ TEST_F(NiDAQmxDriverApiTests, ChannelWithDoneEventRegistered_RunCompleteFiniteAc
   EXPECT_EQ(DAQmxSuccess, response.status());
 }
 
+TEST_F(NiDAQmxDriverApiTests, ChannelWithDoneEventRegisteredTwice_RunCompleteFiniteAcquisition_DoneEventResponseIsReceived)
+{
+  create_ai_voltage_chan(0.0, 1.0);
+  ::grpc::ClientContext reader_context;
+  auto reader = register_done_event(reader_context);
+  reader->WaitForInitialMetadata();
```
astarche, line 735: An unfortunate side effect of the separate completion queue for async methods is that ordering of the initial call is not guaranteed. So, you can get intermittent "can't register events on task that is already started" errors. `WaitForInitialMetadata` on the client side seems like an effective workaround. I'm not sure if there's a way to address that in the server.

source/codegen/templates/service.h.mako
```diff
@@ -37,26 +43,45 @@ resource_repository_ptr = f"std::shared_ptr<{resource_repository_type}>"
...
 public:
   using ResourceRepositorySharedPtr = ${resource_repository_ptr};
 
   ${service_class_prefix}Service(${service_class_prefix}LibraryInterface* library, ResourceRepositorySharedPtr session_repository);
   virtual ~${service_class_prefix}Service();
   
+%if has_async_functions:
+  void register_async_functions(::grpc::ServerCompletionQueue* completion_queue);
+%endif
+% for function in async_functions.keys():
```
astarche, line 64: TODO: shouldn't just copy/paste this. Maybe I can just put the `using` statements next to the method.

source/codegen/service_helpers.py
```diff
@@ -188,14 +188,27 @@ def is_custom_close_method(function_data):
...
 def requires_checked_conversion(parameters):
     return any([is_input_array_that_needs_coercion(p) for p in parameters])
-  
+
+def get_request_type(method_name):
+    return f"{method_name}Request"
+
 
 def get_request_param(method_name, function_data):
     return f'const {method_name}Request* request'
```
gregstoll, line 197: seems like we could use `get_request_type()` here

source/codegen/service_helpers.py
```diff
@@ -188,14 +188,27 @@ def is_custom_close_method(function_data):
...
 
 
+def get_response_type(method_name):
+    return f'{method_name}Response'
+
 def get_response_param(method_name, function_data):
     is_streaming = common_helpers.has_streaming_response(function_data)
     response_type = f'{method_name}Response'
```
gregstoll, line 205: seems like we could use `get_response_type()` here

gregstoll:
Overall I think this seems like a reasonable direction, although I'll admit I don't follow 100% all of the code. But I don't see any red flags in particular.

gregstoll (approved):

source/codegen/templates/service.h.mako
```diff
@@ -18,6 +18,12 @@ if len(config["custom_types"]) > 0:
 resource_handle_type = config.get("resource_handle_type", "ViSession")
 resource_repository_type = f"nidevice_grpc::SessionResourceRepository<{resource_handle_type}>"
 resource_repository_ptr = f"std::shared_ptr<{resource_repository_type}>"
+
+async_functions = service_helpers.get_async_functions(functions)
+has_async_functions = any(async_functions)
+base_class_name = f"{service_class_prefix}::Service"
+for async_function in async_functions.keys():
+  base_class_name = f"{service_class_prefix}::WithAsyncMethod_{async_function}<{base_class_name}>"
```
reckenro, line 26: I couldn't find the `WithAsyncMethod` documentation but is it fine to just use the last `async_function` key here for the `base_class_name`? I know right now we only have `RegisterDoneEvent` but are we planning on adding any more async methods and will changing this be fine? Does it just have to be one of the async_function names?

reckenro (approved):
Similar to Greg. I didn't see anything alarming but again didn't 100% follow all of the code. I need to read up on async grpc and this "hybrid" type of approach.

source/codegen/templates/service.h.mako
astarche: The idea is that they're supposed to nest, so each time through the loop, we will wrap the previous result.

So, it should come out as 

`class DAQService: WithAsyncMethod_RegisterDoneEvent<WithAsyncMethod_RegisterEveryNSamplesEvent<Service>>

Here are [some examples](https://grep.app/search?current=3&q=WithAsyncMethod_%5BA-Za-z%5D%2B%3C&regexp=true). But there's definitely not much documentation on it.

reckenro: Oh, I just misread it. That makes sense.

source/tests/system/nidaqmx_driver_api_tests.cpp
astarche: Actually, I'm not sure that this is because of the separate completion queue. It could just be caused by using an async client.

source/server/async_processing_loop.h
astarche: Actually, that doesn't work. You can't do that before the service is registered/started.

Leaving as is.

source/custom/nifake_non_ivi_service.custom.cpp
```diff
@@ -1,8 +1,94 @@
 #include <nifake_non_ivi/nifake_non_ivi_service.h>
+#include <server/async_operation.h>
+#include <server/async_processing_loop.h>
 #include <server/callback_router.h>
 
+#include <memory>
+#include <thread>
 namespace nifake_non_ivi_grpc {
 
+class ReadStreamOperation : public nidevice_grpc::CompletionQueueElement {
```
astarche, line 10: Note: When I add the other 2 DAQ Register callback methods, I will implement codegen for these "operation" classses and `register_async_functions` but *not* `process_XYZ`.

So this block will be pulled out.

source/server/async_method_context.h
```diff
@@ -0,0 +1,83 @@
...
+namespace nidevice_grpc {
+using AsyncOperationToken = void*;
+// Type erasing abstract base class to allow managing driver-specific operation
+// registration in an AsyncMethodContext.
+struct AsyncOperationRegistration {
+  virtual ~AsyncOperationRegistration() {}
+  virtual AsyncOperationToken token() { return reinterpret_cast<AsyncOperationToken>(0); }
+};
+
+struct AsyncMethodContext {
```
astarche, line 19: Note: This baseclass/interface allows many of the operations to no longer be templatized. This means:
1. They can move to `async_operation.cpp`
2. They will no longer get unique instantiations for each method type.

Note that some operations (i.e., `WriteOperation`) still need `TResponse` or `TRequest` and will use `AsyncMethodContextT` below.

source/server/async_server_write_queue.h
```diff
@@ -0,0 +1,58 @@
+#ifndef NIDEVICE_GRPC_DEVICE_ASYNC_SERVER_WRITE_QUEUE_H
+#define NIDEVICE_GRPC_DEVICE_ASYNC_SERVER_WRITE_QUEUE_H
+#include <grpcpp/grpcpp.h>
+
+#include <deque>
+#include <mutex>
+#include <queue>
+
+namespace nidevice_grpc {
```
astarche, line 9: Note: I had to add this because the fake test I wrote would crash consistently by queueing up multiple writes. This will also be an issue with other callbacks (but not RegisterDone, which only happens once).

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <chrono>
+#include <memory>
+namespace ni {
+namespace tests {
+namespace integration {
+
+using namespace nifake_non_ivi_grpc;
+
+// Test NiFakeNonIviService using a client stub.
```
astarche, line 14: The async implementation is so dependent on grpc infrastructure (completion queues, async writers, etc) that it doesn't make sense to try to test it in the "normal" fake unit test style.

Instead, this is more like the system tests, but it's using a fake service implementation.

The other advantage is that this doesn't depend too much on internals. That's especially useful right now because I'm not sure this is a final design.

source/server/core_server.cpp
```diff
@@ -171,9 +172,11 @@ static void RunServer(const ServerConfiguration& config)
   nidevice_grpc::logging::log(
       nidevice_grpc::logging::Level_Info,
       "Security is configured with %s%s.", security_description, tls_description);
+
+  nidevice_grpc::run_async_processing_loop(completion_queue.get(), nidaqmx_service);
   // This call will block until another thread shuts down the server.
   server->Wait();
-  // This code is currently unreachable, but if the call to wait exits, we need to clean up the service here.
```
astarche, line 176: TODO: still need to test this in the daemonized RT shutdown case.

source/server/async_operation.h
astarche: I think this is somewhat clearer now. And I may refine it some when I take another pass at codegen-ing some of the operation boiler plate.

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,81 @@
...
+    return server;
+  }
+
+  std::unique_ptr<NiFakeNonIvi::Stub>& stub()
+  {
+    return stub_;
+  }
+};
+
+TEST_F(NiFakeNonIviServiceTests_EndToEnd, ReadStream_ReturnsExpectedSequence)
```
astarche, line 61: TODO: Oy! This test passes locally but fails a grpc assert when running the workflow. I need to figure out what that's about. I think the other integration tests have the same dependencies (i.e., "I need to run an in process server; Don't need drivers"). But maybe there's something wrong with my setup code.

PR #278, opened 2021-07-28
Reimplement DAQmx callbacks with experimental callback service

astarche:

Reimplement callbacks with experimental callback service.

The first pass implementation of DAQmx callbacks used synchronous streams. This has the problem of blocking a thread while the callback is registered and requiring polling to check for cancellation.

I attempted to use the traditional grpc async service API in #275. This API is difficult to use and my implementation had persistent race condition/quality issues.

Grpc recently "de-experimentalized" the async callback/reactor API for async services. This is much easier to implement than the previous API and a simple implementation is much more reliable than my previous attempt.

[Callback API proposal](https://github.com/grpc/proposal/blob/master/L67-cpp-callback-api.md)
[Callback example](https://github.com/grpc/grpc/blob/master/examples/cpp/route_guide/route_guide_callback_server.cc)

Improve behavior and performance of `RegisterDoneEvent` callback.
Unblocks additional callbacks.
Adds integration testing for some of the async server glue code.

Ran and passed all DAQ tests.
Ran and passed new tests 500+ times.

### DIFF
--- a/generated/nidaqmx/nidaqmx_service.h
+++ b/generated/nidaqmx/nidaqmx_service.h
@@ -21,7 +21,7 @@
 
 namespace nidaqmx_grpc {
 
-class NiDAQmxService final : public NiDAQmx::Service {
+class NiDAQmxService final : public NiDAQmx::ExperimentalWithCallbackMethod_RegisterDoneEvent<NiDAQmx::Service> {
 public:
   using ResourceRepositorySharedPtr = std::shared_ptr<nidevice_grpc::SessionResourceRepository<TaskHandle>>;
 
@@ -187,7 +187,7 @@ public:
   ::grpc::Status ReadDigitalU32(::grpc::ServerContext* context, const ReadDigitalU32Request* request, ReadDigitalU32Response* response) override;
   ::grpc::Status ReadDigitalU8(::grpc::ServerContext* context, const ReadDigitalU8Request* request, ReadDigitalU8Response* response) override;
   ::grpc::Status ReadRaw(::grpc::ServerContext* context, const ReadRawRequest* request, ReadRawResponse* response) override;
-  ::grpc::Status RegisterDoneEvent(::grpc::ServerContext* context, const RegisterDoneEventRequest* request, ::grpc::ServerWriter<RegisterDoneEventResponse>* writer) override;
+  ::grpc::experimental::ServerWriteReactor<RegisterDoneEventResponse>* RegisterDoneEvent(::grpc::CallbackServerContext* context, const RegisterDoneEventRequest* request) override;
   ::grpc::Status ReserveNetworkDevice(::grpc::ServerContext* context, const ReserveNetworkDeviceRequest* request, ReserveNetworkDeviceResponse* response) override;
   ::grpc::Status ResetDevice(::grpc::ServerContext* context, const ResetDeviceRequest* request, ResetDeviceResponse* response) override;
   ::grpc::Status SelfTestDevice(::grpc::ServerContext* context, const SelfTestDeviceRequest* request, SelfTestDeviceResponse* response) override;
--- a/source/server/async_operation.h
+++ b/source/server/async_operation.h
@@ -39,7 +39,7 @@ class ServerWriterReactor : public grpc::ServerWriteReactor<TResponse> {
   {
     LockGuard lock(mutex_);
     done_ = true;
-    Finish(::grpc::Status::OK);
+    this->Finish(::grpc::Status::OK);
   }
 
   void queue_write(const TResponse& response)
@@ -57,7 +57,7 @@ class ServerWriterReactor : public grpc::ServerWriteReactor<TResponse> {
     if (!done_ && write_ready_ && !pending_send_.empty()) {
       write_ready_ = false;
       auto& response = pending_send_.front();
-      StartWrite(&response);
+      this->StartWrite(&response);
     }
   }
 
--- a/source/server/async_operation.h
+++ b/source/server/async_operation.h
@@ -16,23 +16,29 @@ class ServerWriterReactor : public grpc::ServerWriteReactor<TResponse> {
   using LockGuard = std::lock_guard<std::recursive_mutex>;
 
  public:
-  virtual ~ServerWriterReactor() {}
+  virtual ~ServerWriterReactor()
+  {
+  }
 
   void OnDone() override
   {
     delete this;
   }
 
-  void OnWriteDone(bool /*ok*/) override
+  void OnWriteDone(bool ok) override
   {
     LockGuard lock(mutex_);
-    write_ready = true;
+    write_ready_ = true;
     pending_send_.pop();
-    next_write();
+    if (ok) {
+      next_write();
+    }
   }
 
   void OnCancel() override
   {
+    LockGuard lock(mutex_);
+    done_ = true;
     Finish(::grpc::Status::OK);
   }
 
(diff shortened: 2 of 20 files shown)


source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,235 @@
...
+  auto reader = stub()->ReadStream(&context, request);
+
+  ReadStreamResponse response;
+  for (auto i = START; i <= STOP / 2; ++i) {
+    reader->Read(&response);
+    EXPECT_EQ(response.value(), i);
+  }
+}
+
+TEST_F(NiFakeNonIviServiceTests_EndToEnd, ReadTwoStreamsAlternating_ReturnsExpectedSequences)
```
gregstoll, line 97: Instead of "Alternating", could we use "Consecutively" or something? I was confused when we didn't go back and forth between `reader` and `second_reader` like the "Interleaved" test below.

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,235 @@
...
+    EXPECT_EQ(response.value(), i + OFFSET);
+  }
+}
+
+TEST_F(NiFakeNonIviServiceTests_EndToEnd, StartReadStream_ShutdownBeforeRead_CleanShutdown)
+{
+  ::grpc::ClientContext context;
+  const auto START = 0;
+  const auto STOP = 10;
+  const auto OFFSET = 100;
```
gregstoll, line 192: `OFFSET` is unused, can delete it.

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  reader.reset();
+}
+
+TEST_F(NiFakeNonIviServiceTests_EndToEnd, WaitForMetadata_Cancel_CleanShutdown)
+{
+  ::grpc::ClientContext context;
+  const auto START = 0;
+  const auto STOP = 10;
+  const auto OFFSET = 100;
```
gregstoll, line 206: `OFFSET` is unused, can delete it.

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  reader.reset();
+}
+
+TEST_F(NiFakeNonIviServiceTests_EndToEnd, WaitForMetadata_ShutdownAndRead_CleanShutdown)
+{
+  ::grpc::ClientContext context;
+  const auto START = 0;
+  const auto STOP = 10;
+  const auto OFFSET = 100;
```
gregstoll, line 221: `OFFSET` is unused, can delete it.

gregstoll (approved):

generated/nidaqmx/nidaqmx_service.h
```diff
@@ -21,7 +21,7 @@
 
 namespace nidaqmx_grpc {
 
-class NiDAQmxService final : public NiDAQmx::Service {
+class NiDAQmxService final : public NiDAQmx::ExperimentalWithCallbackMethod_RegisterDoneEvent<NiDAQmx::Service> {
```
astarche, line 24: Note: These have been "de-experimentalized" in the latest grpc release. I've started #281 which will pull this in. But I would like to resolve the [AMD build issues](https://bugs.chromium.org/p/boringssl/issues/detail?id=423) first.

source/server/async_operation.h
```diff
@@ -0,0 +1,74 @@
...
+#include <grpcpp/grpcpp.h>
+
+#include <deque>
+#include <memory>
+#include <mutex>
+#include <queue>
+
+namespace nidevice_grpc {
+
+template <typename TResponse, typename TJoinable>
```
astarche, line 14: TODO: I'm using "thread"/TJoinable in this api. But the in-product implementation uses a "callback registration". I was thinking that I might use a `join` method to ensure that it was detached, but that's just happening in the destructor (hence `AutoJoinThread`).

I think `producer` or `registration` would be better terms.

source/server/async_operation.h
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  std::recursive_mutex mutex_;
+  // This class relies on: std::deque does not invalidate on push/pop.
+  std::queue<TResponse, std::deque<TResponse>> pending_send_;
+  bool write_ready_{true};
+  bool done_{false};
+
+ public:
```
astarche, line 70: TODO: make this private with a setter.

Document that the position of this is important. It's last in the member list because it must be destroyed *first*. Otherwise the producer may still try to `queue_write` after `mutex_` etc have been destroyed.

source/server/server_reactor.h
```diff
@@ -0,0 +1,79 @@
...
+  virtual ~ServerWriterReactor()
+  {
+  }
+
+  void OnDone() override
+  {
+    delete this;
+  }
+
+  void OnWriteDone(bool ok) override
```
reckenro, line 29: Is it possible to update this argument name or does the name need to keep in sync with the `grpc::ServerWriteReactor`? If so, something a little more telling might be nice like `continue` or verbosely `start_next_write`.

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
```diff
@@ -0,0 +1,232 @@
...
+  ReadStreamRequest request;
+  request.set_start(START);
+  request.set_stop(STOP);
+  auto reader = stub()->ReadStream(&context, request);
+  request.set_start(START + OFFSET);
+  request.set_stop(STOP + OFFSET);
+  ::grpc::ClientContext second_context;
+  auto second_reader = stub()->ReadStream(&second_context, request);
+  reader->WaitForInitialMetadata();
```
reckenro, line 111: Do we need to call `WaitForInitialMetadata`? I don't see us explicitly sending initial metadata from this ni fake service handler unless it's done behind the scenes somewhere. And if it is needed, for things with two stream do we need to call `WaitForInitialMetadata` for both readers?

reckenro (approved):

SparkingSpork (approved):

source/tests/integration/ni_fake_non_ivi_service_tests_endtoend.cpp
astarche: DAQ cares about `WaitForInitialMetadata` because it requires that you set register before calling `DAQmxStartTask`. This test doesn't have an ordering dependency like that. Also, because it starts updating immediately, it doesn't really need a SendInitialMetadata because the first write will do that.

I focused more on stress testing the stream/lifetime code than trying to copy the DAQ case.

reckenro: Yeah, I understood it for the DAQ case and was curious if we needed to care/know about it for these tests. That makes sense though and is more clear. It seems fine as is. I'll mark this resolved.

source/server/server_reactor.h
astarche: I think `write_succeeded` could be good.

I'm not sure how important it is to match. `ok` is used on the grpc "completion queue" API and overloaded to mean a bunch of different things (depending on what operation you're dequeuing on) that are documented in a sort of intricate way throughout the API. In that context, it's important because the meaning of the variable emerges from all that stuff.

This is definitely inherited from all of that. But in the new API the status is scoped to the specific operation so it seems reasonable to assign it a specific meaning. Having the tie-in to the old API was helpful to me for understanding it, but I'm not sure if that will help anyone else. Hopefully no one else will have to mess with completion queues, though I'm sure we'll run into grpc's `ok` params elsewhere.

PR #280, opened 2021-07-29
Clean up exceptions in DAQ betterproto examples

astarche:

Fix DAQ `betterproto` examples to print better exceptions, especially in common error cases.

Note that `betterproto` does not report `GRPCError`s for connection failures. Those are reported as `OSError` or `TimeouError`, which do not provide much structure.

Sample output:

Connection errors (different cases):
```
> python digital-input-betterproto.py localhost 31763 "gRPCSystemTestDAQ/port0"                                                                                                                                                               
Multiple exceptions: [Errno 10061] Connect call failed ('::1', 31763, 0, 0), [Errno 10061] Connect call failed ('127.0.0.1', 31763)

> python digital-input-betterproto.py 0.0.0.0 31763 "gRPCSystemTestDAQ/port0"                                                                                                                                                                 
[Errno 10049] Connect call failed ('0.0.0.0', 31763)

> python digital-input-betterproto.py 10.0.62.17 31763 "gRPCSystemTestDAQ/port0"
[Errno 10060] Connect call failed ('10.0.62.17', 31763)
```

UNIMPLEMENTED error:
```
> python digital-input-betterproto.py 10.0.62.17 31763 "gRPCSystemTestDAQ/port0"
The operation is not implemented or is not supported/enabled in this service
```

Other GRPC error:
```
> python digital-input-betterproto.py 10.0.62.17 31763 "gRPCSystemTestDAQ/port0"
GRPCError: (<Status.SOMETHINGELSE: 12345>, None, None)
```

Error from DAQ/`raise_if_error`:
```
> python digital-input-betterproto.py 10.0.62.17 31763 "gRPCSystemTestDAQ/not-a-port"
Error: Device identifier is invalid.
```

Simple implementation to provide reasonable errors for betterproto examples.

Ran examples in various success and failure cases.

### DIFF
--- a/examples/nidaqmx/analog-input-betterproto.py
+++ b/examples/nidaqmx/analog-input-betterproto.py
@@ -30,6 +30,7 @@ import asyncio
 from nidevice import nidaqmx_grpc
 import sys
 from grpclib.client import Channel
+from grpclib.exceptions import GRPCError
 
 
 server_address = "localhost"
@@ -98,6 +99,13 @@ async def main():
 
         print(f"Acquired {response.samps_per_chan_read} samples.")
         print(f"First 5 samples: {response.read_array[:5]}")
+    except GRPCError as e:
+        if e.status.name == "UNIMPLEMENTED":
+            print("The operation is not implemented or is not supported/enabled in this service")
+        else:
+            print(f"GRPCError: {str(e)}")
+    except Exception as e:
+        print(str(e))
     finally:
         if task:
             await daq_service.stop_task(task=task)
--- a/examples/nidaqmx/digital-input-betterproto.py
+++ b/examples/nidaqmx/digital-input-betterproto.py
@@ -30,6 +30,7 @@ import asyncio
 from nidevice import nidaqmx_grpc
 import sys
 from grpclib.client import Channel
+from grpclib.exceptions import GRPCError
 
 server_address = "localhost"
 server_port = "31763"
@@ -76,6 +77,13 @@ async def main():
         await raise_if_error(response)
 
         print(f"Data acquired: {hex(response.read_array[0])}")
+    except GRPCError as e:
+        if e.status.name == "UNIMPLEMENTED":
+            print("The operation is not implemented or is not supported/enabled in this service")
+        else:
+            print(f"GRPCError: {str(e)}")
+    except Exception as e:
+        print(str(e))
     finally:
         if task:
             await daq_service.stop_task(task=task)


examples/nidaqmx/analog-input-betterproto.py
```diff
@@ -98,6 +99,13 @@ async def raise_if_error(response):
 
         print(f"Acquired {response.samps_per_chan_read} samples.")
         print(f"First 5 samples: {response.read_array[:5]}")
+    except GRPCError as e:
```
astarche, line 102: There's a different implementation of this for the other [betterproto ](https://github.com/ni/grpc-device/blob/0771571470857a699b2562fafa0479fdabe31ee0/examples/niscope/graph-measurement-betterproto.py#L200) examples.

But that doesn't work for me at all (either when I copied the code or running those examples directly). It requires an exception with at least 2 args (the `OSError` and `TimeoutError` that I got for connection failures only had 1) and it looks for a code `22` (where does that come from? I didn't see it). 

@ni-pkhare do you know what's going on with that? Did that logic work for you?

If those are really broken, I can apply the same fix to all of them.

We could also consider an alternative of always just printing `str(e)`. That's the best for readability of the example. But the `UNIMPLEMENTED` error specifically is harder to make sense of (though *very* standard as a grpc error).

gregstoll (approved):

examples/nidaqmx/analog-input-betterproto.py
ni-pkhare: @astarche I am not sure why it did not work for you, because I tried it today as well and it seems to be working for me.
1. err code 22 is for the case when we run the example without running the server. When I debugged for this case, I saw err code 22, thats why I used it so that we could display a proper message.
image

2. I am also a fan of simply printing the error message using str(e). Also the way you have written for Unimplemented case and other cases in general is quite nice. I would go for it.

harsha-bhushan (approved):

examples/nidaqmx/analog-input-betterproto.py
astarche: Interesting!

Thanks for the screenshot. I don't have a complete explanation for that. But, I think that's just another example to go with my 3 slightly different failure modes in the PR description. Error reporting for os-level connection failures is just harder to filter on than the grpc-errors that the standard protoc generator gives you. My implementation is a decent way to deal with that for an example. So, I'll got with this proposal and I created #282 to fix this up.

PR #283, opened 2021-07-30
Add DAQmx Advanced functions to grpc-device

astarche:

Adds [DAQmx Advanced functions](https://zone.ni.com/reference/en-XX/help/370471AM-01/TOC27.htm) to grpc-device.

Note: Calibration is limited to SelfCal.

Adds support for advanced DAQmx features to grpc-device.

Ran and passed all DAQ tests.
Some python test spot checks (including adding a real daq network device).

### DIFF
--- a/generated/nidaqmx/nidaqmx.proto
+++ b/generated/nidaqmx/nidaqmx.proto
@@ -17,8 +17,11 @@ import "session.proto";
 import "google/protobuf/timestamp.proto";
 
 service NiDAQmx {
+  rpc AddCDAQSyncConnection(AddCDAQSyncConnectionRequest) returns (AddCDAQSyncConnectionResponse);
   rpc AddGlobalChansToTask(AddGlobalChansToTaskRequest) returns (AddGlobalChansToTaskResponse);
   rpc AddNetworkDevice(AddNetworkDeviceRequest) returns (AddNetworkDeviceResponse);
+  rpc AreConfiguredCDAQSyncPortsDisconnected(AreConfiguredCDAQSyncPortsDisconnectedRequest) returns (AreConfiguredCDAQSyncPortsDisconnectedResponse);
+  rpc AutoConfigureCDAQSyncConnections(AutoConfigureCDAQSyncConnectionsRequest) returns (AutoConfigureCDAQSyncConnectionsResponse);
   rpc CalculateReversePolyCoeff(CalculateReversePolyCoeffRequest) returns (CalculateReversePolyCoeffResponse);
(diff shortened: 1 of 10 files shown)


generated/nidaqmx/nidaqmx.proto
```diff
@@ -3241,6 +3319,39 @@ message DeleteNetworkDeviceResponse {
...
+message DeleteSavedScaleRequest {
+  string scale_name = 1;
+}
+
+message DeleteSavedScaleResponse {
+  int32 status = 1;
+}
+
+message DeleteSavedTaskRequest {
+  string task_name = 1;
```
reckenro, line 3339: I would've expected the type on the request to mirror `SaveTaskRequest`: that being a `nidevice_grpc.Session` type to represent the task by name or id.

source/codegen/metadata/nidaqmx/enums.py
```diff
@@ -2075,5 +2091,30 @@
                 'value': 10208
             }
         ]
+    },
+    'WriteBasicTEDSOptions': {
+        'values': [
+            {
+                'documentation': {
```
reckenro, line 2098: You can delete the 'blah' documentation from these.

generated/nidaqmx/nidaqmx.proto
reckenro: Nevermind, I see now the SaveTask has a save_as field which I'm guessing what's referenced here. Makes sense.

reckenro (approved):

source/codegen/metadata/nidaqmx/enums.py
astarche: Aw but it's generated!

Yeah, I can fix the tool to filter out any "blah" documentation.

reckenro: Oh yeah, sorry. Forgot you were generating these. Thanks for cleaning it up though!

As a side note, I wonder if we'll ever want to use the documentation in the grpc metadata. Right now I believe we just discard it but we could put it in the generated proto files next to fields. I don't believe proto comments are carried over to generated code but there still might be some value having them in the proto files at least.

astarche: Yeah, that was borrowed from the original hapigen extraction scripts (which pull it from CVI metadata). Yeah, maybe that's something we could try out at some point. But we'd need to look both at the quality of the result (how is it presented?) and the quality of the documentation itself.

The best-case upside could be pretty significant. Certainly it would be possible for a protoc tool to pull comments into doc strings. But, the actual behavior could be ... blah.

reckenro: Yeah, true. The value to noise ratio would be interesting. Anyway, marking this thread resolved.

PR #285, opened 2021-08-03
Add support for remaining DAQmx callbacks

astarche:

Add remaining callback methods to DAQ service:
* DAQmxRegisterSignalEvent
* DAQmxRegisterEveryNSamplesEvent

Implement codegen for async callbacks to:
* Allow easier changes in the future.
* (Semi) Automatically pick up additional feature (i.e., enums).

Update fake non-ivi to use codegen'd callback method instead of hardcoded synchronous version.

Add spotcheck tests.

Adds all the methods for the DAQ callback user story.

Ran and passed all DAQ tests and all included tests.
Some `--gtest_repeat=N` stress testing.


source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -7283,14 +7282,133 @@
...
+            },
+            {
+                'direction': 'in',
+                'name': 'nSamples',
+                'type': 'uInt32'
+            },
+            {
+                'direction': 'in',
+                'name': 'options',
+                'type': 'uInt32'
```
gregstoll, line 7334: Should we make this an enum? (Does gRPC support bitflag-style enums?)

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -7283,14 +7282,133 @@
...
+            {
+                'direction': 'in',
+                'enum': 'Signal2',
+                'name': 'signalID',
+                'type': 'int32'
+            },
+            {
+                'direction': 'in',
+                'name': 'options',
+                'type': 'uInt32'
```
gregstoll, line 7388: (same question about enum)

gregstoll (approved):

source/codegen/templates/service_helpers.mako
```diff
@@ -112,20 +174,21 @@ ${set_response_values(output_parameters=output_parameters)}\
   input_parameters = [p for p in parameters if common_helpers.is_input_parameter(p)]
 %>\
 % for parameter in input_parameters:
-${initialize_input_param(function_name, parameter)}
+${initialize_input_param(function_name, parameter)}\
```
astarche, line 177: Note: basically I inverted the placement of the newline-eating `\` so that no newlines are placed by the for loop but newlines *are* placed by any `initialize_whatever_param` below.

This allows my new `## pass` to no-op-generate instead of generating a newline.

source/codegen/templates/service_helpers.mako
```diff
@@ -85,6 +85,68 @@ ${set_response_values(output_parameters)}\
...
+    }
+
+    ::grpc::Status start(const ${request_type}* request, ${driver_library_interface}* library, const ResourceRepositorySharedPtr& session_repository_)
+    {
+      try {
+        auto handler = CallbackRouter::register_handler(
+          [this](${service_helpers.create_args_for_callback(callback_parameters)}) {
+            ${response_type} callback_response;
+            auto response = &callback_response;
+<%block filter="common_helpers.indent(2)">\
```
astarche, line 115: Note: 
A "filter" in mako does some sort of transform on text (in this case, a block of text). Since these blocks are reusing helpers with a different indentation level, this pushes them out 1-2 more levels to align correctly.

`mako` has a nicer `${val | format}` filter syntax, but I can't get it to work with functions.

source/codegen/templates/service_helpers.mako
```diff
@@ -85,6 +85,68 @@ ${set_response_values(output_parameters)}\
...
+  response_type = service_helpers.get_response_type(function_name)
+  driver_library_interface = common_helpers.get_library_interface_type_name(config)
+%>\
+    using CallbackRouter = nidevice_grpc::CallbackRouter<int32, ${service_helpers.create_param_type_list(callback_parameters)}>;
+    class ${function_name}Reactor : public nidevice_grpc::ServerWriterReactor<${response_type}, nidevice_grpc::CallbackRegistration> {
+    public:
+    ${function_name}Reactor(const ${request_type}& request, ${driver_library_interface}* library, const ResourceRepositorySharedPtr& session_repository)
+    {
+      auto status = start(&request, library, session_repository);
+      if (!status.ok()) {
```
astarche, line 103: Note: the control flow of start has to change to reuse the enum initialization methods that include an early return. So now, rather than returning the `unique_ptr` to the producer and conditionally calling `Finish` internally; `start` will call `set_producer` internally (on success) and return a status.

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -7283,14 +7282,133 @@
             },
             {
                 'callback_params': [
+                    {
```
astarche, line 7285: `RegisterDoneEvent` has two interesting callback params:

`TaskHandle`: typically not part of a response message. We're tracking context through the output stream, so there's no reason to include it in the response. So: `include_in_proto: False`

'status': We auto-add `status` to responses and assume that it is the output of the method. That's true for `Register` itself (which can fail, in which case we early return). But when you're called as a callback, you don't automatically have a `status` output to use. `RegisterDoneEvent` *does* have a status, which we'll use. The other ones *don't* but still have a `default_status_param`.

source/codegen/metadata_mutation.py
```diff
@@ -14,7 +14,7 @@
     'object', 'operator', 'out', 'override',
     'params', 'private', 'protected', 'public',
     'readonly', 'ref', 'return',
-    'sbyte', 'sealed', 'short', 'sizeof', 'stackalloc', 'static', 'status', 'string', 'struct', 'switch',
```
astarche, line 17: Note: I'm not sure how `status` got on the exclude list. If we did have status as an input or extra output, we would need to special case it, though I'm not sure if how we're doing it is right. For this case, `RegisterDoneEvent` has a status callback param/output that I actually (a) do want to be called `status` and share with the default `status` response param and (b) I need to know it's there.

So I removed it from the list. Nothing else changed. If we hit other status params, they may need other handling. The default behavior is that they'll have a compile error for a local name conflict.

source/codegen/proto_helpers.py
```diff
@@ -8,11 +8,6 @@ def should_allow_alias(enums):
       return True
   return False
 
-def filter_parameters_for_grpc_fields(parameters):
```
astarche, line 11: Note: move to common_helpers

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,62 +1,4 @@
 #include <nidaqmx/nidaqmx_service.h>
-#include <server/callback_router.h>
```
astarche, line 2: Note: Delete custom implementation. This is now generated.

source/codegen/metadata/nidaqmx/functions.py
astarche: Great question!

`protobuf` doesn't support bitflags exactly. My general solution for bitflags is (a) make it an enum, (b) if you want to set multiple flags you can use the `set_X_raw` method and | in the 2 enum values (in most languages!) .

[Example](https://github.com/ni/grpc-device/blob/3d674d75127842a08e0d5bbaf991e1e755e97a6b/source/tests/system/nidaqmx_driver_api_tests.cpp#L535)

This only has one flag and 2 possible values, so even the caveats aren't really an issue.

**BUT** the flag is to allow calling the callback on the registering thread, which requires pumping messages on the registering thread. I don't have a way to implement that in our grpc server and I don't think we should mess with it. SO the fix for this is actually to mark them `include_in_proto: False` and hardcode the value to zero. I have that tracked as AB#{1560237}.

generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -8355,6 +8357,200 @@ namespace nidaqmx_grpc {
...
+      catch (nidevice_grpc::LibraryLoadException& ex) {
+         return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
+      }
+
+      return ::grpc::Status::OK;
+    }
+    };
+
+    return new RegisterDoneEventReactor(*request, library_, session_repository_);
+  }
```
astarche, line 8412: TODO: add spacers

```
```

source/codegen/metadata_mutation.py
reckenro: Yeah, that sounds fine to me. I'm not sure how `status` got on the list. I'm guessing its from being overly cautious in our initial prototype. 

What you're doing and plans for future `status` params sounds reasonable, though, if they pop up.

reckenro (approved):

PR #290, opened 2021-08-09
Add support for input varargs functions

gregstoll:

Adds support for functions that take varargs as input parameters. (output parameters will be coming in another PR, but I want to get this in first to make sure it's on the right track) To test the behavior where there are too many parameters in a function, I added one real DAQmx function.

We need to support DAQmx methods that take varargs.

Added unit tests in `nifake_non_ivi`.


source/codegen/service_helpers.py
```diff
@@ -110,7 +127,26 @@ def get_array_param_size(parameter) -> str:
...
-def create_param(parameter):
+def expand_varargs_parameters(parameters):
+    if not common_helpers.has_varargs_parameter(parameters):
+        return parameters
+    # omit the varargs parameter that we're going to expand
+    new_parameters = parameters[:-1]
+    varargs_parameter = parameters[-1]
+    assert common_helpers.is_varargs_parameter(varargs_parameter)
+    max_length = varargs_parameter['max_length']
+    # Many (all?) functions that take varargs take the first set of parameter as
```
astarche, line 138: I think the way you're talking about `var_args` is a little too specific. But it might not matter. The other usage of `var_args` in the DAQ API are for the generic get/set attribute methods, which don't have this convention. But we're going to handle that by expanding the metadata to fill in specific types for the args.

I'd been calling these "more channels and states" methods, which is not that good a name either.

I think it's fine to mostly keep your code as-is. And comments like this can clarify that we're implementing a specific var_args convention where it's an arbitrary repetition of fixed args. And it would probably also work  an arbitrary repetition of args that aren't in the fixed param list (with changes; but we haven't implemented it).

I think it's also reasonable to think of what you're doing as a more specific convention `repeating_var_args` and potentially use that as structure for how it's represented in the metadata. Like the fixed-args could have a tag because they participate in that specific convention. If that's helpful.

source/codegen/common_helpers.py
```diff
@@ -181,6 +191,25 @@ def is_ivi_dance_array_param(parameter):
...
+
+def has_varargs_parameter(parameters):
+  return any(is_varargs_parameter(p) for p in parameters)
+
+# Google Mock can't handle the case where a function has a variable number of arguments
+# and there's also a limit on the max number of arguments.
+def can_mock_function(parameters):
+  # I'm not sure this is exactly right, but it does enough to distinguish between
+  # non-varargs functions and varargs functions that take > 100 parameters.
+  MAX_PARAM_LEN = 20
```
astarche, line 203: MAX_MOCK_PARAM_LEN?

source/codegen/templates/service.cpp.mako
```diff
@@ -12,6 +12,8 @@ if len(config["custom_types"]) > 0:
   custom_types = config["custom_types"]
 (input_custom_types, output_custom_types) = common_helpers.get_input_and_output_custom_types(functions)
 has_async_functions = any(service_helpers.get_async_functions(functions))
+function_names = service_helpers.filter_proto_rpc_functions_to_generate(functions)
+any_non_mockable_functions = any([not common_helpers.can_mock_function(functions[name]['parameters']) for name in function_names])
 %>\
```
astarche, line 17: I think this needs a comment. You could also potentially grab the library name here to link that requirement (though I'm not sure that's all that clear?).

source/codegen/templates/service_helpers.mako
```diff
@@ -187,11 +209,55 @@ ${initialize_enum_input_param(function_name, parameter)}
...
+% for member in parameter['varargs_type']['fields']:
+<%
+  member_c_type = member['type']
+  member_c_type_pointer = member_c_type.replace('[]','*')
+  # In gRPC fields, the names of the fields in the struct are lowercase
+  member_name = member['name'].lower()
+%>\
+      auto get_${member['name']}_if = [](const google::protobuf::RepeatedPtrField<${stripped_grpc_type}>& vector, int n) -> ${member_c_type_pointer} {
+            if (vector.size() > n) {
+% if common_helpers.is_string_arg(member):
```
astarche, line 246: Since these methods don't use the standard initiailze parameter methods, they're not going to work for more interesting types (enums, timestamps,). I think that's fine, and I'm not sure how easy that would be to fix. Just feeling out the boundaries of what the "varargs" convention supports.

astarche (approved):

source/codegen/service_helpers.py
gregstoll: Yeah, I like this. I moved to the "repeated_var_args" convention and added a "repeating_var_arg" tag for the first copy of each argument.

source/codegen/common_helpers.py
```diff
@@ -1,244 +1,321 @@
...
+# and there's also a limit on the max number of arguments.
+
+
+def can_mock_function(parameters):
+    # I'm not sure this is exactly right, but it does enough to distinguish between
+    # non-varargs functions and varargs functions that take > 100 parameters.
+    MAX_MOCK_PARAM_LEN = 20
+    repeated_varargs_parameters = [
+        p for p in parameters if is_repeated_varargs_parameter(p)]
+    first_repating_parameters = len(
```
reckenro, line 241: [Minor] Typo: 
```suggestion
    first_repeating_parameters = len(
```

source/codegen/templates/library.cpp.mako
```diff
@@ -52,13 +53,14 @@ ${service_class_prefix}Library::~${service_class_prefix}Library()
...
 
-% for method_name in service_helpers.filter_api_functions(functions):
+% for method_name in service_helpers.filter_api_functions(functions, only_mockable_functions=False):
 <%
   f = functions[method_name]
   parameters = f['parameters']
   return_type = f['returns']
-  parameter_list = service_helpers.create_params(parameters)
-  argument_list = ', '.join(p['cppName'] for p in parameters)
+  #can_mock = common_helpers.can_mock_function(parameters)
```
reckenro, line 61: I think you can delete this commented out line.

reckenro (approved):

PR #291, opened 2021-08-10
Add Int32 and Double DAQmx Scale attributes

astarche:

Add `Int32` and `Double` DAQmx Scale attributes.

This is the first use of attributes for non-MI types and the for a driver that has multiple "groups" of attributes. Updating the underlying codegen to support that.

De-risking the attribute API to get a first pass and some feedback,.

Ran and passed all DAQ tests.


generated/nidaqmx/nidaqmx.proto
```diff
@@ -238,8 +242,22 @@ service NiDAQmx {
   rpc WriteToTEDSFromFile(WriteToTEDSFromFileRequest) returns (WriteToTEDSFromFileResponse);
 }
 
-enum NiDAQmxAttributes {
-  NIDAQMX_ATTRIBUTE_UNSPECIFIED = 0;
+enum ScaleAttributes {
```
astarche, line 245: I realized from this work that I misunderstood how attributes work. And, particularly, the differences from enums.

I am planning to add the following features (only for DAQ, for now):
1. Generate a separate attributes enum for each datatype: ScaleInt32Attributes, ScaleDoubleAttributes, etc.
   * This gives better auto-complete. This is especially important because DAQ will not report errors if you use the wrong datatype.
 2. Have an `attribute_raw`/oneof for attributes similar to enums.
    * We need this for internal attributes and attributes-we-don't-know-about.

Any concerns about those ideas?

astarche: This would be in a future PR.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -847,6 +865,52 @@ enum WriteBasicTEDSOptions {
   WRITE_BASIC_TEDS_OPTIONS_DO_NOT_WRITE = 12540;
 }
 
+enum ScaleInt32AttributeValues {
```
astarche, line 868: Note: I think it would be helpful to include the enum name in these names. This will be more important for channel attributes where there are *a lot* of different options.

source/codegen/metadata_mutation.py
```diff
@@ -176,7 +176,12 @@ def populate_grpc_types(parameters, config):
             service_class_prefix)
 
 
-def add_attribute_values_enums(enums, attribute_enums_by_type, service_class_prefix):
+def get_short_enum_type_name(type):
```
astarche, line 179: Note: I'll probably have to iterate on this method. Especially for arrays. But this works reasonably well. A good alternative could be to take "everything after the word Attribute" from the method name and use that as a starting point.

Sample DAQ input: `int32`
Sample MI input: `ViInt32`
Desired output in both cases: `Int32`

source/codegen/common_helpers.py
```diff
@@ -316,6 +327,28 @@ def indent_text_to_level(text, level):
...
 def filter_parameters_for_grpc_fields(parameters):
-    """Filter out the parameters that shouldn't be represented by a field on a grpc message.
-       For example, get rid of any parameters whose values should be determined from another parameter."""
-    return [p for p in parameters if p.get('include_in_proto', True)]
+  """Filter out the parameters that shouldn't be represented by a field on a grpc message.
+      For example, get rid of any parameters whose values should be determined from another parameter."""
+  return [p for p in parameters if p.get('include_in_proto', True)]
+
+
+def get_attribute_groups(data):
```
astarche, line 335: Note: this method is key to this change. 

Previously there was always one group and it's name was `${service_class_prefix}`. Now that is the fallback/default.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -651,6 +651,69 @@ TEST_F(NiFakeNonIviServiceTests, OutputTimestamp_KnownValue)
...
+  const auto VALUE = MarbleInt32AttributeValues::MARBLE_INT32_AQUA;
+  EXPECT_CALL(library_, GetMarbleAttributeInt32(_, ATTRIBUTE, _))
+      .WillOnce(DoAll(SetArgPointee<2>(VALUE), Return(kDriverSuccess)));
+  ::grpc::ServerContext context;
+  GetMarbleAttributeInt32Request request;
+  request.set_attribute(ATTRIBUTE);
+  GetMarbleAttributeInt32Response response;
+  service_.GetMarbleAttributeInt32(&context, &request, &response);
+
+  EXPECT_EQ(kDriverSuccess, response.status());
```
reckenro, line 715: I think we can add an `EXPECT_EQ(VALUE, response.value())` here as well.

generated/nidaqmx/nidaqmx.proto
reckenro: Those ideas sound good to me, especially for DAQ because of the reasons you laid out.

reckenro (approved):

doshirohan (approved):

source/codegen/generate_service.py
```diff
@@ -25,11 +25,44 @@ def generate_service_file(metadata, template_file_name, generated_file_suffix, g
...
+  def __init__(self, metadata):
+    self._metadata = metadata
+    self._attribute_type_map = {}
+
+    for group_name, attributes in common_helpers.get_attribute_groups(metadata).items():
+      attribute_enums_by_type = common_helpers.get_attribute_enums_by_type(attributes)
+      metadata_mutation.add_attribute_values_enums(metadata["enums"], attribute_enums_by_type, group_name)
+      self._attribute_type_map[group_name] = attribute_enums_by_type
+
+  def get_used_attribute(self, parameters):
```
doshirohan, line 45: I believe the better name for this function would be `get_used_attribute_group_name`, because essentially that's what it is returning. Corresponding variable name where this function is called should be updated too.

source/codegen/generate_service.py
```diff
@@ -25,11 +25,44 @@ def generate_service_file(metadata, template_file_name, generated_file_suffix, g
...
+    param_types = (param["grpc_type"] for param in parameters)
```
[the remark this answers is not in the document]
doshirohan: Ah, its not looking for value parameter. It is actually looking for parameter which accepts attribute id and just assumes there would be another parameter for attribute value that needs to be mutated. Sneaky! :)

The reason why this is working for MI drivers is because all attribute params for MI drivers are marked as `ViAttr` type and then `grpc_type` for those is updated [here](https://github.com/ni/grpc-device/blob/c01e3769891e3443fe0add6e6947612a7b6867fb/source/codegen/metadata_mutation.py#L106).

generated/nidaqmx/nidaqmx.proto
astarche: Interesting, I wonder why that's not working for this one. I'll look at it closer.

generated/nidaqmx/nidaqmx.proto
astarche: Good point @doshirohan , I think you're right.

source/codegen/generate_service.py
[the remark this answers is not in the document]
astarche: Thanks for the feedback. The name change you proposed below made this a little clearer. And I added a comment. I'll have to make another pass on this for the data-type-specific attributes. SO, hopefully I can make it less-not-more confusing.

PR #292, opened 2021-08-11
Add hardcoded_value mechanism

astarche:

Add `hardcoded_value` parameter attribute to use in 3 cases:
* Replace the `pass_null` convention on `reserved` params.
* Hardcode `options` to zero on callback methods to disable calling the callback on the registering thread.
* Pass in zero for the size param for scalar attributes.

Standardizes on shared convention and fixes some corner case issues.

Ran and passed all DAQ tests.

### DIFF
--- a/source/tests/system/nidaqmx_driver_api_tests.cpp
+++ b/source/tests/system/nidaqmx_driver_api_tests.cpp
@@ -922,22 +922,19 @@ TEST_F(NiDAQmxDriverApiTests, SetPreScaledUnits_GetPreScaledUnits_ReturnsAttribu
   EXPECT_EQ(ScaleInt32AttributeValues::SCALE_INT32_UNITS_PRE_SCALED_RPM, response.value());
 }
 
-TEST_F(NiDAQmxDriverApiTests, SetScaledUnitsAsDouble_Fails)
+TEST_F(NiDAQmxDriverApiTests, GetScaledUnitsAsDouble_Fails)
 {
   const auto SCALE_NAME = std::string("TestScale");
   auto scale_status = create_lin_scale(SCALE_NAME, 0.5);
-  auto set_request = create_set_scale_attribute_request<SetScaleAttributeDoubleRequest>(
-    SCALE_NAME, 
-    ScaleAttributes::SCALE_ATTRIBUTE_SCALED_UNITS, 
-    ScaleInt32AttributeValues::SCALE_INT32_RPM);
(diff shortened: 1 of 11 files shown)


source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -922,6 +922,21 @@ TEST_F(NiDAQmxDriverApiTests, SetPreScaledUnits_GetPreScaledUnits_ReturnsAttribu
   EXPECT_EQ(ScaleInt32AttributeValues::SCALE_INT32_UNITS_PRE_SCALED_RPM, response.value());
 }
 
+TEST_F(NiDAQmxDriverApiTests, GetScaledUnitsAsDouble_Fails)
```
astarche, line 925: Unfortunately, the trick of using the size=0 as a defensive mechanism does not work as well as I'd hoped. It works for getters using the wrong type, but not for setters.

So I think we will need a different way to handle "wrong type" accesses. Part of that will be enums that are organized by type. We can also generate a attribute-to-type validator to double-check. If we're going to allow for pure "attributes we don't know about", we might leave that behind a toggle. Or allow for plugging into the validator from the config file.

I'm slightly leaning towards keeping the hardcoded-zero-size-param. It's doing some good and provides a nice rule-of-3 third use case for the hardcoded_value mechanism.

reckenro: Ah, I see. I agree though, I'd lean towards keeping it as well. Like you said it's doing some good on its own plus gives some more validity to having the hardcoded_value mechanism.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -922,6 +922,21 @@ TEST_F(NiDAQmxDriverApiTests, SetPreScaledUnits_GetPreScaledUnits_ReturnsAttribu
...
+{
+  const auto SCALE_NAME = std::string("TestScale");
+  auto scale_status = create_lin_scale(SCALE_NAME, 0.5);
+  GetScaleAttributeDoubleResponse response;
+  auto status = get_scale_attribute_double(SCALE_NAME, ScaleAttributes::SCALE_ATTRIBUTE_SCALED_UNITS, response);
+
```
reckenro, line 931: You can get rid of this extra blank line.

reckenro (approved):

PR #293, opened 2021-08-11
Add string and double array scale attributes

astarche:

Add DAQ Scale string and double array attributes.
Regenerating PowerUpStates metadata with automated process.

Expands support and testing for attribute data types.
Completes Get/Set Scale attributes.

Ran and passed all DAQ tests.

### DIFF
--- a/source/codegen/metadata/nidaqmx/enums.py
+++ b/source/codegen/metadata/nidaqmx/enums.py
@@ -1156,19 +1156,17 @@ enums = {
     'PowerUpStates': {
         'values': [
             {
-                'documentation': {
-                    'description': 'Voltage Channel'
-                },
                 'name': 'CHANNEL_VOLTAGE',
                 'value': 0
             },
             {
-                'documentation': {
-                    'description': 'Current Channel'
-                },
                 'name': 'CHANNEL_CURRENT',
                 'value': 1
             },
+            {
+                'name': 'CHANNEL_HIGH_IMPEDANCE',
+                'value': 2
+            }
         ]
     },
     'PressureUnits': {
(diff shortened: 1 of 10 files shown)


source/codegen/metadata/nidaqmx/enums.py
```diff
@@ -1156,19 +1156,17 @@
     'PowerUpStates': {
         'values': [
             {
-                'documentation': {
```
astarche, line 1159: Mentioned in the PR description, I'm not generating these so they look a little different but the output is the same (Except for the extra enum option).

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -914,14 +974,68 @@ TEST_F(NiDAQmxDriverApiTests, SetPreScaledUnits_GetPreScaledUnits_ReturnsAttribu
...
+  EXPECT_EQ(UNITS, response.value());
+}
+
+
+TEST_F(NiDAQmxDriverApiTests, SetPolynomialForwardCoefficients_GetPolynomialForwardCoefficients_ReturnsAttribute)
+{
+  const auto SCALE_NAME = std::string("TestPolynomialScale");
+  const auto INITIAL_COEFFICIENTS = std::vector<double>{1.0, 2.0, 3.0};
+  const auto COEFFICIENTS = std::vector<double>{1.0, 3.0, 8.0};
+  auto scale_status = create_polynomial_scale(SCALE_NAME, INITIAL_COEFFICIENTS, INITIAL_COEFFICIENTS);
```
reckenro, line 1018: Just to make sure I understand, this call for `create_polynomial_scale` is to set the forward and reverse coefficients so later in validation we're making sure we're not getting either of those initial coefficients, right?

reckenro (approved):

source/tests/system/nidaqmx_driver_api_tests.cpp
astarche: Yeah, I'd sort of prefer not to set the initial value at all. But the create method requires that you pass something in. So we initialize to some value. Then assert that we can write-and-read-back some other value.

astarche:
Add DAQ Scale string and double array attributes.
Regenerating PowerUpStates metadata with automated process.

PR #297, opened 2021-08-12
Add support for splitting attribute enums by type

astarche:

Add support for splitting attribute enums by type.

This makes it easier to determine which attributes are valid for typed attribute accessor methods.

Ran and passed all DAQ tests.

### DIFF
--- a/generated/nidaqmx/nidaqmx.proto
+++ b/generated/nidaqmx/nidaqmx.proto
@@ -246,21 +246,30 @@ service NiDAQmx {
   rpc WriteToTEDSFromFile(WriteToTEDSFromFileRequest) returns (WriteToTEDSFromFileResponse);
 }
 
-enum ScaleAttributes {
-  SCALE_ATTRIBUTE_UNSPECIFIED = 0;
+enum ScaleAttributesString {
+  SCALE_STRING_ATTRIBUTE_UNSPECIFIED = 0;
   SCALE_ATTRIBUTE_DESCR = 4646;
+  SCALE_ATTRIBUTE_SCALED_UNITS = 6427;
+}
+enum ScaleAttributesDouble {
+  SCALE_DOUBLE_ATTRIBUTE_UNSPECIFIED = 0;
   SCALE_ATTRIBUTE_LIN_SLOPE = 4647;
   SCALE_ATTRIBUTE_LIN_Y_INTERCEPT = 4648;
   SCALE_ATTRIBUTE_MAP_SCALED_MAX = 4649;
   SCALE_ATTRIBUTE_MAP_SCALED_MIN = 4656;
   SCALE_ATTRIBUTE_MAP_PRE_SCALED_MAX = 4657;
   SCALE_ATTRIBUTE_MAP_PRE_SCALED_MIN = 4658;
+}
+enum ScaleAttributesDoubleArray {
+  SCALE_DOUBLE_ARRAY_ATTRIBUTE_UNSPECIFIED = 0;
   SCALE_ATTRIBUTE_POLY_FORWARD_COEFF = 4660;
   SCALE_ATTRIBUTE_POLY_REVERSE_COEFF = 4661;
   SCALE_ATTRIBUTE_TABLE_SCALED_VALS = 4662;
   SCALE_ATTRIBUTE_TABLE_PRE_SCALED_VALS = 4663;
+}
+enum ScaleAttributesInt32 {
+  SCALE_INT32_ATTRIBUTE_UNSPECIFIED = 0;
   SCALE_ATTRIBUTE_PRE_SCALED_UNITS = 6391;
-  SCALE_ATTRIBUTE_SCALED_UNITS = 6427;
   SCALE_ATTRIBUTE_TYPE = 6441;
 }
 
@@ -3642,7 +3651,7 @@ message GetRefTrigTimestampValResponse {
 
 message GetScaleAttributeDoubleRequest {
   string scale_name = 1;
-  ScaleAttributes attribute = 2;
+  ScaleAttributesDouble attribute = 2;
 }
 
 message GetScaleAttributeDoubleResponse {
@@ -3652,7 +3661,7 @@ message GetScaleAttributeDoubleResponse {
 
 message GetScaleAttributeDoubleArrayRequest {
   string scale_name = 1;
-  ScaleAttributes attribute = 2;
+  ScaleAttributesDoubleArray attribute = 2;
   uint32 size = 3;
 }
 
@@ -3663,7 +3672,7 @@ message GetScaleAttributeDoubleArrayResponse {
 
 message GetScaleAttributeInt32Request {
   string scale_name = 1;
-  ScaleAttributes attribute = 2;
+  ScaleAttributesInt32 attribute = 2;
 }
 
 message GetScaleAttributeInt32Response {
@@ -3673,7 +3682,7 @@ message GetScaleAttributeInt32Response {
 
 message GetScaleAttributeStringRequest {
   string scale_name = 1;
-  ScaleAttributes attribute = 2;
+  ScaleAttributesString attribute = 2;
   uint32 size = 3;
 }
 
@@ -4285,7 +4294,7 @@ message SetFirstSampClkWhenResponse {
 
 message SetScaleAttributeDoubleRequest {
   string scale_name = 1;
-  ScaleAttributes attribute = 2;
+  ScaleAttributesDouble attribute = 2;
   double value = 3;
 }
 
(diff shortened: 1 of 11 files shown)


source/codegen/common_helpers.py
```diff
@@ -141,13 +141,13 @@ def pascal_to_camel(pascal_string):
 
 
 def ensure_pascal_case(pascal_or_camel_string):
```
astarche, line 143: (Sorry) Just an indentation change. This method was out-of-sync with the rest of the file and was causing problems.

source/codegen/common_helpers.py
```diff
@@ -163,20 +163,20 @@ def filter_proto_rpc_functions(functions):
 
 
 def get_attribute_enums_by_type(attributes):
```
astarche, line 165: (Sorry) Just an indentation change. This method was out-of-sync with the rest of the file and was causing problems.

source/codegen/common_helpers.py
```diff
@@ -328,23 +328,147 @@ def filter_parameters_for_grpc_fields(parameters):
...
+    """
+    Used to create an identifier string based on the grpc_type
+    i.e., double -> Double. repeated double -> DoubleArray.
+    """
+    grpc_type = get_grpc_type(data_type, config)
+    grpc_type = re.sub(r'^(repeated )(\w+)$', r'\2Array', grpc_type)
+    return ensure_pascal_case(grpc_type)
+
+
+def get_grpc_type_from_ivi(type, is_array, driver_name_pascal):
```
astarche, line 386: This method moved from metadata_mutation without changes.

source/codegen/metadata/nifake_non_ivi/config.py
```diff
@@ -35,13 +35,14 @@
     'additional_headers': ['custom/nidaqmx_conversions.h'],
     'type_to_grpc_type': {
         'FakeHandle': 'nidevice_grpc.Session',
-        'const char[]': 'string',
+        'char[]': 'string',
```
astarche, line 38: Having to handle `const` and non-const separately adds an extra lookup condition to get grpc_type conversion method. But we always get the same result. So standardizing on non-const here and the conversion method will strip const before the lookup.

source/codegen/metadata_mutation.py
```diff
@@ -214,6 +131,81 @@ def add_attribute_values_enums(enums, attribute_enums_by_type, group_name):
         add_enum(unmapped_enum_name, unmapped_values, enums, enum_value_prefix)
         add_enum(mapped_enum_name, mapped_values, enums, enum_value_prefix, is_mapped=True)
     
+
+AttributeReferencingParameter = namedtuple("AttributeReferencingParameter", ["attribute_group", "parameter"])
+
+class AttributeAccessorExpander:
```
astarche, line 137: Formerly AttributeValueExpander.

Has significant changes.

You can see the move in a separate commit if you look at the history. But it's easier to just read it here as-is.

source/codegen/metadata_mutation.py
```diff
@@ -214,6 +131,81 @@ def add_attribute_values_enums(enums, attribute_enums_by_type, group_name):
...
+                return AttributeReferencingParameter(potential_attribute_name, parameter)
+            return None
+
+        # Assumption: there are 0 or 1 parameters that reference an attribute per function.
+        attribute_resolve_results = (try_resolve_attribute_reference(p) for p in parameters)
+        valid_references = (ref for ref in attribute_resolve_results if ref)
+        return next(valid_references, None)
+
+
+    def patch_attribute_enum_type(self, func):
```
astarche, line 178: This is the new ability of this class.

source/codegen/common_helpers.py
```diff
@@ -328,23 +328,147 @@ def filter_parameters_for_grpc_fields(parameters):
...
+        type = 'fixed64'
+    if 'int' == type:
+        type = 'sint32'
+    if "[]" in type:
+        type = type.replace("[]", "")
+
+    return "repeated " + type if add_repeated else type
+
+
+def get_grpc_type(data_type, config):
```
astarche, line 448: This moved from metadata_mutation and I changed the implementation significantly. (But it still generates the same)

@gregstoll

generated/nifake_non_ivi/nifake_non_ivi.proto
```diff
@@ -36,9 +36,12 @@ service NiFakeNonIvi {
   rpc SetMarbleAttributeInt32(SetMarbleAttributeInt32Request) returns (SetMarbleAttributeInt32Response);
 }
 
-enum MarbleAttributes {
-  MARBLE_ATTRIBUTE_UNSPECIFIED = 0;
+enum MarbleAttributesInt32 {
+  MARBLE_INT32_ATTRIBUTE_UNSPECIFIED = 0;
```
astarche, line 40: If we keep the enum name as `MarbleAttributesInt32` the UNSPECIFIED value should probably be `MARBLE_ATTRIBUTE_INT32_UNSPECIFIED`. I don't really like that, but it doesn't matter much as long as it's just unspecified. (Seem comments in `proto_helpers.mako` on how this works).

source/codegen/common_helpers.py
```diff
@@ -328,23 +328,147 @@ def filter_parameters_for_grpc_fields(parameters):
   return [p for p in parameters if p.get('include_in_proto', True)]
 
 
+class AttributeGroup:
+    def __init__(self, name, attributes, config):
+        self.name = name
+        self.attributes = attributes
+        self._config = config
+
+
```
astarche, line 337: Note: I've started working on `Reset`. That will not be a significant change to the **logic** of this code, but it will change a lot of the terminology.

We'll also need to split out `resettable` attributes into another group that will get its own enum. So, code on that level will be a little more generic (I'm thinking "sub_group"). Then when we're expanding accessors, resetters will be handled separately from typed getters/setters. But will use most of the same patch logic.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -4306,7 +4315,7 @@ message SetScaleAttributeDoubleArrayResponse {
 
 message SetScaleAttributeInt32Request {
   string scale_name = 1;
-  ScaleAttributes attribute = 2;
```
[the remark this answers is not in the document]
reckenro: I have a slight preference to change the `AttributeValues` enum name to match what you have now so the name structure for message, `Attribute` enum, and `AttributeValues` enum matches. That is `ScaleInt32AttributeValues` becomes `ScaleAttributeInt32Values`.

But if that's extra work I agree with switching the `Attributes` enum back to `ScaleInt32Attributes` for the enum names to match.

source/codegen/templates/proto.mako
```diff
@@ -41,8 +41,10 @@ service ${service_class_prefix} {
 % endfor
 }
 
-% for group_name, attributes in common_helpers.get_attribute_groups(data).items():
-${mako_helper.define_attribute_enum(group_name, attributes)}\
+% for group in common_helpers.get_attribute_groups(data):
+% for data_type, attributes in group.get_attributes_split_by_type().items():
+${mako_helper.define_attribute_enum(group.name, data_type, attributes)}\
+% endfor
 % endfor
```
reckenro, line 49: Minor spacing fix. To get empty lines between the enums in the proto file can you remove the blank line on line 49 here. Then in `define_attribute_enum` add a blank line after the closing `}` so each enum is separated from the next.

reckenro (approved):

generated/nidaqmx/nidaqmx.proto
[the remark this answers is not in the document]
astarche: Well,  I'd have to deal with existing code. Either by (a) having a config setting to enable the new behavior or (b) breaking compatibility (and updating existing tests/examples). So, it would add some work. Good point though.

source/codegen/metadata_mutation.py
```diff
@@ -214,6 +131,81 @@ def add_attribute_values_enums(enums, attribute_enums_by_type, group_name):
...
+    group -> type -> attributes
+
+    and implements the metadata_mutations to add_attribute_values_enums, expand_attribute_function_value_param,
+    and patch_attribute_enum_type.
+    """
+    def __init__(self, metadata):
+        self._config = metadata['config']
+        self._enums = metadata['enums']
+        self._attribute_type_map = {}
+        self._spit_by_type = False
```
gregstoll, line 150: should this be "split"_by_type? :-)

doshirohan (approved):

generated/nidaqmx/nidaqmx.proto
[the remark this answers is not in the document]
astarche: Thanks for the input. For this PR I changed it to match AttributeValues. I will see if I can code up a PR changing both to match the accessor methods.

astarche:
This makes it easier to determine which attributes are valid for typed attribute accessor methods.

astarche:
This makes it easier to determine which attributes are valid for typed attribute accessor methods.

PR #299, opened 2021-08-16
Add support for output varargs functions

gregstoll:

Adds support for functions that use varargs to pass output. It builds on #290. To verify that it works with functions with many parameters this also adds a DAQ function `GetDigitalPullUpPullDownStates`.

To do this we now look more at the repeating parameters (the ones marked with `repeating_argument`), which now must match the fields in the repeated vararg.

DAQ has several functions that do this. (for example `DAQmxGetDigitalPullUpPullDownStates()`)

Added unit tests to `nifake_non_ivi`.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -7045,6 +7045,51 @@ namespace nidaqmx_grpc {
...
...tor, 78), get_channelName_if(channel_names, 79), get_state_if(stateVector, 79), get_channelName_if(channel_names, 80), get_state_if(stateVector, 80), get_channelName_if(channel_names, 81), get_state_if(stateVector, 81), get_channelName_if(channel_names, 82), get_state_if(stateVector, 82), get_channelName_if(channel_names, 83), get_state_if(stateVector, 83), get_channelName_if(channel_names, 84), get_state_if(stateVector, 84), get_channelName_if(channel_names, 85), get_state_if(stateVector, 85), get_channelName_if(channel_names, 86), get_state_if(stateVector, 86), get_channelName_if(channel_names, 87), get_state_if(stateVector, 87), get_channelName_if(channel_names, 88), get_state_if(stateVector, 88), get_channelName_if(channel_names, 89), get_state_if(stateVector, 89), get_channelName_if(channel_names, 90), get_state_if(stateVector, 90), get_channelName_if(channel_names, 91), get_state_if(stateVector, 91), get_channelName_if(channel_names, 92), get_state_if(stateVector, 92), get_channelName_if(channel_names, 93), get_state_if(stateVector, 93), get_channelName_if(channel_names, 94), get_state_if(stateVector, 94), get_channelName_if(channel_names, 95), get_state_if(stateVector, 95));
```
astarche, line 7079: I missed this last time: there needs to be 1 extra nullptr at the end to ensure that it's terminated when you pass in the max.

source/codegen/service_helpers.py
```diff
@@ -130,47 +146,50 @@ def get_array_param_size(parameter) -> str:
...
-    elif common_helpers.is_repeated_varargs_parameter(parameter):
+    if common_helpers.is_repeated_varargs_parameter(parameter):
         if expand_varargs:
-            max_length = parameter['max_length']
-            if any_not_include_in_proto:
-                max_length -= 1
+            # All functions that take repeated varargs take the first set of parameter as
+            # non-varargs, so we need one fewer set of parameters in the varargs
+            # section.
+            max_length = parameter['max_length'] - 1
```
astarche, line 177: Consider: this shows up twice with the same block comment. Can it be a helper function somewhere? get_expanded_var_args_length(param)?

astarche (approved):

generated/nidaqmx/nidaqmx_service.cpp
gregstoll: oh, good catch! That means we need to pass one additional set of arguments.

source/codegen/service_helpers.py
gregstoll: This went away since now we're passing an extra set of arguments.

source/codegen/templates/service_helpers.mako
```diff
@@ -420,14 +432,39 @@ ${initialize_standard_input_param(function_name, parameter)}
...
 %>\
-%   if common_helpers.is_array(parameter['type']):
+%   if common_helpers.is_repeating_parameter(parameter):
+      auto get_${parameter_name}_if = [](std::vector<${underlying_param_type}>& vector, int n) -> ${underlying_param_type}* {
+            if (vector.size() > n) {
+## Note that this code will not handle every datatype, but it works for all
+## the ones we currently use with repeated varargs.
+                  return &(vector[n]);
+            }
+            return nullptr;
```
astarche, line 463: Is there a good place to document that terminating with a nullptr is part of the convention and required (hence the extra pair of args)? Maybe around the this get_if code that implements it? Or possibly around the core/common helpers describing the mechanism?

It's OK if it's not obvious to a casual reader. As long as someone who really needs to figure it out has something to hang on to.

gregstoll: I did document it in `create_args_for_varargs()` in `source/codegen/service_helpers.py` - I think that's good enough; let me know if you agree.

reckenro (approved):

PR #300, opened 2021-08-16
Add example for DAQ callback API

astarche:

Add example for DAQ callback API.
Use grpc asyncio API to demonstrate the `async for` syntax for reading streams.

Demonstrates how to use DAQ callbacks in a similar style to the C API examples.
* This would enable porting existing code or writing code that doesn't block server threads while waiting for data.

Demonstrates how to use the grpc asyncio API in general, which allows writing code that doesn't block *client* threads while waiting for data.

Ran example with expected result.
Tested some failure cases.

### DIFF
--- a/examples/nidaqmx/analog-input-everynsamples.py
+++ b/examples/nidaqmx/analog-input-everynsamples.py
@@ -1,0 +1,132 @@
+# Demonstrates how to acquire analog data using the grpc asyncio API and RegisterEveryNSamplesEvent.
+#
+# The gRPC API is built from the C API. NI-DAQmx documentation is installed with the driver at:
+# C:\Program Files (x86)\National Instruments\NI-DAQ\docs\cdaqmx.chm
+#
+# Getting Started:
+#
+# To run this example, install "NI-DAQmx Driver" on the server machine.
+# Link: https://www.ni.com/en-us/support/downloads/drivers/download.ni-daqmx.html
+#
+# For instructions on how to use protoc to generate gRPC client interfaces, see our "Creating a gRPC Client" wiki page.
+# Link: https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+#
+# Running from command line:
+#
+# Server machine's IP address, port number, and physical channel name can be passed as separate command line arguments.
+#   > python analog-input.py <server_address> <port_number> <physical_channel_name>
+# If they are not passed in as command line arguments, then by default the server address will be "localhost:31763", with "Dev1/ai0" as the physical channel name
+import asyncio
+import grpc
+import sys
+
+import nidaqmx_pb2 as nidaqmx_types
+import nidaqmx_pb2_grpc as grpc_nidaqmx
+
+server_address = "localhost"
+server_port = "31763"
+physical_channel = "Dev1/ai0"
+
+if len(sys.argv) >= 2:
+    server_address = sys.argv[1]
+if len(sys.argv) >= 3:
+    server_port = sys.argv[2]
+if len(sys.argv) >= 4:
+    physical_channel = sys.argv[3]
+
+
+async def main():
+    client = None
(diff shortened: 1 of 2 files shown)


examples/nidaqmx/analog-input-everynsamples.py
```diff
@@ -0,0 +1,132 @@
...
+# To run this example, install "NI-DAQmx Driver" on the server machine.
+# Link: https://www.ni.com/en-us/support/downloads/drivers/download.ni-daqmx.html
+#
+# For instructions on how to use protoc to generate gRPC client interfaces, see our "Creating a gRPC Client" wiki page.
+# Link: https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+#
+# Running from command line:
+#
+# Server machine's IP address, port number, and physical channel name can be passed as separate command line arguments.
+#   > python analog-input.py <server_address> <port_number> <physical_channel_name>
```
gregstoll, line 17: this should be `analog-input-everynsamples.py`

examples/nidaqmx/analog-input-everynsamples.py
```diff
@@ -0,0 +1,132 @@
...
+    client = None
+    task = None
+
+    async with grpc.aio.insecure_channel(f"{server_address}:{server_port}") as channel:
+        try:
+            client = grpc_nidaqmx.NiDAQmxStub(channel)
+
+            async def raise_if_error(response):
+                if response.status:
+                    response = await client.GetErrorString(
```
gregstoll, line 48: Just curious - is there a reason we're using `await` here? `GetErrorString()` is usually synchronous, I thought...

examples/nidaqmx/analog-input-everynsamples.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+            )
+            task = create_response.task
+
+            await raise_if_error_async(
+                client.CreateAIVoltageChan(
+                    nidaqmx_types.CreateAIVoltageChanRequest(
+                        task=task,
+                        physical_channel=physical_channel,
+                        min_val=1.0,
```
gregstoll, line 71: Did you copy this from another example? I thought the standard range was usually 0-10 V, although it's not a big deal...

gregstoll:
I'm not very familiar with Python async stuff, but this looks reasonable to me :-)

gregstoll (approved):

examples/nidaqmx/analog-input-everynsamples.py
astarche: using the `grpc.aio` channel causes all the methods on the stub to be async. Those methods are actually defined as:

```python
        self.GetErrorString = channel.unary_unary(
                '/nidaqmx_grpc.NiDAQmx/GetErrorString',
                request_serializer=nidaqmx__pb2.GetErrorStringRequest.SerializeToString,
                response_deserializer=nidaqmx__pb2.GetErrorStringResponse.FromString,
                )
```

So basically we have to `await` them because that's the way the aio channel implements it.

There are a few reasons that I used the asyncio api here:
1. I think the `async for`/generator implementation for streams is nicer than the `future` implementation on a standard channel.
2. callbacks are intended for an `async` programming style, so I expect users interested in callbacks to be interested in the asyncio api (and may actually realize that they actually want is an asyncio client without the callbacks).
3. It's nice to cover `asyncio` somewhere in the examples.

examples/nidaqmx/analog-input-everynsamples.py
astarche: Oops. The other ones are (-10,10). I updated to that. Anyway it should include zero.

harsha-bhushan:
nitpick: the example can be named as analog-input-every-n-samples.py for better readability

harsha-bhushan (approved):

PR #301, opened 2021-08-16
Fix some C++ warnings

gregstoll:

Fix some C++ warnings.

I was hoping to fix all warnings, but there's one caused by gRPC itself, and a few caused by delay loading some libraries we don't currently use in `ni_grpc_device_server` (`nidcpower.dll`, `niScope.dll`, and `niswitch.dll`). I'm assuming we will use those libraries at some point so I'm not removing them now.

Note that each commit in this PR fixes one category of warning, if that's easier to read.

Fewer warnings makes it easier to see when new warnings are introduced.

Ran unit tests.

### DIFF
--- a/generated/nidcpower/nidcpower_service.cpp
+++ b/generated/nidcpower/nidcpower_service.cpp
@@ -524,7 +524,7 @@ namespace nidcpower_grpc {
       ViSession vi = session_repository_->access_session(vi_grpc_session.id(), vi_grpc_session.name());
       auto channel_name = request->channel_name().c_str();
       auto sequence_name = request->sequence_name().c_str();
-      ViInt32 attribute_id_count = request->attribute_ids().size();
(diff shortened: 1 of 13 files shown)


astarche (approved):

reckenro (approved):

bkeryan:
> a few caused by delay loading some libraries we don't currently use in ni_grpc_device_server (nidcpower.dll, niScope.dll, and niswitch.dll)

I think these warnings are due to the use of LabWindows/CVI user protection libraries instead of MSVC import libraries: https://github.com/ni/grpc-device/issues/678

`ni_grpc_device_server.exe` does load `nidcpower_64.dll`, `niScope_64.dll`, and `niswitch_64.dll`, but it doesn't use `/DELAYLOAD` and `delayimp.lib` to do it.

PR #306, opened 2021-08-18
Add enum array inputs for DAQmx Watchdog functions

astarche:

The Watchdog `ExpirStates` functions were missing the enums for input params.
* Add missing enums to API.
* Update grpc-device to support enum array inputs.

Because protobuf does not support `oneof` for `repeated` fields this bypasses the usual enum decomposition that splits out a `_raw` field and (for inputs) wraps it in a `oneof`. Enum array inputs will just be enum array inputs.

Makes it easier to determine correct inputs for DAQ `ExpirStates` methods.
Adds support for enum array inputs that we need in other places (DAQ attributes, PowerUpStates).

Ran and passed all DAQ tests.


gregstoll (approved):

source/codegen/proto_helpers.py
```diff
@@ -39,15 +39,30 @@ def get_enum_definitions(enums_to_define, enums):
...
+
+def is_array_input(parameter: dict):
+  return common_helpers.is_array(parameter["type"]) and common_helpers.is_input_parameter(parameter)
+
+
+def is_decomposable_enum(parameter: dict):
+  """
+  Enums are typically decomposed from a single param into an enum param and an _raw param.
+  The exception is array_inputs which are left as single enum param.
+  This is because protobuf does not support oneof on repeated types, so the standard
```
reckenro, line 51: Yeah, you can get around this by making a `message` with the `oneof` field and then use that message as the field type with `repeated` but that gets kind of ugly and more complicated. This sounds good.

reckenro (approved):

PR #309, opened 2021-08-18
Fix race condition in IVI dance (fixes #247)

### LINKED ISSUE #247: gRPC Server's IVI-Dance has race conditions
See code in https://github.com/ni/grpc-device/blob/main/generated/nidcpower/nidcpower_service.cpp#L2400

In IVI based APIs, the way clients retrieve strings is by calling the function to get the size, allocating, then calling again with the actual buffer (aka ["the IVI Dance"](https://github.com/ni/nimi-python/wiki/API-Metadata#the-ivi-dance)).

It is being done incorrectly: the recursive session lock isn't grabbed, so it is possible for the size of the string to grow between the time in which the size is queried and the time in which the string is copied onto the client buffer.

There are two ways to fix it:

1. Acquire the session lock. This is how NI's own C# and Python APIs do it internally.
2. Keep growing the buffer in a loop until it's big enough, which is usually just the one time. This is how NI's own Digital Pattern Editor does it internaly.

[AB#1579918](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1579918)

gregstoll:

Fixes a race condition if the size of a buffer changes between the initial sizing call and the "real" call afterwards.

This code assumes that if you pass in too small a buffer to a vanilla `ivi-dance` method it either returns the necessary size, or returns `kErrorReadBufferTooSmall`. For `ivi-dance-with-a-twist` methods, passing in too small a buffer will return `kErrorReadBufferTooSmall`.

Race conditions are bad - not fixing this could lead to spurious errors when calling ivi-dance APIs.

Added new unit tests for ivi-dance and ivi-dance-with-a-twist methods.

### DIFF
--- a/generated/nidcpower/nidcpower_service.cpp
+++ b/generated/nidcpower/nidcpower_service.cpp
@@ -2160,20 +2160,26 @@ namespace nidcpower_grpc {
       auto vi_grpc_session = request->vi();
       ViSession vi = session_repository_->access_session(vi_grpc_session.id(), vi_grpc_session.name());
 
-      auto status = library_->ExportAttributeConfigurationBuffer(vi, 0, nullptr);
-      if (status < 0) {
+      while (true) {
+        auto status = library_->ExportAttributeConfigurationBuffer(vi, 0, nullptr);
+        if (status < 0) {
+          response->set_status(status);
+          return ::grpc::Status::OK;
+        }
+        ViInt32 size = status;
+      
+        response->mutable_configuration()->Resize(size, 0);
+        ViAddr* configuration = reinterpret_cast<ViAddr*>(response->mutable_configuration()->mutable_data());
+        status = library_->ExportAttributeConfigurationBuffer(vi, size, configuration);
+        if (status > size) {
+          // buffer is now too small, try again
+          continue;
+        }
         response->set_status(status);
+        if (status == 0) {
+        }
         return ::grpc::Status::OK;
       }
-      ViInt32 size = status;
-
-      response->mutable_configuration()->Resize(size, 0);
-      ViAddr* configuration = reinterpret_cast<ViAddr*>(response->mutable_configuration()->mutable_data());
-      status = library_->ExportAttributeConfigurationBuffer(vi, size, configuration);
-      response->set_status(status);
-      if (status == 0) {
-      }
-      return ::grpc::Status::OK;
     }
     catch (nidevice_grpc::LibraryLoadException& ex) {
       return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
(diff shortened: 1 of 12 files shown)


source/codegen/templates/service_helpers.mako
```diff
@@ -41,22 +41,30 @@ ${initialize_input_params(function_name, parameters)}
...
+        }
+        ${size_param['type']} ${common_helpers.camel_to_snake(size_param['cppName'])} = status;
+      
+<%block filter="common_helpers.indent(1)">\
 ${initialize_output_params(output_parameters)}\
-      status = library_->${function_name}(${service_helpers.create_args(parameters)});
-      response->set_status(status);
+</%block>\
+        status = library_->${function_name}(${service_helpers.create_args(parameters)});
+        if (status == kErrorReadBufferTooSmall || status > ${common_helpers.camel_to_snake(size_param['cppName'])}) {
```
gregstoll, line 56: From looking at some examples I _think_ that ivi-dance functions can either return the "buffer too small" error or just return the required buffer size. In either case to make the logic simple we go around the loop again.

generated/nidcpower/nidcpower_service.cpp
```diff
@@ -11,6 +11,7 @@
 #include <iostream>
 #include <atomic>
 #include <vector>
+const auto kErrorReadBufferTooSmall = -200229;
```
astarche, line 14: Should this go inside the namespace?

Do the drivers not already define this? Or they define the same value under different names?

gregstoll: They use the same value under different names (and for some reason in this repo only `DAQmx.h` does, not sure why the other drivers don't...)

Yeah, I'll move it in the namespace, thanks!

source/codegen/templates/service.cpp.mako
```diff
@@ -16,6 +16,10 @@ function_names = service_helpers.filter_proto_rpc_functions_to_generate(function
 # If there are any non-mockable functions, we need to call the library directly, which
 # means we need another include file
 any_non_mockable_functions = any([not common_helpers.can_mock_function(functions[name]['parameters']) for name in function_names])
+# Define the constant for buffer too small if we have any of these functions.
+any_ivi_dance_functions = any(
+  [common_helpers.has_ivi_dance_with_a_twist_param(functions[name]['parameters']) or
```
astarche, line 21: I think you could replace this with a generator comprehension (instead of list comprehension) to avoid evaluating the whole list.

Although DAQ will still be the worst case by far, unless/until we add ivi-dance-for-daq

source/codegen/templates/service_helpers.mako
```diff
@@ -69,20 +77,30 @@ ${set_response_values(output_parameters)}\
...
+          response->set_status(status);
+          return ::grpc::Status::OK;
+        }
+<%block filter="common_helpers.indent(1)">\
 ${initialize_output_params(array_output_parameters)}\
-      status = library_->${function_name}(${service_helpers.create_args(parameters)});
-      response->set_status(status);
+</%block>\
+        status = library_->${function_name}(${service_helpers.create_args(parameters)});
+        if (status == kErrorReadBufferTooSmall) {
```
astarche, line 90: SO ivi-dance-with-a-twist is always a too-small error (for non zero-but-too-small inputs) but plain-ivi is either error or new size (depends on driver, I guess)?

Can you spell that out in the PR description (and commit description)?

And I think you pointed me to your source on ivi-dance-with-a-twist, although I don't know if that's one example or intrinsic to the twist.

astarche (approved):

generated/nidcpower/nidcpower_service.cpp
```diff
@@ -14,6 +14,8 @@
 
 namespace nidcpower_grpc {
```
astarche, line 15: Run the MI system tests (that work on simulation) or let me know and I can run them

gregstoll: (@astarche ran these and reported they passed)

source/codegen/templates/service_helpers.mako
gregstoll: updated the PR description

gregstoll:
> 
> 
> This looks good.
> 
> I don't know if you want to tackle this with this PR or a separate one but there are several custom ivi-dance function service implementations between the drivers. For instance, scope has several dance with a twist ones in its [niscope_service.custom.cpp](https://github.com/ni/grpc-device/blob/main/source/custom/niscope_service.custom.cpp) and nitclk seems to have one ivi-dance custom implementation for [GetAttributeViString](https://github.com/ni/grpc-device/blob/main/source/custom/nitclk_service.custom.cpp).

I'll do this in a separate PR - thanks!

gregstoll:
> 
> 
> > This looks good.
> > I don't know if you want to tackle this with this PR or a separate one but there are several custom ivi-dance function service implementations between the drivers. For instance, scope has several dance with a twist ones in its [niscope_service.custom.cpp](https://github.com/ni/grpc-device/blob/main/source/custom/niscope_service.custom.cpp) and nitclk seems to have one ivi-dance custom implementation for [GetAttributeViString](https://github.com/ni/grpc-device/blob/main/source/custom/nitclk_service.custom.cpp).
> 
> I'll do this in a separate PR - thanks!

Done in #312

PR #311, opened 2021-08-19
Use IVI dance for error string methods - fix #271

### LINKED ISSUE #271: DAQ's GetErrorString() should remove trailing NULL characters in the error_string
Since we know this parameter is a string, we should remove trailing NULLs for this and all strings in the API.

[AB#1560643](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1560643)

gregstoll:

Uses the "IVI dance" to determine how big the string buffer should be to pass to `GetErrorString` and `GetExtendedErrorInfo`. Also update Python examples.

Removes the need to specify how big the buffer is and strip trailing NULL characters from the response.

Verified that the buffer_size is no longer a member of the gRPC Request. Modified the DAQ driver API test to ensure that the error string we get back exactly matches what we expect (with no trailing NULLs). Ran a modified Python example and triggered an error and verified that the error string looked correct.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -7272,17 +7287,30 @@ namespace nidaqmx_grpc {
...
+          return ::grpc::Status::OK;
+        }
+        uInt32 buffer_size = status;
+      
+        std::string error_string;
+        if (buffer_size > 0) {
+            error_string.resize(buffer_size-1);
+        }
+        status = library_->GetExtendedErrorInfo((char*)error_string.data(), buffer_size);
+        if (status == kErrorReadBufferTooSmall || status > buffer_size) {
```
astarche, line 7304: In DAQ there's also a `DAQmxErrorBufferTooSmallForString`. Is it possible that could be returned here?

Or do these actually have the behavior where they don't return an error?

We can test these specific methods with the old implementation by passing in a too-small-non-zero size.

examples/nidaqmx/analog-output.py
```diff
@@ -44,9 +44,8 @@
 def RaiseIfError(response):
     if response.status != 0:
         response = client.GetErrorString(nidaqmx_types.GetErrorStringRequest(
```
astarche, line 46: @harsha-bhushan : we figured out that the `ivi-dance` works for a lot of DAQ methods. This removed the need to pass in a size and also avoids the bug with passed-in array sizes where you have to strip trailing null characters.

What do you think about adding this to the DAQ server?

If we *don't* add this. One problem is that users would implement the ivi-dance themselves by first calling with a zero size and then calling again with the returned size. So, if we tried to add this feature later (by making size optional/deprecated and doing the ivi-dance if it's zero) then we would break those existing usages. Of course, it would be easy for them to catch/fix.

It should be easy to automatically detect *most* of the cases where this would work. We'd have to do some double-checking to make sure that the messages we mark actually do support the convention. And there are cases where multiple arrays are sized together (or separately) that are a little trickier.

A variation would be to leave the size parameter on the proto request. Then we'd use that size if it's passed in (non-zero), but auto-detect the size if it wasn't. The biggest advantage of that is that the method would still be usable if we were wrong about it supporting the ivi-dance.

astarche (approved):

generated/nidaqmx/nidaqmx_service.cpp
gregstoll: I tested it it out and it actually returns a warning: `kWarningCAPIStringTruncatedToFitBuffer`. This is used in some common places where we're returning a string from an API, so I think it makes sense to handle it explicitly. (note that the code as is will probably handle this correctly since it's a positive fairly large value, but I guess it would fail for a > 200000 character string...)

examples/nidaqmx/analog-output.py
harsha-bhushan: I agree. This would greatly improve the usability. 
We should also run this by/keep informed someone from the team that owns/maintains DAQmx API especially for using ivi-dance on other methods.

harsha-bhushan:
Do update better proto examples too.

reckenro (approved):

PR #313, opened 2021-08-19
avoid signed/unsigned comparison warning in ivi-dance

gregstoll:

A signed/unsigned comparison warning was introduced in #309. This fixes the warning by explicitly casting, and explains in a comment why it is safe to do so.

We should try to keep compiler warnings at a minimum.

Still builds and unit tests pass.

### DIFF
--- a/source/codegen/templates/service_helpers.mako
+++ b/source/codegen/templates/service_helpers.mako
@@ -57,7 +57,9 @@ ${initialize_input_params(function_name, non_ivi_params)}\
 ${initialize_output_params(output_parameters)}\
 </%block>\
         status = library_->${function_name}(${service_helpers.create_args(parameters)});
-        if (status == kErrorReadBufferTooSmall || status > ${common_helpers.camel_to_snake(size_param['cppName'])}) {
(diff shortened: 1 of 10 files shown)


source/codegen/templates/service_helpers.mako
```diff
@@ -57,7 +57,9 @@ ${initialize_input_params(function_name, non_ivi_params)}\
 ${initialize_output_params(output_parameters)}\
 </%block>\
         status = library_->${function_name}(${service_helpers.create_args(parameters)});
-        if (status == kErrorReadBufferTooSmall || status > ${common_helpers.camel_to_snake(size_param['cppName'])}) {
+        // We cast status into ${common_helpers.camel_to_snake(size_param['cppName'])} above, so it's safe to cast
```
astarche, line 60: I'd vote to change this to a mako comment, but it's OK like this.

astarche (approved):

source/codegen/templates/service_helpers.mako
gregstoll: Yeah, I went back and forth on this. I'll change it.

reckenro (approved):

PR #314, opened 2021-08-19
Add support for DAQ Device attributes

astarche:

Add support for DAQ Device attributes

This requires a new feature in codegen: enum array outputs.

Since these use checked conversion, they also convert to UNSPECIFIED if the output is not in the enum, in which case it can be read from the array.

Differentiates enum values from raw values and give some context on enum values.

Ran and passed all DAQ tests.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -7125,6 +7125,221 @@ namespace nidaqmx_grpc {
...
+  {
+    if (context->IsCancelled()) {
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      auto device_name = request->device_name().c_str();
+      int32 attribute = request->attribute();
+      auto size = 0U;
+      bool32 value {};
+      auto status = library_->GetDeviceAttributeBool(device_name, attribute, &value, size);
```
gregstoll, line 7140: I couldn't find documentation on `DAQmxGetDeviceAttribute()` - I'm guessing the `size` parameter is the size of the array or 0 if it's a scalar property?

source/codegen/common_helpers.py
```diff
@@ -74,11 +74,23 @@ def strip_repeated_from_grpc_type(grpc_type):
     return grpc_type[len('repeated '):]
 
 
-def get_underlying_type_name(parameter_type):
+def get_underlying_type_name(parameter_type: str) -> str:
```
gregstoll, line 77: (thanks for adding type annotations; I'll try to be better about adding them in the future!)

gregstoll (approved):

generated/nidaqmx/nidaqmx_service.cpp
astarche: Size is not needed if it's a scalar property. That's the same sort of "surplus var_args" trick that we're using for channels and states.

I added this as a defensive mechanism in case someone passes in an array attribute to a scalar getter. So instead of crashing reading the size (or reading random memory) it would get a zero size and return an error. My original plan was to use a combination of that and padded stack buffers (i.e., to the size of a CVIAbsoluteTimestamp) for "don't crash" support for wrong-type attributes. That works fairly well for getters where the variable sized argument is passed as a void* but it *does not* work for setters where the variable-sized argument is a var_arg (That's [undefined behavior](https://en.cppreference.com/w/cpp/utility/variadic/va_arg)). So I'm going to need another plan for supporting "attributes we don't know about" (internal or new attributes). I'm thinking that will be requiring that the attribute id be validated, but adding a config option to extend the list of valid attributes-by-datatype.

We decided to leave the defensive zero-sizes for now, but I might remove them.

reckenro (approved):

PR #315, opened 2021-08-20
Catch up metadata generation

astarche:

Regenerating metadata to include features recently added manually:
* Channels and states methods.
* Some `const` integer arrays that the generator was not keeping track of.

This causes several additional changes:
* Update enum names to match daq api (`ResistorState` and `PowerUpStates`)
* Update parameter names to match C API docs or DAQ header (whichever is more correct).
    * channelsAndStates setters -> `channelNames`
    * channelsAndStates getters -> `channelName` (can only get one channel per state)
    * Watchdog -> `lines`
 * Alphabetizing names.
 * `double` -> `float64` (for consistency with DAQ API)
 * More array inputs are marked `const` that we didn't notice before.

We need the meatadata generator in sync to easily generate more metadata and make sweeping changes if necessary.

Ran and passed all DAQ tests.


source/codegen/metadata/nidaqmx/custom_proto.mako
```diff
@@ -5,23 +5,23 @@ message AnalogPowerUpState {
 }
 
 message ChannelNamesAndExpState {
-  string channelNames = 1;
+  string lines = 1;
   DigitalLineState expState = 2;
 }
```
astarche, line 11: These names are a little off (Especially PowerUpType) and inconsistent in where channel/state/function are in the name. Since it's kind of obnoxious to change them, I want to get on the same page before fixing them.

I like to keep "channel" before "state". I think it makes sense to use "Channels" when they accept multiple (and that should match "channelName" vs "channelNames"). I think it's more discoverable with function at the beginning although it's more affinitized to the state than the channel:

AnalogPowerUpChannelsAndState
WatchdogExpChannelsAndState
DigitalPowerUpChannelAndState
DigitalPowerUpChannelsAndState 
DigitalPullIUpPullDownChannelsAndState

Does that seem OK?

source/codegen/metadata/nidaqmx/enums.py
```diff
@@ -1797,6 +1760,26 @@
...
                 'name': 'CHANNEL_CURRENT',
                 'value': 1
+            },
+            {
+                'name': 'CHANNEL_HIGH_IMPEDANCE',
+                'value': 2
+            }
+        ]
+    },
+    'PowerUpStates': {
```
astarche, line 1770: These are the names from the internal daq api metadata. They aren't used in the C header, but do have the same names listed as "civName" and "lvName". So maybe it's public somewhere.

source/codegen/metadata/nidaqmx/custom_proto.mako
gregstoll: Yeah, I like that. Thanks for consistent-ifying these!

gregstoll (approved):

source/codegen/metadata/nidaqmx/enums.py
reckenro: I see them here: https://zone.ni.com/reference/en-XX/help/370469AP-01/lvdaqmx/mxsetpwrupstates/#DAQmxSetPowerUpStates(Digital)

So it's public in the documentation at least.

reckenro (approved):

PR #318, opened 2021-08-23
Add DAQmx Timing and Triggering attributes

astarche:

* Add DAQmx Timing and Triggering attributes.
* Add spot test.
* Fix to `get_grpc_type_name_for_identifier` to handle namespaced types (for `google.protobuf.Timestamp`)

Ran and passed all DAQ tests.


source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1863,6 +1912,25 @@ TEST_F(NiDAQmxDriverApiTests, TaskWithChannel_GetTaskAttributes_ReturnsCorectRes
...
 }
 
+TEST_F(NiDAQmxDriverApiTests, AIChannelWithSampleClock_ReconfigureRate_UpdatesRateSuccessfully)
+{
+  const auto INITIAL_RATE = 100.0;
+  const auto RECONFIGURED_RATE = 200.0;
+  create_ai_voltage_chan(0.0, 10.0);
+  auto cfg_request = create_cfg_samp_clk_timing_request(INITIAL_RATE, Edge1::EDGE1_RISING, AcquisitionType::ACQUISITION_TYPE_CONT_SAMPS, 1U);
+  cfg_samp_clk_timing(cfg_request);
+
```
astarche, line 1922: I'm getting tired of the status checking/etc in the previous approach to this. I originally used by-ref-outputs for responses thinking that there would be more cases where we want to test expected-bad statuses. But, that's not very common at this level of testing. So, I'm going to switch over to raising an exception on bad status and returning the response.

I'd like to switch more of the tests over to this style, but I'm not currently working on that.
It would be nice, long/medium term (for RF?) to generate client wrappers to use in tests in this style. It would be easier to build that up from a small service because it will require a few special cases (I.e., oneofs and streaming responses)

reckenro (approved):

PR #320, opened 2021-08-23
Determine array sizes for passed-in arrays in DAQ

gregstoll:

Determines array size and passes that in automatically.

Note that this is just a proof of concept - I have not looked through the rest of the DAQmx functions to see which other ones could take advantage of this. This will hopefully let us get a sense for how hard this is and how much value it provides.

Eases the burden on clients to have to pass in the array size and removes the possibility that a wrong size will be passed in, which can cause memory corruption.

Added an automated test.

### DIFF
--- a/generated/nidaqmx/nidaqmx.proto
+++ b/generated/nidaqmx/nidaqmx.proto
@@ -2952,7 +2952,6 @@ message CfgAnlgMultiEdgeStartTrigRequest {
   string trigger_sources = 2;
   repeated int32 trigger_slope_array = 3;
   repeated double trigger_level_array = 4;
-  uint32 array_size = 5;
 }
 
 message CfgAnlgMultiEdgeStartTrigResponse {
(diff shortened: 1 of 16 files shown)


astarche (approved):
Looks good!

Technically I don't think that wrong-sized arguments would corrupt memory. At least not directly. They would cause DAQ to read past the end of the buffer and (most likely) either (a) use bad data (for what?!) or (b) crash.

astarche:
@harsha-bhushan @gregstoll 

This is a similar issue to #311 but this time for array *inputs* instead of array outputs. 

Compared to that issue:
* How annoying is it when the customer does it right? -> Equally annoying (they have to pass in a size param).
* How bad is it when the customer does it wrong? -> Worse if they pass in a too-large size (causes undefined behavior) the same if they forget to pass a size.
* How hard is it to fix later? -> This one is easier. We can deprecate the size field and basically ignore it and the behavior will be unchanged for normal usages (if they were passing in a non-zero too-large or too-small size, the behavior will change). We might need more code do "something reasonable" when the user passes in a size (return error, use size if smaller, etc)
* How easy is it to detect all usages? -> Probably the same. We'll have to find all array inputs and figure out what their size param is. Mostly that is straightforward and there are common patterns of array/size pairs.

harsha-bhushan:
Thanks for the detailed comparison. This looks promising and improves usability. 
Please get this reviewed by an expert from DAQmx team to identify all usages & (if any places this can cause issues) also the behavior when size is passed.

reckenro (approved):

PR #323, opened 2021-08-24
Add AIImpedance mapped enum

astarche:

AIImpedance is a double enum that contains values that cannot be represented in an int32.

* Add the enum with the `generate-mappings` tag.
* Updates to codegen:
   * Support int64_t as the underlying type (double would also work, but is not great for map lookups and requires munging the metadata)
   * Rename the attribute value enum to `Double` instead of `Float64` to match the reset of the DAQ API.
* Add similar enums to fake non ivi service for tests.

Makes it easier to determine correct values for AIImpedance.

Ran and passed all DAQmx and unit tests.
Some manual debugging with AIImpedance (I couldn't get the test device configured to support this property).


generated/nidaqmx/nidaqmx.proto
```diff
@@ -7331,7 +7339,10 @@ message SetChanAttributeDoubleRequest {
   nidevice_grpc.Session task = 1;
   string channel = 2;
   ChannelDoubleAttributes attribute = 3;
```
astarche, line 7341: @harsha-bhushan @doshirohan @reckenro @gregstoll 

I coded this up, but I actually think this change should be **REJECTED**.

#### WHY is this change bad?
There are >150 channel double attributes (?!?!). Only *one* of them has an associated mapped enum. So we're adding a convenience parameter that will not apply to the vast majority of callsites. As a consequence, all callers will have to set `value_raw` instead of just `value`. And look at the mapped enum param to determine that it's not relevant/helpful.

#### PROPOSAL:
Do nothing. Allow setting AIImpedance as a double with no associated enum. Users of AIImpedance can easily find the valid values in the help (which is much easier than learning anything else about how to use the attribute ...).

#### ALTERNATIVE:
We could add support for a `raw_is_typical` flag in the attribute function metadata and, when that is true, remove the `_raw` suffix and make the oneof:

```proto
  oneof value_enum {
    ChannelDoubleAttributeValuesMapped value_mapped = 4;
    double value = 5;
  }
```

I think that's fine, but this is already a lot of work for one attribute.


#### QUESTION
Does anyone have a strong preference to keep the change in this PR? Or to implement the ALTERNATIVE?

PR #325, opened 2021-08-24
JUST A PROTOTYPE: detect correct array size for ReadAnalogF64

gregstoll:

Here is some code that calculates the correct array size for `ReadAnalogF64`. This is just a proof of concept - in my quick tests it seems to calculate the size correctly. (note that it doesn't build off of origin/main because it requires some attributes that haven't made it in yet)

To productize this we would have to:
- Move the new code into a helper method that could be called by a bunch of Read*() functions.
- Hopefully figure out how to make the Read*() functions call this without requiring us to write a custom implementation for each one.
- Decide when to trigger the auto-detection of array size. The option implemented here is to do it when the passed-in array size is 0 (which is also what happens if no values is specified)
- Decide how to represent `READ_ALL_AVAILABLE` - right now this is the constant -1.

It shouldn't, this is just for us to take a look at and discuss.

Minimal manual testing.


source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,4 +1,106 @@
...
+        break;
+      }
+      case nidaqmx_grpc::ReadAnalogF64Request::FillModeEnumCase::FILL_MODE_ENUM_NOT_SET: {
+        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The value for fill_mode was not specified or out of range");
+        break;
+      }
+    }
+
+    uInt32 array_size_in_samps = request->array_size_in_samps();
+    if (array_size_in_samps == 0) {
```
gregstoll, line 33: The custom code starts here.

astarche: This seems pretty reasonable to me. Implementing the codegen might be a challenge, especially if the different flavors of read are different. And also that we'd need a fair amount of testing to cover the different codepaths. This is probably "nice to have" and "something we could add later", but maybe we can get it in.

If the implementations for different read methods *are* the same, is it possible we could pull this out into helper methods? So the generated code could just be:

```c++
auto num_samps_per_chan = calculate_num_samps_per_chan(library, task, request->num_samps_per_chan());
auto array_size_in_samps = calculate_array_size_in_samps(library, task, num_samps_per_chan, request->array_size_in_samps());
```
I guess we'd have to use exceptions to handle errors, but that's probably doable. A FailedStatusException could be useful in other places.

In that case we *could* use something like the `hardcoded` or `custom-code` mechanisms to just code the one-liner in the metadata. But it would also be fine to just have a special `daq-read-samps-per-chan` and `daq-array-size-in-samps` mechanisms in the metadata/mako but have the implementation be very simple.

Also, this is sort of 2 separate features right? We could minimally just have something like:

```c++
auto array_size_in_samps = request->array_size_in_samps() 
    ? request->array_size_in_samps() 
    : (num_samps_per_chan * get_number_of_channels(library, task);
```

gregstoll:
Hi @harsha-bhushan - my plan is to productize this next. Does that sound OK, or should I work on something else?

This seems like a nice-to-have - it makes it easier for users, who don't have to pass in a read buffer size. It is something that we could add later without breaking things, though.

gregstoll:
Closing this proof of concept PR

PR #326, opened 2021-08-24
Create client wrapper generator for grpc-device tests

astarche:

Generate client wrappers for grpc-device APIs.

General philosophy: mimic betterproto by generating wrappers that take in all request fields as params and returning the response.

Note: these are experimental wrappers intended for test code. They are namespaced to reflect that.

Testing grpc-device requires a lot of boiler plate client code. This is tedious and has become messy as we add different shortcut options.

This has also gotten worse with DAQ attribute testing where there are many APIs with similar inputs/outputs.

Ran and passed all DAQ tests after updating several tests to use the client wrappers.


source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -8,11 +8,13 @@
 
 #include "device_server.h"
 #include "enumerate_devices.h"
```
astarche, line 10: Note: if you look at one file for a "Positive" impression of the change: look at the tests.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -88,10 +92,9 @@ class NiDAQmxDriverApiTests : public Test {
     if (!driver_session_) return;
 
     ::grpc::ClientContext context;
```
gregstoll, line 94: nit: can get rid of this line

gregstoll (approved):
Yeah, this seems like it will make tests easier to read, which is something I've been getting concerned about (particularly with the ones I've been writing :-) )  Thanks for doing this!

generated/nidaqmx/nidaqmx_client.h
```diff
@@ -0,0 +1,7767 @@
...
+#include <cstdint>
+#include <stdexcept>
+#include <vector>
+
+namespace nidaqmx_grpc::experimental::client {
+
+using namespace nidaqmx_grpc;
+
+template <typename TFirst, typename TSecond>
+class simple_variant
```
astarche, line 23: This is a trick to make the oneofs for raw enums work. Since this type supports implicit conversion from either TFirst or TSecond, methods taking a simple_variant will accept either argument and do the right thing. That's the theory anyway ...

This is the same behavior std::variant would have but std::variant doesn't work with out linux rt compiler.

The easiest alternative would be to only support the non-raw params, although that's limiting for the attribute accessors that often use both (and where we have more work planned for raw attribute IDs).

Pass both by pointer would work, but it's annoying for callsites. I think passing both as std::optional would also work, but we'd probably need a simple_optional for the same reason as simple_variant.

astarche:
@reckenro we're leaning towards "just going for it" and submitting a change that's more-or-less just this (I will start another PR though). Even though there are some downsides to the header-only-template-excess implementation, that's outweighed by the benefits of "stopping the bleeding" on the problems we've had in system test client code.

The one big change that I'm leaning towards making is trying to limit this to just DAQ. Then we can avoid having a ton of headers that aren't even included/built anywhere.

SparkingSpork (approved):

generated/nidaqmx/nidaqmx_client.cpp
```diff
@@ -0,0 +1,8422 @@
...
+#include <nidaqmx.grpc.pb.h>
+
+#include <cstdint>
+#include <memory>
+#include <stdexcept>
+#include <vector>
+
+namespace nidaqmx_grpc::experimental::client {
+
+AddCDAQSyncConnectionResponse
```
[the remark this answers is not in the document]
astarche: Good point. It's more of a speed bump in the header and also makes the return type *more* emphasized instead of less. Fixed.

PR #328, opened 2021-08-30
Fix up GetErrorString test to test the full message

astarche:

This test is failing with:
```
nidaqmx_driver_api_tests.cpp(1485): error: Value of: response.error_string()
Expected: is equal to "Requested value is not a supported value for this property."
Actual: "Requested value is not a supported value for this property. The property value may be invalid because it conflicts with another property."
```

This originally used `HasSubstr` but we want to do a stricter match to test `ivi-dance`. Update to match the entire string and explicitly test for extra padded null characters.

Fixes a test.

Ran and passed all DAQ tests.


gregstoll:
Whoops, sorry, I broke this. Is there a way to assert there are no spaces or NULLs at the end of the string? (that's why I made the change)

gregstoll:
(I broke this in #311, fwiw)

reckenro (approved):

PR #332, opened 2021-08-30
Use decltype for function pointers

astarche:

Change the code that declares grpc-device library function pointers to use `decltype` instead of generating the signature from function parameter metadata.

Note that `private` methods still use the old logic because they're signature is not available in the public header.

Currently all of the generic attribute accessors are broken on linux because the function pointers use the elaborated static types instead of `var_args` (this is "undefined behavior"). 

We could fix this with a new metadata mechanism, but using `decltype` has the same result without adding a new mechanism. It's also logically more similar to the windows implementation where we link against the actual method and delay load.

Ran all grpc-device tests on CentOS with no regressions.
Ran and passed all DAQ tests on CentOS.
Ran and passed all DAQ tests on Windows.


source/codegen/templates/library.h.mako
```diff
@@ -43,8 +44,14 @@ class ${service_class_prefix}Library : public ${namespace_prefix}::${service_cla
   f = functions[method_name]
   parameters = f['parameters']
   return_type = f['returns']
+  c_name = service_helpers.get_cname(functions, method_name, c_function_prefix)
 %>\
+## "Private" methods aren't in the public header, so generate the params from the metadata.
+% if service_helpers.is_private_method(f):
```
astarche, line 50: Note: Context. There's currently [one private method](https://github.com/ni/grpc-device/blob/9b335069be8ec0b4f5fb7b76dfb28f600426cdc0/source/codegen/metadata/nidcpower/functions.py#L2282) in grpc-device. There's a similar special case (for the same reason) [in library.cpp.mako](https://github.com/ni/grpc-device/blob/ba49f0d3fd6ad33bf8e3d6c7c4c2b595f37df08e/source/codegen/templates/library.cpp.mako#L70).

@gregstoll we could consider pulling out the `expand_varargs=False` condition/support. This isn't adding a ton of dead code, though.

gregstoll: Yeah, once this goes in I can clean up the `expand_varargs` stuff

gregstoll (approved):

reckenro (approved):

PR #336, opened 2021-08-31
Add documentation about size mechanisms

gregstoll:

Adds internal documentation about the size mechanisms.

Hopefully makes it clearer to developers how the size mechanisms work.

None, just added a comment.

### DIFF
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -109,11 +109,14 @@ def is_unsupported_parameter(parameter):
 # - ivi-dance: This is a two-step process. On the first call, pass in NULL, and the function will return
 #              how big the array should be (instead of an error as usual). Then, the service creates an array
 #              of that size and passes it in. Should only be used for output arrays. The size of the array
-#              (which is still passed in) is specified in the 'value' member.
+#              (which is still passed in) is specified in the 'value' member. Note that it is possible that the
+#              second call will return a "buffer too small" error (if the underlying value in the driver has changed
+#              size between calls), in which case the "dance" will start again by passing NULL.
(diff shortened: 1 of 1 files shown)


source/codegen/common_helpers.py
```diff
@@ -100,7 +100,24 @@ def is_unsupported_parameter(parameter):
...
+# These are the mechanisms that are used to specify the size of arrays/strings. Here's what they mean:
+# - fixed: the array is of a constant size. The size is specified in the 'value' member.
+# - len: the array's length needs to be passed into the function. The parameter to pass it in is specified
+#        in the 'value' member. Should only be used for input arrays. Multiple input arrays can use this
+#        mechanism with the same 'value' member - in that case, the service will enforce that those arrays
+#        are all the same length.
+# - ivi-dance: This is a two-step process. On the first call, pass in NULL, and the function will return
+#              how big the array should be (instead of an error as usual). Then, the service creates an array
+#              of that size and passes it in. Should only be used for output arrays. The size of the array
+#              (which is still passed in) is specified in the 'value' member.
```
astarche, line 112: Would it be useful to add a comment about the looping behavior and status checking? Key concern: if anyone opts into this in a future driver, we want to make sure that they "be careful" about verifying that the driver behavior matches the ivi-dance implementation. I think just mentioining the looping behavior (and maybe that it's somewhat driver-dependent) is enough to get people to consider it. At least that should work as well as anything.

astarche (approved):

source/codegen/common_helpers.py
gregstoll: I added some words to that effect - thanks!

reckenro (approved):

PR #337, opened 2021-08-31
Support legacy sessions in dcpower tests

astarche:

Update dcpower tests to be more tolerant of older versions that don't have `InitializeWithIndependentChannels`.

For tests that run with the old `InitializeWithChannels` API: fallback to that init method and continue.
For tests that don't work with `InitializeWithChannels` or are explicitly written for `InitializeWithIndependentChannels`: Skip the test.

InitializeWithIndependentChannels is a new API that hasn't been released yet on Linux. This makes it harder to test release drivers on Linux.

Ran and passed all dcpower tests on Windows.

Ran and passed all dcpower tests on CentOS with the following skips:
```
[  SKIPPED ] NiDCPowerDriverApiTest.SetAttributeViReal64_GetAttributeViReal64ReturnsSameValue
[  SKIPPED ] NiDCPowerDriverApiTest.SetAttributeViBoolean_GetAttributeViBooleanReturnsSameValue
[  SKIPPED ] NiDCPowerDriverApiTest.SetAttributeViString_GetAttributeViStringReturnsSameValue
[  SKIPPED ] NiDCPowerDriverApiTest.SetMeasureWhenAndInitiate_FetchMultiple_FetchesSuccessfully
[  SKIPPED ] NiDCPowerDriverApiTest.VoltageLevelConfiguredAndExportedToBuffer_ResetAndImportConfigurationFromBuffer_ConfigurationIsImportedSuccessfully
[  SKIPPED ] NiDCPowerSessionTest.InitializeSessionWithDeviceAndSessionName_CreatesDriverSession
[  SKIPPED ] NiDCPowerSessionTest.InitializeSessionWithDeviceAndNoSessionName_CreatesDriverSession
[  SKIPPED ] NiDCPowerSessionTest.InitializedSession_CloseSession_ClosesDriverSession
```

### DIFF
--- a/source/tests/system/nidcpower_driver_api_tests.cpp
+++ b/source/tests/system/nidcpower_driver_api_tests.cpp
@@ -1,6 +1,7 @@
 #include <gtest/gtest.h>
 
 #include "device_server.h"
+#include "nidcpower/nidcpower_client.h"
 #include "nidcpower/nidcpower_service.h"
 
 namespace ni {
@@ -8,6 +9,7 @@ namespace tests {
 namespace system {
 
 namespace dcpower = nidcpower_grpc;
+namespace client = nidcpower_grpc::experimental::client;
 
 const int kdcpowerDriverApiSuccess = 0;
 
@@ -44,19 +46,31 @@ class NiDCPowerDriverApiTest : public ::testing::Test {
 
   void initialize_driver_session()
   {
-    ::grpc::ClientContext context;
-    dcpower::InitializeWithIndependentChannelsRequest request;
-    request.set_resource_name("FakeDevice");
-    request.set_option_string("Simulate=1, DriverSetup=Model:4147; BoardType:PXIe");
-    request.set_session_name("");
-    request.set_reset(false);
-    dcpower::InitializeWithIndependentChannelsResponse response;
+    try {
+      auto independent_response = client::initialize_with_independent_channels(
+          GetStub(),
+          "FakeDevice",
+          false,
+          "Simulate=1, DriverSetup=Model:4147; BoardType:PXIe");
+      driver_session_ = std::make_unique<nidevice_grpc::Session>(independent_response.vi());
+    }
+    catch (std::runtime_error&) {
+      auto legacy_response = client::initialize_with_channels(
+          GetStub(),
+          "FakeDevice",
+          "0",
+          false,
+          "Simulate=1, DriverSetup=Model:4147; BoardType:PXIe");
 
-    ::grpc::Status status = GetStub()->InitializeWithIndependentChannels(&context, request, &response);
-    driver_session_ = std::make_unique<nidevice_grpc::Session>(response.vi());
+      driver_session_ = std::make_unique<nidevice_grpc::Session>(legacy_response.vi());
+      ASSERT_EQ(kdcpowerDriverApiSuccess, legacy_response.status());
+      is_legacy_session_ = true;
+    }
+  }
 
-    ASSERT_TRUE(status.ok());
-    ASSERT_EQ(kdcpowerDriverApiSuccess, response.status());
+  bool is_legacy_session()
+  {
+    return is_legacy_session_;
   }
 
   void initiate()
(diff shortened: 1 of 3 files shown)


source/tests/system/nidcpower_driver_api_tests.cpp
```diff
@@ -1,13 +1,15 @@
 #include <gtest/gtest.h>
 
 #include "device_server.h"
+#include "nidcpower/nidcpower_client.h"
```
astarche, line 4: @doshirohan I added a client generator in #326, that's what the new code in this class is using.

source/tests/system/nidcpower_driver_api_tests.cpp
```diff
@@ -300,8 +314,15 @@ class NiDCPowerDriverApiTest : public ::testing::Test {
   DeviceServerInterface* device_server_;
   std::unique_ptr<::nidevice_grpc::Session> driver_session_;
   std::unique_ptr<dcpower::NiDCPower::Stub> nidcpower_stub_;
+  bool is_legacy_session_{false};
 };
 
+// Some tests don't work with the legacy InitializeWithChannels API
+// because of differences in how channels are configured OR because they
+// test features of the independent channel API.
+#define GTEST_SKIP_IF_LEGACY() \
```
astarche, line 323: GTEST_SKIP includes an early return. So it needs to be a macro so we return from the test method (otherwise the later failures override the skip result).

source/tests/utilities/client_helpers.h
```diff
@@ -49,7 +49,7 @@ class simple_variant {
 inline void raise_if_error(const ::grpc::Status& status)
 {
   if (!status.ok()) {
-    throw new std::runtime_error(status.error_message());
+    throw std::runtime_error(status.error_message());
```
astarche, line 52: Oops!

source/tests/system/nidcpower_driver_api_tests.cpp
```diff
@@ -44,19 +46,31 @@ class NiDCPowerDriverApiTest : public ::testing::Test {
...
-
-    ASSERT_TRUE(status.ok());
-    ASSERT_EQ(kdcpowerDriverApiSuccess, response.status());
+    try {
+      auto independent_response = client::initialize_with_independent_channels(
+          GetStub(),
+          "FakeDevice",
+          false,
+          "Simulate=1, DriverSetup=Model:4147; BoardType:PXIe");
+      driver_session_ = std::make_unique<nidevice_grpc::Session>(independent_response.vi());
```
doshirohan, line 55: Add `ASSERT_EQ(kdcpowerDriverApiSuccess, independent_response.status());`

source/tests/system/nidcpower_session_tests.cpp
```diff
@@ -83,12 +83,18 @@ class NiDCPowerSessionTest : public ::testing::Test {
...
+
 TEST_F(NiDCPowerSessionTest, InitializeSessionWithDeviceAndSessionName_CreatesDriverSession)
 {
   dcpower::InitializeWithIndependentChannelsResponse response;
-  ::grpc::Status status = call_initialize_with_independent_channels(kTestRsrc, kOptionsString, kTestSession, &response);
+  auto status = call_initialize_with_independent_channels(kTestRsrc, kOptionsString, kTestSession, &response);
 
-  EXPECT_TRUE(status.ok());
+  GTEST_SKIP_IF_UNSUPPORTED(status);
   EXPECT_EQ(0, response.status());
```
doshirohan, line 98: We should still add `EXPECT_TRUE(status.ok())` to catch any other errors apart from NOT_FOUND for this and other tests below.

source/tests/utilities/client_helpers.h
doshirohan: uhmm.. where is this used? Or is it something indirectly used by new client generator pattern that you are using in dcpower?

doshirohan (approved):

reckenro (approved):

source/tests/utilities/client_helpers.h
astarche: Yes, that's used by the client generator. The generated tests throw on bad status so they can just return a response. I'll probably add a more specific exception class at some point, but not very many tests care about specific bad statuses.

source/tests/system/nidcpower_driver_api_tests.cpp
astarche: Two good catches! Thanks.

PR #339, opened 2021-09-01
Add support for service-specific feature_toggles.

astarche:

Add support for service-specific feature_toggles.

`is_enabled` is now detected within the service rather than in core_server.

Add `CodeReadiness` to feature toggles so that we can handle different default states.

Services can declare specific toggles that are populated in the service-specific toggle struct.

Enables the `allow_undefined_attributes` feature toggle.

Can be used as staging for future work to generate client factor methods by removing special-case logic in `core_server`.

Ran and passed all DAQ tests.
Ran and passed all integration/unit tests.
Manual testing of DAQ service with and without toggle.
Manual testing with scope example to ensure that other services are still active.


generated/nidaqmx/nidaqmx_service.h
```diff
@@ -406,9 +410,22 @@ class NiDAQmxService final : public NiDAQmx::WithCallbackMethod_RegisterSignalEv
...
   NiDAQmxLibraryInterface* library_;
   ResourceRepositorySharedPtr session_repository_;
+
+  struct NiDAQmxFeatureToggles
+  {
+    using CodeReadiness = nidevice_grpc::FeatureToggles::CodeReadiness;
+    NiDAQmxFeatureToggles(const nidevice_grpc::FeatureToggles& feature_toggles);
+
+    bool is_enabled;
+    bool is_allow_undefined_attributes_enabled;
```
astarche, line 425: Note: This will be the toggle for accessing undefined raw attributes. I haven't implemented that logic yet.

gregstoll (approved):

generated/nifake_non_ivi/nifake_non_ivi_service.cpp
```diff
@@ -673,5 +676,16 @@ namespace nifake_non_ivi_grpc {
...
 
+  bool NiFakeNonIviService::is_enabled()
+  {
+    return feature_toggles_.is_enabled;
+  }
+
+  NiFakeNonIviService::NiFakeNonIviFeatureToggles::NiFakeNonIviFeatureToggles(
+    const nidevice_grpc::FeatureToggles& feature_toggles)
+    : is_enabled(
+        feature_toggles.is_feature_enabled("nifake_non_ivi", CodeReadiness::kNextRelease))
```
reckenro, line 687: Just checking, this one is marked `NextRelease` since it's testing DAQ specific non-ivi types, right?

reckenro (approved):

generated/nifake_non_ivi/nifake_non_ivi_service.cpp
astarche: Yeah, I marked both daq and fake non ivi

```python
    'code_readiness': 'NextRelease',
```

Just so we can use fake non ivi to test that state.

PR #343, opened 2021-09-01
RFC: First pass splitting cmake into libraries

astarche:

Split out `ni_grpc_device_server_core` and `ni_grpc_device_server_core_proto` into separate libraries in separate CMakeLists.txt files.

Move grpc/protoc helpers to `GenerateGrpcSources.cmake`

GOALS:
* Enable auto-generating CMakeLists.txt for each driver
    * Removes a lot of special case/complex code from top-level CmakeLists.txt
    * Enables more easily adding features that generate different files per driver (i.e., a driver that generates multiple services).
* Reduce build time by only building files once (especially for services and pb.cc files)
* More easily allow alternate output types (i.e., a server/test-app that only includes one driver).
* Steps towards cleaning up cmake:
   * No duplicate specifications.
   * MOST CMakeLists are simple or have a single complex element to deal with.
   * (Hopefully) intuitive separation of build specification logic.

Builds ran a few tests.

### DIFF
--- a/source/server/CMakeLists.txt
+++ b/source/server/CMakeLists.txt
@@ -1,0 +1,25 @@
+add_library(ni_grpc_device_server_core
+  "device_enumerator.cpp"
+  "feature_toggles.cpp"
+  "logging.cpp"
+  "semaphore.cpp"
+  "server_configuration_parser.cpp"
+  "server_security_configuration.cpp"
+  "session_repository.cpp"
+  "session_utilities_service.cpp"
+  "shared_library.cpp"
+  "syscfg_library.cpp")
+
+target_include_directories(ni_grpc_device_server_core
+  PUBLIC ${CMAKE_SOURCE_DIR}/imports/include
+  PUBLIC "../" # HACK should be able to sanitize this, but include statements expect it to be server/<whatever>.h
+)
+
+target_link_libraries(ni_grpc_device_server_core
+  ni_grpc_device_server_core_proto
+  ${_REFLECTION}
+  ${_GRPC_GRPCPP}
+  ${_PROTOBUF_LIBPROTOBUF}
+  ${CMAKE_DL_LIBS}
+  nlohmann_json::nlohmann_json
+  )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,29 +10,8 @@ endif()
 #----------------------------------------------------------------------
 # Use the grpc targets directly from this build, only when not cross-compiling.
 #----------------------------------------------------------------------
-if(CMAKE_CROSSCOMPILING)
-  find_program(_PROTOBUF_PROTOC protoc)
-  find_program(_GRPC_CPP_PLUGIN_EXECUTABLE grpc_cpp_plugin)
-
-  if(NOT _GRPC_DEVICE_NILRT_LEGACY_TOOLCHAIN)
-    find_package(gRPC REQUIRED)
-    find_library(_REFLECTION grpc++_reflection)
-    find_library(_GRPC_GRPCPP grpc++)
-    find_library(_PROTOBUF_LIBPROTOBUF protobuf)
-  else()
-    add_subdirectory(third_party/grpc ${CMAKE_CURRENT_BINARY_DIR}/grpc EXCLUDE_FROM_ALL)
-    set(_REFLECTION grpc++_reflection)
-    set(_GRPC_GRPCPP grpc++)
-    set(_PROTOBUF_LIBPROTOBUF libprotobuf)
-  endif()
-
-else()
+if(NOT CMAKE_CROSSCOMPILING OR _GRPC_DEVICE_NILRT_LEGACY_TOOLCHAIN)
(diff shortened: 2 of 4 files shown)


source/server/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+  "server_configuration_parser.cpp"
+  "server_security_configuration.cpp"
+  "session_repository.cpp"
+  "session_utilities_service.cpp"
+  "shared_library.cpp"
+  "syscfg_library.cpp")
+
+target_include_directories(ni_grpc_device_server_core
+  PUBLIC ${CMAKE_SOURCE_DIR}/imports/include
+  PUBLIC "../" # HACK should be able to sanitize this, but include statements expect it to be server/<whatever>.h
```
astarche, line 15: Note: these libraries don't *need* to be in separate CmakeLists.txt files. They could be moved back to the top-level CMakeLists.txt.

I think the split worked a little better for core_proto than the server. For this to really be coherent, we would probably need to move the source folders around some (?). Is an ooch change like this worth it? Or is it better to leave it in the top-level unless/until we spec out how we want to organize it?

Obviously any auto-generated driver-specific CMakeLists.txt files *would* need to be separate files.

astarche: Also, any auto-generated CMakeLists.txt files would need to be generated at-or-before cmake configure time. Not during the build like the current codegen pass.

That’s something to work through, but I think it’s ok. The rest of the codegen can still run during the build. So youd only need to reconfigure if you’re adding a new generated output file or dependency. Right now, to do that, you need to (a) update the metadata, (b) update cmakelists, (c) re-run configure. 

So we wouldn’t be adding a step there and would make the process easier in the normal case (add a new service or opt into an existing feature) and more powerful in novel cases.

source/server/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+)
+
+target_link_libraries(ni_grpc_device_server_core
+  ni_grpc_device_server_core_proto
+  ${_REFLECTION}
+  ${_GRPC_GRPCPP}
+  ${_PROTOBUF_LIBPROTOBUF}
+  ${CMAKE_DL_LIBS}
```
astarche, line 23: Note: I think only the targets that include the driver libraries needs CMAKE_DL_LIBS (at this point that's the server exe and the test exes; eventually they may have their own library/libraries).

SparkingSpork (commented):
Now that I've spent some time looking at this further, I have a couple of comments:
- It would be interesting to see if it's possible to separate the code generation into its own CMakeLists.txt as well. Specifically, using the `add_custom_command` with `add_custom_target` and having other targets depend on the custom target instead of the lists `nidriver_service_srcs` and `nidriver_client_srcs`. I'm not 100% sure that would even work, but it would help simplify the main CMakeLists,txt further. That might be better done on a driver level instead of the current catch-all. 
- Have you tried using `include` instead of `add_subdirectory`? As far as I understand, `include` works such that the CMake files added in that way share the caller's variable scope whereas `add_subdirectory` creates a new variable scope for each. It might make some of the module reuse easier.
- If you're adding a `cmake` directory, it probably makes sense to move the `scripts/cmake/nilrt-x86_64.cmake` into that directory as well for consistency. That should be a separate PR from this since it would require updates to the nilrt workflow as well.

I'm excited to see how this ends up. It seems like not only will this potentially split the drivers into their own CMakeLists.txt, but it could make it really easy to transition if we ever decide to use dynamic libraries instead of static for some of this.

source/server/CMakeLists.txt
SparkingSpork: We could have the codegen run as part of our workflow, or even add a script that did the generation and CMake configure instead. Having the generated files be created by CMake was a trick to make sure they were always up to date after metadata changes but it might not have been the best approach.

astarche: Yeah, that's a good point. Running during build is definitely useful during development. And it semi-enforces that the generated source is up-to-date, which is also useful for development. I could imagine creating a separate "build.py" script that generates code and then invokes cmake. Especially if we started doing more python stuff (venvs, pytest, etc) so that it's more awkward for cmake to drive the process. And then we'd just have to retrain ourselves to use that instead of cmake (or to use a custom build action instead of the cmake vscode plugin). ... And figure out how to handle the different cmake configurations we use (cross-compile, ninja generator, -j, etc).

We've also been lucky (?) that codegen is still relatively cheap to run (cheaper than protoc, cheaper than compiling the generated code).

For generating cmakelists. It's pretty easy to add an `execute_process` to run during configure. I think it would also be worthwhile to re-run the codegen during build and "fail on change" or something like that.

SparkingSpork: For the "fail on change" comment - I've considered something like that in the past. Specifically, I think our PR builds should fail if the codegen files change during the build (e.g., someone didn't check-in the changed code-gen files). It's usually easy to catch but that would be another nice to have in the future.

PR #345, opened 2021-09-02
Add GetTaskAttributeUInt32 to analog-input example

astarche:

Read `TASK_ATTRIBUTE_NUM_CHANS` to calculate `array_size_in_samps` in the analog-input example. 

We want to include attribute accessors in the examples as a way of documenting that API.

Ran included example, double-checked the response value was correct (`1`).

### DIFF
--- a/examples/nidaqmx/analog-input-betterproto.py
+++ b/examples/nidaqmx/analog-input-betterproto.py
@@ -26,8 +26,8 @@
 #
 # Server machine's IP address, port number, and physical channel name can be passed as separate command line arguments.
 #   > python analog-input-betterproto.py <server_address> <port_number> <physical_channel_name>
+# To acquire data from multiple channels, pass in a list or range of channels (i.e., Dev1/ai0:3).
 # If they are not passed in as command line arguments, then by default the server address will be "localhost:31763", with "Dev1/ai0" as the physical channel name
-
 import asyncio
 from nidevice import nidaqmx_grpc
 import sys
@@ -87,17 +87,25 @@ async def main():
         )
 
         await raise_if_error(await daq_service.start_task(task=task))
+        
+        response = await daq_service.get_task_attribute_u_int32(
+            task=task,
+            attribute_raw=nidaqmx_grpc.TaskUInt32Attributes.TASK_ATTRIBUTE_NUM_CHANS)
+        await raise_if_error(response)
+        number_of_channels = response.value
 
         response = await daq_service.read_analog_f64(
             task=task,
             num_samps_per_chan=100,
-            array_size_in_samps=100,
+            array_size_in_samps=number_of_channels * 100,
             fill_mode_raw=nidaqmx_grpc.GroupBy.GROUP_BY_GROUP_BY_CHANNEL,
             timeout=10.0,
         )
         await raise_if_error(response)
 
-        print(f"Acquired {response.samps_per_chan_read} samples.")
+        print(
+            f"Acquired {len(response.read_array)} samples.",
+            f"({response.samps_per_chan_read} samples per channel)")
         print(f"First 5 samples: {response.read_array[:5]}")
     except GRPCError as e:
         if e.status.name == "UNIMPLEMENTED":
--- a/examples/nidaqmx/analog-input.py
+++ b/examples/nidaqmx/analog-input.py
@@ -17,7 +17,8 @@
 #
 # Server machine's IP address, port number, and physical channel name can be passed as separate command line arguments.
 #   > python analog-input.py <server_address> <port_number> <physical_channel_name>
-# If they are not passed in as command line arguments, then by default the server address will be "localhost:31763", with "Dev1/ai0" as the physical channel name
+# To acquire data from multiple channels, pass in a list or range of channels (i.e., Dev1/ai0:3).
+# If they are not passed in as command line arguments, then by default the server address will be "localhost:31763", with "Dev1/ai0" as the physical channel name.
 import grpc
 import sys
 
--- a/examples/nidaqmx/analog-input.py
+++ b/examples/nidaqmx/analog-input.py
@@ -72,6 +72,8 @@ try:
 
     RaiseIfError(client.StartTask(nidaqmx_types.StartTaskRequest(task=task)))
 
+    # Get the number of channels. This may be greater than 1 if the physical_channel
+    # parameter is a list or range of channels.
     response = client.GetTaskAttributeUInt32(
         nidaqmx_types.GetTaskAttributeUInt32Request(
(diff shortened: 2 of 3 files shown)


examples/nidaqmx/analog-input.py
```diff
@@ -73,10 +72,16 @@ def RaiseIfError(response):
 
     RaiseIfError(client.StartTask(nidaqmx_types.StartTaskRequest(task=task)))
```
astarche, line 74: @harsha-bhushan should we add this to all of the input examples?

examples/nidaqmx/analog-input.py
```diff
@@ -73,10 +72,16 @@ def RaiseIfError(response):
...
 
+    response = client.GetTaskAttributeUInt32(
```
gregstoll, line 75: Do you think it's worth adding a comment here that this isn't necessary for this example (since we know there's only one channel), but just a demonstration of how to do it?

gregstoll (approved):

examples/nidaqmx/analog-input.py
astarche, line 74: @harsha-bhushan @gregstoll I tried to think of something for the basic examples that is easy to understand. Hopefully someone can look at this and understand (a) how attribute accesses work and (b) what this particular attribute access does and whether they would need to use it in their program.

I also don't want to add a lot of noise to these examples by reading/writing a bunch of attributes.

I can't think of good examples for setting attributes. The Buffer attributes are easy ones that might be useful. But if this were an example for a true beginner, they might think that setting those attributes is recommended or necessary for simple acquisition (it's not). We could set many of the attributes that are part of the initial config of the sample clock or the channel (AI_MAX, channel name, Rate, etc). 

That's redundant but mostly harmless. Some of the Read attrbutes would be reasonable:
```cpp
DAQmxSetReadAutoStart
DAQmxSetReadOverWrite
```
But those have the same problem as the buffer attributes.

Greg's advanced example has a lot of attribute accesses.
We could also consider an example *just* for accessing attributes that doesn't actually do any acquisition.

examples/nidaqmx/analog-input.py
astarche: Yes, I think so. Good idea.

examples/nidaqmx/analog-input.py
harsha-bhushan: Sorry if my email was confusing. 
As long as any of the shipping example contains accessing attributes, that should be good. Yes, I agree with your point that we should not forcefully add attributes where it is not needed as that can misguide new users. If the advanced examples are covering it that is good enough, we need not have to put them in these examples or create an example just to showcase attribute access.

harsha-bhushan: We anyway plan to add some documentation on generic attribute get/set in the DAQmx page of the wiki. we can include this scenario too & link to the example that contains the attribute access. Since we have quite a few things to document, it might be good to have one section on 'attributes' in the DAQmx page of the wiki instead of putting it under an already existing section.

harsha-bhushan: You could revert back this change as I looked at the triggering example that nicely covers the usage of get/set attributes. Between that example & documentation users should be able to understand the usage. We can keep the analog input simple.

harsha-bhushan: The more I'm looking at the usage here, the more I'm feeling we should not be doing this :p 
Just thinking out loud a few options of creating genuine scenarios:
1.Tweak one of the examples to have multiple channels & then use this way to determine the array? (every-n-samples example or the synchronous input output example that we are yet to create) 
2.Duplicate this example but have multiple channels & name it analog-input-multi-channel & show this usage of setting array size
Any other options too. We can pick whatever is simple & quick.

astarche: Hah. This change actually has a use case *in this script*, I just didn't realize it. You can pass in a "list or range of channels" to the create channel APIs!

With my slight change to the `print` out in the latest update:

```
(.venv) λ python analog-input.py 10.0.62.17 31763 gRPCSystemTestDAQ/ai0:3
Acquired 4000 samples (1000 samples per channel)
```

astarche: PS: thanks for taking the time to provide detailed feedback!

harsha-bhushan: great. then it would be worth adding this in other examples too. and in line 20 - code documentation, we can add "to acquire from 'n' consecutive channels, pass 'Dev1/ai0:n' as parameter". you could rephrase to something better.

What is your thought on this?

astarche: Good idea.

The standard phrasing in the [C API docs ](https://zone.ni.com/reference/en-XX/help/370471AM-01/daqmxcfunc/daqmxcreateaivoltagechan/) is "The names of the physical channels to use to create virtual channels. You can specify a [list or range](https://zone.ni.com/reference/en-XX/help/370466AH-01/mxcncpts/physchannames/) of physical channels." There are several different valid syntaxes.

I think I will at least use the "list or range" terminology and give an example like "i.e., Dev/ai0:n specifies a range of channels from 0 to n".

If we can actually link the help concept, that would be great. I don't know if those links are stable (and ofcourse we're expecting the help to change soon ...)

astarche: I updated all of the analog input examples. With digital input, you often end up with only one "channel" even when you pass in a list/range of inputs. So, the same mechanism can be used, but it's not as intuitive or as common. I don't want to make those examples more complicated or potentially misleading.

examples/nidaqmx/analog-input-betterproto.py
```diff
@@ -87,17 +87,25 @@ async def raise_if_error(response):
         )
 
         await raise_if_error(await daq_service.start_task(task=task))
+        
+        response = await daq_service.get_task_attribute_u_int32(
```
gregstoll, line 91: sigh, it seems like this should be `get_task_attribute_uint32` (with no underscore between `u` and `int32`), but I'm not sure it's worth fixing - what do you think?

gregstoll: oh, this is generated by `betterproto`, isn't it? Nevermind then :-)

gregstoll (approved):
This is an elegant solution for adding attribute usage to our examples!

examples/nidaqmx/analog-input-betterproto.py
astarche: Yeah, unfortunately. Can't really blame them for this one because `UInt32` a pure special-case. (They also `_` between acronyms, which is closer to a bug). We could change the method on our end to `Uint32`. But `UInt32` is pretty ingrained at NI and, I guess for that reason, `Uint32` looks weird to me.

PR #347, opened 2021-09-02
Remove -j 2 on msvc builds

astarche:

Remove the `-j 2` cmake flag on msvc builds.

There have been several file permission errors on windows workflow builds since #344 .

That change made a big difference in linux build times, but did not have a clear benefit for msvc (which significantly faster before).

CMake's`-j` flag on linux has a clear mapping to the very common `-j` flag for `make`. On msvc, the mapping is less clear (and apparently less effective).

Build succeeds. Verified that `-j 2` is passed to cmake on linux but not windows.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -25,13 +25,17 @@ jobs:
             name: "Windows Build", artifact: "Windows.tar.xz",
             os: windows-latest,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
-            cmake_platform: "-A x64"
+            cmake_platform: "-A x64",
+            # On msvc, the -j flag does not help build times and causes intermittent
+            # file permission errors.
+            cmake_number_of_jobs: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -25,13 +25,17 @@ jobs:
             name: "Windows Build", artifact: "Windows.tar.xz",
             os: windows-latest,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
-            cmake_platform: "-A x64"
```
astarche, line 28: Note: `cmake_platform` is another os-specific cmake param (to configure instead of build), so I'm using the same technique for number_of_jobs.

astarche: @gregstoll Charlie is out this week and this is failing ~50%. Can you sanity check buddy this and I'll check back with Charlie next week.

gregstoll (approved):

PR #348, opened 2021-09-02
Add feature toggle to allow undefined attributes with attribute_raw

astarche:

Add a feature toggle to the DAQ service to allow accessing undefined attributes with `attribute_raw`.

Accessing undefined attributes can potentially crash the server if it's a real attribute with the wrong datatype. But we'd like to leave open the option of accessing hidden/missing attributes if a customer needs it. Adding a toggle for access through `attribute_raw` make the server "safe-by-default" but gives us a fallback option.

Ran and passed all DAQ tests.
Ran new fake non-ivi test covering this case.
Manual testing with DAQ server with toggle enabled.
* Note: this is tricky to test with out `DeviceServerInterface::Singleton` implementation because we only init the server once per process. I think it would work to migrate the singleton to a test fixture without a huge performance penalty. But I don't think we want to test with multiple active services. So we would have to migrate all tests to use fixtures for that.

### DIFF
--- a/source/tests/unit/ni_fake_non_ivi_service_tests.cpp
+++ b/source/tests/unit/ni_fake_non_ivi_service_tests.cpp
@@ -955,6 +955,30 @@ TEST_F(NiFakeNonIviServiceTests, SetMarbleAttributeRawWithInvalidAttribute_Passe
   EXPECT_EQ(kDriverSuccess, response.status());
 }
 
+struct NiFakeNonIviServiceAllowUndefinedAttributesTests : public NiFakeNonIviServiceTests {
+  NiFakeNonIviServiceAllowUndefinedAttributesTests()
+      : NiFakeNonIviServiceTests(
+            nidevice_grpc::FeatureToggles({{"nifake_non_ivi.allow_undefined_attributes", true}}))
+  {
+  }
+};
+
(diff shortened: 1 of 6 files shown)


source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -955,6 +955,30 @@ TEST_F(NiFakeNonIviServiceTests, SetMarbleAttributeRawWithInvalidAttribute_Passe
...
 
+struct NiFakeNonIviServiceAllowUndefinedAttributesTests : public NiFakeNonIviServiceTests {
+  NiFakeNonIviServiceAllowUndefinedAttributesTests()
+      : NiFakeNonIviServiceTests(
+            nidevice_grpc::FeatureToggles({{"nifake_non_ivi.allow_undefined_attributes", true}}))
+  {
+  }
+};
+
+TEST_F(NiFakeNonIviServiceAllowUndefinedAttributesTests, ServiceWithAllowUndefinedAttributes_SetMarbleAttributeRawWithInvalidAttribute_PassesZeroAttributeToLibrary)
```
reckenro, line 966: The Then clause in the test name should be `PassesRawAttributeToLibrary`.

reckenro (approved):

PR #353, opened 2021-09-14
Add support for in/out ivi-dance-with-a-twist

astarche:

Add support for `ivi-dance-with-a-twist` when `value` and `value_twist` are the same. In this case, a single in/out parameter is used instead of a separate input and output,

RFSA uses this convention on several methods. [Example](https://zone.ni.com/reference/en-XX/help/372058U-01/rfsacref/cvinirfsa_getrelayoperationscount/).

Ran and passed included test.
Builds with some RFSA entry points as part of a larger change.

### DIFF
--- a/source/codegen/service_helpers.py
+++ b/source/codegen/service_helpers.py
@@ -75,8 +75,6 @@ def create_args(parameters):
             result = f'{result}({type_without_brackets}*){parameter_name}.data(), '
         elif parameter['type'] in {"ViBoolean[]", "ViSession[]", "ViInt16[]"}:
             result = f'{result}{parameter_name}.data(), '
-        elif parameter.get('is_size_param', False) and is_twist_mechanism:
-            result = f'{result}{twist_value_name}, '
         elif 'callback_params' in parameter:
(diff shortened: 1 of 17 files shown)


source/codegen/service_helpers.py
```diff
@@ -75,8 +75,6 @@ def create_args(parameters):
             result = f'{result}({type_without_brackets}*){parameter_name}.data(), '
         elif parameter['type'] in {"ViBoolean[]", "ViSession[]", "ViInt16[]"}:
             result = f'{result}{parameter_name}.data(), '
-        elif parameter.get('is_size_param', False) and is_twist_mechanism:
```
astarche, line 78: Note: I thought this special case in `create_args` was pretty confusing and it also broke this change. This is replaced with the new bit of mako:


```
% if not is_in_out_twist:
        auto ${size_param_name} = ${twist_param_name};
% endif
```
Which allows create_args to treat the size_param as a regular param, instead of special-case swapping out the twist param.

reckenro (approved):

PR #356, opened 2021-09-17
Update workflow to use ubuntu 18.04 runner

astarche:

Update workflow to use ubuntu 18.04 runner.
Change artifact names to reflect the system glibc for 18.04 (2.27)

GH is dropping support for ubuntu 16.04 runners.
We don't know of current clients that depend on older glibc. If we have requests, we can either have them build from source or set up a containerized workflow to do that build.

Standard pre-submit build.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -32,7 +32,7 @@ jobs:
           }
         - {
             name: "Ubuntu Build", artifact: "Linux.tar.xz",
-            os: ubuntu-16.04,
+            os: ubuntu-18.04,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
             cmake_platform: "",
             cmake_number_of_jobs: "-j 2"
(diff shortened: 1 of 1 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -116,7 +116,7 @@ jobs:
     - name: Tar Linux Server Binaries
       if: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/releases')) && (runner.os == 'Linux') }}
       run: >-
-        tar -cvzf ni-grpc-device-server-linux-glibc2_23-x64.tar.gz
+        tar -cvzf ni-grpc-device-server-linux-glibc2_27-x64.tar.gz
```
SparkingSpork, line 119: Definitely a choice I made previously that was short-sighted, but we should probably consider storing the glibc version in a variable to make updating easier in the future.

SparkingSpork (approved):

PR #357, opened 2021-09-20
Support mapped enums in generated test clients

astarche:

Fix the test client codegen to support mapped enums.
Add fake and fake-non-ivi clients to the IntegrationTestsRunner component to build these clients.

Currently mapped enums are only used in the fake service. The clients for the fake services are generated but are not part of the build. The existing implementation does not work for mapped enums.

RFSA and RFSG have several mapped enums for string datatypes. Adding this codegen will allow us to build the test clients for those drivers when they are added to grpc-device.

Ran and passed included tests.
* A new fake end-to-end tests component. The primary fake testing strategy calls directly into the service, so we need a different setup to test client wrappers.
Builds RFSA/RFSG as part of a larger change.

### DIFF
--- a/.github/workflows/build_nilrt.yml
+++ b/.github/workflows/build_nilrt.yml
@@ -25,13 +25,6 @@ jobs:
       with:
         access_token: ${{ github.token }}
 
-    # If 2 large cpp files are compiled in parallel, we can run out of memory.
-    # Increase swap space to protect against this.
-    - name: Set Swap Space
-      uses: pierotofy/set-swap-space@49819abfb41bd9b44fb781159c033dba90353a7c
-      with:
-        swap-size-gb: 8
-
     - name: Checkout Repo
       uses: actions/checkout@v2
 
@@ -91,7 +84,7 @@ jobs:
       shell: cmake -P {0}
       run: |
         execute_process(
-          COMMAND cmake --build $ENV{GITHUB_WORKSPACE}/third_party/grpc/build -j 2
+          COMMAND cmake --build $ENV{GITHUB_WORKSPACE}/third_party/grpc/build
           RESULT_VARIABLE result
           OUTPUT_VARIABLE output
           ERROR_VARIABLE output
@@ -128,7 +121,7 @@ jobs:
       shell: cmake -P {0}
       run: |
         execute_process(
-          COMMAND cmake --build build --config $ENV{BUILD_TYPE} -j 2
+          COMMAND cmake --build build --config $ENV{BUILD_TYPE}
           RESULT_VARIABLE result
           OUTPUT_VARIABLE output
           ERROR_VARIABLE output
--- a/.github/workflows/build_nilrt.yml
+++ b/.github/workflows/build_nilrt.yml
@@ -25,6 +25,13 @@ jobs:
       with:
         access_token: ${{ github.token }}
 
+    # If 2 large cpp files are compiled in parallel, we can run out of memory.
+    # Increase swap space to protect against this.
+    - name: Set Swap Space
+      uses: pierotofy/set-swap-space@49819abfb41bd9b44fb781159c033dba90353a7c
+      with:
+        swap-size-gb: 8
+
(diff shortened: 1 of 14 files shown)


source/codegen/client_helpers.py
```diff
@@ -20,8 +21,7 @@ class ParamMechanism(Enum):
   ])
 
 
-def create_params(func: dict) -> List[str]:
-  client_params = get_client_parameters(func)
+def to_parameter_list(client_params: List[ClientParam]) -> List[str]:
```
astarche, line 24: Changing methods to take in `List[ClientParam]` avoids needing to pass enums into every method on the API. We only need the enums to generate the `ClientParam`s.

source/codegen/common_helpers.py
```diff
@@ -604,3 +604,18 @@ def get_split_attributes_by_type(config):
 
 def supports_raw_attributes(config: dict) -> bool:
     return config.get('supports_raw_attributes', False)
+
+
+def get_enum_value_cpp_type(enum: dict) -> str:
```
astarche, line 609: This method was moved from `service_helpers` (because it is now used in the client) and renamed from `python_to_c` (because that was too low level of a name)

source/codegen/proto_helpers.py
```diff
@@ -1,12 +1,11 @@
 import common_helpers
 
-def should_allow_alias(enums):
-  if enums.get("generate-mappings", False):
+def should_allow_alias(enum):
```
astarche, line 3: Note: the above code can be removed because this code has been updated to handle the duplicate-non-zero values. we previously had 2 versions of this code that each handled a different special case.

CMakeLists.txt
```diff
@@ -255,6 +255,7 @@ endif()
 # Link test executable against gtest
 add_executable(IntegrationTestsRunner
     "source/tests/utilities/run_all_tests.cpp"
+    "source/tests/integration/ni_fake_service_tests_endtoend.cpp"
```
SparkingSpork, line 258: Should we consider a `source/tests/endtoend` or `source/tests/acceptance` instead of appending `endtoend` to the test files? And possibly a new test runner? Or will this be the only `endtoend` test source?

SparkingSpork (approved):

CMakeLists.txt
astarche: Good question. I think the `integration` test category that these are currently in is a good one. In this context, end-to-end is a subcategory of Integration tests that run an in-process server and connect a client to it. Unlike system tests (which are all end-to-end) they don't have external deps. There might be a more-perfect name for that category. I think your suggestions are less differentiated from 'system', which I don't like.

These files technically don't need an end-to-end suffix. There's another file with the same name, but it's in a different directory. I actually think that's helpful though. Both because it's easier for navigation and because it quickly (if partially) answers the question "what is a fake integration test?"

For fake tests, we only use end-to-end tests when our primary unit-testing strategy doesn't work. This is pretty rare. If we decided to make this the primary fake testing strategy, maybe like a FakeEndToEnd runner would be a better option.

.github/workflows/build_nilrt.yml
```diff
@@ -25,6 +25,13 @@ jobs:
       with:
         access_token: ${{ github.token }}
 
+    # If 2 large cpp files are compiled in parallel, we can run out of memory.
```
astarche, line 28: TODO: This change tickled the build so that the nilrt build is running out of memory. This is something we've been pushing up against since I enabled `-j2`. It just depends on which objects get built in parallel. I'm going to try out this fix, which has worked in other branches and then re-ping owners on the proposal.

reckenro (approved):

.github/workflows/build_nilrt.yml
astarche: I'm going to remove `-j2` instead of doing the more swap option. Running with `-j2` and more swap takes the same amount of time. Discussed with @SparkingSpork

PR #361, opened 2021-09-21
RFSG initial service

gregstoll:

Adds the RFSG service behind a feature toggle. Add basic system tests for the RFSG session APIs.

Initial work on adding RFSG support.

Ran new system tests.


SparkingSpork (approved):

source/codegen/metadata/nirfsg/functions.py
```diff
@@ -0,0 +1,98 @@
...
+                'direction': 'in',
+                'type': 'ViStatus'
+            },
+            {
+                'name': 'errorMessage',
+                'direction': 'out',
+                'type': 'ViChar[]',
+                'size': {
+                    'mechanism': 'fixed',
+                    'value': 256
```
astarche, line 92: The examples use `niRFSG_GetError`. In the RFSA header, this `error_message` variant is in a section for "Required VXIplug&play Functions". And, there are several idiosyncracies to those methods. Including not having a size mechanism. And I don't think any of them are in CVI. I have user stories to audit these as part of making sure we exclude everything that should be excluded. Or ... choose a legitimate fixed max size.

TL;DR: I think we're going to remove this method. Try using GetError instead,

source/tests/system/nirfsg_session_tests.cpp
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  virtual ~NiRFSGSessionTest() {}
+
+  std::unique_ptr<rfsg::NiRFSG::Stub>& GetStub()
+  {
+    return nirfsg_stub_;
+  }
+
+  ::grpc::Status call_init_with_options(const char* resource_name, const char* option_string, const char* session_name, rfsg::InitWithOptionsResponse* response)
```
astarche, line 37: Are you not using the generated clients because these session tests have some expected grpc-status errors? Or is this just based on the existing session tests?

For the [RFSA tests](https://github.com/astarche/grpc-device/blob/4b56406b7864898850c6ff066bf4a4d161115292/source/tests/system/nirfsa_driver_api_tests.cpp) I did a couple things different:
* Skip the session tests class (just put a simple open/close in the api tests)
* Use the generated client for everything
* Have a very light test class fixture (just managing the stub).

We don't touch the session tests classes much after their written. And it also makes some sense for them to be "more low-level" (though I don't know if that's required). *So I think it's fine to leave this one as-is. *

But I'd like to use the generated client more in the driver api tests to avoid this mass of request building code. That could still have some convenience wrappers for commonly-used-methods-with-uncommonly-used-params (like [this one](https://github.com/ni/grpc-device/blob/361d29b2e906b163ad9e29e04158d5bacb32955d/source/tests/integration/ni_fake_service_tests_endtoend.cpp#L67))

source/codegen/metadata/nirfsg/functions.py
```diff
@@ -0,0 +1,98 @@
...
+            {
+                "name": "reset",
+                "direction": "in",
+                "type": "ViBoolean"
+            },
+            {
+                "name": "vi",
+                "direction": "out",
+                "type": "ViSession",
+                "xxgrpc_type": "nidevice_grpc.Session"
```
astarche, line 26: xx? I think this isn't needed.

astarche (approved):

source/codegen/metadata/nirfsg/functions.py
gregstoll: Ah, OK. Size issue aside, I like `error_message` better because it takes in an error code, instead of just returning the last one the session hit. But I switched this to `GetError` and updated the tests - thanks!

source/codegen/metadata/nirfsg/functions.py
gregstoll: hah, yup, should have used `TODOTODO` or something :-)

source/tests/system/nirfsg_session_tests.cpp
gregstoll: Yeah, I was trying to copy another IVI session test class while using the generated client, and somehow I got mixed up and failed. I'll leave this as-is but definitely will use the generated client stuff for the driver API tests. Thanks!

source/codegen/metadata/nirfsg/functions.py
astarche: Oops. I missed that. Yeah, I like that too. We'll figure it out.

reckenro (approved):

PR #365, opened 2021-09-22
Fix output custom structs that are ivi-dance-with-a-twist

gregstoll:

Fixes some issues with custom struct code generation. These will be needed for some RFSA and RFSG methods.

Will be necessary for RFSA/G.

Added tests in `nifake` that failed before this change (actually, `nifake` failed to generate) and now pass.


source/codegen/templates/service_helpers.mako
```diff
@@ -672,6 +672,14 @@ ${copy_to_response_with_transform(source_buffer=parameter_name, parameter_name=p
         response->mutable_${parameter_name}()->resize(${actual_size_param_name});
 %       elif common_helpers.is_string_arg(parameter):
         response->mutable_${parameter_name}()->resize(${actual_size_param_name}-1);
+%       elif common_helpers.is_struct(parameter):
+        {
+          auto shrunk_size = ${actual_size_param_name};
+          auto current_size = response->mutable_${parameter_name}()->size();
+          if (shrunk_size != current_size) {
+            response->mutable_${parameter_name}()->DeleteSubrange(shrunk_size, current_size - shrunk_size);
```
astarche, line 680: I feel like it'd be helpful to have a mako comment here but ... I'm not sure of a concise and useful way to explain it.

`RepeatedField` has a `Resize` that just adjusts the index in the "make smaller" case.
`RepeatedPtrField` owns all the pointed-to structs so it can't implement the same API (at least not the same way?): it needs to call destructors.

It would have been nice if RepeatedPtrField just implemented resize by doing what you're doing, but I guess they have their reasons.

Maybe just "RepeatedPtrField doesn't support Resize, use DeleteSubrange to Delete any extra elements"?

astarche (approved):

reckenro (approved):

PR #366, opened 2021-09-22
Add initial RFSG metadata

gregstoll:

Adds RFSG methods that don't have custom struct parameters.

Progress towards completing RFSG.

Added a few system tests, more will come later.


source/codegen/common_helpers.py
```diff
@@ -304,6 +304,8 @@ def get_size_mechanism(parameter):
 
 def get_size_expression(parameter):
     size_mechanism = get_size_mechanism(parameter)
+    if 'size' not in parameter:
```
gregstoll, line 307: note: I added this to give a better error message when an out array parameter is missing its size.

source/codegen/service_helpers.py
```diff
@@ -313,7 +313,7 @@ def to_cpp_readiness(user_readiness: str) -> str:
     return f'CodeReadiness::k{user_readiness}'
     
 
-def get_enums_to_map(functions: dict, enums: dict) -> dict:
+def get_enums_to_map(functions: dict, enums: dict) -> List[str]:
```
gregstoll, line 316: note: fixed this wrong type annotation

source/codegen/templates/service_helpers.mako
```diff
@@ -10,7 +10,7 @@
   output_parameters = [p for p in parameters if common_helpers.is_output_parameter(p)]
   session_output_param = next((parameter for parameter in output_parameters if parameter['grpc_type'] == 'nidevice_grpc.Session'), None)
   resource_handle_type = session_output_param['type']
-  session_output_var_name = session_output_param['cppName']
+  session_output_var_name = common_helpers.camel_to_snake(session_output_param['cppName'])
```
gregstoll, line 13: note: ran into this problem for the first time so I fixed it.

source/tests/system/nirfsg_driver_api_tests.cpp
```diff
@@ -0,0 +1,130 @@
...
+        false,
+        OPTIONS);
+
+    ASSERT_EQ(krfsgDriverApiSuccess, response.status());
+    driver_session_ = std::make_unique<nidevice_grpc::Session>(response.vi());
+  }
+
+  void close_driver_session()
+  {
+    ::grpc::ClientContext context;
```
astarche, line 64: I guess you added these before our previous thread? You shouldn't need these anymore. I'd also recommend a more concise name for `GetSessionId` so that you can inline client calls (at least most of the time): `auto response = client::whatever(stub(), session(), ...)`

astarche (approved):

source/tests/system/nirfsg_driver_api_tests.cpp
gregstoll: Yeah, you're right. I cleaned this all up, thanks!

reckenro (approved):

astarche:
Bypassing @SparkingSpork . Simple CmakeLists change that's required for follow on changes.

PR #368, opened 2021-09-24
Support codegen for off-by-one strlen-bug APIs

astarche:

Add support for a `strlen-bug` tag on size mechanisms to ensure that we allocate a bigger buffer for driver APIs that incorrectly report strlen instead of adding 1 for the required C string buffer size.

Update codegen to include the standard `- 1` in the size_expression so that it doesn't always require a special case and so it can be conditional on the above tag.

Add `trim_trailing_nulls` to trim down the string when the strlen bug has been fixed, currently only implemented for ivi-dance-with-a-twist.

Fixes [AB#1639825](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1639825)

`niRFSA_GetRelayName` has this bug which currently overwrites the buffer and truncates the string. They are planning to fix the bug so we need to support the current behavior and future behavior. See [AB#1639788](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1639788).

Ran and passed included tests.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -94,7 +95,7 @@ namespace nidaqmx_grpc {
       
         std::string device_name_out;
         if (device_name_out_buffer_size > 0) {
-            device_name_out.resize(device_name_out_buffer_size-1);
+            device_name_out.resize(device_name_out_buffer_size - 1);
```
astarche, line 98: Sorry: half the size expressions had spaces and half didn't.

generated/nidcpower/nidcpower_service.cpp
```diff
@@ -11,6 +11,7 @@
 #include <iostream>
 #include <atomic>
 #include <vector>
+#include <server/converters.h>
```
astarche, line 14: I'm including my helper unconditionally even though only a few services use it now. I'd like it to be easier to add standard utilities there since I think it's easier to grok than generating them inline.

generated/nidigitalpattern/nidigitalpattern_service.cpp
```diff
@@ -1760,7 +1761,7 @@ namespace nidigitalpattern_grpc {
       
         std::string value;
         if (buffer_size > 0) {
```
astarche, line 1763: Note this is a case where we need both the actual buffer_size and the C++ string size expression. So the codegen differentiates those  2.

generated/nifake/nifake_service.cpp
```diff
@@ -600,7 +601,46 @@ namespace nifake_grpc {
...
+        auto buffer_size = actual_size;
+        status = library_->GetAnIviDanceWithATwistStringStrlenBug(buffer_size, (ViChar*)string_out.data(), &actual_size);
+        if (status == kErrorReadBufferTooSmall || status == kWarningCAPIStringTruncatedToFitBuffer) {
+          // buffer is now too small, try again
+          continue;
+        }
+        response->set_status(status);
+        if (status == 0) {
+          response->set_string_out(string_out);
+          nidevice_grpc::trim_trailing_nulls(*(response->mutable_string_out()));
```
astarche, line 643: Note: here's the gencode for the new case. The initial resize doesn't do the `- 1` and the cleanup resize clears trailing nulls.

source/codegen/common_helpers.py
```diff
@@ -302,7 +306,30 @@ def get_size_mechanism(parameter):
     return size.get('mechanism', None)
 
 
-def get_size_expression(parameter):
+def has_size_mechanism_tag(parameter: dict, tag: str) -> bool:
```
gregstoll, line 309: Cool, I like this tag mechanism!

generated/nifake/nifake_service.cpp
```diff
@@ -644,7 +645,46 @@ namespace nifake_grpc {
...
+      ViInt32 actual_size {};
+      while (true) {
+        auto status = library_->GetAnIviDanceWithATwistStringStrlenBug(0, nullptr, &actual_size);
+        if (status < 0) {
+          response->set_status(status);
+          return ::grpc::Status::OK;
+        }
+        std::string string_out;
+        if (actual_size > 0) {
+            string_out.resize(actual_size);
```
gregstoll, line 676: I realize this is tricky because of how the code is structured, but it would be really great if we could manage to codegen a comment here explaining why it's not `actual_size - 1` here. If there's not a good way to do it, though, that's OK.

gregstoll (approved):

SparkingSpork (approved):

reckenro (approved):

PR #369, opened 2021-09-24
Add basic code coverage/API tests for RFSA

astarche:

Add tests that exercise the different codegen features in the RFSA service to make sure we get reasonable results.

Improve validation of RFSA service.

Ran and passed included tests. Except `GetRelayName_ReturnsRelayName`, which currently fails due to [AB#1639825](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1639825).

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -26,16 +26,12 @@ jobs:
             os: windows-latest,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
             cmake_platform: "-A x64",
-            # On msvc, the -j flag does not help build times and causes intermittent
-            # file permission errors.
-            cmake_number_of_jobs: ""
           }
         - {
             name: "Ubuntu Build", artifact: "Linux.tar.xz",
             os: ubuntu-18.04,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
             cmake_platform: "",
-            cmake_number_of_jobs: "-j 2",
             glibc_version: "2_27"
           }
 
@@ -86,7 +82,7 @@ jobs:
       shell: cmake -P {0}
       run: |
         execute_process(
-          COMMAND cmake --build build --config $ENV{BUILD_TYPE} ${{ matrix.config.cmake_number_of_jobs }}
+          COMMAND cmake --build build --config $ENV{BUILD_TYPE}
           RESULT_VARIABLE result
           OUTPUT_VARIABLE output
           ERROR_VARIABLE output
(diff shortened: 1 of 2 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -26,16 +26,12 @@ jobs:
...
-            # On msvc, the -j flag does not help build times and causes intermittent
-            # file permission errors.
-            cmake_number_of_jobs: ""
           }
         - {
             name: "Ubuntu Build", artifact: "Linux.tar.xz",
             os: ubuntu-18.04,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
             cmake_platform: "",
-            cmake_number_of_jobs: "-j 2",
```
astarche, line 38: Same with nilrt, ubuntu went over the edge.

The "good news" is that it's increasingly clear that the problems are with core_server and device_server including everything. We can fix that, and I have a plan for it. So, as long as we can implement that and keep future services DAQ-size-or-smaller, we can turn this back on.

gregstoll (approved):

.github/workflows/build_cmake.yml
gregstoll: I'm pretty sure this is causing the build on #370 to fail as well, so I'll wait for this to go in first.

astarche: I'll just separate this out and push it through

source/tests/system/nirfsa_driver_api_tests.cpp
```diff
@@ -21,38 +40,340 @@ class NiRFSADriverApiTests : public Test {
...
+  EXPECT_SUCCESS(session, response);
+  EXPECT_THAT(response.frequencies(), Each(Not(Eq(0))));
+  EXPECT_THAT(response.phase_response(), Each(Eq(0)));
+  EXPECT_THAT(response.magnitude_response(), Each(Eq(0)));
+  EXPECT_EQ(EXPECTED_SIZE, response.frequencies().size());
+  EXPECT_EQ(EXPECTED_SIZE, response.phase_response().size());
+  EXPECT_EQ(EXPECTED_SIZE, response.magnitude_response().size());
+}
+
+// TODO: AB#1639825. This currently fails because of an off-by-one error.
```
reckenro, line 161: Does that this mean that this test should be skipped or commented out for now then?

reckenro (approved):

source/tests/system/nirfsa_driver_api_tests.cpp
reckenro: Oh nevermind, these are system tests so they're run manually.

astarche: It looks like the gtest standard is a DISABLED_ prefix. I can do that.

astarche:
Bypassing @SparkingSpork . More j2. Need to fix the build.

PR #371, opened 2021-09-24
Move codegen functions to cmake scripts

astarche:

Move codegen functions and related imports into cmake scripts instead of the top-level CMakeLists file.

Our CMakeLists file is pretty noisy, just moving stuff out is useful.

We need to do some refactoring to make it easier to add new drivers and particularly to split drivers into separate services. To enable that, we will need to have these as separate helpers that can be pulled in my separate component builds.

Built it.


CMakeLists.txt
```diff
@@ -7,34 +7,14 @@ if (CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
...
-else()
-  add_subdirectory(third_party/grpc ${CMAKE_CURRENT_BINARY_DIR}/grpc EXCLUDE_FROM_ALL)
-  set(_PROTOBUF_PROTOC $<TARGET_FILE:protobuf::protoc>)
-  set(_REFLECTION grpc++_reflection)
-  set(_GRPC_CPP_PLUGIN_EXECUTABLE $<TARGET_FILE:grpc_cpp_plugin>)
-  set(_GRPC_GRPCPP grpc++)
-  set(_PROTOBUF_LIBPROTOBUF libprotobuf)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+if(NOT CMAKE_CROSSCOMPILING OR _GRPC_DEVICE_NILRT_LEGACY_TOOLCHAIN)
+  add_subdirectory(${CMAKE_SOURCE_DIR}/third_party/grpc ${CMAKE_CURRENT_BINARY_DIR}/grpc EXCLUDE_FROM_ALL)
```
SparkingSpork, line 13: Is the `${CMAKE_SOURCE_DIR}` fully necessary here? If so, consider using `${CMAKE_CURRENT_SOURCE_DIR}` instead. While they should be the same in a top-level CMakeLists.txt file, it should ensure that it evaluates correctly if this is ever run from another project.

cmake/GenerateGrpcSources.cmake
```diff
@@ -0,0 +1,21 @@
...
+function(GenerateGrpcSources proto_file proto_path proto_srcs proto_hdrs grpc_srcs grpc_hdrs)
+  get_filename_component(proto_out_path "${proto_srcs}" PATH)
+  add_custom_command(
+    OUTPUT "${proto_srcs}" "${proto_hdrs}" "${grpc_srcs}" "${grpc_hdrs}"
+    COMMAND ${_PROTOBUF_PROTOC}
+    ARGS --grpc_out "${proto_out_path}"
+      --cpp_out "${proto_out_path}"
+      -I "${proto_path}"
+      -I ${CMAKE_SOURCE_DIR}/third_party/grpc/third_party/protobuf/src/
+      -I ${CMAKE_SOURCE_DIR}/source/protobuf
```
SparkingSpork, line 17: Similarly to above, make sure we're using the right value between `${CMAKE_SOURCE_DIR}` and `${CMAKE_CURRENT_SOURCE_DIR}` based on your intended behavior.

SparkingSpork: To be clear, I have trouble fully understanding when to use each so this may be correct as it is.

SparkingSpork (approved):
Looks good to me aside from some nitpicking.

cmake/GenerateGrpcSources.cmake
astarche: Yeah, these will be `CMAKE_SOURCE_DIR` to get back to the root even when they're included from elsewhere.

CMakeLists.txt
astarche: Oh, hm. Thanks for bringing this up.

Yeah, these are not needed. I added some of that when I was trying to move the add_subdirectory into ImportGrpc, but that didn't work.

Pragmatically, it's easiest to use `${CMAKE_SOURCE_DIR}` to reference files in fixed locations off of the root. When it's in a cmake script that is shared between multiple cmakelists, I'm not sure there's an easy alternative (since they each may be in different relative locations to the root). But that does prevent building them as a separate top-level projects. Since, that's not specifically a goal of this effort, I think I'll call that "out of scope" and stick with `${CMAKE_SOURCE_DIRECTORY}` in `cmake` scripts. 

In `CMakeLists.txt` files it make sense to use `{CMAKE_CURRENT_SOURCE_DIR}` since you have a fixed starting location (unless you move the file ...), Although combining it with `../../` is kind of clunky and I don't think that's adding any real portability. So, I'd probably prefer `${CMAKE_SOURCE_DIR}` for accessing fixed dirs off of the root there as well.

PR #373, opened 2021-09-27
Revert "Move codegen functions to cmake scripts"

astarche:
Reverts ni/grpc-device#371

Reason: This change broke codegen.


CMakeLists.txt
```diff
@@ -114,11 +134,10 @@ foreach(api ${nidrivers})
       ${nidriver_client_srcs}
       "${service_output_dir}/${api}/${api}_client.cpp")
   endif()
-  GenerateServices(
```
astarche, line 117: Something got messes up with the dependencies in GenerateServices. I think this worked in my original change (which was pretty different ...) but it's possible I missed it. Not sure what the problem is, but obviously we need codegen to work.

SparkingSpork (approved):

PR #375, opened 2021-09-27
Add support for scalar custom types

astarche:

Add the ability to read and write scalar custom types (previously only repeated custom types were supported).

Move towards a standardized `convert_to_grpc`/`convert_from_grpc` specialization syntax. This allows supporting scalars/arrays without adding yet another mako condition block.

Scalar custom types are used by RFSA.

Ran and passed new tests for `Get`/`SetCustomData`.
Ran and passed scope/DAQ/Digital system tests.

### DIFF
--- a/generated/nidcpower/nidcpower_service.cpp
+++ b/generated/nidcpower/nidcpower_service.cpp
@@ -11,9 +11,11 @@
 #include <iostream>
 #include <atomic>
 #include <vector>
-#include <server/converters.h>
 
 namespace nidcpower_grpc {
+
+  using nidevice_grpc::converters::convert_from_grpc;
+  using nidevice_grpc::converters::convert_to_grpc;
 
   const auto kErrorReadBufferTooSmall = -200229;
   const auto kWarningCAPIStringTruncatedToFitBuffer = 200026;
@@ -28,13 +30,6 @@ namespace nidcpower_grpc {
 
   NiDCPowerService::~NiDCPowerService()
   {
-  }
-
-  void NiDCPowerService::Copy(const std::vector<ViBoolean>& input, google::protobuf::RepeatedField<bool>* output) 
-  {
-    for (auto item : input) {
-      output->Add(item != VI_FALSE);
-    }
   }
 
   //---------------------------------------------------------------------
@@ -2272,7 +2267,7 @@ namespace nidcpower_grpc {
       auto status = library_->FetchMultiple(vi, channel_name, timeout, count, voltage_measurements, current_measurements, in_compliance.data(), &actual_count);
       response->set_status(status);
       if (status == 0) {
-        Copy(in_compliance, response->mutable_in_compliance());
+        convert_to_grpc(in_compliance, response->mutable_in_compliance());
         response->set_actual_count(actual_count);
       }
       return ::grpc::Status::OK;
--- a/generated/nidcpower/nidcpower_service.cpp
+++ b/generated/nidcpower/nidcpower_service.cpp
@@ -11,6 +11,7 @@
 #include <iostream>
 #include <atomic>
 #include <vector>
+#include <server/converters.h>
 
 namespace nidcpower_grpc {
 
--- a/generated/nidigitalpattern/nidigitalpattern_service.cpp
+++ b/generated/nidigitalpattern/nidigitalpattern_service.cpp
@@ -11,9 +11,11 @@
 #include <iostream>
 #include <atomic>
 #include <vector>
-#include <server/converters.h>
 
 namespace nidigitalpattern_grpc {
+
+  using nidevice_grpc::converters::convert_from_grpc;
+  using nidevice_grpc::converters::convert_to_grpc;
 
   const auto kErrorReadBufferTooSmall = -200229;
   const auto kWarningCAPIStringTruncatedToFitBuffer = 200026;
@@ -28,13 +30,6 @@ namespace nidigitalpattern_grpc {
 
   NiDigitalService::~NiDigitalService()
   {
-  }
-
-  void NiDigitalService::Copy(const std::vector<ViBoolean>& input, google::protobuf::RepeatedField<bool>* output) 
-  {
-    for (auto item : input) {
-      output->Add(item != VI_FALSE);
-    }
   }
 
   template <typename TEnum>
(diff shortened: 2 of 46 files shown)


generated/nidcpower/nidcpower_service.cpp
```diff
@@ -30,13 +33,6 @@ namespace nidcpower_grpc {
   {
   }
 
-  void NiDCPowerService::Copy(const std::vector<ViBoolean>& input, google::protobuf::RepeatedField<bool>* output)
```
astarche, line 33: These were all identical and are moved to converters.h (with a new name: convert_from_grpc)

generated/nidigitalpattern/nidigitalpattern_service.cpp
```diff
@@ -1478,9 +1474,11 @@ namespace nidigitalpattern_grpc {
         if (status == 0) {
           CopyBytesToEnums(expected_pin_states, response->mutable_expected_pin_states());
           response->set_expected_pin_states_raw(expected_pin_states);
+          response->mutable_expected_pin_states()->Resize(actual_num_pin_data, 0);
```
astarche, line 1477: I accidentally exposed a missing case here. Arrays of enum-byte-arrays were not going through the ivi-dance resizing. That got fixed by rejiggering that codepath.

generated/nifake/nifake_service.cpp
```diff
@@ -46,38 +42,6 @@ namespace nifake_grpc {
     }
   }
 
-  void NiFakeService::Copy(const CustomStruct& input, nifake_grpc::FakeCustomStruct* output)
```
astarche, line 49: The scalar copy methods are now generated as `convert_to/from_grpc` methods in the same file but now in the converters namespace (at the end of the file) instead of as class members.

The Array copy methods are no longer generated. They were identical and are now in converters.h.

source/codegen/metadata/nirfsa/functions.py
```diff
@@ -1471,11 +1471,6 @@
             {
                 'direction': 'out',
                 'name': 'numberOfGainReferenceCalConstants',
-                'size': {
```
astarche, line 1474: TODO fix metadata extraction, This is mistagged.

source/codegen/templates/service_helpers.mako
```diff
@@ -659,20 +657,29 @@ ${copy_to_response_with_transform(source_buffer=raw_response_field, parameter_na
...
 %   elif common_helpers.is_array(parameter['type']):
-%     if common_helpers.is_string_arg(parameter):
+### pass: other array types don't need to copy.
+%   else:
         response->set_${parameter_name}(${parameter_name});
-%     elif common_helpers.is_struct(parameter) or parameter['type'] == 'ViBoolean[]':
-        Copy(${parameter_name}, response->mutable_${parameter_name}());
-%     endif
+%   endif
+### Handle ivi-dance-with-a-twist resizing.
```
astarche, line 674: The `elif`s get broken by this change if I leave this ivi-twist-resize in the original switch statement. (actually they were broken before, but for a corner case: bytes enums). So I moved it to a separate if block after the initial copy.

source/custom/nidaqmx_conversions.h
```diff
@@ -1,20 +1,14 @@
+#ifndef NIDEVICE_GRPC_DEVICE_NIDAQMX_CONVERSIONS_H
+#define NIDEVICE_GRPC_DEVICE_NIDAQMX_CONVERSIONS_H
+
 // Needed for the definition of CVIAbsoluteTime
 #include <google/protobuf/util/time_util.h>
+#include <server/converters.h>
 
 #include "NIDAQmx.h"
 
-template <typename CType, typename GrpcType>
```
astarche, line 6: Moved to converters.h

source/server/converters.h
```diff
@@ -1,9 +1,14 @@
 #ifndef NIDEVICE_GRPC_DEVICE_CONVERTERS_H
```
astarche, line 1: Here's converters.h!

source/codegen/common_helpers.py
```diff
@@ -48,13 +48,16 @@ def any_function_uses_timestamp(functions):
...
+
+
 def get_input_and_output_custom_types(functions):
     '''Returns a set of custom types used by input and output parameters separately.'''
     input_custom_types = set()
     output_custom_types = set()
     for function in functions:
-        struct_array_params = [p for p in functions[function]
-                               ["parameters"] if is_struct(p) and is_array(p["type"])]
+        struct_array_params = [p for p in functions[function]["parameters"] if is_struct(p)]
```
gregstoll, line 60: Can we rename this variable now that it's no longer just arrays?

source/codegen/templates/service.cpp.mako
```diff
@@ -203,3 +154,34 @@ ${mako_helper.define_simple_method_body(function_name=function_name, function_da
...
+  output->set_${field["grpc_name"]}(input.${field["name"]});
+%       endfor
+}
+
+%     endif
+%     if custom_type["name"] in input_custom_types:
+template <>
+${custom_type["name"]} convert_from_grpc(const ${namespace_prefix}${custom_type["grpc_name"]}& input) 
+{
+  ${custom_type["name"]}* output = new ${custom_type["name"]}();
```
gregstoll, line 175: Is there a reason this is `new`d instead of allocating it on the stack and relying on a copy constructor? I'm worried that this will lead to memory leaks when we call this and don't put it in a `RepeatedPtrField`.

(my C++ is pretty rusty though :-( )

source/codegen/templates/service_helpers.mako
```diff
@@ -659,20 +657,29 @@ ${copy_to_response_with_transform(source_buffer=raw_response_field, parameter_na
...
 ${initialize_response_buffer(parameter_name=parameter_name, parameter=parameter)}\
 ${copy_to_response_with_transform(source_buffer=parameter_name, parameter_name=parameter_name, transform_x="x")}\
+%   elif common_helpers.is_struct(parameter) or parameter['type'] == 'ViBoolean[]':
+        convert_to_grpc(${parameter_name}, response->mutable_${parameter_name}());
+%   elif common_helpers.is_string_arg(parameter):
+        response->set_${parameter_name}(${parameter_name});
+%   elif parameter['type'] == 'ViSession':
+        auto session_id = session_repository_->resolve_session_id(${parameter_name});
+        response->mutable_${parameter_name}()->set_id(session_id);
+%   elif parameter['type'] == 'CVIAbsoluteTime':
```
gregstoll, line 667: I forgot I had to add a special case for `CVIAbsoluteTime` here. It would be nice to get rid of it (maybe tag the field in metadata), but we don't have to do that in this change, it's already pretty big.

gregstoll (approved):

source/codegen/templates/service.cpp.mako
astarche: Ooh, yikes. Yeah, that's a bug. The array implementation of `convert_to_grpc` uses new + AddAllocated to transfer ownership. That's weird but it works. This ... is just a leak. I think in either the array or the scalar case. It's returning a copy of the heap allocated struct and never freeing it.

source/codegen/templates/service_helpers.mako
astarche: I was thinking, short term, I would add a "supports_standard_conversion" helper. But yeah, we could also make that more pluggable.

reckenro (approved):

PR #376, opened 2021-09-30
Validate metadata as a part of build

gregstoll:

Adds a step to the build that validates that the metadata is well-formed. Validates some common problems and adds a way to suppress some rules where necessary. Also cleans up a few issues that the validation found.

When metadata is badly-formed, usually we get an error in `mako` that has the exception and file/function that it occurred in, but not a line number. This can make problems frustrating to track down. The hope is that this validation step will catch these problems earlier and make it much clearer what's wrong.

Additionally, this step can catch issues that won't cause errors but will lead to a suboptimal interface. For example, input arrays shouldn't use the `passed-in` size, because the gRPC layer can just determine the size of the array. (there were a few entry points we shipped this way and we can't change now; I added suppressions for those)

Here's an example of what it looks like when validation fails:

```
[build] [1/9  11% :: 0.737] Generating proto file and service for nidigitalpattern
[build] FAILED: ../generated/nidigitalpattern/nidigitalpattern_client.cpp ../generated/nidigitalpattern/nidigitalpattern_client.h ../generated/nidigitalpattern/nidigitalpattern_library_interface.h ../generated/nidigitalpattern/nidigitalpattern_library.cpp ../generated/nidigitalpattern/nidigitalpattern_library.h ../generated/nidigitalpattern/nidigitalpattern_mock_library.h ../generated/nidigitalpattern/nidigitalpattern.proto ../generated/nidigitalpattern/nidigitalpattern_service.cpp ../generated/nidigitalpattern/nidigitalpattern_service.h 
[build] cmd.exe /C "cd /D C:\dev\grpc-device\build && C:\dev\grpc-device\build\venv\Scripts\python.exe C:/dev/grpc-device/source/codegen/generate_service.py C:/dev/grpc-device/source/codegen/metadata/nidigitalpattern/ -o C:/dev/grpc-device/generated/"
[build] Traceback (most recent call last):
[build]   File "C:\dev\grpc-device\source\codegen\metadata_validation.py", line 168, in validate_function
[build]     validate_parameter_size(parameter, function_name, metadata)
[build]   File "C:\dev\grpc-device\source\codegen\metadata_validation.py", line 236, in validate_parameter_size
[build]     raise Exception(f"parameter {parameter['name']} is an array but has no size!")
[build] Exception: parameter sessions is an array but has no size!
[build] 
[build] The above exception was the direct cause of the following exception:
[build] 
[build] Traceback (most recent call last):
[build]   File "C:\dev\grpc-device\source\codegen\metadata_validation.py", line 21, in validate_metadata
[build]     validate_function(function_name, metadata)
[build]   File "C:\dev\grpc-device\source\codegen\metadata_validation.py", line 194, in validate_function
[build]     raise Exception(f"Failed to validate function {function_name}") from e
[build] Exception: Failed to validate function WriteSequencerFlagSynchronized
[build] 
[build] The above exception was the direct cause of the following exception:
[build] 
[build] Traceback (most recent call last):
[build]   File "C:\dev\grpc-device\source\codegen\generate_service.py", line 111, in <module>
[build]     generate_all(args.metadata, "." if args.output is None else args.output)
[build]   File "C:\dev\grpc-device\source\codegen\generate_service.py", line 83, in generate_all
[build]     metadata_validation.validate_metadata(metadata)
[build]   File "C:\dev\grpc-device\source\codegen\metadata_validation.py", line 31, in validate_metadata
[build]     raise Exception(
[build] Exception: Failed to validate nidigitalpattern
```

Verified that everything still builds with validation turned on, and unit tests still pass. Verified that no shipping generated code has changed.


CMakeLists.txt
```diff
@@ -43,12 +45,12 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 #----------------------------------------------------------------------
-# Find python3 for the build system
+# Set up virtual environment for the build system
```
gregstoll, line 48: note: Since we use the Python package `schema` to do validation, we need set up a virtual environment to do the build in that has that package.

source/codegen/metadata/nifake/functions.py
```diff
@@ -743,7 +743,7 @@
                 'name': 'numberOfElements',
                 'size': {
                     'mechanism': 'custom-code',
-                    'value': None
+                    'value': '',
```
gregstoll, line 746: note: all of these get mocked out anyway, but we don't generally allow `None` as a custom-code value so I changed it here.

source/codegen/metadata/nifake/functions_addon.py
```diff
@@ -4,4 +4,20 @@
 functions_override_metadata = {
 }
 
-
+functions_validation_suppressions = {
+    "AcceptViSessionArray": {
+        "parameters": {
+            "sessionArray": ["INPUT_ARRAY_SHOULD_NOT_HAVE_PASSED_IN_SIZE"]
```
gregstoll, line 10: note: These are just in `nifake` so I could have "fixed" them, but it's nice to have some tests that make sure we can still generate code for cases like this.

source/codegen/metadata/nifake_non_ivi/functions.py
```diff
@@ -455,7 +455,11 @@
                 'direction': 'in',
                 'enum': 'BeautifulColor',
                 'name': 'colors',
-                'type': 'int32[]'
+                'type': 'int32[]',
```
gregstoll, line 458: note: Again this is in a fake service, and it's not really necessary, but I didn't want to add a suppression for something that should never happen for real (an array without a size type).

source/codegen/metadata/niscope/functions.py
```diff
@@ -2009,7 +2009,7 @@
             },
             {
                 'size': {
-                    'mechanism': 'ivi_dance_with_a_twist',
+                    'mechanism': 'ivi-dance-with-a-twist',
```
gregstoll, line 2012: note: doesn't matter because the function is CustomCode, but fixed it anyway.

source/codegen/metadata_validation.py
```diff
@@ -0,0 +1,287 @@
+from typing import Any, Dict, Iterable, List, Set
+from schema import Schema, And, Optional, Or, Use
+import common_helpers
+
+
+# Rules that can be suppressed
```
gregstoll, line 6: note: These are the rules that can be suppressed. If we find cases where other rules need to be suppressed, we can add them as we discover them.

source/codegen/metadata_validation.py
```diff
@@ -0,0 +1,287 @@
...
+SIZE_SCHEMA = Schema(
+    {
+        'mechanism': And(str, lambda s: not common_helpers.is_unsupported_size_mechanism_type(s)),
+        'value': Or(str, int),
+        Optional('value_twist'): str,
+        Optional('tags'): [str],
+    },
+)
+
+PARAM_SCHEMA = Schema(
```
gregstoll, line 52: note: I decided to explicitly list the keys that are valid even if we don't use most of them. My idea was that this will prevent us from accepting typos in keys that we don't want. Sadly the list of "keys we use somewhere" is a bit longer than I expected...

source/codegen/metadata_validation.py
```diff
@@ -0,0 +1,287 @@
...
+def validate_metadata(metadata: dict):
+    try:
+        for function_name in metadata['functions']:
+            validate_function(function_name, metadata)
+        for attribute_group in common_helpers.get_attribute_groups(metadata):
+            for attribute_id in attribute_group.attributes:
+                validate_attribute(attribute_group.attributes[attribute_id], metadata)
+        function_enums = get_function_enums(metadata['functions'])
+        attribute_enums = get_attribute_enums(metadata)
+        used_enums = function_enums.union(attribute_enums)
```
gregstoll, line 27: note: Fun fact - we only generate enums from `enums.py` that are actually used in a function or attribute. So pass those in to `validate_enum` so we can know how much to check them.

.github/workflows/build_cmake.yml
```diff
@@ -70,7 +73,7 @@ jobs:
           COMMAND cmake
             -S .
             -B build
-            -D CMAKE_BUILD_TYPE=$ENV{BUILD_TYPE}
```
SparkingSpork, line 73: Was this line intentionally removed?

CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server C CXX)
 
+include(CreateVirtualEnvironment.cmake)
```
SparkingSpork, line 5: I noticed that you placed this file in the repo top-level. Consider placing it in a `cmake` or subdirectory as Alex did in #343 and adding that directory to the `CMAKE_MODULE_PATH`. Note that you also shouldn't need the `*.cmake` file extension here.

CMakeLists.txt
```diff
@@ -43,11 +45,21 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
+# Set up virtual environment for the build system
+# This is only needed for local builds - GitHub builds install the
+# necessary dependencies in the .yml files.
 #----------------------------------------------------------------------
-find_package(Python3 COMPONENTS Interpreter)
-if(NOT ${Python3_FOUND})
-  message(WARNING "Python3 not found. Python3 is required for code generation.")
+if (NOT IS_GITHUB_BUILD)
+  CreateVirtualEnvironment(venv
+    REQUIREMENTS_TXT ${CMAKE_SOURCE_DIR}/requirements.txt
```
SparkingSpork, line 54: Nitpick: Rename `requirements.txt` to something more specific to indicate it's specifically python modules.

requirements.txt
```diff
@@ -0,0 +1,32 @@
+backports-datetime-fromisoformat==1.0.0
+betterproto==1.2.5
+black==21.8b0
```
astarche, line 3: This seems like it's pulling in a lot of extra stuff
betterproto: only used in examples (separate venv? I'd say defer this and just don't add their deps)
black: for sure we don't use (but we should!)
pyperclip: random useful dev tool?

Is this just stuff that you happen to have used in a venv locally?

Is there a way we can get this down to mako + schema (and their deps)?

source/codegen/generate_service.py
```diff
@@ -73,11 +74,14 @@ def mutate_metadata(metadata: dict):
 def generate_all(metadata_dir, gen_dir):
     sys.path.append(metadata_dir)
     init_file = os.path.join(metadata_dir, "__init__.py")
+
     spec = importlib.util.spec_from_file_location("metadata", init_file)
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
 
     metadata = module.metadata
+    metadata_validation.validate_metadata(metadata)
```
SparkingSpork, line 83: It might be nice to add a command line option to just validate metadata. That way changes can be validated without starting the full generation.

SparkingSpork (approved):

CMakeLists.txt
```diff
@@ -43,11 +45,21 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
 #----------------------------------------------------------------------
-# Find python3 for the build system
+# Set up virtual environment for the build system
+# This is only needed for local builds - GitHub builds install the
+# necessary dependencies in the .yml files.
 #----------------------------------------------------------------------
-find_package(Python3 COMPONENTS Interpreter)
-if(NOT ${Python3_FOUND})
-  message(WARNING "Python3 not found. Python3 is required for code generation.")
+if (NOT IS_GITHUB_BUILD)
```
astarche, line 52: Why not do this with a github build? It would be nice to keep the process consistent and to make sure we're validating it on both OSs.

source/codegen/common_helpers.py
```diff
@@ -301,7 +307,7 @@ def has_enum_array_string_out_param(functions):
     return False
 
 
-def get_size_mechanism(parameter):
+def get_size_mechanism(parameter: dict) -> Union[str, None]:
```
astarche, line 310: `Optional[str]` I think.

source/codegen/metadata/nidaqmx/functions_addon.py
```diff
@@ -1,2 +1,36 @@
 functions_override_metadata = {
 }
+
+functions_validation_suppressions = {
```
astarche, line 4: The suppressions look good. There aren't as many as I feared and the syntax makes sense.

requirements.txt
gregstoll: Yeah, "a bunch of random stuff" is a good summary. I started clean and just installed mako and schema and shrunk this a lot.

.github/workflows/build_cmake.yml
gregstoll: Yup, must have happened while I was flailing about with cmake. Fixed it.

source/codegen/metadata_validation.py
astarche: Were there a lot of violations in unused enums? I'd run into that fun fact before and been annoyed by it. I'm not sure what to do about it (there's also some value in making fewer changes to nimi-python-derived metadata; so I'm not sure we should torch the unused ones). We could make "allow-unused-enums" be optional or something.

astarche: (not part of this PR, just curious)

gregstoll: I'm not sure - because validation stops as soon as it finds a problem (this is something that could be improved), all I know is it came up at least once in a few different drivers.

source/codegen/metadata_validation.py
```diff
@@ -0,0 +1,287 @@
...
+
+PARAM_SCHEMA = Schema(
+    {
+        'direction': And(str, lambda s: s in ('in', 'out')),
+        'name': str,
+        Optional('type'): str,
+        Optional('grpc_type'): str,
+        Optional('documentation'): DOCUMENTATION_SCHEMA,
+        Optional('enum'): str,
+        # This should be SIZE_SCHEMA but don't want to validate if codegen_method is no
```
astarche, line 60: You have this a couple of places and it seems consequential to me. This is adding quite a few lines-of-code to the validate methods.

Can the conditional move up to the top-level?

I was first thinking that you would handle this with an Or syntax.

`Functions: [Or(FUNCTION_SCHEMA, NO_CODEGEN_SCHEMA)]`
where NO_CODEGEN_SCHEAM would constrain `codegen _method` to `no` and validate nothing else. And `FUNCTION_SCHEMA` would have the opposite constraint and be more strict.

But the way it's written seems more imperative than that. What about swapping:

```python
        FUNCTION_SCHEMA.validate(function)
        if function.get('codegen_method', 'public') != 'no':
```

with

```python
        if function.get('codegen_method', 'public') != 'no':
          FUNCTION_SCHEMA.validate(function)
```

Would that allow moving size and params to schema constraints instead of custom code?

It looks like part of the purpose of the custom validation code is to give better errors (though maybe they're pretty good by default?).

source/codegen/metadata_validation.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def validate_enum(enum_name: str, used_enums: Set[str], metadata: dict):
+    try:
+        enum: Dict[str, Any] = metadata['enums'][enum_name]
+        ENUM_SCHEMA.validate(enum)
+        if enum_name in used_enums:
+            generate_mappings = enum.get('generate-mappings', False)
+            value_types = set([type(value['value']) for value in enum['values']])
+            if not generate_mappings:
```
astarche, line 218: Maybe another case where the conditional could be moved up and have 2 schemas?

source/codegen/metadata_validation.py
```diff
@@ -0,0 +1,287 @@
...
+                DOCUMENTATION_SCHEMA.validate(function['documentation'])
+            for parameter in function['parameters']:
+                validate_parameter_size(parameter, function_name, metadata)
+                if 'type' not in parameter:
+                    if 'grpc_type' not in parameter:
+                        raise Exception(f"parameter {parameter['name']} has no type or grpc_type!")
+                    if not parameter.get('repeated_var_args', False):
+                        raise Exception(f"parameter {parameter['name']} has no type and repeated_var_args is not set!")
+                if 'enum' in parameter:
+                    if parameter['enum'] not in metadata['enums']:
```
astarche, line 175: Is it possible to mutate the schema or have it be dynamic in a way that catches this?
(not sure that's a good idea)

astarche (approved):
I like it! Thanks for doing this.

The one fix I definitely want is using the venv on the GH build (or having some mitigation for that).

My schema questions do not need to be resolved here (but if they are, great!). 

I want to make sure I understand how the SCHEMAs work, what the tradeoffs you've considered are, and discover any consensus (or disagreement!) in what could be improved.

I generally prefer "more declarative SCHEMA" to "more validation code" but I appreciate that there are potentially several pragmatic tradeoffs (ease of implementation; quality of errors; *potentially* understandability of complex schemas). Also, I'm not as concerned with the custom validation code as it is today, as much as the possibility that "custom validation code" will end up being the default method of extending this code. So, if we can make it easier to choose SCHEMA when extending: great!

source/codegen/metadata_validation.py
gregstoll: I think something like using NO_CODEGEN_SCHEMA would work, but if we're not going to validate anything else anyway I think it makes sense to just skip it like the example you gave. I think I can make this work.

source/codegen/metadata_validation.py
gregstoll: We could do that here, but that means we'd have to duplicate most of ENUM_SCHEMA, which I think would be less clear/maintainable than doing it this way.

source/codegen/metadata_validation.py
gregstoll: (I assume you're talking about the `repeated_var_args` part of this)

I can't think of a way to do this that would be an improvement over what we have. We could have two schemas - one where `type` is required and `repeated_var_args` is not present or False, and one where type is not present and `repeated_var_args` is True. That way everything would validate under one of those schemas. But these would have to be two additional schemas unless we wanted to copy all the other parts of the original schema into the second schema, which is bad. And I think in this case it's clearer to just do the check in code rather than adding two additional schemas for this purpose.

reckenro (approved):

source/codegen/metadata_validation.py
astarche: Oh, actually no. I was thinking of the list of enums. Like 

`Optiona('enum'): And(str, lambda s: s in _get_actual_list_of_enums())`

Curious if that works.

I definitely don't want to worry more about repeated_var_args!

PR #416, opened 2021-11-03
Generate service registrars

astarche:

Add codegen to generate per-service registrar methods and `register_all_services`.

Updates to `core_server` and `device_server` to use these methods.

Restore `-j 2` flag on linux builds because "peak memory usage" for above files is lowered.

This has several advantages:
* Avoid long pole files that include lots of big headers (`core_server` and `device_server`)
   * Significantly improves build time on GH agents.
   * Worst-case obj file no longer grows with each new driver.
* Remove duplicate code between `core_server` and `device_server`. Improves consistency between tests and production server.
* Standardize on toggle logic and short-circuit service creation earlier.
* Makes it easier to add new services.

Ran and passed system tests for several drivers.
Manual testing of toggle behavior in app.


generated/nidaqmx/nidaqmx_service.h
```diff
@@ -23,14 +23,23 @@
 
 namespace nidaqmx_grpc {
 
+struct NiDAQmxFeatureToggles
```
astarche, line 26: Note: this code is moved around slightly so that the toggles can be created (and checked) in the registrar methods before creating the service.

generated/nidaqmx/nidaqmx_service_registrar.cpp
```diff
@@ -0,0 +1,53 @@
...
+        &library, 
+        resource_repository, 
+        feature_toggles) {
+  }
+  NiDAQmxLibrary library;
+  NiDAQmxService service;
+};
+}
+
+std::shared_ptr<void> register_service(
```
astarche, line 32: Note: `std::shared_ptr<void>` is being used like `std:any` here, but we can't use `any` because it's not supported in the gcc used for nilrt. Basically it's type-erasing the `LibraryAndService` class so that we can transfer lifetime ownership without an intrusive empty interface (and we don't actually want to expose a different interface to interact with the service once it's created).

generated/register_all_services.cpp
```diff
@@ -0,0 +1,114 @@
...
+  const nidevice_grpc::FeatureToggles& feature_toggles,
+  nidevice_grpc::FeatureToggles::CodeReadiness code_readiness)
+{
+  auto service_vector = std::make_shared<std::vector<std::shared_ptr<void>>>();
+
+  auto session_repository = std::make_shared<nidevice_grpc::SessionRepository>();
+  auto core_service = nidevice_grpc::register_core_service(server_builder, session_repository);
+  service_vector->insert(
+    service_vector->end(), 
+    {session_repository, core_service});
```
astarche, line 39: Note: adding the `session_repository` to the `vector` is kind of a hack. `std::vector` doesn't guarantee any kind of destructor ordering on elements, so it could get destroyed before the refers. I started another change to always use a `shared_ptr` for that class so that it can be jointly-owned by the services. That touches a lot of files, so I'm holding that back as a separate change.

We don't really have a clean shutdown procedure other than needing it to work in tests where we've already gone into an idle state. If we were shutting down a live server, this would not be the only potential ordering issue.

source/codegen/generate_service.py
```diff
@@ -1,56 +1,26 @@
-import sys
```
astarche, line 1: Note: a lot of code moves from here to `template_helpers.py`

source/codegen/template_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    old_contents = _read_file_contents(output_file_path)
+    if old_contents != new_contents:
+        with open(output_file_path, "w+", newline="") as f:
+            f.write(new_contents)
+
+
+def load_metadata(metadata_dir: str, ) -> dict:
+    metadata_path = Path(metadata_dir)
+    module = import_module("metadata." + metadata_path.name)
```
astarche, line 31: Note: Tada! simple `import_module` code (see `__init__` comment above).

source/server/feature_toggles.cpp
```diff
@@ -17,11 +17,9 @@ bool FeatureToggles::is_feature_enabled(
     FeatureToggles::CodeReadiness feature_readiness) const
 {
   auto toggle_state = get_feature_state(feature_name);
-  switch (feature_readiness) {
-    case CodeReadiness::kRelease:
-      return toggle_state != FeatureState::kDisabled;
-    default:
-      return toggle_state == FeatureState::kEnabled;
+  if (feature_readiness >= app_readiness_) {
```
astarche, line 20: Note: This is allowing `device_server` to get pre-release services by default.

source/server/session_repository.h
```diff
@@ -1,14 +1,18 @@
...
-#include <session.grpc.pb.h>
-
 #include <atomic>
+#include <functional>
+#include <map>
 #include <shared_mutex>
 
 #include "semaphore.h"
 
+namespace grpc {
```
astarche, line 11: I did a little bit of `include` file cleanup and forward-declares etc. Nothing especially made a different, but good policy.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -6,11 +7,11 @@
 #include <cstring>
 #include <random>
 #include <stdexcept>
+#include <vector>
 
 #include "device_server.h"
 #include "enumerate_devices.h"
 #include "nidaqmx/nidaqmx_client.h"
-#include "nidaqmx/nidaqmx_service.h"
```
astarche, line 13: Note: tests don't need to include the service because that's all internal to the `register_all_services` implementation. It doesn't make a huge difference because the `client.h` files include `grpc.pb.h` (needed for request/response messages and the Stub). If we have issues with long-pole test files, we can try to address that. The easiest improvement would be encapsulating the `stub` so that we only need `pb.h` (unfortunately the stub is a nested class, so we can't just forward declare it ...).

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -1154,22 +1158,20 @@ TEST_F(NiFakeNonIviServiceTests, SetColors_PassesColorsArrayToLibrary)
...
-struct NiFakeNonIviServiceEnabledToggleTests : public NiFakeNonIviServiceTests {
-  NiFakeNonIviServiceEnabledToggleTests()
-      : NiFakeNonIviServiceTests(
-            nidevice_grpc::FeatureToggles({{"nifake_non_ivi", true}}))
-  {
-  }
-};
-
-TEST_F(NiFakeNonIviServiceEnabledToggleTests, UnreleasedServiceWithToggleEnabled_IsEnabled_ReturnsTrue)
+TEST_F(NiFakeNonIviServiceTests, UnreleasedServiceWithToggleEnabled_IsEnabled_ReturnsTrue)
```
astarche, line 1161: Note: rewriting these tests, since `is_enabled` is now used external to the service (in the registrar).

source/codegen/metadata/nidaqmx/__init__.py
```diff
@@ -1,8 +1,8 @@
-from functions import functions
```
astarche, line 1: Note: I'm change all of these to "relative" imports so that they actually import from the module. Previously they were un-qualified absolute/global imports, which relied on adding this folder to `sys.path`. The immediate problem this solves is that the previous implementation wouldn't allow loading 2 modules in the same interpeter because there can only be one module with the absolute name `functions`, etc.

Also note that the new `importlib` code is both more correct and simpler. This is how importing modules is supposed to work and the old code was hacking around a misunderstanding of module imports.

Normally it is preferred to use absolute imports to import from the module. In this case `nidaqmx.functions`, etc. Since these are pure glue code I think it's useful to preserve the consistency/simplicity/paste-ability of the relative imports.

source/codegen/template_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+    This prevents downstream recompiles when codegen runs but does not change
+    a given file.
+    """
+    old_contents = _read_file_contents(output_file_path)
+    if old_contents != new_contents:
+        with open(output_file_path, "w+", newline="") as f:
+            f.write(new_contents)
+
+
+def load_metadata(metadata_dir: str, ) -> dict:
```
gregstoll, line 29: I think the `, ` here is unneeded?

gregstoll (approved):
I'm not going to pretend I 100% understand all of this, but it looks like this will solve our problems and I assume problems will be obvious if there are any. Thanks for doing this!

SparkingSpork (approved):
Changes look good to me. All I needed to convince me is that this is already shaving 20-30 minutes off build times!

reckenro (approved):

PR #420, opened 2021-11-08
Make all attribute enum names singular

### LINKED ISSUE #401: Attribute enum values starting with S get mangled by C# protoc generator
Reported by @smooresni

PecificDriverDescription

This happens because our attribute prefix (`NIRFSA_ATTRIBUTE`) does not match the attribute name (`NiRfsaAttributes`). Normally this would defeat the nice-to-have prefix stripping in the C# client generator. But when the value starts with `S` the prefix-stripping *works* and takes the `S` off with it.

astarche:

Change all attribute enum names to be singular to match the attribute prefix.

I.e.:

`NiRFSAAttribute::NIRFSA_ATTRIBUTE_DEEMBEDDING_TYPE`

NOT

`NiRFSAAttributes::NIRFSA_ATTRIBUTE_DEEMBEDDING_TYPE`

Fixes #401 .

1. Prefixing enums is a protobuf best practice because some languages don't support enum name scoping.
2. The de facto standard-within-a-standard is to prefix enum values with the `ENUM_NAME` in upper snake. This is what is done in the [proto style docs](https://developers.google.com/protocol-buffers/docs/style#enums) and [what linters check for](https://docs.buf.build/lint/rules#enum_value_prefix).
3. It is mostly harmless to use a different prefix or no-prefix-at-all as long as there are no conflicts. It just breaks linting.
4. BUT the C# generator has special code that matches and removes redundant enum prefixes. This causes a problem with the plural/singular case: if the enum value starts with an `S` it will match the prefix and split the word.

#### Alternativess:
1. Only fix RFSA/G and future drivers. This is an easy fix, but does not help C# clients of existing drivers.
2. Fix/wait for protocolbuffers/protobuf#9138 . I am not sure that protobuf will be willing to allow that change, especially if they don't get any other reports.

Builds.
TODO: test with examples and spot check system tests.


examples/nidaqmx/dsa-shared-timebase-and-trig-analog-input-and-output.py
```diff
@@ -221,40 +221,40 @@ async def raise_if_error_async(awaitable_call):
                 # your hardware device manual for further information on whether this method
                 # of synchronization is supported for your particular device.
                 await raise_if_error_async(client.SetTimingAttributeString(nidaqmx_types.SetTimingAttributeStringRequest(
-                    task=leader_input_task, attribute=nidaqmx_types.TimingStringAttributes.TIMING_ATTRIBUTE_REF_CLK_SRC,
```
astarche, line 224: NOTE: DAQ is one of the 2 cases where the prefix still doesn't match. This isn't susceptible to the false match, so it won't hit the C# bug.

We could fix these up too "while we're at it". The known benefits are fewer linter suppressions to deal with and shorter names in C# (because the prefix matching works). The cost is longer names in C++/python.

This would mean changing the enum **VALUE_NAME** which is a little more significant in python/C++. Many of our python examples just use the value name, without the enum name. Which is why a lot of examples don't show up in this change (the value name didn't change).

gregstoll: I guess it's a breaking change either way, but avoiding changing those is a little less breaking :-)

gregstoll (approved):

harsha-bhushan (approved):

examples/nidaqmx/dsa-shared-timebase-and-trig-analog-input-and-output.py
harsha-bhushan: All are valid points. Only a particular BU users are using C# as of now. 
But if this helps in having some consistency in the names format across drivers in any given language, then we can fix it.

reckenro (approved):

examples/nidaqmx/dsa-shared-timebase-and-trig-analog-input-and-output.py
astarche: I'll go with the change as-is for now so we can fix the key RFSA/G C# issue. I'll try and code up the other change and we can consider including it in the release as a general improvement.

PR #422, opened 2021-11-08
Make client artifacts toggle-aware

astarche:

Moves artifact staging to a python script, which can use driver metadata to check toggle state and only include proto/example files for release-ready drivers.

Allows us to release with drivers that are (cleanly) toggled off.

Several test builds on my private repo and locally.

### DIFF
--- a/.github/workflows/create_client_artifacts.yml
+++ b/.github/workflows/create_client_artifacts.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
       - 'releases/**'
-  pull_request:
-    types: [opened, synchronize, reopened]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/create_client_artifacts.yml
+++ b/.github/workflows/create_client_artifacts.yml
@@ -1,0 +1,67 @@
+name: Create Client Artifacts
+
+on:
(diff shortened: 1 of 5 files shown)


.github/workflows/create_client_artifacts.yml
```diff
@@ -0,0 +1,65 @@
+name: Create Client Artifacts
```
astarche, line 1: Note: Here's a good run https://github.com/astarche/grpc-device/actions/runs/1436376237

gregstoll (approved):

.github/workflows/create_client_artifacts.yml
SparkingSpork, line 1: This approach leaves open the potential for the client files to not match the client files used for the compiled binaries and tests. Specifically, in the case that the *.mako files change but the updated *.proto and similar are not checked in afterwards. That case could result in the main build (which will regenerate the files) using different versions of the files than what is in the repo.

Ideally we that would never happen since we would catch it in code reviews, but making it fool-proof couldn't hurt.

SparkingSpork (approved):

reckenro (approved):

PR #423, opened 2021-11-09
Fail workflow if files are dirtied during build

astarche:

Update `build_cmake.yml` and `build_nilrt.yml` to fail if files are changed during build/codegen.

This ensures that our generated source code is up-to-date. This is important to (a) ensure that our review/tracking process for generated code is sound and (b) ensure that other workflow runs that operate on the generated source get the same inputs.

Testing of workflow results with induced failures (and fixing those failures).

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -100,6 +100,13 @@ jobs:
           message(FATAL_ERROR "Build failed")
         endif()
 
+    # The generated source in git must match the output of workflow builds.
+    # If this step fails, something is changing during build. Either:
+    # 1. Fix non-deterministic codegen, OR
+    # 2. Build locally to ensure generated files are up-to-date.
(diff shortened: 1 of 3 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -100,6 +100,13 @@ jobs:
           message(FATAL_ERROR "Build failed")
         endif()
 
+    # The generated source in git must match the output of workflow builds.
+    # If this step fails, something is changing during build. Either:
+    # 1. Fix non-deterministic codegen, OR
+    # 2. Build locally to ensure generated files are up-to-date.
+    - name: Check for files dirtied by codegen
```
astarche, line 107: Sample failure: https://github.com/astarche/grpc-device/runs/4144845039?check_suite_focus=true

SparkingSpork (approved):
Looks good!

reckenro (approved):

PR #425, opened 2021-11-10
Fix venv creation on GH runners

astarche:

Fix venv creation on GH runners and some other windows configurations. Note that previously this code was disabled because we couldn't get it to work.

The `CreateVirtualEnvironment.cmake` script was failing to install pip in the venv. This is because cmake's `FindPython` finds `python3.exe` on the GH runners (and some other windows machines), which triggers a bug in cpython code. See python/cpython#25216 and actions/virtual-environments#2690.

Update our create venv code to do the following:
* Significantly simplify the cmake file.
* FIX: in `CreateVirtualEnvironment.cmake`, use `--without-pip` to separate `venv` creation and `ensurepip` so that we can call `ensurepip` with a python.exe that actually exists.
* Clean up cmake and GH actions to remove old conditional branches.

Recently a developer hit this bug on their dev machine. This fixes the issue for developers and also updates our CI to ensure that the dev workflow is tested (granted not on a clean install).

Running build with tracing on GH workflow.

Local build with ninja.


.github/workflows/build_cmake.yml
```diff
@@ -54,11 +54,8 @@ jobs:
       with:
         python-version: '3.x'
 
-    - name: Install build dependencies
-      run: python -m pip install -r python_build_modules.txt
-
     - name: Update Submodules
-      run: git submodule update --init --recursive
+      run: git submodule update --init --recursive --depth=1
```
astarche, line 58: Note: this is a bonus change because I got tired of waiting so long for each successive build failure. Doesn't improve overall build time by much, but time-to-early-failure is a nice improvement.

cmake/CreateVirtualEnvironment.cmake
```diff
@@ -3,7 +3,7 @@
 CreateVirtualEnvironment
 ------------------------
 Function that creates a virtual Python environment with specific requirements.
-From https://schemingdeveloper.com/2020/07/02/how-to-create-a-new-python-virtual-environment-in-cmake/
+Inspired by https://schemingdeveloper.com/2020/07/02/how-to-create-a-new-python-virtual-environment-in-cmake/
```
astarche, line 6: Note: I chopped this up a lot while trying to debug it, plus I needed to move the core logic to python for the fix (and ultimately that also helped diagnose the problem).

The original script is nice, but had a lot of features we didn't need that made it hard for me to follow and also fueled a lot of bogus theories.

python_build_requirements.txt
```diff
@@ -1,4 +1,4 @@
 contextlib2==21.6.0
```
astarche, line 1: I renamed this because `requirements` is more standard and I keep forgetting the name. As long as requirements.txt is in there, I'm happy with it.

gregstoll (approved):

SparkingSpork (approved):

PR #430, opened 2021-11-16
Initial RFmxSpecAn implementation

gregstoll:

Adds RFmxSpecAn service with Initialize, Close, and a few GetError* functions.

Start work on RFmxSpecAn.

Session tests pass on my machine.


source/codegen/metadata/nirfmxspecan/config.py
```diff
@@ -0,0 +1,57 @@
...
+    'service_class_prefix': 'NiRFmxSpecAn',
+    'java_package': 'com.ni.grpc.nirfmxspecan',
+    'csharp_namespace': 'NationalInstruments.Grpc.NiRFmxSpecAn',
+    'namespace_component': 'nirfmxspecan',
+    'close_function': 'Close',
+    'custom_types': [],
+    'additional_headers': {},
+    'type_to_grpc_type': {
+        'niRFmxInstrHandle': 'nidevice_grpc.Session',
+        'char[]': 'string',
```
gregstoll, line 15: I just took the rest of this from DAQmx, not sure if we'll need all of these or not.

source/codegen/metadata/nirfmxspecan/config.py
```diff
@@ -0,0 +1,57 @@
...
+    'feature_toggles': {},
+    'driver_name': 'NI-RFMXSPECAN',
+    'extra_errors_used': [
+    ],
+    'init_function': 'Initialize',
+    'resource_handle_type': 'niRFmxInstrHandle',
+    'library_info': {
+        'Linux': {
+            '64bit': {
+                'name': 'nirfmxspecan',
```
gregstoll, line 37: This is just a guess. Actually I guess this isn't supported on Windows, so maybe I should just get rid of this?

source/codegen/metadata/nirfmxspecan/config_addon.py
```diff
@@ -0,0 +1,6 @@
+# We need to maintain the version here since it needs to be updated by the build process on GitHub
+# TODO
+config_additional_config = {
+    'module_version': '1.3.3.dev0',
```
gregstoll, line 4: I'm not sure what this needs to be - is this the version of grpc-device?

source/codegen/metadata/nirfmxspecan/functions.py
```diff
@@ -0,0 +1,103 @@
...
+    'Close': {
+        'parameters': [
+            {
+                'direction': 'in',
+                'name': 'instrumentHandle',
+                'type': 'niRFmxInstrHandle'
+            },
+            {
+                'direction': 'in',
+                'enum': 'Boolean',
```
gregstoll, line 11: I left this as an enum'd boolean for now, FYI.

source/codegen/metadata/nirfmxspecan/functions.py
```diff
@@ -0,0 +1,103 @@
...
+                    'mechanism': 'ivi-dance',
+                    'value': 'errorDescriptionBufferSize'
+                },
+                'type': 'char[]'
+            }
+        ],
+        'returns': 'int32'
+    },
+    'Initialize': {
+        'custom_close': 'Close(id, RFMXSPECAN_VAL_FALSE)',
```
gregstoll, line 77: I had to add this because the `Close()` method takes that extra parameter.

source/codegen/templates/service_helpers.mako
```diff
@@ -17,6 +18,7 @@
   session_field_name = next(explicit_session_params, 'session_name')
 %>\
 ${initialize_input_params(function_name, parameters)}
+${initialize_output_params(output_parameters_to_initialize)}\
```
gregstoll, line 21: Since `Initialize()` returns an extra output parameter we need to initialize it.

source/codegen/metadata/nirfmxspecan/config_addon.py
astarche: I think this is leftover from nimi-python.

source/codegen/metadata/nirfmxspecan/config.py
astarche: fgen and digitalpattern also don't support linux and also went with "just guess". I think that's fine. If removing it does something reasonable, that's fine too. But I'm guessing it would so something weird.

source/codegen/common_helpers.py
```diff
@@ -312,6 +313,9 @@ def pascal_to_camel(pascal_string):
 
     # Normal case: NormalCase -> normalCase
     match = re.fullmatch(r"([A-Z])(.*)", pascal_string)
+    if not match:
```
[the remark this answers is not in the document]
astarche: I think you mean it was "implicitly assuming that the resource_handle_type was **pascal** case", that tripped me up.

Consider: `camel_to_snake` starts with a `is_actually_pascal` check. You could do sort of the inverse here and early return `if not is_actually_pascal()`

astarche (approved):

SparkingSpork (approved):

reckenro (approved):

PR #433, opened 2021-11-22
codegen changes that will be needed for RFmxSpecAn

gregstoll:

Codegen changes that will be needed to build RFmxSpecAn. This change doesn't include adding the RFmxSpecAn metadata - that will come in a separate change.

Needed for RFmxSpecAn.

Added some unit tests, verify they all pass.


source/codegen/common_helpers.py
```diff
@@ -495,7 +495,10 @@ def get_twist_value(parameters):
     for p in parameters:
         if is_array(p['type']):
             size = p.get('size', {})
-            return size.get('value_twist', None)
+            value_twist = size.get('value_twist', None)
```
gregstoll, line 498: This function was looking for the first array in the list of parameters and assuming that was the input that had a `value_twist`. Now we only return a non-None `value_twist` (if there is one).

source/codegen/proto_helpers.py
```diff
@@ -50,15 +50,16 @@ def is_decomposable_enum(parameter: dict):
   This is because protobuf does not support oneof on repeated types, so the standard
   input decomposition does not work for arrays.
   """
-  return common_helpers.is_enum(parameter) and not is_array_input(parameter)
+  return common_helpers.is_enum(parameter) and not (is_array_input(parameter) and parameter["grpc_type"] != "string")
```
gregstoll, line 53: There are a handful of changes like this where we were treating `char[]` parameters to functions as arrays. I think we got away with it because most places these are marked as `ViString` and `is_array_input()` returns False for these. Note that I added unit tests for this case.

source/codegen/service_helpers.py
```diff
@@ -55,6 +55,33 @@ def create_args_for_callback(parameters):
     ])
 
 
+def create_standard_arg(parameter):
```
gregstoll, line 58: Extracted this method so we could call it from the initial call in the `ivi_dance` case.

source/codegen/service_helpers.py
```diff
@@ -55,6 +55,33 @@ def create_args_for_callback(parameters):
...
+        return f'({type_without_brackets}*){parameter_name}.data(), '
+    elif parameter['type'] in {"ViBoolean[]", "ViSession[]", "ViInt16[]"}:
+        return f'{parameter_name}.data(), '
+    elif 'callback_params' in parameter:
+        return f'CallbackRouter::handle_callback, '
+    elif 'callback_token' in parameter:
+        return f'handler->token(), '
+    elif not is_output and common_helpers.is_pointer_parameter(parameter) and 'hardcoded_value' not in parameter:
+        return f'&{parameter_name}_copy, '
+    elif common_helpers.is_enum(parameter) and parameter["type"] == 'char[]':
```
gregstoll, line 74: This is a new case. For enums that are strings, we get a temporary variable that's a `std::string`, so we need to pass in `.data()`. See `InputStringValuedEnum()` in `ni_fake_non_ivi_service.cpp` for an example of how this all looks in the generated code.

source/codegen/service_helpers.py
```diff
@@ -174,6 +182,15 @@ def expand_varargs_parameters(parameters):
     return new_parameters
 
 
+def create_library_argument(parameter):
+    parameter_type = parameter.get('type', '')
+    # This is a hack. niRFmxInstr.h typdefs int8 as "char" (instead of "signed char"), but our
```
gregstoll, line 187: Hopefully the comment explains this well enough - just calling it out in the PR.

source/codegen/templates/service_helpers.mako
```diff
@@ -352,13 +352,16 @@ ${initialize_standard_input_param(function_name, parameter)}
...
+  variable_type = "std::string" if parameter['type'] == 'char[]' else parameter['type']
 %>\
-      ${parameter['type']} ${parameter_name};
+      ${variable_type} ${parameter_name};
       switch (request->${field_name}_enum_case()) {
 % if has_unmapped_enum:
         case ${one_of_case_prefix}::k${pascal_field_name}: {
 %   if parameter['type'] in ["ViConstString", "ViString"]:
           ${parameter_name} = const_cast<${parameter['type']}>((${enum_request_snippet}).c_str());
+%   elif parameter['type'] == "char[]":
```
astarche, line 363: Why aren't these all merged into the above cases for ViString/ViConstString? And could those all be:

`if parameter['grpc_type'] == 'string'`

?

source/codegen/service_helpers.py
astarche: What happens if we change the `signed char[]` to `int8[]` in the extraction code? Does that just work?

I'm interpreting this as "2 wrongs":
1. CppHeaderParser doesn't preserve the nominal types for those semi-standard int types (in this case `int8`).
2. rfmx defines `int8` in a weird way.

For DAQ, we already compensate for (1) by correcting those types during extraction (for `int32` and `uint32`. So it seems reasonable to do that for `int8`. So `type` always matches the library type. But ... maybe it would still need to do something in the service/coercion code.

source/codegen/service_helpers.py
astarche: Why does this check `parameter["type"] == 'char[]'` instead of `grpc_type == string`?

Per offline discussion, I think the `const` cast will no longer be necessary when we have everything on C++17 (https://en.cppreference.com/w/cpp/string/basic_string/data). Consider a comment here on that.

source/codegen/service_helpers.py
astarche: `standard` as opposed to what? Is this just differentiating from the size mechanism args?

(Probably the best way to address this is a doc comment. But if a great name comes to you: even better)

astarche (approved):

source/codegen/service_helpers.py
gregstoll: Yeah, I was able to do this in scrapigen and I think it looks a bit cleaner here.

source/codegen/service_helpers.py
gregstoll: Yeah, I was able to refactor things a bit here to share this logic so this case is gone. Thanks!

reckenro (approved):

PR #435, opened 2021-11-30
add basic spectrum example for RFmxSpecAn

gregstoll:

Adds a basic spectrum example for RFmxSpecAn, modeled after a C example.

We haven't decided exactly what examples we are going to write, but this is a good one to start with (and a good proof point that the API works)

Ran it on my machine, and I get the results:

> min frequency: 999.500 MHz: -51.0 dBm
> midpoint frequency: 1.000 GHz: -8.5 dBm
> max frequency: 1.001 GHz: -51.0 dBm

which lines up with the graphs seen in the LabVIEW and C examples.

### DIFF
--- a/examples/nirfmxspecan/spectrum-basic.py
+++ b/examples/nirfmxspecan/spectrum-basic.py
@@ -1,0 +1,151 @@
+# This program demonstrates the use of RFmx SpecAn to read a spectrum. The steps are:
+# 1. Open a new RFmx session
+# 2. Configure the basic signal properties  (Center Frequency, Reference Level and External Attenuation)
+# 3. Configure Spectrum Span
+# 4. Configure Spectrum RBW filter
+# 5. Configure Spectrum Averaging
+# 6. Read Spectrum Measurement Results
+# 7. Close the RFmx Session
+#
+# The gRPC API is built from the C API. RFmx SpecAn documentation is installed with the driver at:
+# C:\Program Files (x86)\National Instruments\RFmx\SpecAn\Documentation\rfmxspecancvi.chm
+#
+# Getting Started:
+#
+# To run this example, install "RFmx SpecAn" on the server machine.
+# Link: # https://www.ni.com/en-us/support/downloads/software-products/download.rfmx-specan.html
+#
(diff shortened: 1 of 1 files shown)


examples/nirfmxspecan/spectrum-basic.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+        )
+    )
+
+    raise_if_error(
+        client.SpectrumCfgRBWFilter(
+            nirfmxspecan_types.SpectrumCfgRBWFilterRequest(
+                instrument_handle=instr, selector_string="",
+                rbw_auto=nirfmxspecan_types.SPECTRUM_RBW_AUTO_BANDWIDTH_TRUE,
+                rb_w=100e3,
```
astarche, line 115: (Not necessarily for this PR): Looks like we need to fix this param name. Not sure what happens when this goes through the various transformations. But I think the fix is probably: "RBW" -> "rbw" in the metadata.

examples/nirfmxspecan/spectrum-basic.py
```diff
@@ -0,0 +1,151 @@
...
+        f /= 1000.0
+    return f"{f:.3f} G{suffix}"
+
+
+# Raise an exception if an error was returned
+def raise_if_error(response):
+    if response.status != 0:
+        error_response = client.GetError(
+            nirfmxspecan_types.GetErrorRequest(
+                instrument_handle=instr,
```
astarche, line 75: (Not necessarily for this PR): This `instrument_handle` name seems too long. The obvious thing is to drop handle and go with just `instrument`, which is what we did for DAQ `taskHandle`s. That seems pretty good. `instr` is appealing too, but it's better to err on the side of complete words these days. Any other ideas?

examples/nirfmxspecan/spectrum-basic.py
gregstoll: Yeah, the metadata has the parameter as "rbW". Filed #436 for this.

examples/nirfmxspecan/spectrum-basic.py
gregstoll: Yeah, I like `instrument` - filed #437.

PR #440, opened 2021-12-01
add the beginning of SpecAn driver API tests, fix some issues with attributes

gregstoll:

Add a few driver API tests for RFmxSpecAn.  This found a few issues:
- We were not correctly generating attribute value enums. This was because (among other things) the attributeID parameters to the get/set/reset functions did not have an enum specified.
- This revealed that `split_attributes_by_type` doesn't work right now, because there are many types that have no attributes of that type (for example: uint32, int32[]), and so those functions use an enum type that we don't generate. I think ideally we would like to split them by type since there are so many, so we should figure out whether we're missing something or there really are no attributes of that type and we can get rid of those versions of GetAttribute* and SetAttribute*. (it is possible that RfmxInstr affects this somehow)

Progress towards writing driver API tests for RFmxSpecAn.

New tests pass.


source/codegen/metadata_mutation.py
```diff
@@ -248,7 +248,7 @@ def expand_attribute_function_value_param(function, enums, attribute_enums_by_ty
             value_param['name'] = value_param['name'] + "_raw"
 
 def get_attribute_function_value_param(function):
-    return next((param for param in function["parameters"] if param["name"] in {"value", "attributeValue"}), None)
+    return next((param for param in function["parameters"] if param["name"] in {"value", "attributeValue", "attrVal"}), None)
```
gregstoll, line 251: Another attribute problem is that the setter attribute functions have the value parameter named `attrVal`.

SparkingSpork (approved):

astarche (approved):

reckenro (approved):

Issue #444, opened 2021-12-07
Enum aliases cause problems in json replay style use cases


astarche:
Avoiding enum aliases is considered a best practice because the JSON representation has to just "pick one". This issue is described [here](https://docs.buf.build/lint/rules#enum_no_allow_alias).

We have some enums where the C API has a zero value but we also generate an UNSPECIFIED zero value. These are marked as allow_alias to allow for that duplicate. When you do a json trace of a gRPC session (i.e., in grpc-dump or with an interceptor), you often get UNSPECIFIED when the actual sent value was something else.

Unfortunately, using UNSPECIFIED for the zero value of an enum is **also** a [best practice](https://docs.buf.build/lint/rules#enum_zero_value_suffix). So, these 2 conventions are in conflict, assuming that we want to preserve the mapping to the C API.

We have been discussing using gRPC json based APIs for SCPI style workflows, which would include trace/copy/paste. The aliases **break that workflow** because you end up capturing bogus "UNSPECIFIED" values. (OR does this work but just look weird?)

We need some research to consider our options:
1. Can we remove the UNSPECIFIED field when there is a zero value in the C API? What is the effect on backwards compatibility (assuming users aren't intentionally sending UNSPECIFIED)?
2. Can we use "mapped" enums in this case? So that the enum values aren't necessarily the same as the C API values.
3. Are "raw" values an acceptable alternative? Or does this work "well enough"?

[AB#1758247](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1758247)

astarche:
AtributeValues have some aliasing even wwithout the UNSPECIFIED attribute, just because there can be overlap in the various enums that are pulled into the aggregate enum. 


SpecAn is actually the worst:

```
  NIRFMXSPECAN_INT32_ACP_AMPLITUDE_CORRECTION_TYPE_RF_CENTER_FREQUENCY = 0;
  NIRFMXSPECAN_INT32_ACP_AMPLITUDE_CORRECTION_TYPE_SPECTRUM_FREQUENCY_BIN = 1;
  NIRFMXSPECAN_INT32_ACP_AVERAGING_ENABLED_FALSE = 0;
  NIRFMXSPECAN_INT32_ACP_AVERAGING_ENABLED_TRUE = 1;
  NIRFMXSPECAN_INT32_ACP_AVERAGING_TYPE_RMS = 0;
[...]

PR #445, opened 2021-12-07
Change TRUE/FALSE enums to bools in RFmxSpecAn

astarche:

Replace all enums where the only values are TRUE/FALSE with bools.

Update converters.h to handle copy conversion of arrays of booleans in requests.

TRUE/FALSE enums are clunky and have the aliasing problem from #444 

RFmx C# APIs also use bools for bools.

Ran and passed all RFmx tests and new fake_non_ivi test.


gregstoll:
There are a few cases in the example too :-(

gregstoll (approved):

Issue #446, opened 2021-12-08
Coerced scalar inputs do not implement range checking


astarche:
This is implemented for arrays and structs but not for scalars. This affects `int16`, `uInt16`, and `int8` params. All of these are represented as 32-bit types in protobuf. If a number is passed in that doesn't fit in the driver type range, it will be truncated.

[AB#1759778](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1759778)

PR #448, opened 2021-12-08
Add system tests for RFmxSpecAn

astarche:

Add system level tests based on examples and exercising datatypes that require conversion.

Fills out acceptance table for initial SpecAn extraction feature.

Ran and passed all RFmx Tests.

Stepped through debugger on attribute conversion tests to ensure that the values were correct when passed to the driver.

### DIFF
--- a/generated/nirfmxspecan/nirfmxspecan_service.cpp
+++ b/generated/nirfmxspecan/nirfmxspecan_service.cpp
@@ -11703,6 +11703,9 @@ namespace nirfmxspecan_grpc {
     catch (nidevice_grpc::LibraryLoadException& ex) {
       return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
     }
+    catch (nidevice_grpc::ValueOutOfRangeException& ex) {
+      return ::grpc::Status(::grpc::OUT_OF_RANGE, ex.what());
+    }
   }
 
   //---------------------------------------------------------------------
@@ -11992,6 +11995,9 @@ namespace nirfmxspecan_grpc {
     }
     catch (nidevice_grpc::LibraryLoadException& ex) {
       return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
+    }
+    catch (nidevice_grpc::ValueOutOfRangeException& ex) {
+      return ::grpc::Status(::grpc::OUT_OF_RANGE, ex.what());
     }
   }
 
--- a/source/tests/system/nirfmxspecan_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxspecan_driver_api_tests.cpp
@@ -1,3 +1,4 @@
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 #include <thread>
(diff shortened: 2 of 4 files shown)


generated/nirfmxspecan/nirfmxspecan_service.cpp
```diff
@@ -11703,6 +11703,9 @@ namespace nirfmxspecan_grpc {
     catch (nidevice_grpc::LibraryLoadException& ex) {
       return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
     }
+    catch (nidevice_grpc::ValueOutOfRangeException& ex) {
```
astarche, line 11706: Note: scalar coercions like these are not actually implemented. Filed #446

gregstoll (approved):

source/tests/system/nirfmxspecan_driver_api_tests.cpp
```diff
@@ -106,10 +160,153 @@ TEST_F(NiRFmxSpecAnDriverApiTests, SpectrumBasicFromExample_DataLooksReasonable)
...
+      client::set_attribute_u16(stub(), session, "", NiRFmxSpecAnAttribute::NIRFMXSPECAN_ATTRIBUTE_NF_EXTERNAL_PREAMP_PRESENT, 400));
+}
+
 TEST_F(NiRFmxSpecAnDriverApiTests, SetAndGetAttributeInt32_Succeeds)
 {
   auto session = init_session(stub(), PXI_5663);
-  EXPECT_SUCCESS(session, client::set_attribute_i32(stub(), session, "", NiRFmxSpecAnAttribute::NIRFMXSPECAN_ATTRIBUTE_NF_EXTERNAL_PREAMP_PRESENT, NiRFmxSpecAnInt32AttributeValues::NIRFMXSPECAN_INT32_NF_EXTERNAL_PREAMP_PRESENT_TRUE));
+  EXPECT_SUCCESS(
+      session,
+      client::set_attribute_i32(stub(), session, "", NiRFmxSpecAnAttribute::NIRFMXSPECAN_ATTRIBUTE_NF_EXTERNAL_PREAMP_PRESENT, NiRFmxSpecAnInt32AttributeValues::NIRFMXSPECAN_INT32_NF_EXTERNAL_PREAMP_PRESENT_TRUE));
```
gregstoll, line 309: Hmm, I thought we got rid of all the "TRUE" enum values?

astarche: Attributes are different. That's a big "attribute_values" enum that includes bool enums as well as non-bool enums. So the simple swap doesn't work.

We could add a `get/set_attribute_bool` wrapper that calls into `get/set_attribute_i32`. I think that would be cool, although nothing would stop you from calling it for some other int32 attribute that's not a bool (though is that worse than allowing passing in cross-enums?).

gregstoll: Yeah, the bool wrapper does sound kinda nice but I'm not sure it's worth the complication...

reckenro (approved):

PR #455, opened 2021-12-10
API cleanup and fixes for RFmxSpecAn service

astarche:
### What does this Pull Request accomplish?

Fixes #451 
Fixes #450

* Promote `BuildSpurString2` to `BuildSpurString`
* Fix handling of ivi-twist arrays when there is also an input array.
* Fix direction of DpdApplyPredistortion methods.
* Add tests for changed code and support for loading test data sets extracted from C API examples.

### Why should this Pull Request be merged?

Incremental progress on RFmx SpecAn feature.

### What testing has been done?

Ran and passed all RFmx tests and new UnitTest for size mechanism.

PR #456, opened 2021-12-10
Add support for optional linked array inputs

### LINKED ISSUE #452: SpecAn SpurCfgRangeAbsoluteLimit has optional array parameters that are not supported by grpc-device
C API docs: https://zone.ni.com/reference/en-XX/help/374264N-01/rfmxspecancvi/cvirfmxspecan_spurcfgrangeabso

We don't implement the convention of "Specify NULL if you do not want to set a value for this array", so all arrays must be specified (empty will report an error for mismatched sizes).

I think we can address this b introducing the concept of an optional field and making sure that our codegen supports passing it as null.

[AB#1761383](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1761383)

astarche:

Fixes #452 

Add support for optional arrays among linked input arrays.

At least one method in RFmxSpecAn needs to support optional arrays, where "no value" is signalled as NULL. Note that `repeated optional` does not exist in protobuf and protobuf arrays always have a default, non-null empty value. So I'm handling this as a special case of linked arrays when:
1. All non-zero sizes are the same.
2. At least one array has a zero size.
3. Set that param to null.

Ran and passed new tests and all RFmx tests.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -301,10 +312,18 @@ namespace nidaqmx_grpc {
       auto trigger_sources = request->trigger_sources().c_str();
       auto trigger_slope_array = reinterpret_cast<const int32*>(request->trigger_slope_array().data());
       auto trigger_level_array = const_cast<const float64*>(request->trigger_level_array().data());
-      if (request->trigger_slope_array().size() != request->trigger_level_array().size()) {
-        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The sizes of repeated fields trigger_slope_array and trigger_level_array do not match");
+      auto array_size_determine_from_sizes = std::array<int, 2>
```
astarche, line 315: The "no optional" version of the code is not an improvement on the original. But I figure it's better to have one implementation of this check.

source/server/converters.h
```diff
@@ -26,6 +26,44 @@ inline void trim_trailing_nulls(std::string& s)
       s.end());
 }
 
+enum MatchState {
```
astarche, line 29: Here's the C++ code that implements the check.

generated/nirfmxspecan/nirfmxspecan_service.cpp
```diff
@@ -12749,13 +12942,25 @@ namespace nirfmxspecan_grpc {
...
+        request->absolute_limit_stop_size()
+      };
+      const auto number_of_elements_size_calculation = calculate_linked_array_size(number_of_elements_determine_from_sizes, true);
+
+      if (number_of_elements_size_calculation.match_state == MatchState::MISMATCH) {
+        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The sizes of repeated fields TODO do not match");
       }
-      if (request->absolute_limit_start().size() != request->absolute_limit_stop().size()) {
-        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The sizes of repeated fields absolute_limit_start and absolute_limit_stop do not match");
+      // NULL out optional params with zero sizes.
```
astarche, line 12956: This is what the code looks like when optionals need to be handled.

source/codegen/metadata_mutation.py
```diff
@@ -64,7 +64,12 @@ def mark_non_proto_params(parameters):
             if mechanism == 'len':
                 if 'determine_size_from' not in size_param:
                     size_param['determine_size_from'] = []
+                    size_param['linked_params_are_optional'] = True
                 size_param['determine_size_from'].append(param['name'])
+                is_optional = common_helpers.is_optional(param)
+                if is_optional and not size_param['linked_params_are_optional']:
+                    raise Exception("Code generator does not support linked params that aare a mix of optional and required", size_param)
```
gregstoll, line 71: typo: "are"

source/codegen/metadata_mutation.py
```diff
@@ -64,7 +64,12 @@ def mark_non_proto_params(parameters):
             if mechanism == 'len':
                 if 'determine_size_from' not in size_param:
                     size_param['determine_size_from'] = []
+                    size_param['linked_params_are_optional'] = True
```
gregstoll, line 67: I'm a little confused - is there a reason to set this here since we're unconditionally setting it below?

gregstoll: Oh, I see, because you're using the value to make sure there's not a mix. Nevermind.

source/codegen/metadata_mutation.py
```diff
@@ -64,7 +64,12 @@ def mark_non_proto_params(parameters):
...
                 size_param['determine_size_from'].append(param['name'])
+                is_optional = common_helpers.is_optional(param)
+                if is_optional and not size_param['linked_params_are_optional']:
```
gregstoll, line 70: This won't detect the case when the first linked parameter is optional and the second linked parameter is not optional, right? Maybe it makes more sense to initialize `size_param['linked_params_are_optional']` to `None` above instead of True, then you can do something like

```python
if size_param['linked_params_are_optional'] is not None and size_param['linked_params_are_optional'] != is_optional:
```

source/server/converters.h
```diff
@@ -26,6 +26,44 @@ inline void trim_trailing_nulls(std::string& s)
...
+  int size = 0;
+  MatchState match_state = MatchState::MISMATCH;
+};
+
+template <size_t N>
+inline LinkedArraySize calculate_linked_array_size(std::array<int, N> sizes, bool allow_optional)
+{
+  auto result = LinkedArraySize{};
+
+  if (std::adjacent_find(sizes.begin() + 1, sizes.end(), std::not_equal_to<>()) == sizes.end()) {
```
gregstoll, line 45: Why does this start at `sizes.begin() + 1`? Won't this miss the case where the first two elements are non-zero but different? I would think it should start at `sizes.begin()`...

gregstoll (approved):
Yeah, I like this. I was a little skeptical of the need for a helper function, but after seeing what that function looks like it's definitely better to avoid inlining that everywhere :-)

source/server/converters.h
astarche: Oops. Yeah, there are alternate implementations where you compare everything to the first element, which start with `begin() + 1`. But `adjacent_find` needs to start at the beginning.

generated/nirfmxspecan/nirfmxspecan_service.cpp
```diff
@@ -15,8 +15,10 @@
 
 namespace nirfmxspecan_grpc {
 
+  using nidevice_grpc::converters::calculate_linked_array_size;
```
astarche, line 18: TODO: merge this with #455 and update the test case that exercises this.

source/codegen/metadata_mutation.py
astarche: Yikes, that wasn't even close. I addressed this a little bit differently.

```python
            if mechanism == 'len':
                if 'determine_size_from' not in size_param:
                    size_param['determine_size_from'] = []
                size_param['determine_size_from'].append(param['name'])
                is_optional = common_helpers.is_optional(param)
                if is_optional != size_param.get('linked_params_are_optional', is_optional):
                    raise Exception("Code generator does not support linked params that are a mix of optional and required", size_param)
                size_param['linked_params_are_optional'] = is_optional
```

Basically `size_param.get('linked_params_are_optional', is_optional)` is shorthand for "missing is considered a match".

I also did some manual testing around this and it looked OK.

source/tests/unit/converters_tests.cpp
```diff
@@ -32,6 +36,56 @@ TEST(ConvertersTests, StringResizedToAddTrailingNulls_TrimTrailingNulls_StringNo
   EXPECT_EQ(DATA, copy);
 }
 
+TEST(ConvertersTests, AllEqualSizesAllowOptional_Match)
+{
+  constexpr auto SIZES = std ::array<int, 3>{3, 3, 3};
+  const auto calculation = calculate_linked_array_size(SIZES, true);
+
+  EXPECT_EQ(3, calculation.size);
+  EXPECT_EQ(true, calculation.match_state);
```
reckenro, line 45: Should this be `MatchState::Match` instead of `true`?

PR #460, opened 2021-12-14
Add ability to override grpc name for parameters

astarche:
### What does this Pull Request accomplish?

* Add ability to override grpc name for parameters
* Use for:
   * Workaround capitalization of `dB` breaking `camel_to_snake`.
   * Translation of `*handle*` to `instrument` in RFmx.
* Remove all direct usages of `camel_to_snake` and call into more descriptive helper methods.

### Why should this Pull Request be merged?

Fixes #449 

Direct calls to `camel_to_snake` cause a lot of confusion in the codebase.

Specifically overriding `grpc_name` rather than patching the `name` in the metadata preserves the connection to the header file and avoids coupling to string conversion implementation (i.e., if we changed `dB` to `Db`).

### What testing has been done?

Ran and passed all RFmx tests.

PR #461, opened 2021-12-15
Add support for cross-driver sessions

astarche:

Add support for cross-driver sessions in grpc-device to enable RFmx methods that access or initialize from ivi ViSessions. This change allows using grpc-device sessions from the corresponding drivers in those APIs.

Steps:
1. Pass in the session resource repository for all cross-driver sessions required by a service.
2. Ensure that code taking those sessions as inputs looks them up with `access_session` on the correct repository.
3. Methods that return sessions from other drivers are "cross driver init methods": they need to register that session with the correct repository.
4. Ensure that sessions created with "cross driver init methods" are removed from the repository when the initiating session is closed (the handles are closed internally by the driver).

This was flagged as an important feature by the RFmx teams and the SEBU. It allows intermixing RFmx code with functionality in other drivers.

Ran and passed all RFmx and unit tests, including new tests for cross-driver sessions.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -30,7 +30,9 @@ namespace nidaqmx_grpc {
       NiDAQmxLibraryInterface* library,
       ResourceRepositorySharedPtr session_repository, 
       const NiDAQmxFeatureToggles& feature_toggles)
-      : library_(library), session_repository_(session_repository), feature_toggles_(feature_toggles)
```
astarche, line 33: Splitting these to separate lines just makes it easier to conditionally codegen additional arguments

generated/nirfmxinstr/nirfmxinstr_service.cpp
```diff
@@ -26,8 +26,12 @@ namespace nirfmxinstr_grpc {
   NiRFmxInstrService::NiRFmxInstrService(
       NiRFmxInstrLibraryInterface* library,
       ResourceRepositorySharedPtr session_repository, 
+      ViSessionResourceRepositorySharedPtr vi_session_resource_repository,
```
astarche, line 29: RFmxInstr is a good example of the new C++ code that (a) passes through the repository pointer and (b) implements the access/init methods.

source/codegen/metadata/nifake_non_ivi/functions.py
```diff
@@ -10,6 +10,24 @@
         ],
         'returns': 'int32'
     },
+    'GetCrossDriverSession': {
```
astarche, line 13: All of the tags to make this work are a little verbose. I wanted to keep the codegen logic a little simpler, so I didn't code up mutations to infer that "cross_driver_session" outputs should be tagged as init_methods with a Session grpc_type. Feedback welcome though.

source/codegen/metadata/nirfmxinstr/config.py
```diff
@@ -9,7 +9,7 @@
     'namespace_component': 'nirfmxinstr',
     'close_function': 'Close',
     'custom_types': [],
-    'additional_headers': {},
+    'additional_headers': { "ivi.h": ["service.h"] },
```
astarche, line 12: For `ViSession`

source/codegen/templates/service_helpers.mako
```diff
@@ -39,6 +39,34 @@ ${initialize_output_params(output_parameters_to_initialize)}\
...
 </%def>
 
+<%def name="define_cross_driver_init_method_body(function_name, function_data, parameters)">\
+<%
+  initiating_driver_input_param = next(p for p in parameters if common_helpers.is_input_parameter(p) and p['grpc_type'] == 'nidevice_grpc.Session')
+  output_parameters = [p for p in parameters if common_helpers.is_output_parameter(p)]
+  session_output_param = next(p for p in output_parameters if p['grpc_type'] == 'nidevice_grpc.Session')
+  output_parameters_to_initialize = [p for p in output_parameters if p['grpc_type'] != 'nidevice_grpc.Session']
+  cross_driver_dep = service_helpers.get_cross_driver_session_dependency(session_output_param)
+  session_output_var_name = common_helpers.camel_to_snake(session_output_param['cppName'])
```
astarche, line 49: TODO: merge with #460, No raw `camel_to_snake`.

source/codegen/metadata/nifake_non_ivi/functions.py
gregstoll: Yeah, we're not talking about a lot of functions like this so I think I like the more verbose approach to keep the codegen simpler.

gregstoll: One thing to think about is if we can add some metadata validation rules to this - for example, if a parameter has a `cross_driver_session` then the function should be an `init_method` and things like that. At least then we will know when we get the boilerplate wrong...

gregstoll (approved):

reckenro (approved):

PR #466, opened 2022-01-12
Create initial RFmx NR service with Init/Close.

reckenro:

Adds handcoded metadata for RFmx NR to have a simple service with Initialize and Close methods.

Closes [AB#1788691](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1788691)

We want to get in an initial version of the RFmx NR service with Init/Close to ensure the framework is set up correctly in grpc-device to communicate with the driver.

Ran the added SystemTests locally with RFmx NR installed.


reckenro:
I wanted to run these on a Windows VM but couldn't get that quite set up so I just resolved to install the driver locally on my dev machine. I'll get the Windows VM set up, though, so future system tests and manual testing later on with Examples can be done with it as the server.

image

generated/nirfmxnr/nirfmxnr.proto
```diff
@@ -0,0 +1,57 @@
...
+
+message InitializeRequest {
+  string session_name = 1;
+  string resource_name = 2;
+  string option_string = 3;
+}
+
+message InitializeResponse {
+  int32 status = 1;
+  nidevice_grpc.Session handle_out = 2;
```
reckenro, line 54: I just used the same param names as what was in the C Header. Figured when we update the handcoded metadata with the generated metadata I can fix up the session tests to use whatever parameter names we land on.

source/tests/system/nirfmxnr_session_tests.cpp
```diff
@@ -0,0 +1,117 @@
+#include <gtest/gtest.h>
```
reckenro, line 1: Note: Since this was so similar to the `nirfmxspecan_session_tests` I figured I'd give these the same versions to get that coverage. I also wanted to have the `driver_api_tests` set up as well for future System Level tests so I went ahead and added it with a simple `Init_Close_Succeeds` test. I can delete the Initialize test in the `driver_api_tests` when we add more, but thought it'd be good to have at least one test in that style up front to have the `experimental::client` framework validated for NR.

generated/nirfmxnr/nirfmxnr.proto
astarche: Yeah, I think we're pretty set on `instrument` for both `handle_out` and `instrument_handle` in. And any other changes would be special cases (bad snake_case conversions mostly).

But you can apply that later.

source/codegen/metadata/nirfmxnr/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .functions import functions
```
astarche, line 1: ^^^ Did you use the lib from the installer and/or my branch?

source/tests/system/nirfmxnr_session_tests.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include "device_server.h"
+#include "nirfmxnr/nirfmxnr_service.h"
+
+namespace ni {
+namespace tests {
+namespace system {
+
```
astarche, line 9: Also anonymous namespace here.

astarche (approved):

source/codegen/metadata/nirfmxnr/__init__.py
reckenro: Yeah, I just took it from your branch before I had it set up and updated it in this commit: https://github.com/ni/grpc-device/pull/466/commits/7ab5c264c54642546becdb0a95cda5260f2df266.

source/codegen/metadata/nirfmxnr/config.py
```diff
@@ -0,0 +1,61 @@
...
+            '64bit': {
+                'name': 'niRFmxNR.dll',
+                'type': 'cdll'
+            }
+        }
+    },
+    'metadata_version': '0.1',
+    'module_name': 'nirfmxnr',
+    'session_class_description': 'An NI-RFmxNR instrument handle',
+    'session_handle_parameter_name': 'instrumentHandle'
```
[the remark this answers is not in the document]
reckenro: Yeah, these aren't used. I removed it from NR, SpecAn, and Instr `config.py` files.

source/tests/system/nirfmxnr_driver_api_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "nirfmxnr/nirfmxnr_client.h"
+
+using namespace ::testing;
+using namespace nirfmxnr_grpc;
+namespace client = nirfmxnr_grpc::experimental::client;
+
+namespace ni {
+namespace tests {
+namespace system {
+
```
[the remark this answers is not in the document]
reckenro: Ah, that makes sense. Added the unnamed namespace here and in the session tests to uniquify things.

source/tests/system/nirfmxnr_session_tests.cpp
reckenro: Done.

source/tests/system/nirfmxnr_session_tests.cpp
```diff
@@ -0,0 +1,119 @@
...
+    ::grpc::Status status = GetStub()->Initialize(&context, request, response);
+    return status;
+  }
+
+ private:
+  DeviceServerInterface* device_server_;
+  std::unique_ptr<rfmxnr::NiRFmxNR::Stub> nirfmxnr_stub_;
+};
+
+TEST_F(NiRFmxNRSessionTest, InitializeSessionWithDeviceAndSessionName_CreatesDriverSession)
```
mgumble, line 53: I made a comment on your work item, but we should capture these tests in the Acceptance table.

I also would like to test initializing a 2nd session to the same device

SparkingSpork (approved):

generated/nirfmxnr/nirfmxnr_service.cpp
```diff
@@ -0,0 +1,119 @@
...
+      if (status_ok(status)) {
+        response->mutable_handle_out()->set_id(session_id);
+      }
+      return ::grpc::Status::OK;
+    }
+    catch (nidevice_grpc::LibraryLoadException& ex) {
+      return ::grpc::Status(::grpc::NOT_FOUND, ex.what());
+    }
+  }
+
```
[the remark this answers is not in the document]
reckenro: I think it's just some line endings are messed up between the different if blocks at the end of the `service.cpp.mako` file that generates this file per service. I see this in several other service.cpp files, but interestingly not all. I'll look into this as a separate cleanup change.

PR #471, opened 2022-01-14
Fix initializing two or more services to same device on RFmx.

reckenro:

Fixes issue with session_resource_repository in relation to RFmx sessions.

- Our session repository assumes 1 driver session maps to 1 grpc_session (which is stored in our session repository)
- RFmx breaks this assumption. Specifically when a second initialize is called on the same device it will overwrite the previous key/value pair in the `reverse_map_` with the same handle but new session_id.
- Then when you try to close the sessions you get an error because it is accessing the session id's through the `reverse_map_` which isn't accurate.

Closes [AB#1792683](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1792683)

Currently, initializing two or more sessions on the same device in an RFmx service causes problems with session management.

System level test for RFmx NR passes (Two sessions initialized on same device -> Close -> Close succeeds on both calls).
Added unit test for `session_resource_repository` changes.

All other Unit, Integration, and System level tests pass locally (except a couple pre-existing System level test failures).

### Follow up changes
After BT and WLAN services go in, we'll need to apply similar updates to their config.py file, regenerate, and add a two session on one device test. These are captured in [AB#1794497](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1794497) and [AB#1794286](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1794286)


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -1161,7 +1161,7 @@ namespace nidaqmx_grpc {
     try {
       auto task_grpc_session = request->task();
       TaskHandle task = session_repository_->access_session(task_grpc_session.id(), task_grpc_session.name());
-      session_repository_->remove_session(task);
+      session_repository_->remove_session(task_grpc_session.id(), task_grpc_session.name());
```
reckenro, line 1164: Why the change?
`remove_session` wasn't working for the RFmx services. There is no correct mapping from RFmx handle to grpc session. This is because if two sessions are "initialized" on the same device it will be one RFmx handle but two grpc_device sessions.

Why is it safe?
In each case, we have the grpc_session as input so we can just use those in a similar way to `access_session` to figure out the session_id to clear out of the `session_repository` and `session_resource_repository`.

generated/nifake_non_ivi/nifake_non_ivi_service.cpp
```diff
@@ -75,7 +75,7 @@ namespace nifake_non_ivi_grpc {
       auto handle_grpc_session = request->handle();
       FakeHandle handle = session_repository_->access_session(handle_grpc_session.id(), handle_grpc_session.name());
 
-      auto initiating_session_id = session_repository_->resolve_session_id(handle);
+      auto initiating_session_id = session_repository_->access_session_id(handle_grpc_session.id(), handle_grpc_session.name());
```
reckenro, line 78: Why the change?
There can be a one to many relationship between (RFmx handle, grpc_session). Therefore, we can't reliably look up a session_id by the instrument_handle for RFmx services.

Why is it safe?
Similar to the `remove_session` comment, we have the `grpc_session` input from the client so we should just be able to use that to get the actual session_id stored in the session_repository/session_resource_repository.

source/tests/unit/session_resource_repository_tests.cpp
```diff
@@ -58,6 +58,32 @@ TEST(SessionResourceRepositoryTests, AddSessionResource_ResourceIsAdded)
       repository.access_session(session_id, ""));
 }
 
+TEST(SessionResourceRepositoryTests, AddTwoSessionsWithSameResourceHandle_RemoveOneSession_ResourceStillAccessible)
```
reckenro, line 61: This test used to fail on the last line where it expected the second close call's `close_response_two` to be good. It was getting an invalid session error though.

The flow through the test previously was
```c++
Init session one
    Session_Resource mappings
        map [{1, 0x01}]
        reverse_map [{0x01, 1}]
    Session repository mappings
        [{Id: 1, {1, }}]
        [Name: "SessionOne", {1, ...}]
Init session two
    Session_Resource mappings
        map [{1, 0x01}, {2, 0x01}]
        reverse_map [{0x01, 2}] // Overwrites pair with key 0x01 with new session id: 2
    Session repository mappings
        [{Id: 1, {1, }}, {Id: 2, {2, ...}}]
        [{Name: "SessionOne", {1, ...}}, {Name: "SessionTwo", {2, ...}}]
Close session one
    Session_Resource mappings
        map [{1, 0x01}] // Here, it gets session id of handle (0x01) which returns 2. So 2 key is removed.
        reverse_map [] // Empty, there was only one key/value pair in here.
    Session repository mappings
        [{Id: 1, {1, }}}]
        [{Name: "SessionOne", {1, ...}}] // Removes wrong one since it gets wrong one through Handle => session map
Close session two
     // Can't find session with id 2, it was removed.
     // So it goes through with a 0 session id and 0 for the session handle causing failure.
```

source/codegen/metadata/nirfmxspecan/config.py
```diff
@@ -58,4 +58,5 @@
     'metadata_version': '0.1',
     'module_name': 'nirfmxspecan',
     'session_class_description': 'An NI-RFmxSpecAn instrument handle',
+    'disallow_session_reverse_lookup': True
```
reckenro, line 61: This will cause generation to fail if there's any method that has an output session from the driver API that's not marked as a `init_method`. Essentially, this scenario would be broken for RFmx because there is no mapping currently between a session handle and grpc session since it can be one to many.

I added this to RFSG temporarily and confirmed it caused generation to fail for RFSG's `GetAttributeViSession` as expected.

source/codegen/metadata_validation.py
```diff
@@ -228,7 +231,11 @@ def validate_function(function_name: str, metadata: dict):
                     ivi_dance_with_a_twist_params.append(parameter['name'])
                     ivi_dance_with_a_twist_sizes.add(size['value'])
                     ivi_dance_with_a_twist_twists.add(size['value_twist'])
-
+                if disallow_session_reverse_lookup:
+                    if not is_init_method:
+                        if parameter['direction'] == 'out' and parameter['type'] == service_helpers.get_resource_handle_type(metadata['config']):
+                            raise  Exception(
+                               f"Since reverse session lookup is disallowed for this driver, {function_name} is not valid since it's not marked as an init_method but has a session output!")
```
astarche, line 238: I don't see anything in the code about "why" reverse lookup would be disallowed. Either from the perspective of someone adding a new driver (that might want to toggle that flag) or someone trying to understand a driver that disallows it (not sure why they would care, but could happen).

Some ideas:
* Comment in metadata.
* Update this message saying why reverse lookup is disallowed.
* Update the name of the flag to be something like "duplicate_resource_handles_allowed" which implies reverse lookup is not allowed.
* Comments in tests.

I think putting stuff in this file is probably the most useful. Even though the schema is not exactly purely declarative, it does specifically exist to explain what the metadata means. So it's a good place for more info.

astarche (approved):

mgumble (approved):

source/codegen/metadata_validation.py
reckenro: Good point. I went with your second and third idea. Change the flag to 'duplicate_resource_handles_allowed' and update the Exception message in validation to explain why  it failed.

For rfsg, the new message would look like
```
Exception: nirfsg allows duplicate resource handles in the session repository. Therefore, the handle we'd get for "value" can't be mapped back to a Session to provide to the client!
...
Exception: Failed to validate function GetAttributeViSession
```
which I think gives a good idea of what's going on.

PR #472, opened 2022-01-18
Add RFmxInstr enums and attributes

astarche:

Adds enums and attributes to the RFmxInstr service.

Improve usability of the RFmxInstr service by adding the enums and attributes defined in the C API.

Fixes [AB#1781494](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1781494).

Ran and passed all RFmx tests including new tests.

Some additional free form testing with RFmxInstr attributes.


source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -132,7 +143,30 @@ TEST_F(NiRFmxInstrDriverApiTests, NoActiveList_GetListNames_ReturnsEmptyLists)
...
+  initiate_to_enter_committed_state(session);
+  const auto response = client::get_attribute_i32(stub(), session, "", NiRFmxInstrAttribute::NIRFMXINSTR_ATTRIBUTE_TUNING_SPEED);
+
+  EXPECT_SUCCESS(session, response);
+  EXPECT_EQ(response.attr_val(), NiRFmxInstrInt32AttributeValues::NIRFMXINSTR_INT32_TUNING_SPEED_FAST);
+}
+
+TEST_F(NiRFmxInstrDriverApiTests, SetAndGetFrequencyReferenceSource_ReturnsFrequencyReferenceSource)
+{
+  constexpr auto MIXER_LEVEL = 5.5e7;
```
astarche, line 159: TODO Remove bogus constant

source/codegen/metadata/nirfmxinstr/enums.py
```diff
@@ -1,41 +1,244 @@
 enums = {
```
astarche, line 1: TODO sync with scrapigen PR.

source/codegen/metadata/nirfmxinstr/enums_addon.py
```diff
@@ -4,3 +4,6 @@
 enums_override_metadata = {
 }
 
+enums_validation_suppressions = {
+    "FrequencyReferenceExportedTerminal": ["ENUMS_SHOULD_NOT_HAVE_DUPLICATE_VALUES"]
```
reckenro, line 8: What is this bit about? I might be missing something but I don't see this 'enums_validation_suppressions' string referenced anywhere except the `__init__.py` and `enums_addon.py` for a few drivers.

reckenro (approved):

mgumble (approved):

source/codegen/metadata/nirfmxinstr/enums_addon.py
astarche: Yeah! This was tricky to track down.

1. There's a rule in metadata_validation to require no duplicate enum values.
2. There's a way to override that rule.
3. BUT how does it work!?

I'm guessing that the bit you missed was the same one that took me forever to find: [rule_is_suppressed](https://github.com/ni/grpc-device/blob/d6fddbe4f42eb9f4c36fd3d037385cf7dc1a6088/source/codegen/metadata_validation.py#L155).
 So basically the "enums" prefix means it applies to the `enums.py` metadata file

PR #474, opened 2022-01-19
Add function enums to RFmxInstr

astarche:

Add function enums to RFmxInstr.

Fixes [AB#1799016](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1799016). Fixes [AB#1781495](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1781495).

Improves usability of RFmxInstr service.

Ran and passed all RFmx tests.


generated/nirfmxinstr/nirfmxinstr.proto
```diff
@@ -211,6 +211,106 @@ enum NiRFmxInstrAttribute {
   NIRFMXINSTR_ATTRIBUTE_THERMAL_CORRECTION_TEMPERATURE_RESOLUTION = 120;
 }
 
+enum Boolean {
```
astarche, line 214: @harsha-bhushan this is the second change adding enums to grpc-device for RFmxInstr. You can see all of the enums and attributes now in the proto file.

Does this look OK? Is it what you expected? Any questions/comments/concerns?

source/codegen/metadata/nirfmxinstr/functions.py
```diff
@@ -1720,10 +1738,12 @@
             },
             {
                 'direction': 'out',
+                'enum': 'Boolean',
                 'name': 'selfCalibrateValid',
                 'type': 'int32'
             },
             {
+                'bitfield_as_enum_array': 'SelfCalStep',
```
reckenro, line 1746: Should we be marking the output parameters using the `Personality` enum with this tag? `Personality` also looks like a bitfield enum but I'm not sure if there's some nuances between the methods returning it vs. this one with `SelfCalStep`.

astarche: Good question. `Personality` is a bitfield, but most functions that accept it are just passing or returning a single value (or so it seems).

There is one exception: `personalityFilter`. That's definitely a bitfield input.

`bitfield_as_enum_array` currently only works for outputs. That seems surprising. But the reason for that is that using regular-old-enums for bitfield *inputs* actually works OK. In most client languages (i.e., python, C++) you can `or` together the enum values and pass them for the `_raw` field and you get a pretty reasonable experience. We've been doing that since the MI services. 

Trying to do that with outputs is ... less obvious. We'd have to always return the raw field and would never know how to populate the non-raw. And the client would have to do bitwise logic on the result with enum values that don't actually match the field. So, I added `bitfield_as_enum_array` the first time I hit a bitfield output.

I was still thinking about adding a task to use `bitfield_as_enum_array` for `personalityFilter`, which would mean adding support for `bitfield_as_enum_array` inputs. That should be a better experience with a possible exception for passing a single value (still not bad). But it is a little extra work, and we have already decided (multiple times) that regular enums were "good enough for this case.

source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -165,6 +167,48 @@ TEST_F(NiRFmxInstrDriverApiTests, SetAndGetFrequencyReferenceSource_ReturnsFrequ
...
+  EXPECT_SUCCESS(session, specan_client::cfg_rf(specan_stub, session, "", 1e9, 0, 0));
+  EXPECT_SUCCESS(session, client::cfg_frequency_reference(stub(), session, "", FrequencyReferenceSource::FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK, 10e6));
+  EXPECT_SUCCESS(session, specan_client::spectrum_cfg_span(specan_stub, session, "", 1e6));
+  EXPECT_SUCCESS(session, specan_client::spectrum_cfg_rbw_filter(specan_stub, session, "", true, 100e3, nirfmxspecan_grpc::SpectrumRbwFilterType::SPECTRUM_RBW_FILTER_TYPE_GAUSSIAN));
+  EXPECT_SUCCESS(session, specan_client::spectrum_cfg_averaging(specan_stub, session, "", false, 10, nirfmxspecan_grpc::SpectrumAveragingType::SPECTRUM_AVERAGING_TYPE_RMS));
+
+  auto read_response = specan_client::spectrum_read(specan_stub, session, "", 10.0);
+
+  EXPECT_SUCCESS(session, read_response);
+  int midpoint_x = 1005 / 2;
```
reckenro, line 207: Can you add a `EXPECT_EQ(1005, read_response.spectrum_size());` or equivalent above this line to make it clear 1005 is the size?

source/codegen/metadata/nirfmxinstr/functions.py
reckenro: Ah okay, that all makes sense. Looks good then.

And yeah, I think it's fine to keep it as is with the "regular enums are good enough for input bit fields" rationale. Good to keep in mind, though, if we do ever decide the improved clarity for clients filling out input fields is worth it.

reckenro (approved):

generated/nirfmxinstr/nirfmxinstr.proto
```diff
@@ -1001,7 +1148,10 @@ message SelectActiveExternalAttenuationTableResponse {
 message SelfCalibrateRequest {
   nidevice_grpc.Session instrument = 1;
   string selector_string = 2;
-  int32 steps_to_omit = 3;
+  oneof steps_to_omit_enum {
+    SelfCalStep steps_to_omit = 3;
+    int32 steps_to_omit_raw = 4;
```
mgumble, line 1153: Question: Do you have any background on why some self-cal steps are int 64 and some int 32? Based on the functions I see both flavor APIs

source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -165,6 +167,48 @@ TEST_F(NiRFmxInstrDriverApiTests, SetAndGetFrequencyReferenceSource_ReturnsFrequ
...
+  const auto session = init_session(stub(), PXI_5663E);
+
+  EXPECT_SUCCESS(session, client::cfg_frequency_reference(stub(), session, "", FrequencyReferenceSource::FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK, 10e6));
+  EXPECT_SUCCESS(session, client::cfg_mechanical_attenuation(stub(), session, "", MechanicalAttenuationAuto::MECHANICAL_ATTENUATION_AUTO_TRUE, 10.0));
+}
+
+TEST_F(NiRFmxInstrDriverApiTests, CallCheckMethods_SucceedsWithReasonableResponseValues)
+{
+  const auto session = init_session(stub(), PXI_5663E);
+
```
mgumble, line 181: not-pick remove this new line

mgumble (approved):

generated/nirfmxinstr/nirfmxinstr.proto
astarche: It's just RFSA/G that have 64-bit self cal steps, right? And RFmxInstr is 32-bit? I assume that it's because RFSA/G conservatively reserved a ton of bits for self cal steps and RFmx decided not to. Not sure why that was but `protobuf`-compatibility turns out to be a good example of a reason to stick with 32-bits for enum-like-encodings if you can get away with it: it represents enums as int32s.

generated/nirfmxinstr/nirfmxinstr.proto
```diff
@@ -858,7 +994,10 @@ message GetSelfCalibrateLastDateAndTimeResponse {
 message GetSelfCalibrateLastTemperatureRequest {
   nidevice_grpc.Session instrument = 1;
   string selector_string = 2;
-  int64 self_calibrate_step = 3;
+  oneof self_calibrate_step_enum {
+    SelfCalStep self_calibrate_step = 3;
+    int64 self_calibrate_step_raw = 4;
```
mgumble, line 999: @astarche Here is a int 64 self-cal step in RFInstru thats why I was curious

astarche: Yikes thanks. I don't know why that is. I created AB#1804296 to make sure we sign this off in API review. I'm going to see if I can find the G code that implements these methods and send Nagashree an email.

PR #476, opened 2022-01-19
Add missing tests for RFmxInstr

astarche:

Add tests for `CVIAbsoluteTime` and attributes with unknown (not known in metadata) types in RFmxInstr.

Fixes [AB#1801138](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1801138).

Completes missing rows in AcceptanceTable and ensures that those features/assumptions of the API work.

### DIFF
--- a/source/tests/system/nirfmxinstr_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxinstr_driver_api_tests.cpp
@@ -165,6 +165,28 @@ TEST_F(NiRFmxInstrDriverApiTests, SetAndGetFrequencyReferenceSource_ReturnsFrequ
   EXPECT_EQ(response.attr_val(), "OnboardClock");
 }
 
+TEST_F(NiRFmxInstrDriverApiTests, GetModelName_ReturnsModelName)
+{
+  const auto session = init_session(stub(), PXI_5663E);
+  const auto response = client::get_attribute_string(stub(), session, "", NiRFmxInstrAttribute::NIRFMXINSTR_ATTRIBUTE_INSTRUMENT_MODEL);
+
+  EXPECT_SUCCESS(session, response);
+  EXPECT_EQ(response.attr_val(), "NI PXIe-5663E");
+}
+
+TEST_F(NiRFmxInstrDriverApiTests, TimestampFromValuesRoundTrip_SucceedsWithOriginalValues)
+{
+  constexpr auto SECONDS_SINCE_1970 = 10000;
+  constexpr auto FRACTIONAL_SECONDS = .75;
+  const auto timestamp_response = client::timestamp_from_values(stub(), SECONDS_SINCE_1970, FRACTIONAL_SECONDS);
+  const auto values_response = client::values_from_timestamp(stub(), timestamp_response.timestamp());
+
+  ni::tests::system::EXPECT_SUCCESS(timestamp_response);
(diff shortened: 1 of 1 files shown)


reckenro (approved):

source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -165,6 +165,28 @@ TEST_F(NiRFmxInstrDriverApiTests, SetAndGetFrequencyReferenceSource_ReturnsFrequ
...
+  const auto session = init_session(stub(), PXI_5663E);
+  const auto response = client::get_attribute_string(stub(), session, "", NiRFmxInstrAttribute::NIRFMXINSTR_ATTRIBUTE_INSTRUMENT_MODEL);
+
+  EXPECT_SUCCESS(session, response);
+  EXPECT_EQ(response.attr_val(), "NI PXIe-5663E");
+}
+
+TEST_F(NiRFmxInstrDriverApiTests, TimestampFromValuesRoundTrip_SucceedsWithOriginalValues)
+{
+  constexpr auto SECONDS_SINCE_1970 = 10000;
```
mgumble, line 179: Super nit-picky but maybe just name this UNIX_TIMESTAMP, since technically this low of a UNIX code is still Jan 1st 1970

source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -165,6 +165,28 @@ TEST_F(NiRFmxInstrDriverApiTests, SetAndGetFrequencyReferenceSource_ReturnsFrequ
...
+  constexpr auto FRACTIONAL_SECONDS = .75;
+  const auto timestamp_response = client::timestamp_from_values(stub(), SECONDS_SINCE_1970, FRACTIONAL_SECONDS);
+  const auto values_response = client::values_from_timestamp(stub(), timestamp_response.timestamp());
```
mgumble, line 182: Question: I am assuming this is a simulation hardware where internal time doesn't change?

mgumble (approved):

source/tests/system/nirfmxinstr_driver_api_tests.cpp
astarche: That's a good question. I might be misunderstanding (chances are good since time is involved).

Neither of these APIs are meant to give hardware/current/os time as far as I know. In the C API, they're convenience methods for converting input values to/from `CVIAbsoluteTime`, which is a standard NI time representation. We convert `CVIAbsoluteTime` to [protobuf Timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp), so I'm trying to validate that that (double) conversion is working.

I'm honestly not sure why these utility function are in RFmxInstr, because I don't see APIs in the other personalities that are dealing with either time representation (maybe there's an attribute that I'm missing). But it's great for testing roundtrip conversion because we don't need to rely on any state. I wish we'd started with this API when adding it in the first place (Hi @gregstoll )

source/tests/system/nirfmxinstr_driver_api_tests.cpp
astarche: Hah, I guess that's more concise than `SECONDS_SINCE_MIDNIGHT_JANUARY_FIRST_1970`

PR #477, opened 2022-01-19
Sync up SpecAn metadata from scrapigen

astarche:

Sync up SpecAn with metadata from scrapigen.

Update CMakeLists.txt to ensure that register_all_service is regenerated on changes to any metadata file.

We have changed over to using only API json metadata for extraction. This process caused several superficial changes to the metadata files and a handful of functional ones (removing APIs that RFmx does not generally support). We need to get these to codebases in sync to apply those changes and allow building on top of them.

In making this change, I noticed that `register_all_services` was not regenerating on changes to SpecAn metadata. This is because it was only using the metadata from the last loop of codegening services.

Ran and passed all SpecAn tests.

Manual testing on codegen invalidation.


mgumble:
It would be nice to have a documented screen shot of passing System tests, because we do not have any automated results yet

astarche:
@mgumble Read em and weep

image

reckenro (approved):

PR #479, opened 2022-01-20
Add support for InitializeFromNIRFSASessionArray to RFmxInstr

astarche:

Add support for InitializeFromNIRFSASessionArray to RFmxInstr.
* Fixup grpc-device codegen to support arrays of cross-driver sessions as inputs.

Fixes [AB#1804470](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1804470).

Improve support for cross-driver sessions in RFmx.

Unit and system tests added for cross-driver session arrays.


source/codegen/service_helpers.py
```diff
@@ -62,6 +62,25 @@ def create_args_for_callback(parameters):
     ])
 
 
+def _is_array_requiring_conversion(parameter):
```
astarche, line 65: Note I pulled out this helper because the control flow in `create_standard_arg` was getting totally nuts and we're losing track of what the point of the special cases are. So this collects several such cases into one and gives it a name.

The functional change here is using grpc_type to match `repeated nidevice_grpc.Session` instead of the pre-existing code that only matched `ViSession[]` for ivi drivers.

Testing done: all other drivers output the same code.

astarche:
image

source/codegen/service_helpers.py
reckenro: Thanks for the cleanup! This is much easier to read through.

source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -125,13 +130,26 @@ TEST_F(NiRFmxInstrDriverApiTests, GetNIRFSASession_SelfTest_Succeeds)
...
-  client::self_calibrate(stub(), session, "", 0);
+  EXPECT_SUCCESS(session, client::self_calibrate(stub(), session, "", 0));
+}
+
+TEST_F(NiRFmxInstrDriverApiTests, InitializeFromNIRFSAArray_SelfCalibrate_Succeeds)
+{
+  auto rfsa_stub = create_stub<nirfsa_grpc::NiRFSA>();
+  auto init_rfsa_response = init_rfsa(rfsa_stub, "Sim");
+  auto init_rfsa_response_2 = init_rfsa(rfsa_stub, "Sim2");
+  ni::tests::system::EXPECT_SUCCESS(init_rfsa_response);
```
reckenro, line 147: Minor: Should we go ahead and add a matching `EXPECT_SUCCESS` for the second init rsponse?

reckenro (approved):

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -1508,6 +1514,36 @@ TEST_F(NiFakeNonIviServiceTests, InitFromCrossDriverSession)
...
+TEST_F(NiFakeNonIviServiceTests, InitFromCrossDriverSessionArray_PassesSessionResourcesAndSucceeds)
+{
+  constexpr auto CROSS_DRIVER_HANDLES = std::array<FakeCrossDriverHandle, 2>{1234, 5678};
+  const auto CROSS_DRIVER_SESSION_NAMES = std::array<std::string, 2>{"cross_driver_session", "cross_driver_session_2"};
+  add_session(cross_driver_resource_repository_, CROSS_DRIVER_HANDLES[0], CROSS_DRIVER_SESSION_NAMES[0]);
+  add_session(cross_driver_resource_repository_, CROSS_DRIVER_HANDLES[1], CROSS_DRIVER_SESSION_NAMES[1]);
+  EXPECT_CALL(library_, InitFromCrossDriverSessionArray(_, _, _))
+      .With(Args<0, 1>(ElementsAreArray(CROSS_DRIVER_HANDLES.data(), CROSS_DRIVER_HANDLES.size())))
+      .WillOnce(DoAll(
+          SetArgPointee<2, FakeHandle>(4567), Return(kDriverSuccess)));
```
mgumble, line 1538: Question: Why 4567?

mgumble (approved):

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
astarche: Can you check out [this commit](https://github.com/ni/grpc-device/pull/479/commits/41ba1b8975b062f4d33f7f114a4b81233358407f) and see if it helps explain it? 

This is the internal driver session (not the cross-driver session). For this test, it just needs to be a non-zero integer so that there's something to add to the map. If we were to call other methods with the grpc session returned by init, it would resolve to that session and pass it to the driver C API (or mock library in this case). 

For this test, we're more concerned that the right stuff gets passed into the init method (the EXPECT above) and what happens after that is covered more in other tests. But ... I agree its confusing to have a magic number, so I tried adding some asserts/constants that explain it.

source/codegen/service_helpers.py
```diff
@@ -70,22 +89,19 @@ def create_standard_arg(parameter):
         type_without_brackets = common_helpers.get_underlying_type_name(
             parameter['type'])
         return f'({type_without_brackets}*){parameter_name}.data(), '
-    elif parameter['type'] in {"ViBoolean[]", "ViSession[]", "ViInt16[]"}:
+    elif _is_array_requiring_conversion(parameter):
```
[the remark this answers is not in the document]
astarche: We don't know that it's an array yet though.

What about: `_is_array_that_requires_conversion`?

Could also be:

`elif is_array and _requires_conversion(parameter)`

Although my helper is probably not as comprehensive a list of all types that require conversion, so that could be misleading.

PR #485, opened 2022-01-24
Add support for GetNIRFSASessionArray to RFmxInstr

astarche:

Add support for GetNIRFSASessionArray to RFmxInstr.

Fixes AB#1781503.

Enables accessing multiple RFSA sessions from a grpc-device RFmx session and giving those sessions a name.

### DIFF
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -609,6 +609,18 @@ def is_init_method(function_data):
     return function_data.get('init_method', False)
 
 
+def _get_session_output_param(function_data):
+    return next(
+        p 
+        for p in function_data["parameters"] 
+        if p["direction"] == "out" and "nidevice_grpc.Session" in p["grpc_type"]
+    )
+
+
+def is_init_array_method(function_data):
+    return is_init_method(function_data) and "repeated" in _get_session_output_param(function_data)["grpc_type"]
+
+
 def is_cross_driver_init_method(function_data: dict) -> bool:
     return (
         is_init_method(function_data) 
--- a/source/codegen/metadata/nirfmxinstr/functions.py
+++ b/source/codegen/metadata/nirfmxinstr/functions.py
@@ -1500,6 +1500,8 @@ functions = {
         'returns': 'int32'
     },
     'GetNIRFSASessionArray': {
+        'codegen_method': 'CustomCode',
+        'init_method': True,
         'parameters': [
             {
                 'direction': 'in',
@@ -1508,7 +1510,9 @@ functions = {
                 'type': 'niRFmxInstrHandle'
             },
             {
+                'cross_driver_session': 'ViSession',
                 'direction': 'out',
+                'grpc_type': 'repeated nidevice_grpc.Session',
                 'name': 'nirfsaSessions',
                 'size': {
                     'mechanism': 'ivi-dance-with-a-twist',
--- a/source/custom/nirfmxinstr_service.custom.cpp
+++ b/source/custom/nirfmxinstr_service.custom.cpp
@@ -1,2 +1,65 @@
+#include <nirfmxinstr/nirfmxinstr_service.h>
+#include <server/converters.h>
+
+#include <sstream>
+
 namespace nirfmxinstr_grpc {
+const auto kErrorReadBufferTooSmall = -200229;
+const auto kWarningCAPIStringTruncatedToFitBuffer = 200026;
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+::grpc::Status NiRFmxInstrService::GetNIRFSASessionArray(::grpc::ServerContext* context, const GetNIRFSASessionArrayRequest* request, GetNIRFSASessionArrayResponse* response)
+{
+  if (context->IsCancelled()) {
+    return ::grpc::Status::CANCELLED;
+  }
+  try {
+    auto instrument_grpc_session = request->instrument();
+    auto instrument = session_repository_->access_session(instrument_grpc_session.id(), instrument_grpc_session.name());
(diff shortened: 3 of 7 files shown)


source/custom/nirfmxinstr_service.custom.cpp
```diff
@@ -1,2 +1,65 @@
...
+
+#include <sstream>
+
 namespace nirfmxinstr_grpc {
+const auto kErrorReadBufferTooSmall = -200229;
+const auto kWarningCAPIStringTruncatedToFitBuffer = 200026;
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+::grpc::Status NiRFmxInstrService::GetNIRFSASessionArray(::grpc::ServerContext* context, const GetNIRFSASessionArrayRequest* request, GetNIRFSASessionArrayResponse* response)
```
astarche, line 12: Note: I use `CustomCode` pretty rarely. This would be a pretty gnarly bit of codegen:
* The only `init_method` that uses an array size mechanism (`ivi-dance-with-a-twist`).
* Error checking for multiple session names input.
* The only method that doesn't actually call a driver API in it's `init_lambda`.

Adding complex codegen can be a real problem for maintainability.

I searched around and I'm pretty sure that this is the only method in this stile (at least in RFmx).

There are a couple of other reasons to prefer codegen to custom code:
* Better testability (implement in fake service).
* (theoretically) easier to keep up-to-date with other codegen changes and features.

We may decide to codegen this eventually (that's what we did for callbacks in DAQ), or we may leave it as-is (quite a few MI methods still use custom_code and have been fine).

How does this work?

The codgenerator generates "everything exept the service implementation": proto file, library wrapper files, client files. Then the service implementation is added in this cpp file.

source/tests/system/nirfmxinstr_driver_api_tests.cpp
```diff
@@ -102,6 +102,43 @@ nidevice_grpc::Session init_session(const client::StubPtr& stub, const std::stri
...
+::grpc::Status get_rfsa_session_array(const client::StubPtr& stub, const nidevice_grpc::Session& session, const std::vector<std::string>& session_names, GetNIRFSASessionArrayResponse& response)
+{
+  auto request = GetNIRFSASessionArrayRequest{};
+
+  for (const auto& name : session_names) {
+    request.add_session_names(name);
+  }
+  request.mutable_instrument()->CopyFrom(session);
+  ::grpc::ClientContext context;
+  // Note: this can't use the experimental::client APIs, because those omit the session name
```
astarche, line 114: I considered adding session names to the client API. This is actually pretty easy to add, but does require that all init methods use it and all calls pass in something (empty string is fine). I decided against it though. That API is deliberately streamlined to make it easy to test common use cases. Adding session names wouldn't be a huge deal, but it does work against that.

astarche:
image

source/codegen/metadata/nirfmxinstr/functions.py
```diff
@@ -1500,6 +1500,8 @@
         'returns': 'int32'
     },
     'GetNIRFSASessionArray': {
+        'codegen_method': 'CustomCode',
```
reckenro, line 1503: Will this go into the metadata extraction repo after these changes go in here? Has that been the process for keeping the versions of the metadata in sync? Get manual updates in in grpc-device and then in scrapigen?

source/codegen/common_helpers.py
```diff
@@ -609,6 +609,18 @@ def is_init_method(function_data):
     return function_data.get('init_method', False)
 
 
+def _get_session_output_param(function_data):
```
reckenro, line 612: Do we want to keep the changes in source/codegen since we went with the custom implementation and there aren't any generated changes? I think the methods and updates look fine, was just wondering if we still want/need them.

source/codegen/metadata/nirfmxinstr/functions.py
astarche: Yes, that will need to go in. I'll leave this comment open until this is up.

reckenro (approved):

source/codegen/metadata/nirfmxinstr/functions.py
astarche: Oh, you asked more questions, sorry.

> Has that been the process for keeping the versions of the metadata in sync? 

I try to do them "about the same time". When the codgen is new/complicated, the grpc-device part tends to go first (partly to vet the PR). When the change is mostly extraction, sometimes scrapigen goes first. We'll need to be more disciplined about keeping them in sync with more people working on it.

@DavidCurtiss has a script to sync them for a local dev workflow. We can consider making that a workflow on scrapigen if we can figure out how to make it work. I guess from a workflow we can clone another repo, copy over rfmx metadata, and then check for diffs. We could make that non-blocking, but we'd want to trigger on changes either direction (and we'd expect it to be broken fairly frequently; but hopefully settle to "green" pretty quickly too,.

DavidCurtiss: I do have a simple python script for copying all of the rfmx dirs (and I can share that if someone would find it useful), but mostly I just use my NR-specific .bat file, since that's all I'm normally editing:
```
@mkdir ..\grpc-device\source\codegen\metadata\nirfmxnr\ 2>NUL
cp -u out/rfmxnr/export/* ../grpc-device/source/codegen/metadata/nirfmxnr/
```
P.S. I also have .bat files for "building" (generating) and testing, so I don't have to think about the commands. More useful, though, are my slightly-more-complicated .bat files for grpc-device, which I've set up to run my ninja-based builds.

mgumble (approved):

PR #487, opened 2022-01-25
Exclude cross-driver sessions from reverse lookup

astarche:

Exclude cross-driver sessions from reverse lookup.

Fixes [AB#1807465](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1807465)

If cross-driver sessions can be added to the reverse lookup, it can introduce ambiguity about which session should be returned from reverse lookup methods if there are multiple sessions with the same resource handle. If we exclude cross-driver sessions from reverse lookup, only primary sessions can be returned.

Methods that reverse lookup sessions are very rare. In RFSA, for example, there is a `GetAttributeViSession` method, but no actual public `ViSession` attributes. We could attempt more complicated features that make it possible to access cross-driver sessions under some conditions. But this provides little value. By restricting reverse lookups completely, we have more options to change/add behavior in the future.

Ran and passed all RFmx and RFSA test.

### DIFF
--- a/source/server/session_repository.cpp
+++ b/source/server/session_repository.cpp
@@ -8,6 +8,16 @@ SessionRepository::SessionRepository()
 {
 }
 
+SessionRepository::~SessionRepository()
+{
+  // Do not cleanup external resources during shutdown. At this stage, all of the
+  // driver Library objects have been cleaned up so it's not safe to call them.
+  // To fix that, we'd have to change up our resource management. But right now we
+  // don't have a structured cleanup process and will let the driver figure out how
+  // to handle ctrl+c shutdown.
+  reset_server(/* cleanup_external_resources =*/false);
+}
+
 // Either returns (via session_id) an existing session with the specified name or initializes and
 // adds a new session using the init_func provided. The init_func is only called when an existing
 // session with session_name is not found. It is expected to return a tuple, where the first value
@@ -220,15 +230,15 @@ void SessionRepository::clear_reservations()
   }
 }
 
-bool SessionRepository::reset_server()
+bool SessionRepository::reset_server(bool cleanup_external_resources)
 {
   std::unique_lock<std::shared_mutex> lock(repository_lock_);
   clear_reservations();
-  bool is_server_reset = close_sessions();
+  bool is_server_reset = close_sessions(cleanup_external_resources);
   return is_server_reset && reservations_.empty();
 }
 
-bool SessionRepository::close_sessions()
+bool SessionRepository::close_sessions(bool cleanup_external_resources)
 {
   named_sessions_.clear();
   // Copy sessions for cleanup to avoid invalidating iterators when dependent sessions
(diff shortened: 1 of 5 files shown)


source/server/session_resource_repository.h
```diff
@@ -74,15 +76,17 @@ class SessionResourceRepository {
   };
 
   SessionRepository* session_repository_;
-  ResourceMap resource_map_;
+  // shared_ptr to allow sharing ownership with cleanup delegate in register_dependent_session.
```
astarche, line 79: Note: my test for this change uncovered a bug in the original design. The dependent sessions can get removed after resource_repository is cleaned up (the session_repository lives longer). Making this a shared_ptr lets us pass off shared ownership to the cleanup delegate so this can still be cleaned up later.

astarche:
image

source/server/session_resource_repository.h
astarche: Actually, this is a mess. I'm planning to stick with the general idea of excluding cross-driver sessions from the revere lookup. But the notifications/synchronization for cross-driver sessions in general are in worse shape than I thought.

source/server/session_repository.cpp
```diff
@@ -8,6 +8,16 @@ SessionRepository::SessionRepository()
...
 }
 
+SessionRepository::~SessionRepository()
+{
+  // Do not cleanup external resources during shutdown. At this stage, all of the
+  // driver Library objects have been cleaned up so it's not safe to call them.
+  // To fix that, we'd have to change up our resource management. But right now we
+  // don't have a structured cleanup process and will let the driver figure out how
+  // to handle ctrl+c shutdown.
+  reset_server(/* cleanup_external_resources =*/false);
```
astarche, line 18: This fixes the second mess. Trying to clean up the data structures in the destructor with default ordering was not working well and my tests exposed that (introduced by earlier cross-driver sessions change). `reset_server` can handle this ordering, so I switched to that. That needed one additional change of suppressing the external cleanup (calling back into the driver API, because those have already been unloaded).

source/server/session_resource_repository.h
astarche: Update. I'm OK with the state as of the last couple commits. So, please review as normal.

reckenro (approved):

source/server/session_resource_repository.h
```diff
@@ -229,8 +237,8 @@ TResourceHandle SessionResourceRepository<TResourceHandle>::ResourceMap::remove_
 {
   MapLock lock(map_mutex_);
   auto handle = get_handle(session_id);
+  map_.erase(session_id);
```
mgumble, line 240: Not sure I understand this change.

mgumble (approved):

source/server/session_resource_repository.h
astarche: Yeah, I had that confused. I thought that `get_handle` used the reverse map. And we needed to erase the session even if that mapping wasn't present. But `get_handle` uses the "forward map" so it should still always be there. Anyway it's not impacted by this change.

Also a good reminder I should add a test for removing dependent sessions to check this code. Or at least double check that it'd getting covered.

PR #488, opened 2022-01-26
Add some session and cross-driver session test coverage for RFmx NR.

reckenro:
### What does this Pull Request accomplish?

- Fixes NR metadata to denote `InitializeFromNIRFSASession`'s `nirfsaSession` parameter as a cross driver session param.
- Adds test coverage for the one cross-driver init function and adds an additional session test.

### Why should this Pull Request be merged?

We want to be confident in the session management and cross-driver session handling for RFmx NR.
Fixes [AB#1794284](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1794284)

### What testing has been done?

NR system-level tests (including new ones) pass.

### Follow-up work
- Get updated metadata generating in scrapigen
  - I'll update BT and WLAN at the same time
  - Done with [this PR](https://github.com/ni/grpc-device-scrapigen/pull/28) in scrapigen
- Make similar test additions for BT and WLAN after their cross-driver init methods are in grpc-device
  - Will be handled with [AB#1794516](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1794516) and [AB#1814907](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1814907)

PR #489, opened 2022-01-26
Add Mose to specific code owner patterns.

reckenro:

Adds Mose to specific code owner patterns so he's picked up as a code owner always.

We intended Mose to be a code owner for the repo since he's reviewing everything.

None.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,13 +2,13 @@
 * @astarche @mgumble
 
 # Owners for the build files
-/CMakeLists.txt      @SparkingSpork @astarche
-/.github/**/*.yml    @SparkingSpork @astarche
-/scripts/            @SparkingSpork @astarche
+/CMakeLists.txt      @SparkingSpork @astarche @mgumble
+/.github/**/*.yml    @SparkingSpork @astarche @mgumble
+/scripts/            @SparkingSpork @astarche @mgumble
 
 # Codegen
-/generated           @reckenro @astarche
-/source/codegen/     @reckenro @astarche
+/generated           @reckenro @astarche @mgumble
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -2,13 +2,13 @@
 * @astarche @mgumble
 
 # Owners for the build files
-/CMakeLists.txt      @SparkingSpork @astarche
-/.github/**/*.yml    @SparkingSpork @astarche
-/scripts/            @SparkingSpork @astarche
+/CMakeLists.txt      @SparkingSpork @astarche @mgumble
```
astarche, line 5: Ah my bad. So these are like overrides?

astarche (approved):

CODEOWNERS
```diff
@@ -2,13 +2,13 @@
 * @astarche @mgumble
```
reckenro, line 2: According to the GitHub docs about CodeOwners I think this works as a catch all. So if there was a file that didn't match the below patterns, Mose would be added. But if all the files in a PR matched the below patterns previously Mose wouldn't be as a code owner on the PR.

PR where all below patterns match and Mose was added manually (not marked code owner): #488 
vs.
This PR where the CODEOWNERS file doesn't match any of the specific patterns below and thus goes to this one (lowest matching pattern gets precedence) and Mose was added as an owner.

These changes should make Mose an owner on all PRs.

CODEOWNERS
reckenro: Yeah, the last pattern in CODEOWNERS that a given file matches adds the owners listed for that pattern.

mgumble (approved):

Issue #465, opened 2022-01-11
Misnamed parameters in RFmxSpecAn and RFmxInstr


astarche:
`duts_parameters_frequency` -> `dut_sparameters_frequency `or `dut_s_parameters_frequency`

`s_parameter_*` -> CONSIDER `sparameter_*` (matches RFSA/G, but both are reasonable).

`s2_p_file_path` -> `s2p_file_path`

`seconds_since1970` -> CONSIDER `seconds_since_1970`

`NIRFMXSPECAN_INT32_IMIF_OUTPUT_POWER_OFFSET_AUTO_FALSE` -> `NIRFMXSPECAN_INT32_IM_IF_OUTPUT_POWER_OFFSET_AUTO_FALSE` 

`NIRFMXSPECAN_INT32_NFY_FACTOR_MODE_MEASURE` -> `NIRFMXSPECAN_INT32_NF_Y_FACTOR_MODE_MEASURE` 

`pre_cfrpapr` -> `pre_cfr_papr`

`PERSONALITY_CDMA2_K` -> `PERSONALITY_CDMA2K`

[AB#1790788](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1790788)

PR #490, opened 2022-01-27
Add WLAN generated metadata

danestull:

Add initial metadata to grpc-device to add most functionality from the driver

This allows the WLAN RFmx personality to be more concrete with actual functions.

Build completes successfully and tests pass. Review of Proto file shows expected additional content.

Screen Shot 2022-01-27 at 2 50 14 PM


generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+  OFDM_MOD_ACC_EVM_UNIT_UNSPECIFIED = 0;
+  OFDM_MOD_ACC_EVM_UNIT_OFDMMODACC_EVM_UNIT_PERCENTAGE = 0;
+  OFDM_MOD_ACC_EVM_UNIT_OFDMMODACC_EVM_UNIT_DB = 1;
+}
+
+enum OfdmModAccFrequencyErrorEstimationMethod {
+  option allow_alias = true;
+  OFDM_MOD_ACC_FREQUENCY_ERROR_ESTIMATION_METHOD_UNSPECIFIED = 0;
+  OFDM_MOD_ACC_FREQUENCY_ERROR_ESTIMATION_METHOD_OFDMMODACC_FREQUENCY_ERROR_ESTIMATION_METHOD_DISABLED = 0;
+  OFDM_MOD_ACC_FREQUENCY_ERROR_ESTIMATION_METHOD_OFDMMODACC_FREQUENCY_ERROR_ESTIMATION_METHOD_INITIAL_PREAMBLE = 1;
```
astarche: Note: not part of this change, but this is a good example of what we're up against with these acronyms.

`OFDM_MOD_ACC_FREQUENCY_ERROR_ESTIMATION_METHOD_OFDMMODACC_FREQUENCY_ERROR_ESTIMATION` Would be a redundant-or-annoying enum prefix. We need to strip off the prefix in enums.py, because a redundant one will be added by the grpc-device codegen. Those can be addressed in the API review story, although it's more efficient to do it before writing examples and tests. That logic is in scrapigen.


`OFDMMODACC_` vs `OFDM_MOD_ACC_` is something that we have to keep an eye on and maybe make changes for. The .NET API has good enum names (i.e.,  `OfdmModAccFrequencyErrorEstimationMethod`) but they chose to tokenize some of those acronyms differently from the enum values in the C API (i.e., `RFMXWLAN_ATTR_OFDMMODACC_RESULTS_POWER_PPDU_PEAK_POWER_MAXIMUM`). The `grpc-device` enum value prefixer uses the enum-name (`OfdmModAccFrequencyErrorEstimationMethod`) as the input and converts it to UPPER_SNAKE, which introduces the extra `_`s.

generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_ACQUISITION_LENGTH_MODE_DSSSMODACC_ACQUISITION_LENGTH_MODE_AUTO = 1;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_AVERAGING_ENABLED_FALSE = 0;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_AVERAGING_ENABLED_TRUE = 1;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_BURST_START_DETECTION_ENABLED_FALSE = 0;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_BURST_START_DETECTION_ENABLED_TRUE = 1;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_CHIP_CLOCK_ERROR_CORRECTION_ENABLED_FALSE = 0;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_CHIP_CLOCK_ERROR_CORRECTION_ENABLED_TRUE = 1;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_DATA_DECODING_ENABLED_FALSE = 0;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_DATA_DECODING_ENABLED_TRUE = 1;
+  NIRFMXWLAN_INT32_DSSS_MOD_ACC_DATA_MODULATION_FORMAT_DSSSMODACC_DATA_MODULATION_FORMAT_DSSS1_MBPS = 0;
```
astarche: Note: not part of this change (should be added as a task either on this story or the api review).

These are iffy. 

I.e., 

RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_DSSS1MBPS

`NIRFMXWLAN_INT32_DSSS_MOD_ACC_DATA_MODULATION_FORMAT_DSSSMODACC_DATA_MODULATION_FORMAT_DSSS1_MBPS `

I'm concerned that the C API does use `_`s in some of these, but not the same ones we have.

I'm not sure where we're introducing the extra `_`s relative to the C API. If it's in RFmx metadata it's OK. If it's in one of our identifier conversion methods then that's not great.

Here's the full block of defines from the C API:

```c
// Values for RFMXWLAN_ATTR_DSSSMODACC_RESULTS_DATA_MODULATION_FORMAT
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_DSSS1MBPS                                    0
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_DSSS2MBPS                                    1
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_CCK5_5MBPS                                   2
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_CCK11MBPS                                    3
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_PBCC5_5MBPS                                  4
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_PBCC11MBPS                                   5
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_PBCC22MBPS                                   6
#define RFMXWLAN_VAL_DSSSMODACC_DATA_MODULATION_FORMAT_PBCC33MBPS                                   7
```

generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+message DSSSModAccFetchEVMRequest {
+  nidevice_grpc.Session instrument = 1;
+  string selector_string = 2;
+  double timeout = 3;
+}
+
+message DSSSModAccFetchEVMResponse {
+  int32 status = 1;
+  double rmsevm_mean = 2;
+  double peak_evm80211_2016_maximum = 3;
```
astarche: Note: not part of this change.

We should track these as CONSIDER changing to `peak_evm_80211_2016_maximum`. The default doesn't tokenize for numbers (some works as their own token and some don't).

generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+  nidevice_grpc.Session instrument = 1;
+  string selector_string = 2;
+  double timeout = 3;
+}
+
+message OFDMModAccFetchChainDataRMSEVMPerSymbolMeanTraceResponse {
+  int32 status = 1;
+  double x0 = 2;
+  double dx = 3;
+  repeated float chain_data_rmsevm_per_symbol_mean = 4;
```
astarche: Note: not part of this change.

`rmsevm` -> `rms_evm`.

Here's an example bug/task on how to track these: https://github.com/ni/grpc-device/issues/465

You don't have to make it a GH issue, it can be an azdo task. But you can use the same laundry list approach.

generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+
+message OFDMModAccFetchFrequencyErrorCCDF10PercentRequest {
+  nidevice_grpc.Session instrument = 1;
+  string selector_string = 2;
+  double timeout = 3;
+}
+
+message OFDMModAccFetchFrequencyErrorCCDF10PercentResponse {
+  int32 status = 1;
+  double frequency_error_ccdf10_percent = 2;
```
astarche: Makes more sense as `ccdf_10_percent`

generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+
+message OFDMModAccFetchPreambleAveragePowers80211acRequest {
+  nidevice_grpc.Session instrument = 1;
+  string selector_string = 2;
+  double timeout = 3;
+}
+
+message OFDMModAccFetchPreambleAveragePowers80211acResponse {
+  int32 status = 1;
+  double vhtsiga_average_power_mean = 2;
```
astarche: there are several of these that I think need to be split. `vht_siga`, `he_sigb`, etc.

generated/nirfmxwlan/nirfmxwlan.proto
```diff
@@ -31,27 +590,3421 @@ enum Boolean {
...
+
+message OFDMModAccFetchSIGCRCStatusRequest {
+  nidevice_grpc.Session instrument = 1;
+  string selector_string = 2;
+  double timeout = 3;
+}
+
+message OFDMModAccFetchSIGCRCStatusResponse {
+  int32 status = 1;
+  OfdmModAccSigCrcStatus sigcrc_status = 2;
```
astarche: `sig_crc`

astarche (approved):
Looks good. I just recorded all the weird names while going through the proto, but none of that needs to be part of this change.

mgumble (approved):

generated/nirfmxwlan/nirfmxwlan.proto
reckenro: Just to tack on to this. I think we'll need to consider this same scenario for some of the DSSSMODACC defines/enum values as well .

reckenro (approved):

generated/nirfmxwlan/nirfmxwlan.proto
astarche: @DavidCurtiss and I discussed this problem (and the related one for param names) and he's started looking at scripts for fixing this stuff.

We are thinking that we want to use `DSSS_MOD_ACC` and equivalent everywhere instead of `DSSSMODACC`. This will be more consistent with enum names and function names. And also just-plain-better. We'll have to see how that works in the implementation. Some stuff will naturally work that way (like the grpc-device code that adds enum value prefixes) and some might need to be corrected (attribute names?) but we'll have some options on how to do that. For example there are some PascalCaseNames in the related metadata: 
image

And we'll be building up a list of back-to-back acronym tokens either way.

danestull:
FYI @astarche I've captured fixes/confirming enums to the beautify story in this task #1816468
https://dev.azure.com/ni/DevCentral/_workitems/edit/1816468

PR #493, opened 2022-01-28
Remove UNSPECIFIED aliases for enums with natural zero values

### LINKED ISSUE #444: Enum aliases cause problems in json replay style use cases

astarche:

Remove UNSPECIFIED aliases for enums with natural zero values

Fixes #444.

Enum aliases make it impossible to map protobuf data back to one-or-the-other specific alias. This specifically causes problems and/or confusion in replay/capture and other grpc-json use cases.

Normally the `UNSPECIFIED` zero-value enum is used as a convention to ensure that the default value of an enum is defined to be invalid/meaningless. This isn't effective when it is aliased to a meaningful zero value from the C API. Also, we use enums along with `oneof`s which are a different way of ensuring that a value is set. Ultimately, avoiding aliases is more useful than injecting a conventional `UNSPECIFIED` value that doesn't serve it's intended purpose.

### Alternatives Considered

**Use mapped enum values to avoid valid zero values.** The reason we have zero valued enums is because we're using the enum values from the C API, some of which have zero values. Mapped enums provide a way to avoid this. ***Reason Rejected***: This would be massively API breaking for existing drivers and also deviate from the C API. We can consider offering a "mapped" alternative if there is some case where it is especially useful (consider attribute values, which have a lot of overlap), but we should stay-the-course of supporting an API that uses C enum values directly and that solution is sufficient for now.

### Breaking change

This change **Does not** break binary compatibility.

It will break "source compatibility" for existing clients that intentionally use the `UNSPECIFIED` enum value for enums that have a value default value. Those clients will need to make some sort of edit after re-running protoc (Very rare. Either a bug or a test.).

Ran all RF, Scope, Switch, DCPower, and DAQmx tests. One pre-existing failure.


source/codegen/metadata/niscope/CHANGES.md
```diff
@@ -208,4 +208,9 @@ The following functions were tagged with `'init_method': True` to ensure their g
 The following parameters were marked as `'include_in_proto': False` because they are calculated in the service handler and shouldn't be passed in.
 - `bufferSize` in `GetNormalizationCoefficients`
 - `bufferSize` in `GetScalingCoefficients`
-- `measurementWaveformSize` in `FetchArrayMeasurement`
\ No newline at end of file
+- `measurementWaveformSize` in `FetchArrayMeasurement`
+
+## enums.py
+
+Move all zero enums to the beginning of the list. proto requires that the first enum entry be zero. Moving only the zeor
```
astarche, line 215: TODO spelling

reckenro (commented):
Found a few non-attribute enums that still had the `allow_alias` and UNSPECIFIED value due to the other 0 value not being directly after it:
- BT
  - AcpResultsMeasurementStatus
  - PacketType
- Instr
  - Boolean
  - LinearInterpolationFormat
  - Personality
  - SelfCalStep
- WLAN
  - OfdmLtfSize
  - OfdmModAccLSigParityCheckStatus 
  - OfdmModAccSigBCrcStatus 
  - OfdmModAccSigCrcStatus 
- RFSG
  - RoutedSignal

source/codegen/proto_helpers.py
```diff
@@ -5,7 +5,8 @@ def should_allow_alias(enum):
   if enum.get("generate-mappings", False):
     return False
   enum_values = [e["value"] for e in enum["values"]]
-  return (0 in enum_values) or (len(enum_values) != len(set(enum_values)))
+  # aliases are allowed for zero-values if they are not first in the list.
+  return 0 in enum_values[1:] or len(enum_values) != len(set(enum_values))
```
reckenro, line 9: Do you think we should add metadata validation to make sure we don't get into the first condition with a regular (non-attribute) enum? I'm assuming for all regular integer enums in the metadata, we always expect the 0 value (if it's present) to be the first value.

If so, I played with
```python
if type(values[0]) is int:
                if 0 in values and values[0] != 0:
                    raise Exception(f"0 value in integer enum needs to be the first value!")
```
at the end of `validate_enum` in metadata_validation.py and it seemed like it worked.

astarche: Yeah, I think that's a good idea. 

But looking harder at this it seems like we should be able to remove the `0 in ...` check. We're expecting that all aliases are now true aliases in the input value set. Not aliases introduced by adding `UNSPECIFIED` (since we explicitly don't want to do that).

mgumble (approved):

source/codegen/proto_helpers.py
astarche: I'm now handling this in metadata_mutation. Also, if this ever sneaks in for an enum that is not attribute values, there will be an error from protoc. So I think this is covered now.

I also reverted the hand-edits.

source/codegen/metadata_mutation.py
```diff
@@ -285,3 +285,18 @@ def add_enum(enum_name, enum_values, enums, enum_value_prefix, is_mapped=False):
         'values': values
     }
     enums.update({enum_name: new_enum})
+
+
+def move_zero_enums_to_front(enums: dict) -> None:
+    """
+    protobuf requires that the first enum value be zero. For enums that have to zero value,
```
reckenro, line 292:
```suggestion
    protobuf requires that the first enum value be zero. For enums missing a zero value,
```

reckenro (approved):

PR #498, opened 2022-02-01
Add AcpBasic system-level test for NI-RFmx Bluetooth

tdunkle:

This pull request adds one of two system-level tests for basic data-fetching operations in the gRPC API for NI-RFmx Bluetooth. This is the first of two pull requests that will fix [AB#1784736](https://ni.visualstudio.com/DevCentral/_workitems/edit/1784736).

We want some basic system-level tests that confirm that fetching data with the gRPC API for NI-RFmx Bluetooth returns reasonable-looking results.

grpc-device locally builds successfully with the new test, and the new test passes on a VM with NI-RFSA and NI-RFmx Bluetooth installed:

image

### DIFF
--- a/source/tests/system/nirfmxbt_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxbt_driver_api_tests.cpp
@@ -137,48 +137,40 @@ TEST_F(NiRFmxBTDriverApiTests, AcpBasicFromExample_DataLooksReasonable)
   EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MeasurementTypes::MEASUREMENT_TYPES_ACP, Boolean::BOOLEAN_TRUE));
   EXPECT_SUCCESS(session, client::acp_cfg_burst_synchronization_type(stub(), session, "", AcpBurstSynchronizationType::ACP_BURST_SYNCHRONIZATION_TYPE_PREAMBLE));
   EXPECT_SUCCESS(session, client::acp_cfg_averaging(stub(), session, "", AcpAveragingEnabled::ACP_AVERAGING_ENABLED_FALSE, 10));
-  AcpOffsetChannelMode offsetChannelMode = AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_SYMMETRIC;
-  EXPECT_SUCCESS(session, client::acp_cfg_offset_channel_mode(stub(), session, "", offsetChannelMode));
-  if (offsetChannelMode == AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_SYMMETRIC)
-  {
-    EXPECT_SUCCESS(session, client::acp_cfg_number_of_offsets(stub(), session, "", 5));
-  }
-  else if (offsetChannelMode == AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_INBAND)
-  {
-    EXPECT_SUCCESS(session, client::cfg_channel_number(stub(), session, "", 0));
-  }
+  EXPECT_SUCCESS(session, client::acp_cfg_offset_channel_mode(stub(), session, "", AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_SYMMETRIC));
+  EXPECT_SUCCESS(session, client::acp_cfg_number_of_offsets(stub(), session, "", 5));
   EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
 
-  auto referenceChannelPower = client::acp_fetch_reference_channel_power(stub(), session, "", 10.0);
-  ni::tests::system::EXPECT_SUCCESS(referenceChannelPower);
-  auto measurementResponse = client::acp_fetch_offset_measurement_array(stub(), session, "", 10.0);
-  ni::tests::system::EXPECT_SUCCESS(measurementResponse);
-  auto maskTrace = client::acp_fetch_mask_trace(stub(), session, "", 10.0);
-  ni::tests::system::EXPECT_SUCCESS(maskTrace);
-  auto absolutePowerTrace = client::acp_fetch_absolute_power_trace(stub(), session, "", 10.0);
-  ni::tests::system::EXPECT_SUCCESS(absolutePowerTrace);
-  auto fetchedSpectrum = client::acp_fetch_spectrum(stub(), session, "", 10.0);
-  ni::tests::system::EXPECT_SUCCESS(fetchedSpectrum);
+  auto reference_channel_power = client::acp_fetch_reference_channel_power(stub(), session, "", 10.0);
+  EXPECT_SUCCESS(session, reference_channel_power);
+  auto measurement_response = client::acp_fetch_offset_measurement_array(stub(), session, "", 10.0);
+  EXPECT_SUCCESS(session, measurement_response);
+  auto mask_trace = client::acp_fetch_mask_trace(stub(), session, "", 10.0);
+  EXPECT_SUCCESS(session, mask_trace);
+  auto absolute_power_trace = client::acp_fetch_absolute_power_trace(stub(), session, "", 10.0);
+  EXPECT_SUCCESS(session, absolute_power_trace);
+  auto fetched_spectrum = client::acp_fetch_spectrum(stub(), session, "", 10.0);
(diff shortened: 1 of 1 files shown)


source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -104,6 +125,62 @@ TEST_F(NiRFmxBTDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  ni::tests::system::EXPECT_SUCCESS(absolutePowerTrace);
+  auto fetchedSpectrum = client::acp_fetch_spectrum(stub(), session, "", 10.0);
+  ni::tests::system::EXPECT_SUCCESS(fetchedSpectrum);
+
+  EXPECT_GT(referenceChannelPower.reference_channel_power(), 0.0);
+  EXPECT_THAT(measurementResponse.lower_absolute_power(), Each(Ne(0.0)));
+  EXPECT_THAT(measurementResponse.upper_absolute_power(), Each(Ne(0.0)));
+  EXPECT_THAT(measurementResponse.lower_relative_power(), Each(Ne(0.0)));
+  EXPECT_THAT(measurementResponse.upper_relative_power(), Each(Ne(0.0)));
+  EXPECT_GT(measurementResponse.actual_array_size(), 0.0);
```
astarche, line 168: nit: I think all of the size comparisons should be GT `0` (integer value) not `0.0` (float value).

DavidCurtiss: Why compare float to int?

reckenro: I think he's just talking about the assertions on array sizes, all of which should be ints.

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -104,6 +125,62 @@ TEST_F(NiRFmxBTDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  EXPECT_SUCCESS(session, client::cfg_iq_power_edge_trigger(stub(), session, "", "0", IQPowerEdgeTriggerSlope::IQ_POWER_EDGE_TRIGGER_SLOPE_RISING_SLOPE, -20.0, 0.0, TriggerMinimumQuietTimeMode::TRIGGER_MINIMUM_QUIET_TIME_MODE_AUTO, 100e-6, IQPowerEdgeTriggerLevelType::IQ_POWER_EDGE_TRIGGER_LEVEL_TYPE_RELATIVE, Boolean::BOOLEAN_TRUE));
+  EXPECT_SUCCESS(session, client::cfg_packet_type(stub(), session, "", PacketType::PACKET_TYPE_DH1));
+  EXPECT_SUCCESS(session, client::cfg_data_rate(stub(), session, "", 1000000));
+  EXPECT_SUCCESS(session, client::cfg_payload_length(stub(), session, "", PayloadLengthMode::PAYLOAD_LENGTH_MODE_AUTO, 10));
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MeasurementTypes::MEASUREMENT_TYPES_ACP, Boolean::BOOLEAN_TRUE));
+  EXPECT_SUCCESS(session, client::acp_cfg_burst_synchronization_type(stub(), session, "", AcpBurstSynchronizationType::ACP_BURST_SYNCHRONIZATION_TYPE_PREAMBLE));
+  EXPECT_SUCCESS(session, client::acp_cfg_averaging(stub(), session, "", AcpAveragingEnabled::ACP_AVERAGING_ENABLED_FALSE, 10));
+  AcpOffsetChannelMode offsetChannelMode = AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_SYMMETRIC;
```
astarche, line 140: nit: pretty much all locals and functions in grpc-device are `snake_case`. Classes are `PascalCase`. And constants are mostly `UPPER_SNAKE_CASE`.

@mgumble @reckenro we use the "google" configuration for clang auto-format. Naming/casing is something it doesn't mess with. I've always used the google style guide for anything ambiguous (this [case](https://google.github.io/styleguide/cppguide.html#Variable_Names)) because (a) I figured it would be in line with the clang formatter, (b) existing style in grpc-device when I started was closer to google than NI C++ codebases I've worked in, (c) it's public, detailed and I like it. You're welcome to take a different approach or we can talk about ways to extend/document style guidelines or extend our use of auto-formatters and linters.

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -104,6 +125,62 @@ TEST_F(NiRFmxBTDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  EXPECT_SUCCESS(session, client::acp_cfg_offset_channel_mode(stub(), session, "", offsetChannelMode));
+  if (offsetChannelMode == AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_SYMMETRIC)
+  {
+    EXPECT_SUCCESS(session, client::acp_cfg_number_of_offsets(stub(), session, "", 5));
+  }
+  else if (offsetChannelMode == AcpOffsetChannelMode::ACP_OFFSET_CHANNEL_MODE_INBAND)
```
astarche, line 146: This is dead code right? I typically just pick a case and delete the other when adapting examples. You can also parameterize the test case (ping me for docs if you're interested).

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -104,6 +125,62 @@ TEST_F(NiRFmxBTDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  {
+    EXPECT_SUCCESS(session, client::cfg_channel_number(stub(), session, "", 0));
+  }
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  auto referenceChannelPower = client::acp_fetch_reference_channel_power(stub(), session, "", 10.0);
+  ni::tests::system::EXPECT_SUCCESS(referenceChannelPower);
```
astarche, line 153: This is sort of subtle. There are 2 `EXPECT_SUCCESS` variants. One of them takes in a `session` and will use that to get a detailed error message. The other just checks the status code. When possible, it's better to use the one that takes in the session because you get better error messages (see above).

astarche (approved):
Looks good!

source/tests/system/nirfmxbt_driver_api_tests.cpp
DavidCurtiss: Ah, right.

reckenro (approved):

source/tests/system/nirfmxbt_driver_api_tests.cpp
astarche: Yeah, bad lack of emphasis. The ***SIZE*** comparisons should be ints. Everything else is correct as floats (I think).

source/tests/system/nirfmxbt_driver_api_tests.cpp
mgumble: We're not exhaustively testing these drivers, not sure if there is enough benefit to test INBAND, since it appears that mode does less for this driver. I would suggest simply deleting it

mgumble (approved):

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: I've changed the locals to snake case.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: I've updated the test to just stick with the SYMMETRIC case.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: Fixed.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: Changed the array size comparisons to use an integer value.

PR #499, opened 2022-02-01
Split apart acronyms appropriately

DavidCurtiss:

Splits apart acronyms and other word combos appropriately.
Adds some attributes that weren't present previously, but that are used by examples.

Regen from scrapigen

Built


generated/nirfmxnr/nirfmxnr_client.cpp
```diff
@@ -102,14 +102,14 @@ acp_cfg_noise_compensation_enabled(const StubPtr& stub, const nidevice_grpc::Ses
 }
 
 ACPCfgNumberOfENDCOffsetsResponse
-acp_cfg_number_of_endc_offsets(const StubPtr& stub, const nidevice_grpc::Session& instrument, const pb::string& selector_string, const pb::int32& number_of_endc_offsets)
+acp_cfg_number_of_endc_offsets(const StubPtr& stub, const nidevice_grpc::Session& instrument, const pb::string& selector_string, const pb::int32& number_of_en_dc_offsets)
```
astarche, line 105: Just caught this here. ENDC is all-one-word.

Source: https://zone.ni.com/reference/en-XX/help/377910C-01/rfmxnrdotnet/html/m_nationalinstruments_rfmx_nrmx_rfmxnrmxacpconfiguration_getnumberofendcoffsets/

I added AB#1825782 to track this. You can resolve this if your happy with that.

I figure it makes sense to get in this big change and then patch up stuff afterwards. But if you'd rather fix up in place (and start up a scrapigen PR) that's fine too.

astarche (approved):

generated/nirfmxnr/nirfmxnr_client.cpp
reckenro: Ah, good catch. Yeah, I think we'll have to do something to note ENDC as a "sanctioned" acroynym so it doesn't get split up into EN and DC (since DC is also a valid acronym). I think passing ENDC as an additional acronym for NR and passing along those `additional_acronyms` to the resolve step could work so if we come across one we don't try to split it down any further.

reckenro (approved):

mgumble (approved):

generated/nirfmxnr/nirfmxnr_client.cpp
DavidCurtiss: Restored ENDC.

PR #500, opened 2022-02-01
Add some system level tests based on examples for WLAN.

reckenro:

Adds four system level tests based on some of the WLAN C Examples.

Related work item: AB#1794514.

We want to spot check the WLAN service and doing so with examples seems like a good way.

WLAN Driver Api system tests pass:

image

### Up next for WLAN

I think we need to go through the AC for the story to determine if we want any more example based tests, general tests, or if the remaining should be manual spot check on concerning methods / params + reliance on SpecAn/Instr tests covering similar patterns.


source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -16,13 +16,20 @@ namespace system {
 namespace {
 
 const auto PXI_5663E = "5663E";
+const int kRisingEdgeDetectionFailedWarning = 379206;
```
reckenro, line 19: I"m still trying to figure this out, but the TXP related tests gave this warning but still got results. The full message was "379206: Rising edge detection failed for the acquired waveform.". I'm trying to figure out if this is problematic or not.

astarche: I think it's OK. I've hit the same type of warning running on hardware that's not set up with the right cables/external-config etc.

Also can you use `UPPER_SNAKE` for constants, at least in files that don't already use `kWhatever`. I know I've been ignoring the session tests ...

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -104,6 +129,232 @@ TEST_F(NiRFmxWLANDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  OFDMModAccFetchCompositeRMSEVMResponse ofdm_mod_acc_fetch_composite_rmsevm_response;
+  TXPFetchMeasurementResponse txp_fetch_measurement_response;
+  ofdm_mod_acc_fetch_composite_rmsevm_response = client::ofdm_mod_acc_fetch_composite_rmsevm(stub(), session, "", 10.0);
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_composite_rmsevm_response);
+  txp_fetch_measurement_response = client::txp_fetch_measurement(stub(), session, "", 10.0);
+  EXPECT_WARNING(txp_fetch_measurement_response, kRisingEdgeDetectionFailedWarning);
+  
+  EXPECT_LT(0.0, ofdm_mod_acc_fetch_composite_rmsevm_response.composite_rmsevm_mean());
```
astarche, line 156: Some of these field names will be broken by David's change right? Or is wlan not included in that yet?

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -104,6 +129,232 @@ TEST_F(NiRFmxWLANDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  EXPECT_SUCCESS(session, client::cfg_standard(stub(), session, "", STANDARD_802_11_AG));
+  EXPECT_SUCCESS(session, client::cfg_channel_bandwidth(stub(), session, "", 20e6));
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MEASUREMENT_TYPES_TXP|MEASUREMENT_TYPES_OFDMMODACC, true));
+  EXPECT_SUCCESS(session, client::ofdm_mod_acc_cfg_measurement_length(stub(), session, "", 0, 16));
+  EXPECT_SUCCESS(session, client::ofdm_mod_acc_cfg_averaging(stub(), session, "", OFDM_MOD_ACC_AVERAGING_ENABLED_FALSE, 10));
+  EXPECT_SUCCESS(session, client::txp_cfg_averaging(stub(), session, "", TXP_AVERAGING_ENABLED_FALSE, 10));
+  EXPECT_SUCCESS(session, client::txp_cfg_maximum_measurement_interval(stub(), session, "", 1e-3));
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  OFDMModAccFetchCompositeRMSEVMResponse ofdm_mod_acc_fetch_composite_rmsevm_response;
```
astarche, line 149: These are declared separately because their codegenerated? I'll allow it. That's fine.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -104,6 +129,232 @@ TEST_F(NiRFmxWLANDriverApiTests, InitializeFromNIRFSA_Close_Succeeds)
...
+  
+  EXPECT_WARNING(txp_fetch_power_trace_response, kRisingEdgeDetectionFailedWarning);
+  EXPECT_GT(0.0, txp_fetch_power_trace_response.x0());
+  EXPECT_LT(0.0, txp_fetch_power_trace_response.dx());
+  EXPECT_EQ(25250, txp_fetch_power_trace_response.power_size());
+  EXPECT_EQ(25250, txp_fetch_power_trace_response.power().size());
+  EXPECT_LT(0.0, txp_fetch_power_trace_response.power(0));
+  EXPECT_WARNING(txp_fetch_measurement_response, kRisingEdgeDetectionFailedWarning);
+  EXPECT_LT(0.0, txp_fetch_measurement_response.average_power_mean());
+  EXPECT_LT(0.0, txp_fetch_measurement_response.peak_power_maximum());
```
astarche, line 355: Nice chunk of tests!

astarche (approved):
Nice chunk of tests!

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Yeah, you're right. I think his is getting close so I'll wait for #499 to go in and then merge main and fix up the names.

DavidCurtiss: I can also regen after the update, so Ryan can merge that in rather than doing it manually.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
DavidCurtiss: I intentionally generated variable declarations at the top of the block in which they're first used. If you'd rather they all be at the top of the function, I can do that.
(The examples define all the variables at the top of the function because they're in C [vs. C++], which requires that.)

astarche: Oh no, this is way better than the C examples! That style is a nightmare for transcription for one thing.

My preference is:

```c++
const auto ofdm_mod_acc_fetch_composite_rmsevm_response = client::ofdm_mod_acc_fetch_composite_rmsevm(stub(), session, "", 10.0);
```

That allows it to be const and all-on-one-line. I think of it as avoiding the extra default initialization, but that might be optimizer-dependent anyway (though some of that is more official in c++17).

DavidCurtiss: Ah, right. I agree. I had another look at the generator, and it was easy enough to change. @reckenro I pushed another change that can be merged in, to inline the declarations when possible. You (or I) can probably just include it in a future update.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
DavidCurtiss: Done. You can find the updates at the usual place, @reckenro -- also rebased on top of main.

reckenro: Updated from your branch. Got the updates for the acronym names and the field names on one line.

PR #501, opened 2022-02-02
Add WLAN N Waveforms Size Mechanism

danestull:

Created a basic implementation of a size mechanism for validating information on two-dimensional arrays such as the N waveform ones in WLAN.

This adds back in three required functions for WLAN functionality, now with more appropriate sizing mechanisms.

Created tests for the size mechanism in the nifake tests. These pass. 
Screen Shot 2022-02-01 at 3 35 43 PM

Manually checked the generated proto files and service.cpp for WLAN to ensure everything generated correctly


source/codegen/metadata/nirfmxwlan/functions.py
```diff
@@ -70,6 +70,142 @@
...
+                'name': 'iq',
+                'size': {
+                    'mechanism': 'two-dimension',
+                    'value': 'iqSize'
+                },
+                'type': 'NIComplexSingle[]'
+            },
+            {
+                'direction': 'in',
+                'grpc_name': 'iq_lengths',
```
astarche, line 120: I think these should be `iq_sizes`/etc not lengths. lengths is a fine name, but RFmx uses `size` throughout the API. And sizes is also closer to the original name.

source/codegen/templates/service_helpers.mako
```diff
@@ -485,6 +488,22 @@ ${initialize_standard_input_param(function_name, parameter)}
...
+  parameter_name = common_helpers.get_cpp_local_name(parameter)
+  size_param_name = common_helpers.get_size_param(parameter)
+  size_param = common_helpers.get_param_with_name(parameters, size_param_name)
+  size_field_name = common_helpers.get_grpc_field_name(size_param)
+%>\
+${initialize_standard_input_param(function_name, parameter)}
+      auto total_length = std::accumulate(request->${size_field_name}().cbegin(), request->${size_field_name}().cend(), 0);
+
+      if (total_length != request->${parameter_name}_size()) {
+        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The total size of ${parameter_name} did not match the exected size from ${size_field_name}");
```
astarche, line 503: I'm trying to think of a better message here. Imagine a user that is as confused about this API as we were trying to figure it out.

"did not match the expected sizes" -> Just saying "sizes" instead of "size" helps a little bit.

I think either "sum" or "2D" should be in here somewhere.

> "The total size of the two-dimensional array ${parameter_name} does not match the expected size from the sum of ${size_field_name}"

Something like that.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -1822,6 +1822,64 @@ TEST(NiFakeServiceTests, NiFakeService_AcceptViSessionArray_CallsAcceptViSession
...
+  NiFakeMockLibrary library;
+  auto resource_repository = std::make_shared<FakeResourceRepository>(&session_repository);
+  nifake_grpc::NiFakeService service(&library, resource_repository);
+  std::uint32_t session_id = create_session(library, service, kTestViSession);
+  ViInt32 array[] = {1, 2, 2, 3, 3, 3};
+  ViInt32 array_lengths[] = {1, 2, 3};
+  ViInt32 array_size = 3;
+  // two-dimension call
+  EXPECT_CALL(library, UseATwoDimensionParameter(kTestViSession, _, _, array_size))
+      .WillOnce(Return(kDriverSuccess));
```
astarche, line 1838: Update this to test that the actual data is passed in.

There is a matcher called `ElementsAreArray` that is used in some tests. You still use the `_` for the param matcher, the array validation.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -1822,6 +1822,64 @@ TEST(NiFakeServiceTests, NiFakeService_AcceptViSessionArray_CallsAcceptViSession
   EXPECT_EQ(kDriverSuccess, response.status());
 }
 
+// Test for two-dimension mechanism
```
astarche, line 1825: nit: I think this is a helpful place to emphasize that `two-dimension` ONLY does the size validation check and passes through the data. (Unlike other size mechanisms that do stuff to figure out the size).

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -1822,6 +1822,64 @@ TEST(NiFakeServiceTests, NiFakeService_AcceptViSessionArray_CallsAcceptViSession
...
+  }
+  for (int length : array_lengths) {
+    request.add_array_lengths(length);
+  }
+  nifake_grpc::UseATwoDimensionParameterResponse response;
+  ::grpc::Status status = service.UseATwoDimensionParameter(&context, &request, &response);
+
+  EXPECT_FALSE(status.ok());
+  EXPECT_EQ(::grpc::INVALID_ARGUMENT, status.error_code());
+}
```
astarche, line 1881: Can you check with @reckenro  if any of the system tests use these methods or could use these methods? We at least need a task to try to exercise them with system tests if that's not already covered.

astarche (approved):

source/codegen/common_helpers.py
```diff
@@ -197,11 +197,12 @@ def is_unsupported_size_mechanism(parameter: dict) -> bool:
 #                     elements filled in to the array as long as that is smaller than the passed-in value. (if it is
 #                     larger, the underlying call will return an error)
 #                     Should only be used for output arrays.
+# - two-dimension: The array is two dimensional
```
astarche, line 200: Can you put a little more information here. Basically that the size param is actually a list of sizes and the actual mechanism is just to validate that the 1-d input data is the same size as the sum of the list of sizes.

generated/nirfmxwlan/nirfmxwlan_service.cpp
```diff
@@ -2964,6 +3055,49 @@ namespace nirfmxwlan_grpc {
...
+      if (total_length != request->reference_waveform_size()) {
+        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The total size of reference_waveform did not match the exected size from reference_waveform_lengths");
+      }
+
+      auto reference_waveform_lengths = const_cast<int32*>(reinterpret_cast<const int32*>(request->reference_waveform_lengths().data()));
+      auto array_size_determine_from_sizes = std::array<int, 3>
+      {
+        request->x0_size(),
+        request->dx_size(),
+        request->reference_waveform_lengths_size()
```
mgumble, line 3083: @astarche Dane mentioned to me he went with lengths because this generated code got choked up it whenever it was named _size().

generated/nitclk/nitclk_service.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include <sstream>
 #include <fstream>
+#include <numeric>
```
mgumble, line 11: Why are we including numeric now?

source/codegen/templates/service_helpers.mako
mgumble: I like alex's recommendation

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -1822,6 +1822,64 @@ TEST(NiFakeServiceTests, NiFakeService_AcceptViSessionArray_CallsAcceptViSession
...
+    request.add_array(arrayval);
+  }
+  for (int length : array_lengths) {
+    request.add_array_lengths(length);
+  }
+  nifake_grpc::UseATwoDimensionParameterResponse response;
+  ::grpc::Status status = service.UseATwoDimensionParameter(&context, &request, &response);
+
+  EXPECT_FALSE(status.ok());
+  EXPECT_EQ(::grpc::INVALID_ARGUMENT, status.error_code());
```
mgumble, line 1880: nit: would be nice to check the message too since there are multiple code paths that return invalid_argument would be good to know that we going through expected code path,.

mgumble (approved):

source/codegen/metadata/nirfmxwlan/functions.py
danestull: These have been updated in Scrapigen and the exports moved to apply here

source/codegen/templates/service_helpers.mako
danestull: I've updated this description so as to be more descriptive and helpful to the user.

source/tests/unit/ni_fake_service_tests.cpp
danestull: I had forgotten to go back and update this but have now added a more full description.

generated/nirfmxwlan/nirfmxwlan_service.cpp
danestull: We've updated to `sizes`

source/tests/unit/ni_fake_service_tests.cpp
reckenro: The only one I saw across the C Examples was `RFmxWLANOFDMModAccMIMOAnalysisOnly` which uses `RFmxWLAN_AnalyzeNWaveformsIQ`. So far we've avoided that one, along with others, because they're reading from a TDMS file to get the waveforms.

reckenro (approved):

generated/nitclk/nitclk_service.cpp
astarche: `std::accumulate` is in `numeric`.

danestull: I'm trying to write a helper so it only gets included if we have a two-dimension size mechanism rn

source/tests/unit/ni_fake_service_tests.cpp
astarche: I can help get those working if you ping me. I did that for a SpecAn test. But for this PR you can just add a task and resolve this.

generated/nitclk/nitclk_service.cpp
danestull: I've added some helper functions and we now only include numeric if we have a two dimensional array and need it.

source/tests/unit/ni_fake_service_tests.cpp
danestull: Added a new task #1828053 So I will get this PR in the work with both of yall on getting additional spot tests in for these methods.

source/tests/unit/ni_fake_service_tests.cpp
danestull: Was able to update to use `ElementsAreArray` for the Sizes array but since we don't pass the total size of the array to the function we couldn't get it to work to check the array proper. @reckenro was able to help me create a new Matcher and called `MatchesArray` and use it in the function call in place of the discard.

source/tests/unit/ni_fake_service_tests.cpp
danestull: We now check the message as well!

PR #502, opened 2022-02-02
Add TxpBasic system-level test for NI-RFmx Bluetooth

tdunkle:

This pull request adds a system-level test based on the TxpBasic CVI example for NI-RFmx Bluetooth. This is the second of two pull requests (after #498 ) that solves [AB#1784736](https://ni.visualstudio.com/DevCentral/_workitems/edit/1784736).

We want some basic system-level tests that confirm that fetching data with the gRPC API for NI-RFmx Bluetooth returns reasonable-looking results.

grpc-device locally builds successfully with the new test, and the new test passes on a VM with NI-RFSA and NI-RFmx Bluetooth installed:

image

### DIFF
--- a/source/tests/system/nirfmxbt_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxbt_driver_api_tests.cpp
@@ -197,6 +197,7 @@ TEST_F(NiRFmxBTDriverApiTests, TxpBasicFromExample_DataLooksReasonable)
   EXPECT_SUCCESS(session, client::txp_cfg_averaging(stub(), session, "", TxpAveragingEnabled::TXP_AVERAGING_ENABLED_FALSE, 10));
   EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
 
+  // We expect these actions to produce kPreambleSyncPacketStartDetectionFailedWarning since the test uses simulated hardware.
   const auto fetched_powers_response = client::txp_fetch_powers(stub(), session, "", 10.0);
   EXPECT_WARNING(fetched_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
   const auto fetched_edr_powers_response = client::txp_fetch_edr_powers(stub(), session, "", 10.0);
--- a/source/tests/system/nirfmxbt_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxbt_driver_api_tests.cpp
@@ -18,11 +18,18 @@ namespace system {
 namespace {
 
 const auto PXI_5663E = "5663E";
+const int kPreambleSyncPacketStartDetectionFailedWarning = 686280;
 
 template <typename TResponse>
 void EXPECT_SUCCESS(const TResponse& response)
 {
   EXPECT_EQ(0, response.status());
+}
+
+template <typename TResponse>
+void EXPECT_WARNING(const TResponse& response, const int expected_warning_id)
+{
+  EXPECT_EQ(expected_warning_id, response.status());
 }
 
 class NiRFmxBTDriverApiTests : public Test {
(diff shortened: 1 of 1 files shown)


source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -173,6 +180,36 @@ TEST_F(NiRFmxBTDriverApiTests, AcpBasicFromExample_DataLooksReasonable)
...
+  EXPECT_WARNING(fetched_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
+  auto fetched_edr_powers_response = client::txp_fetch_edr_powers(stub(), session, "", 10.0);
+  EXPECT_WARNING(fetched_edr_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
+  auto fetched_lecte_reference_period_powers_response = client::txp_fetch_lecte_reference_period_powers(stub(), session, "", 10.0);
+  EXPECT_WARNING(fetched_lecte_reference_period_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
+
+  EXPECT_GT(fetched_powers_response.average_power_mean(), 0.0);
+  EXPECT_GT(fetched_powers_response.average_power_maximum(), 0.0);
+  EXPECT_GT(fetched_powers_response.average_power_minimum(), 0.0);
+  EXPECT_GT(fetched_powers_response.peak_to_average_power_ratio_maximum(), 0.0);
```
tdunkle, line 210: As is, this test only contains assertions for `fetched_powers_response`, because it is the only response that contains valid data. Since my testing so far has relied on virtual instruments that are not necessarily configured in the way this API expects (e.g. we may need multiple instruments connected, or some sort of external input signal), we get warnings from all three actions and the members contained in `fetched_edr_powers_response` and `fetched_lecte_reference_period_powers_response` are all nan. Let me know if you have any concerns with this.

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -18,13 +18,20 @@ namespace system {
 namespace {
 
 const auto PXI_5663E = "5663E";
+const int kPreambleSyncPacketStartDetectionFailedWarning = 686280;
```
astarche, line 21: nit: in files that use UPPER_SNAKE for constants, let's stick with that.

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -173,6 +180,36 @@ TEST_F(NiRFmxBTDriverApiTests, AcpBasicFromExample_DataLooksReasonable)
...
+  EXPECT_SUCCESS(session, client::cfg_data_rate(stub(), session, "", 1000000));
+  EXPECT_SUCCESS(session, client::cfg_payload_length(stub(), session, "", PayloadLengthMode::PAYLOAD_LENGTH_MODE_AUTO, 10));
+  EXPECT_SUCCESS(session, client::cfg_le_direction_finding(stub(), session, "", DirectionFindingMode::DIRECTION_FINDING_MODE_DISABLED, 160e-6, 1e-6));
+  EXPECT_SUCCESS(session, client::auto_level(stub(), session, "", 10e-3));
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MeasurementTypes::MEASUREMENT_TYPES_TXP, Boolean::BOOLEAN_TRUE));
+  EXPECT_SUCCESS(session, client::txp_cfg_burst_synchronization_type(stub(), session, "", TxpBurstSynchronizationType::TXP_BURST_SYNCHRONIZATION_TYPE_PREAMBLE));
+  EXPECT_SUCCESS(session, client::txp_cfg_averaging(stub(), session, "", TxpAveragingEnabled::TXP_AVERAGING_ENABLED_FALSE, 10));
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  auto fetched_powers_response = client::txp_fetch_powers(stub(), session, "", 10.0);
```
astarche, line 200: nit/fyi: these can be `const auto`

astarche (approved):

source/tests/system/nirfmxbt_driver_api_tests.cpp
astarche: Retracting, you can use whichever you want. As Ryan pointed out to me, this is what the driver APIs tend to use so it's reasonable to mimic them.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: I implemented this suggestion for both the AcpBasic and TxpBasic tests.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: Either way, the two constants in my file weren't consistent, so I went back and changed `PXI_5663E` to `kPxi5663e`.

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -173,6 +180,36 @@ TEST_F(NiRFmxBTDriverApiTests, AcpBasicFromExample_DataLooksReasonable)
...
+  EXPECT_SUCCESS(session, client::txp_cfg_burst_synchronization_type(stub(), session, "", TxpBurstSynchronizationType::TXP_BURST_SYNCHRONIZATION_TYPE_PREAMBLE));
+  EXPECT_SUCCESS(session, client::txp_cfg_averaging(stub(), session, "", TxpAveragingEnabled::TXP_AVERAGING_ENABLED_FALSE, 10));
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  const auto fetched_powers_response = client::txp_fetch_powers(stub(), session, "", 10.0);
+  EXPECT_WARNING(fetched_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
+  const auto fetched_edr_powers_response = client::txp_fetch_edr_powers(stub(), session, "", 10.0);
+  EXPECT_WARNING(fetched_edr_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
+  const auto fetched_lecte_reference_period_powers_response = client::txp_fetch_lecte_reference_period_powers(stub(), session, "", 10.0);
+  EXPECT_WARNING(fetched_lecte_reference_period_powers_response, kPreambleSyncPacketStartDetectionFailedWarning);
```
mgumble, line 205: Worth adding a simple comment here to explain we are not worried about this error as we expect it with simulated HW

mgumble (approved):

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: Done!

reckenro (approved):

PR #507, opened 2022-02-03
Add another batch of example based tests for WLAN.

reckenro:

Adds some more WLAN example based tests. These use some functionality from the RFmx Instr service as well as WLAN.
Fixes [AB#1828895](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1828895)

Add more test coverage/confidence in WLAN service.

WLAN driver api system-level tests passing:

image

### DIFF
--- a/source/tests/system/nirfmxwlan_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxwlan_driver_api_tests.cpp
@@ -2,12 +2,12 @@
 
 #include "device_server.h"
 #include "niRFmxWLAN.h"
-#include "nirfmxwlan/nirfmxwlan_client.h"
 #include "nirfmxinstr/nirfmxinstr_client.h"
+#include "nirfmxwlan/nirfmxwlan_client.h"
 #include "nirfsa/nirfsa_client.h"
 
 #ifdef WIN32
-#define strncpy strncpy_s
+  #define strncpy strncpy_s
 #endif
 
 using namespace ::testing;
@@ -90,24 +90,21 @@ class NiRFmxWLANDriverApiTests : public Test {
   std::unique_ptr<NiRFmxWLAN::Stub> stub_;
 };
 
-void GetCommaSeparatedStringFromArray(char* arrayOfNames, int numOfNames, int nameLength, char* commaSeparatedName)
+void GetCommaSeparatedStringFromArray(char* array_of_names, int num_of_names, int name_length, char* comma_seperated_name)
 {
-   int i = 0;
-   int j=0;
-   while(i < numOfNames)
-   {
-      int k;
-      for (k = 0; k < nameLength; ++k)
-      {
-         if (*(arrayOfNames + i*nameLength + k) == '\0') break;
-         commaSeparatedName[j++] = *(arrayOfNames + i*nameLength + k);
-      }
-      if (++i != numOfNames)
-      {
-         commaSeparatedName[j++] = ',';
-      }
-   }
-   commaSeparatedName[j] = '\0';
+  int name_index = 0;
+  int comma_index = 0;
+  while (name_index < num_of_names) {
+    int char_index;
+    for (char_index = 0; char_index < name_length; ++char_index) {
+      if (*(array_of_names + name_index * name_length + char_index) == '\0') break;
+      comma_seperated_name[comma_index++] = *(array_of_names + name_index * name_length + char_index);
+    }
+    if (++name_index != num_of_names) {
+      comma_seperated_name[comma_index++] = ',';
+    }
+  }
+  comma_seperated_name[comma_index] = '\0';
 }
 
 InitializeResponse init(const client::StubPtr& stub, const std::string& model, const std::string& resource_name)
@@ -118,7 +115,7 @@ InitializeResponse init(const client::StubPtr& stub, const std::string& model, c
 
 InitializeResponse init(const client::StubPtr& stub, const std::string& model)
 {
-   return init(stub, model, "FakeDevice");
+  return init(stub, model, "FakeDevice");
 }
 
 nidevice_grpc::Session init_session(const client::StubPtr& stub, const std::string& model, const std::string& resource_name)
@@ -183,7 +180,7 @@ TEST_F(NiRFmxWLANDriverApiTests, OFDMModAccTXPCompositeFromExample_FetchData_Dat
   EXPECT_SUCCESS(session, client::cfg_iq_power_edge_trigger(stub(), session, "", "0", IQ_POWER_EDGE_TRIGGER_SLOPE_RISING_SLOPE, -20.0, 0.0, TRIGGER_MINIMUM_QUIET_TIME_MODE_AUTO, 5.0e-6, IQ_POWER_EDGE_TRIGGER_LEVEL_TYPE_RELATIVE, BOOLEAN_TRUE));
   EXPECT_SUCCESS(session, client::cfg_standard(stub(), session, "", STANDARD_802_11_AG));
   EXPECT_SUCCESS(session, client::cfg_channel_bandwidth(stub(), session, "", 20e6));
-  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MEASUREMENT_TYPES_TXP|MEASUREMENT_TYPES_OFDMMODACC, true));
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MEASUREMENT_TYPES_TXP | MEASUREMENT_TYPES_OFDMMODACC, true));
   EXPECT_SUCCESS(session, client::ofdm_mod_acc_cfg_measurement_length(stub(), session, "", 0, 16));
   EXPECT_SUCCESS(session, client::ofdm_mod_acc_cfg_averaging(stub(), session, "", OFDM_MODACC_AVERAGING_ENABLED_FALSE, 10));
   EXPECT_SUCCESS(session, client::txp_cfg_averaging(stub(), session, "", TXP_AVERAGING_ENABLED_FALSE, 10));
@@ -239,7 +236,7 @@ TEST_F(NiRFmxWLANDriverApiTests, SemFromExample_FetchData_DataLooksReasonable)
   EXPECT_EQ(4, sem_fetch_lower_offset_margin_array_response.margin().size());
   EXPECT_GT(0.0, sem_fetch_lower_offset_margin_array_response.margin(0));
   EXPECT_EQ(4, sem_fetch_lower_offset_margin_array_response.margin_frequency_size());
-  EXPECT_EQ( 4, sem_fetch_lower_offset_margin_array_response.margin_frequency().size());
+  EXPECT_EQ(4, sem_fetch_lower_offset_margin_array_response.margin_frequency().size());
   EXPECT_LT(0.0, sem_fetch_lower_offset_margin_array_response.margin_frequency(0));
   EXPECT_EQ(4, sem_fetch_lower_offset_margin_array_response.margin_absolute_power_size());
   EXPECT_EQ(4, sem_fetch_lower_offset_margin_array_response.margin_absolute_power().size());
@@ -277,11 +274,11 @@ TEST_F(NiRFmxWLANDriverApiTests, SemFromExample_FetchData_DataLooksReasonable)
 TEST_F(NiRFmxWLANDriverApiTests, SEMCustomMaskFromExample_FetchData_DataLooksReasonable)
 {
   const auto NUMBER_OF_OFFSETS = 3;
-  std::vector<float64> offsetStartFrequency {9e06, 11e06, 20e06};
-  std::vector<float64> offsetStopFrequency {11e06, 20e06, 40e06};
-  std::vector<int> offsetSideband {RFMXWLAN_VAL_SEM_OFFSET_SIDEBAND_BOTH, RFMXWLAN_VAL_SEM_OFFSET_SIDEBAND_BOTH, RFMXWLAN_VAL_SEM_OFFSET_SIDEBAND_BOTH};
-  std::vector<float64> relativeLimitStart {0.0, -20.0, -28.0};
-  std::vector<float64> relativeLimitStop {-20.0, -28.0, -40.0};
+  std::vector<float64> offsetStartFrequency{9e06, 11e06, 20e06};
+  std::vector<float64> offsetStopFrequency{11e06, 20e06, 40e06};
+  std::vector<int> offsetSideband{RFMXWLAN_VAL_SEM_OFFSET_SIDEBAND_BOTH, RFMXWLAN_VAL_SEM_OFFSET_SIDEBAND_BOTH, RFMXWLAN_VAL_SEM_OFFSET_SIDEBAND_BOTH};
+  std::vector<float64> relativeLimitStart{0.0, -20.0, -28.0};
+  std::vector<float64> relativeLimitStop{-20.0, -28.0, -40.0};
   auto session = init_session(stub(), PXI_5663E);
   EXPECT_SUCCESS(session, client::cfg_frequency_reference(stub(), session, "", FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK, 10e6));
   EXPECT_SUCCESS(session, client::cfg_frequency(stub(), session, "", 2.412e9));
@@ -437,7 +434,7 @@ TEST_F(NiRFmxWLANDriverApiTests, DSSSModAccFromExample_FetchData_DataLooksReason
 }
 
 // TODO AB#1830352:
-// Disabled because it's getting the error during initiate of 
+// Disabled because it's getting the error during initiate of
 // "-1074135024: \nIVI: (Hex 0xBFFA0010) Invalid value for parameter or property.\n\nProperty: IQ Rate\nRequested Value:  80.0e6\nValue Must Be Less Than or Equal to:  75.0e6\nDevice: Simulated_NI_PXIe_5601_Device_31\n\nStatus Code: -200077"
 TEST_F(NiRFmxWLANDriverApiTests, DISABLED_DSSSPowerRampFromExample_FetchData_DataLooksReasonable)
 {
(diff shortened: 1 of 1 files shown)


source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,805 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_EQ(1, dsss_mod_acc_fetch_evm_per_chip_mean_trace_response.dx());
+  EXPECT_EQ(0, dsss_mod_acc_fetch_evm_per_chip_mean_trace_response.evm_per_chip_mean_size());
+  EXPECT_EQ(0, dsss_mod_acc_fetch_evm_per_chip_mean_trace_response.evm_per_chip_mean().size());
+  EXPECT_WARNING(dsss_mod_acc_fetch_constellation_trace_response, EVM_CHIPS_MUST_BE_300_WARNING);
+  EXPECT_EQ(0, dsss_mod_acc_fetch_constellation_trace_response.actual_array_size());
+}
+
+// Disabled because it's getting the error during initiate of 
+// "-1074135024: \nIVI: (Hex 0xBFFA0010) Invalid value for parameter or property.\n\nProperty: IQ Rate\nRequested Value:  80.0e6\nValue Must Be Less Than or Equal to:  75.0e6\nDevice: Simulated_NI_PXIe_5601_Device_31\n\nStatus Code: -200077"
+TEST_F(NiRFmxWLANDriverApiTests, DISABLED_DSSSPowerRampFromExample_FetchData_DataLooksReasonable)
```
reckenro, line 441: I've added comments above all the currently Disabled tests. Each one fails with some error. I'm trying to figure out if it's something related to the simulated hardware setup and can be ignored or not.

I thought I'd put them up in case anyone had come across similar errors and see how we want to handle them.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,805 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_data_constellation_trace_response.data_constellation(0).imaginary());
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response);
+  EXPECT_EQ(-26, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.x0());
+  EXPECT_EQ(1, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.dx());
+  EXPECT_EQ(53, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean_size());
+  EXPECT_EQ(53, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean().size());
+  EXPECT_LT(0.0, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean(0));
+}
+
+// Getting exception on the second internal call of BuildPortString2 (called from BuildPortString on the client):
```
reckenro, line 578: This one's interesting, it seems to be failing on the second internal call to `BuildPortString2` after we've determined the `selector_string_out_length` in `NiRFmxInstrService::BuildPortString`. I'm still working to see what the possible cause could be.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,805 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+
+  const auto dsss_mod_acc_fetch_evm_response = client::dsss_mod_acc_fetch_evm(stub(), session, "", 10.0);
+  const auto dsss_mod_acc_fetch_ppdu_information_response = client::dsss_mod_acc_fetch_ppdu_information(stub(), session, "", 10.0);
+  const auto dsss_mod_acc_fetch_iq_impairments_response = client::dsss_mod_acc_fetch_iq_impairments(stub(), session, "", 10.0);
+  const auto dsss_mod_acc_fetch_evm_per_chip_mean_trace_response = client::dsss_mod_acc_fetch_evm_per_chip_mean_trace(stub(), session, "", 10.0);
+  const auto dsss_mod_acc_fetch_constellation_trace_response = client::dsss_mod_acc_fetch_constellation_trace(stub(), session, "", 10.0);
+
+  EXPECT_WARNING(dsss_mod_acc_fetch_evm_response, EVM_CHIPS_MUST_BE_300_WARNING);
+  EXPECT_EQ(0.0, dsss_mod_acc_fetch_evm_response.rms_evm_mean());
+  EXPECT_EQ(0.0, dsss_mod_acc_fetch_evm_response.peak_evm80211_2016_maximum());
```
astarche, line 413: Doesn't need to be in this change. But I think these are names we'll want to fix.

`peak_evm_80211_2016_maximum`

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,805 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  const auto ofdm_mod_acc_fetch_sigbcrc_status_response = client::ofdm_mod_acc_fetch_sigbcrc_status(stub(), session, "", 10.0);
+  if (PPDUType == NIRFMXWLAN_INT32_OFDM_PPDU_TYPE_MU)
+  {
+    ofdm_mod_acc_fetch_number_of_users_response = client::ofdm_mod_acc_fetch_number_of_users(stub(), session, "", 10.0);
+    MCSIndexArrayLength = numberOfUsers;
+    MCSIndexArray = (int32*)malloc(MCSIndexArrayLength * sizeof(int32));
+    numberOfSpaceTimeStreamsArrayLength = numberOfUsers;
+    numberOfSpaceTimeStreamsArray = (int32*)malloc(numberOfSpaceTimeStreamsArrayLength * sizeof(int32));
+    spaceTimeStreamOffsetArray = (int32*)malloc(numberOfUsers * sizeof(int32));
+    if (MCSIndexArray == NULL || numberOfSpaceTimeStreamsArray == NULL)
```
astarche, line 667: I'm not sure what exactly your workflow is on codegen'ing these tests. If it's important to be able to regen from the tool, that's an OK reason to leave them alone even if the style is not perfect.

That said: this if block is empty. Should probably `FAIL() << "Could not allocate array";`.

Also, I assume that the reason that we're using c-style arrays here is that it allows us to copy the code directly from the C Examples? (that's a good reason).

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,805 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_data_constellation_trace_response.data_constellation(0).imaginary());
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response);
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.x0());
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.dx());
+  EXPECT_EQ(999, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean_size());
+  EXPECT_EQ(999, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean().size());
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean(0));
+}
+
+// Getting exception on the second internal call of BuildPortString2 (called from BuildPortString on the client):
```
astarche, line 975: These are definitely suspicious.

astarche (approved):

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: So, for now I'm just going to add the built port strings manually and comment out the calls to BuildPortString. 

I think we should add a Task somewhere to track it but I found that if I bumped the `selector_string_out_length` in `NiRFmxInstrService::BuildPortString` by one to be `int32 selector_string_out_length = status + 1;` then the tests using BuildPortString would pass. I'm not sure if this indicates something about BuildPortString2 or if it's something on our end but we have a lot of other ivi-dance string outputs so I'd expect us to hit this earlier if it was a general issue. @astarche , thoughts?

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Added AB#1829974 to address this.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Ah, yeah that was weird. I added the failure in the if block.

And yes, I believe it's more straightforward to copy over straight from the C Examples this way.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
astarche: We actually have a size mechanism tag to deal with that: [strlen-bug](https://github.com/ni/grpc-device/search?q=strlen-bug). Although it's also possible that it's "not a bug" and the API expects us to account for the null terminator.

Also, I've noticed that `ivi-dance` typically works for the build string methods, but I'm not sure that it's documented that way. And `ivi-dance` is pretty rare in RFmx. AND I think in examples they use a fixed buffer size. 

So questions we should answer:
1. Do all of the build string APIs have the strlen bug (i.e., returns strlen for the size when strlen + 1 is needed)? Or just BuildPortString.
2. Is `ivi-dance` a cotrrect/good way to call these methods?
3. Is there a documented fixed max size that we can use instead?
4. Should we file a bug to RFmx? (doesn't affect how we resolve it, but if it's a bug we should let them know).

source/tests/system/nirfmxwlan_driver_api_tests.cpp
DavidCurtiss: > If it's important to be able to regen from the tool, ...

My plan is just to codegen the tests once and then throw away the generator. No need to preserve the leftover oddities.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
DavidCurtiss: peak_evm_80211_2016_maximum is already covered in AB#1816468

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: I'll close this thread and keep the convo in the [other one](https://github.com/ni/grpc-device/pull/507#discussion_r798981123) about the investigation on this.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Oh yeah, I realized after the fact and marked the Task I added as Removed and linked to that one.

danestull (approved):

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -70,11 +77,39 @@ class NiRFmxWLANDriverApiTests : public Test {
...
+
  private:
   DeviceServerInterface* device_server_;
   std::unique_ptr<NiRFmxWLAN::Stub> stub_;
 };
 
+void GetCommaSeparatedStringFromArray(char* arrayOfNames, int numOfNames, int nameLength, char* commaSeparatedName)
+{
+   int i = 0;
+   int j=0;
```
mgumble, line 96: inconsistent spacing here. Also while we are at it I would recommend different variable names than ijk for readability

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,794 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_SUCCESS(session, client::cfg_number_of_frequency_segments_and_receive_chains(stub(), session, "", NUMBER_OF_FREQUENCY_SEGMENTS, NUMBER_OF_RECEIVE_CHAINS));
+  for (int i = 0; i < NUMBER_OF_FREQUENCY_SEGMENTS; ++i)
+  {
+    auto segment_string_response = client::build_segment_string(stub(), "", i);
+    EXPECT_SUCCESS(session, segment_string_response);
+    EXPECT_SUCCESS(session, client::cfg_frequency(stub(), session, segment_string_response.selector_string_out(), centerFrequencyArray[i]));
+  }
+  for (int i = 0; i < NUMBER_OF_DEVICES; ++i)
+  {
+    // TODO: Fix build_port_string parameter invalid issue.
```
mgumble, line 605: Need task to track

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,794 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_pilot_constellation_trace_response);
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_pilot_constellation_trace_response.pilot_constellation(0).real());
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_pilot_constellation_trace_response.pilot_constellation(0).imaginary());
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_data_constellation_trace_response);
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_data_constellation_trace_response.data_constellation(0).real());
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_data_constellation_trace_response.data_constellation(0).imaginary());
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response);
+  EXPECT_EQ(-26, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.x0());
+  EXPECT_EQ(1, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.dx());
+  EXPECT_EQ(53, ofdm_mod_acc_fetch_chain_rmsevm_per_subcarrier_mean_trace_response.chain_rms_evm_per_subcarrier_mean_size());
```
mgumble, line 572: I am sure this is just from generation but is there really much value for checking both the size of the array and the size we pass in ? We have unit tests for this

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,794 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_stream_rmsevm_response.stream_pilot_rms_evm_mean());
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_stream_rmsevm_per_subcarrier_mean_trace_response);
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_stream_rmsevm_per_subcarrier_mean_trace_response.x0());
+  EXPECT_EQ(0.0, ofdm_mod_acc_fetch_stream_rmsevm_per_subcarrier_mean_trace_response.dx());
+  EXPECT_EQ(0, ofdm_mod_acc_fetch_stream_rmsevm_per_subcarrier_mean_trace_response.stream_rms_evm_per_subcarrier_mean_size());
+  EXPECT_EQ(0, ofdm_mod_acc_fetch_stream_rmsevm_per_subcarrier_mean_trace_response.stream_rms_evm_per_subcarrier_mean().size());
+  // EXPECT_EQ(0.0, ofdm_mod_acc_fetch_stream_rmsevm_per_subcarrier_mean_trace_response.stream_rms_evm_per_subcarrier_mean(0));
+  EXPECT_SUCCESS(session, ofdm_mod_acc_fetch_pilot_constellation_trace_response);
+  EXPECT_EQ(0, ofdm_mod_acc_fetch_pilot_constellation_trace_response.pilot_constellation().size());
+  // EXPECT_EQ(0.0, ofdm_mod_acc_fetch_pilot_constellation_trace_response.pilot_constellation(0).real());
```
mgumble, line 753: im assuming these were not zero?

source/tests/system/nirfmxwlan_driver_api_tests.cpp
mgumble: We should add a task or maybe better a bug on the feature to investigate these failures or delete them if we decide it isnt worth fixing.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -339,6 +383,794 @@ TEST_F(NiRFmxWLANDriverApiTests, TXPFromExample_FetchData_DataLooksReasonable)
...
+  EXPECT_SUCCESS(session, client::cfg_number_of_frequency_segments_and_receive_chains(stub(), session, "", NUMBER_OF_FREQUENCY_SEGMENTS, NUMBER_OF_RECEIVE_CHAINS));
+  for (int i = 0; i < NUMBER_OF_FREQUENCY_SEGMENTS; ++i)
+  {
+    auto segment_string_response = client::build_segment_string(stub(), "", i);
+    EXPECT_SUCCESS(session, segment_string_response);
+    EXPECT_SUCCESS(session, client::cfg_frequency(stub(), session, segment_string_response.selector_string_out(), centerFrequencyArray[i]));
+  }
+  for (int i = 0; i < NUMBER_OF_DEVICES; ++i)
+  {
+    // TODO: Fix build_port_string parameter invalid issue.
```
mgumble, line 992: Another untracked todo

mgumble (approved):
I like using the generation to get us started in writing tests. I do think we could time later to make the tests less monolithic.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Sounds good. I'll start investigating this. Added a Bug under the WLAN feature to sort this out: AB#1830349.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
DavidCurtiss: The array size is 0 (see the line above), so this query for the first array element would fail if uncommented. I did the same (left in a commented line) in the NR tests. Either we'll figure out how the setup is wrong and fix it, in which case this commented line will be a reminder to test the value, or else we'll never fix the setup, and the commented line will just be pointing out that there's no data values to test. That's _my_ opinion on the matter, anyway.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Added a Bug under the WLAN feature for the two Disabled tests: AB#1830352 to track fixing or removing the tests.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Yeah, this is from the generation. We could possibly trim that down on the generation side with David's script but I'm going to leave them as is for now. Not worth the manual cleanup to remove redundant validation.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Yeah, it's because the array sizes are 0. David made the point in his NR tests PR that he was just going to leave them around and I liked that.

Like David said, either
1. At some point we have array sizes and are reminded to assert on at least on of the array values.
2. It remains forever and is a minor visual reminder that we got an empty array for something.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Cleaned up the spacing. Realize I didn't have format on save turned on so now it should be in the right style throughout. Also, updated the names.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Added Bug AB#1830349 under the WLAN feature to investigate the `build_port_string` issues. Added a reference to that Bug in the three spaces where the `build_port_string` calls are commented out.

source/tests/system/nirfmxwlan_driver_api_tests.cpp
reckenro: Same as the other ones, tracked with AB#1830349 and added reference in todo comments.

PR #508, opened 2022-02-03
Use protobuf bools instead of Boolean enum

astarche:

Use protobuf bools instead of Boolean enum.

Use an enum for `isDone` output on `CheckMeasurementStatus`.

Previously, I overzealously converted all `TRUE`/`FALSE` enums to `bool`s. This was rejected. However, the `Boolean` enum is really just ` bool` and represented that way in both C# and LV.

Ran and passed all RFmx tests.

### DIFF
--- a/generated/nirfmxnr/nirfmxnr.proto
+++ b/generated/nirfmxnr/nirfmxnr.proto
@@ -733,11 +733,6 @@ enum AcpSweepTimeAuto {
   ACP_SWEEP_TIME_AUTO_TRUE = 1;
 }
 
-enum Boolean {
-  BOOLEAN_FALSE = 0;
-  BOOLEAN_TRUE = 1;
-}
-
 enum ChpAveragingEnabled {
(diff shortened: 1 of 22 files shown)


generated/nirfmxnr/nirfmxnr.proto
```diff
@@ -733,11 +733,6 @@ enum AcpSweepTimeAuto {
   ACP_SWEEP_TIME_AUTO_TRUE = 1;
 }
```
astarche, line 735: Note: Bluetoooth and wlan are out-of-sync with scrapigen so I didn't import those.

reckenro (approved):

generated/nirfmxnr/nirfmxnr.proto
mgumble: lets make sure to throw a task on those user stories for Api cleanup

mgumble (approved):

PR #509, opened 2022-02-04
Add system-level test to get and set I32 attribute in NiRFmxBT.

tdunkle:

This PR adds a system-level test that proves that we can get and set I32 attributes in the RFmx Bluetooth service.
Fixes [AB#1830039](https://ni.visualstudio.com/DevCentral/_workitems/edit/1830039).

We want to be certain that getting and setting I32 attributes works as expected in the RFmx Bluetooth service.

grpc-device builds locally with this test, and the test passes on a VM with NI-RFSA and NI-RFmx Bluetooth installed.

### DIFF
--- a/source/tests/system/nirfmxbt_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxbt_driver_api_tests.cpp
@@ -211,6 +211,32 @@ TEST_F(NiRFmxBTDriverApiTests, TxpBasicFromExample_DataLooksReasonable)
   EXPECT_GT(fetched_powers_response.peak_to_average_power_ratio_maximum(), 0.0);
 }
 
+TEST_F(NiRFmxBTDriverApiTests, SetAndGetAttributeInt32_Succeeds)
+{
+  auto session = init_session(stub(), kPxi5663e);
+  EXPECT_SUCCESS(session, client::cfg_frequency_reference(stub(), session, "", FrequencyReferenceSource::FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK, 10e6));
+  EXPECT_SUCCESS(session, client::cfg_frequency(stub(), session, "", 2.402000e9));
+  EXPECT_SUCCESS(session, client::cfg_external_attenuation(stub(), session, "", 0.0));
+  EXPECT_SUCCESS(session, client::cfg_iq_power_edge_trigger(stub(), session, "", "0", IQPowerEdgeTriggerSlope::IQ_POWER_EDGE_TRIGGER_SLOPE_RISING_SLOPE, -20.0, 0.0, TriggerMinimumQuietTimeMode::TRIGGER_MINIMUM_QUIET_TIME_MODE_AUTO, 100e-6, IQPowerEdgeTriggerLevelType::IQ_POWER_EDGE_TRIGGER_LEVEL_TYPE_RELATIVE, Boolean::BOOLEAN_TRUE));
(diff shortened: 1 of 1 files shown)


source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -211,6 +211,32 @@ TEST_F(NiRFmxBTDriverApiTests, TxpBasicFromExample_DataLooksReasonable)
   EXPECT_GT(fetched_powers_response.peak_to_average_power_ratio_maximum(), 0.0);
 }
 
+TEST_F(NiRFmxBTDriverApiTests, SetAndGetAttributeInt32_Succeeds)
+{
+  auto session = init_session(stub(), kPxi5663e);
+  EXPECT_SUCCESS(session, client::cfg_frequency_reference(stub(), session, "", FrequencyReferenceSource::FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK, 10e6));
```
tdunkle, line 217: Since this setup is the same routine as in `AcpBasicFromExample_DataLooksReasonable()`, I want to move the routine to a helper function. However, I have been running into several issues with doing so. Even on successful builds, the test fails with this error when I use a helper function:

`-380413: Invalid Acquisition State Transition.\r\n\nniRFInstr SA HW UTIL Transition Hardware State.vi <append>\nRFmx driver cannot transition to the  Configuring  state from the Acquiring state. The Acquisition must complete or Aborted.\n"`

I think there's probably a way to do this, but after talking to Alex more I want to submit what I have, get VSCode and ninja set up so that I can build more efficiently, and then I'll create a separate task to come back and single-source this setup routine.

astarche: I'd expect this to look more like:

```cpp

  constexpr auto ATTR_ID = NiRFmxBTAttribute::NIRFMXBT_ATTRIBUTE_TXP_WHATEVER;
  constexpr auto ATTR_VAL = WHATEVER;
  auto session = init_session(stub(), kPxi5663e);
  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MeasurementTypes::MEASUREMENT_TYPES_TXP, Boolean::BOOLEAN_TRUE));
  EXPECT_SUCCESS(
      session,
      client::set_attribute_i32(stub(), session, "", ATTR_ID, ATTR_VAL));

  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));

  auto get_response = client::get_attribute_i32(stub(), session, "", ATTR_ID);

  EXPECT_SUCCESS(session, get_response);
  EXPECT_EQ(ATTR_VAL, get_response.attr_val());

```

1. Remove extra configuration.
2. Call set_attribute before initiate and get_attribute after
3. Get/Set an attribute that matches the measurement type (TXP not ACP)
4. (I don't really care about the constants declared at the top; that's just to get the WHATEVER's in one place)

Does something like that work?

astarche (changes requested):
I want so see if the simpler version works or get more info on why not.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: Ugh, it took me way too long to realize that I was mixing up the measurement types. I think that was the root of my difficulty with this change, and originated from a copy/paste error on my part.

Yeah, I tried implementing your suggested approach, and the test passes!

reckenro (approved):

PR #510, opened 2022-02-05
Create examples for NR and SpecAn

DavidCurtiss:

Created desired python examples for NR and SpecAn, based on the C examples

[AB#1781498](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1781498), [AB#1781499](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1781499), [AB#1825336](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1825336), [AB#1825337](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1825337)

black+mypy, and ran them on my test VM

### DIFF
--- a/examples/nirfmxnr/acp-single-carrier.py
+++ b/examples/nirfmxnr/acp-single-carrier.py
@@ -69,11 +69,14 @@ instr = None
 def raise_if_error(response):
     if response.status != 0:
         error_response = client.GetError(
-            nirfmxnr_types.GetErrorRequest(
+            nirfmxwlan_types.GetErrorRequest(
                 instrument=instr,
             )
         )
-        raise RuntimeError(f"Error: {error_response.error_description or response.status}")
+        if response.status < 0:
+            raise RuntimeError(f"Error: {error_response.error_description or response.status}")
+        else:
+            sys.stderr.write(f"Warning: {error_response.error_description or response.status}\n")
 
     return response
 
--- a/examples/nirfmxnr/acp-single-carrier.py
+++ b/examples/nirfmxnr/acp-single-carrier.py
@@ -1,0 +1,309 @@
+# Steps:
+# 1. Open a new RFmx Session.
+# 2. Configure Frequency Reference.
+# 3. Configure Selected Ports.
+# 4. Configure basic signal properties (Center Frequency, RF Attenuation and External Attenuation).
+# 5. Configure Trigger Parameters for IQ Power Edge Trigger.
+# 6. Configure Link Direction, Frequency Range and Carrier Bandwidth and Subcarrier Spacing.
+# 7. Configure Reference Level.
+# 8. Select ACP measurement and enable Traces.
+# 9. Configure Measurement Method.
+# 10. Configure Noise Compensation Parameter.
+# 11. Configure Sweep Time Parameters.
+# 12. Configure Averaging Parameters for ACP measurement.
+# 13. Initiate the Measurement.
+# 14. Fetch ACP Measurements and Traces.
+# 15. Close RFmx Session.
+
+import sys
+
+import grpc
+import nirfmxnr_pb2 as nirfmxnr_types
+import nirfmxnr_pb2_grpc as grpc_nirfmxnr
+
+server_address = "localhost"
+server_port = "31763"
+session_name = "NI-RFSASession"
+
+# Resource name and options for a simulated 5663 client.
+resource = "SimulatedRFSA"
+options = "Simulate=1,DriverSetup=Model:5663"
+
+# Read in cmd args
+if len(sys.argv) >= 2:
+    server_address = sys.argv[1]
+if len(sys.argv) >= 3:
+    server_port = sys.argv[2]
+if len(sys.argv) >= 4:
+    resource = sys.argv[3]
+    options = ""
+
+# Create a gRPC channel + client.
+channel = grpc.insecure_channel(f"{server_address}:{server_port}")
+client = grpc_nirfmxnr.NiRFmxNRStub(channel)
+instr = None
+
+
+# Maximum size of an error message
+MAX_ERROR_DESCRIPTION = 4096
+
+
+# Raise an exception if an error was returned
+def raise_if_error(response):
+    if response.status != 0:
+        error_response = client.GetError(
+            nirfmxspecan_types.GetErrorRequest(
+                instrument=instr,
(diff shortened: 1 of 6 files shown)


examples/nirfmxnr/acp-single-carrier.py
```diff
@@ -0,0 +1,309 @@
...
+
+import sys
+
+import grpc
+import nirfmxnr_pb2 as nirfmxnr_types
+import nirfmxnr_pb2_grpc as grpc_nirfmxnr
+
+server_address = "localhost"
+server_port = "31763"
+session_name = "NI-RFSASession"
```
astarche, line 26: we should change up these throwaway names to avoid RFSA (also "SimulatedRFSA" below)

Good options:
* `RFmxNRSession`
* `ACPSingleCarrierSession`
* `SimulatedDevice`
* `Simulated5663`

Different string-casing conventions are also fine.

astarche (approved):
Looks really good overall!

Feel free to reopen me to check over any changes.

examples/nirfmxspecan/acp-basic.py
```diff
@@ -0,0 +1,322 @@
...
+        )
+    )
+    raise_if_error(
+        client.ACPCfgOffsetArray(
+            nirfmxspecan_types.ACPCfgOffsetArrayRequest(
+                instrument=None,
+                selector_string="",
+                offset_frequency=[1.0e6, 2.0e6],
+                offset_sideband=None,
+                offset_enabled=None,
```
DavidCurtiss, line 246: This call (mirrored from the C example, which passes in NULL for the last two parameters) errors with:
"The sizes of linked repeated fields [offsetFrequency, offsetSideband, offsetEnabled] do not match"
Filed [AB#1836020](https://ni.visualstudio.com/DevCentral/_workitems/edit/1836020).
(But now I'm going to remove this offending example, since I just realized I copied from the Acp example instead of the AcpBasic example.)
(FYI: @astarche)

astarche: I think we want to use this mechanism: #456. Added a comment to the bug. We'll have to double-check the docs that it's really implemented the same way.

These should not be `None`, by the way. In `protobuf` the arrays for repeated fields can be empty but not null, so it's confusing to appear to be passing `None` (granted maybe it does the same thing). Either passing `[]` or not passing the argument at all would be fine. I don't think there's ever a case where we want to pass `None`/null to a protobuf/grpc stub.

examples/nirfmxspecan/acp-basic.py
```diff
@@ -0,0 +1,322 @@
...
+                session_name=session_name, resource_name=resource, option_string=""
+            )
+        )
+    )
+    instr = initialize_response.instrument
+
+    raise_if_error(
+        client.CfgFrequencyReference(
+            nirfmxspecan_types.CfgFrequencyReferenceRequest(
+                instrument=None,
```
astarche, line 105: These should all be `instrument=instr`.

examples/nirfmxnr/acp-single-carrier.py
```diff
@@ -0,0 +1,325 @@
...
+    return response
+
+
+try:
+    auto_level = True
+
+    initialize_response = raise_if_error(
+        client.Initialize(
+            nirfmxnr_types.InitializeRequest(
+                session_name=session_name, resource_name=resource, option_string=""
```
mgumble, line 87: Pass options into option_string we defined about

examples/nirfmxnr/acp-single-carrier.py
```diff
@@ -0,0 +1,325 @@
...
+            )
+        )
+    )
+    raise_if_error(
+        client.SetAttributeString(
+            nirfmxnr_types.SetAttributeStringRequest(
+                instrument=instr,
+                selector_string="",
+                attribute_id=nirfmxnr_types.NIRFMXNR_ATTRIBUTE_SELECTED_PORTS,
+                attr_val_raw="",
```
mgumble, line 108: I would image attr_val is also a string here?

examples/nirfmxnr/acp-single-carrier.py
```diff
@@ -0,0 +1,325 @@
...
+            nirfmxnr_types.InitializeRequest(
+                session_name=session_name, resource_name=resource, option_string=""
+            )
+        )
+    )
+    instr = initialize_response.instrument
+    raise_if_error(
+        client.CfgFrequencyReference(
+            nirfmxnr_types.CfgFrequencyReferenceRequest(
+                instrument=None,
```
mgumble, line 95: @astarche  mentioned it but yest instrument=instr everywhere not sure there is a case where None makes sense

mgumble (changes requested):
We need to ensure these can actually run on a system first, once we have done that I am okay submitting these

examples/nirfmxnr/acp-single-carrier.py
DavidCurtiss: SetAttributeString allows either `attr_val_raw` or `attr_val_mapped`, wherein the latter is a mapped value like FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK (whose value is 1). There's no `attr_val` for string attributes.

source/codegen/generate_shared_service_files.py
```diff
@@ -7,7 +7,7 @@ def generate_register_all_services(metadata_dir: Path, output_dir: Path) -> None
   driver_modules = [
     load_metadata(p)
     for p in sorted(metadata_dir.iterdir())
-    if p.is_dir() and not "fake" in p.name
+    if p.is_dir() and not "fake" in p.name and not p.name.startswith(".")
```
reckenro, line 10: Curious: what is this change about?

reckenro (approved):

source/codegen/generate_shared_service_files.py
DavidCurtiss: Oops, I didn't mean for this to go into this change. I'll revert it.

I made the change, though, because I had a .mypy_cache folder in the same path.

mgumble (approved):

PR #512, opened 2022-02-08
Add system-level tests of complex and unused numeric data types to NiRFmxBT service.

tdunkle:

This pull request adds system-level tests that set attributes with complex numeric data types, and with 8- and 16-bit integer data types. All of these data types are not present in `attributes.py`, but these tests at least exercise the code.
Fixes [AB#1834560](https://ni.visualstudio.com/DevCentral/_workitems/edit/1834560).

We want to be certain that it is theoretically possible to set attributes of these data types in order to be more confident in the quality of the Bluetooth service.

The new tests build locally and pass on a VM with NI-RFSA and NI-RFmx Bluetooth installed:

image

### DIFF
--- a/source/tests/system/nirfmxbt_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxbt_driver_api_tests.cpp
@@ -133,8 +133,8 @@ TComplex complex(TFloat real, TFloat imaginary)
 
 template <typename TFloat, typename TComplex>
 std::vector<TComplex> complex_array(
-    std::vector<TFloat> reals,
-    std::vector<TFloat> imaginaries)
+    const std::vector<TFloat> reals,
+    const std::vector<TFloat> imaginaries)
 {
   auto c = std::vector<TComplex>{};
   c.reserve(reals.size());
@@ -148,8 +148,8 @@ std::vector<TComplex> complex_array(
 }
 
 std::vector<nidevice_grpc::NIComplexNumber> complex_number_array(
-    std::vector<double> reals,
-    std::vector<double> imaginaries)
+    const std::vector<double> reals,
+    const std::vector<double> imaginaries)
 {
   return complex_array<double, nidevice_grpc::NIComplexNumber>(reals, imaginaries);
 }
@@ -258,7 +258,7 @@ TEST_F(NiRFmxBTDriverApiTests, TxpBasicFromExample_DataLooksReasonable)
   EXPECT_GT(fetched_powers_response.peak_to_average_power_ratio_maximum(), 0.0);
 }
 
-// Note: there aren't any complex attributes in attributes.py, but this at least exercises the code.
+// TODO -- AB#1835516: Add system-level test using an API function that uses complex data types.
 TEST_F(NiRFmxBTDriverApiTests, SetAttributeComplex_ExpectedError)
 {
   const auto session = init_session(stub(), kPxi5663e);
--- a/source/tests/system/nirfmxbt_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxbt_driver_api_tests.cpp
@@ -7,6 +7,7 @@
 #include "nirfmxbt/nirfmxbt_service.h"
 #include "nirfsa/nirfsa_client.h"
 
+namespace pb = google::protobuf;
 using namespace ::testing;
 using namespace nirfmxbt_grpc;
 namespace client = nirfmxbt_grpc::experimental::client;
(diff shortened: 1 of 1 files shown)


source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -211,6 +258,48 @@ TEST_F(NiRFmxBTDriverApiTests, TxpBasicFromExample_DataLooksReasonable)
   EXPECT_GT(fetched_powers_response.peak_to_average_power_ratio_maximum(), 0.0);
 }
 
+// Note: there aren't any complex attributes in attributes.py, but this at least exercises the code.
+TEST_F(NiRFmxBTDriverApiTests, SetAttributeComplex_ExpectedError)
```
astarche, line 262: RFmxBT has several APIs that actually use complex numbers: `AnalyzeIQ1Waveform`, `ModAccFetchConstellationTrace`. We should write an example-based test that uses at least one of those. @DavidCurtiss may be able to help generate those. And, if they use TDMS files, @reckenro was looking at pulling out some of that code/data for a different test.

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -211,6 +258,48 @@ TEST_F(NiRFmxBTDriverApiTests, TxpBasicFromExample_DataLooksReasonable)
...
+{
+  const auto session = init_session(stub(), kPxi5663e);
+
+  EXPECT_RFMX_ERROR(
+      -380231, "This attribute is read-only and cannot be written", session,
+      client::set_attribute_ni_complex_double_array(stub(), session, "", NiRFmxBTAttribute::NIRFMXBT_ATTRIBUTE_ACP_RESULTS_REFERENCE_CHANNEL_POWER, complex_number_array({1.2, 2.2}, {1e6, 1.01e6})));
+}
+
+// Note: there aren't any i8 attributes in attributes.py, but this at least exercises the code.
+TEST_F(NiRFmxBTDriverApiTests, SetAttributeInt8_ExpectedError)
```
astarche, line 272: RFmxBT *does not* have any methods with narrow numerics other that the attribute accesors. So you don't need any other tests for these.  I don't think that we need to add these "identical attribute accessors that do nothing; do nothing" tests for every personality. But it's fine to add these. Also double-check that the service implementation for the setters is doing the same coercion code that specan does.

astarche (approved):
These tests are good but don't cover what we need for `NIComplex` tests. If you add those to this PR, you can reopen me to review them. Or you can submit this without the NIComplex example based tests and make a separate PR.

source/tests/system/nirfmxbt_driver_api_tests.cpp
reckenro: It's actually @danestull who is working with TDMS -> JSON files along with a test that uses complex numbers. He posted a PR for the tests using the JSON waveform files he converted from TDMS here: #513

reckenro (approved):

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: Good call. I've created a new work item for this: [AB#1835516](https://ni.visualstudio.com/DevCentral/_workitems/edit/1835516).

source/tests/system/nirfmxbt_driver_api_tests.cpp
```diff
@@ -107,6 +122,38 @@ nirfsa_grpc::InitWithOptionsResponse init_rfsa(const nirfsa_client::StubPtr& stu
...
+{
+  auto c = TComplex{};
+  c.set_real(real);
+  c.set_imaginary(imaginary);
+  return c;
+}
+
+template <typename TFloat, typename TComplex>
+std::vector<TComplex> complex_array(
+    std::vector<TFloat> reals,
```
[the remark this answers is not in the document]
tdunkle: I updated these parameters, as well as the ones in `complex_number_array()`, so that the parameters are const.

source/tests/system/nirfmxbt_driver_api_tests.cpp
tdunkle: It appears that the Bluetooth service performs the same coercion as the SpecAn service. For example:
[Bluetooth casts to I8](https://github.com/ni/grpc-device/blob/b10041818bfa7f4d4d2ec28e17d83c97568e674c/generated/nirfmxbt/nirfmxbt_service.cpp#L3527) the same way that [SpecAn casts to I8](https://github.com/ni/grpc-device/blob/main/generated/nirfmxspecan/nirfmxspecan_service.cpp#L12172).

PR #515, opened 2022-02-08
Add tests for IQFetchData

astarche:

Add tests for IQFetchData "delete record on fetch" behavior.

Fixes [AB#1784631](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1784631).

The default "delete record on fetch" behavior for IQFetchData is a gotcha in the C API, especially if you don't read the entire record. Adding test coverage for the common workarounds:
1. Fetch all available by passing `-1` for `samples_to_read`.
2. Set the `NIRFMXSPECAN_ATTRIBUTE_IQ_DELETE_RECORD_ON_FETCH` attribute before fetching.

### Alternative

We could expose a way to override the default as an additional field in the fetch request message. Reason rejected: this can be added easily later. Existing behavior is good enough and is consistent with the C API.

Ran and passed all IQWFetchData tests.

### DIFF
--- a/source/tests/system/nirfmxspecan_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxspecan_driver_api_tests.cpp
@@ -79,7 +79,7 @@ class NiRFmxSpecAnDriverApiTests : public ::testing::Test {
   }
 
   template <typename TResponse>
-  void EXPECT_SUCCESS(const nidevice_grpc::Session& session, const TResponse& response)
+  TResponse EXPECT_SUCCESS(const nidevice_grpc::Session& session, const TResponse& response)
   {
     ni::tests::system::EXPECT_SUCCESS(response);
     if (response.status() != SUCCESS) {
@@ -87,6 +87,8 @@ class NiRFmxSpecAnDriverApiTests : public ::testing::Test {
       EXPECT_EQ("", error_message_response.error_description());
     }
     check_error(session);
+
+    return response;
   }
 
   template <typename TResponse>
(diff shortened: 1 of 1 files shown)


astarche:
image

source/tests/system/nirfmxspecan_driver_api_tests.cpp
```diff
@@ -445,6 +447,52 @@ TEST_F(NiRFmxSpecAnDriverApiTests, BuildSpurString_ReturnsSpurString)
...
+  constexpr auto FETCH_ALL_AVAILABLE = -1;
+  constexpr auto EXPECTED_RECORD_COUNT = 50000;
+  const auto session = init_session(stub(), PXI_5663);
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MeasurementTypes::MEASUREMENT_TYPES_IQ, true));
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  const auto fetch_response = EXPECT_SUCCESS(session, client::iq_fetch_data(stub(), session, "", 10.0, 0, FETCH_ALL_AVAILABLE));
+
+  EXPECT_THAT(fetch_response.data(), Not(IsEmpty()));
+  EXPECT_EQ(EXPECTED_RECORD_COUNT, fetch_response.data().size());
```
mgumble, line 479: Question: Why is this 50000?

source/tests/system/nirfmxspecan_driver_api_tests.cpp
```diff
@@ -445,6 +447,52 @@ TEST_F(NiRFmxSpecAnDriverApiTests, BuildSpurString_ReturnsSpurString)
...
+}
+
+TEST_F(NiRFmxSpecAnDriverApiTests, DisableDeleteRecordOnFetch_IQFetchData_RecordIsStillAvailable)
+{
+  const auto session = init_session(stub(), PXI_5663);
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MeasurementTypes::MEASUREMENT_TYPES_IQ, true));
+  EXPECT_SUCCESS(session, client::set_attribute_i32(stub(), session, "", NiRFmxSpecAnAttribute::NIRFMXSPECAN_ATTRIBUTE_IQ_DELETE_RECORD_ON_FETCH, 0));
```
mgumble, line 486: Note: What an odd int32 attribute. Do we not have an obvious enum to link to bool? Wouldn't it be  enable/disable record

mgumble (approved):

source/tests/system/nirfmxspecan_driver_api_tests.cpp
astarche: That's just the amount of data you get in the default configuration. I think it's consistent. Really what I care about is that it's a reasonable number that's bigger than the arbitrarily truncated reads in the other tests (`1000`).

reckenro (approved):

source/tests/system/nirfmxspecan_driver_api_tests.cpp
astarche: Yeah, it does have a TRUE/FALSE enum. `i32` is the default for all enums (including T/F) and passing `1` is just a bit less typing.

I updated to use the enum though, good catch.

PR #520, opened 2022-02-09
RFmxSpecAn API fixups

### LINKED ISSUE #505: RFmxSpecan BuildIntermodStringRequest has outputs in request
```proto
message BuildIntermodStringRequest {
  string selector_string = 1;
  int32 intermod_number = 2;
  int32 selector_string_out_length = 3;
  string selector_string_out = 4;
}

message BuildIntermodStringResponse {
  int32 status = 1;
}
```

astarche:

Fixes #505

Fixes [AB#1812491](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1812491)

Fix mistakes and inconsistency in `RFmxSpecAn`.

Ran and passed new `BuildIntermodString` tests and all `RFmxInstr` tests.

### DIFF
--- a/generated/nirfmxinstr/nirfmxinstr.proto
+++ b/generated/nirfmxinstr/nirfmxinstr.proto
@@ -937,7 +937,7 @@ message GetNIRFSASessionRequest {
 
 message GetNIRFSASessionResponse {
   int32 status = 1;
-  nidevice_grpc.Session ni_rfsa_session = 2;
+  nidevice_grpc.Session nirfsa_session = 2;
 }
 
 message GetNIRFSASessionArrayRequest {
--- a/source/codegen/metadata/nirfmxspecan/functions.py
+++ b/source/codegen/metadata/nirfmxspecan/functions.py
@@ -2209,8 +2209,12 @@ functions = {
                 'type': 'int32'
             },
             {
-                'direction': 'in',
+                'direction': 'out',
                 'name': 'selectorStringOut',
+                'size': {
+                    'mechanism': 'ivi-dance',
+                    'value': 'selectorStringOutLength'
+                },
                 'type': 'char[]'
             }
         ],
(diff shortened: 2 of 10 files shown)


source/codegen/metadata/nirfmxspecan/functions.py
```diff
@@ -2209,8 +2209,12 @@
                 'type': 'int32'
```
mgumble, line 2209: Shouldn't selectorStringOutLength be direction out too?

generated/nirfmxinstr/nirfmxinstr.proto
```diff
@@ -937,7 +937,7 @@ message GetNIRFSASessionRequest {
 
 message GetNIRFSASessionResponse {
   int32 status = 1;
-  nidevice_grpc.Session ni_rfsa_session = 2;
+  nidevice_grpc.Session nirfsa_session = 2;
```
mgumble, line 940: Was this the only thing wrong with spelling? or will there be more to the task>?

mgumble (approved):

source/codegen/metadata/nirfmxspecan/functions.py
reckenro: No, it's an input to the Driver API method. These ivi-dance ones get the size from the first call where a positive status indicates the size to use.

1. Call API with 0 for `selectorStringOutLength`, returned status (if positive) is the size to use
2. Call API again with status in step one passed in for `selectorStringOutLength`

reckenro (approved):

generated/nirfmxinstr/nirfmxinstr.proto
astarche: Good question. Everything else was tracked in separate tasks, especially #465, which David has been working on (I think he's got all of the ones listed in the issue done; but keeps finding more stuff and it's been combined with work for different personalities).

I already did the manual proto review as AB#1781514, which added some entries to bug 465. Though I'll probably have to do at least one more pass and maybe I'll find something else.

PR #526, opened 2022-02-10
Add OutputTerminal enum to grpc-device for RFmxInstr

astarche:

Add `OutputTerminal` enum to grpc-device for `RFmxInstr`.

Fixup references to string enums for `DigitalEdgeTriggerSource.

Fixes [AB#1799610](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1799610)

String enums make it easier to set correct values for string params and attributes.

Using a single `OutputTerminal` and `DigitalEdgeTriggerSource` rather than one for each attribute/param because adding the extra enums doesn't add much value. Also, these are constants in C/.NET, so a single shared enum is a more similar implementation.

Ran and passed all `RFmxInstr` tests.


astarche:
image

reckenro (approved):

mgumble (approved):

PR #540, opened 2022-02-15
Update raise_if_error to use the correct package when calling getErrorRequest

mgumble:

Copy paste error where we were using the wlan type in specan and nr

Throwing error method throws its own error
 -> Running example with simulated hardware: acp-single-carrier.py
Reference level (dBm)        : 22.5
Traceback (most recent call last):
  File "build\validate_examples\examples\nirfmxnr\acp-single-carrier.py", line 270, in <module>
    acp_fetch_offset_measurement_array_response = raise_if_error(
  File "build\validate_examples\examples\nirfmxnr\acp-single-carrier.py", line 72, in raise_if_error
    nirfmxwlan_types.GetErrorRequest(
NameError: name 'nirfmxwlan_types' is not defined
 -> Running example with real hardware: txp-single-carrier.py

Now correctly throws error

### DIFF
--- a/examples/nirfmxnr/acp-single-carrier.py
+++ b/examples/nirfmxnr/acp-single-carrier.py
@@ -69,7 +69,7 @@ instr = None
 def raise_if_error(response):
     if response.status != 0:
         error_response = client.GetError(
-            nirfmxwlan_types.GetErrorRequest(
+            nirfmxnr_types.GetErrorRequest(
                 instrument=instr,
             )
         )
--- a/examples/nirfmxnr/txp-single-carrier.py
+++ b/examples/nirfmxnr/txp-single-carrier.py
@@ -65,7 +65,7 @@ instr = None
 def raise_if_error(response):
     if response.status != 0:
         error_response = client.GetError(
-            nirfmxwlan_types.GetErrorRequest(
+            nirfmxnr_types.GetErrorRequest(
                 instrument=instr,
             )
         )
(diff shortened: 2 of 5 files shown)


DavidCurtiss (approved):

reckenro (approved):

examples/nirfmxnr/acp-single-carrier.py
```diff
@@ -69,7 +69,7 @@
 def raise_if_error(response):
     if response.status != 0:
         error_response = client.GetError(
-            nirfmxwlan_types.GetErrorRequest(
+            nirfmxnr_types.GetErrorRequest(
```
astarche, line 72: It's interesting/frustrating that `mypy` didn't catch these.

astarche (approved):

DavidCurtiss:
> @DavidCurtiss does this need to be fixed in your script?

If I'm to generate any more tests with it, yes. Except I looked in the script and it's already fixed?! I must've fixed that after I last generated the code, or maybe I did a bad merge (with an earlier version of my code) or something.

examples/nirfmxnr/acp-single-carrier.py
DavidCurtiss: IIRC, mypy only checks functions that are annotated, so we'd need to put in typing info for the `response` parameter and/or the return value.

PR #541, opened 2022-02-15
Add support for repeated structs and grpc_name in optional array inputs

astarche:

Add support for repeated structs in optional array inputs.
* Add the `nullable_vector` class so that we can (a) assign `nullptr` to the vector and (b) so that `data` returns `nullptr` in the `is_null_` state. This allows structs to work without changing any of the codegen.
* Add `metadata_validation` to disallow `optional` params used as `two-dimension` size params.

Add support for custom `grpc_name` in linked array input code:
* This is required in cases where one or more of the linked params has a `grpc_name`. It requires that we look up the `grpc_name` from the parameter list.

I'm marking all "linked array inputs" as `optional` in the RFmx services, because that matches how those params are handled in the .NET API and is required as a feature for many methods. One of these methods is a repeated `NIComplexSingle` and many have custom `grpc_name`s.

See [AB#1836020](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1836020).

Ran and passed all `RFmx` tests and all unit tests.

Builds with `optional` arrays added to several `RFmx` services.


generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -293,7 +293,7 @@ namespace nidaqmx_grpc {
       const auto array_size_size_calculation = calculate_linked_array_size(array_size_determine_from_sizes, false);
 
       if (array_size_size_calculation.match_state == MatchState::MISMATCH) {
-        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The sizes of linked repeated fields [triggerSlopeArray, triggerLevelArray] do not match");
+        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The sizes of linked repeated fields [trigger_slope_array, trigger_level_array] do not match");
```
astarche, line 296: Bonus: these exceptions now have the correct names.

generated/nifake/nifake_service.cpp
```diff
@@ -1451,28 +1451,31 @@ namespace nifake_grpc {
...
       if (size_size_calculation.match_state == MatchState::MATCH_OR_ZERO) {
-        values1 = request->values1_size() ? values1 : nullptr;
-        values2 = request->values2_size() ? values2 : nullptr;
-        values3 = request->values3_size() ? values3 : nullptr;
-        values4 = request->values4_size() ? values4 : nullptr;
+        values1 = request->values1_size() ? std::move(values1) : nullptr;
+        values2 = request->values2_size() ? std::move(values2) : nullptr;
+        values3 = request->values3_size() ? std::move(values3) : nullptr;
+        values4 = request->values4_size() ? std::move(values4) : nullptr;
+        values5 = request->values5_size() ? std::move(values5) : nullptr;
```
astarche, line 1474: This is the service implementation for the new field. The "trick" is that `nullable_vector` makes it so nothing has to change here even though `values1-4` are pointers into request data and `values5` is a stack allocated `vector` (stored in a `std::vector` now wrapped in a `nullable_vector`).

reckenro (approved):

DavidCurtiss:
From description: "See AB#" -- I think you meant to link to [AB#1836020](https://ni.visualstudio.com/DevCentral/_workitems/edit/1836020).

source/codegen/metadata/nifake/functions.py
```diff
@@ -1984,6 +1984,22 @@
                 },
                 'type': 'ViReal64[]'
             },
+            {
```
reckenro, line 1987: After seeing the scrapigen review, I'd like to see marking the `UseATwoDimensionParameter` function's `arrayLengths` parameter as optional and have a test in relation to the two-dimension size mechanism.

astarche: Added a metadata_validation step to ensure that two-dimension and optional are not used together.

reckenro (approved):

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -1000,24 +1007,38 @@ TEST(NiFakeServiceTests, NiFakeService_MultipleArraysSameSize_OneArrayWithZeroSi
...
   NiFakeMockLibrary library;
   auto resource_repository = std::make_shared<FakeResourceRepository>(&session_repository);
   nifake_grpc::NiFakeService service(&library, resource_repository);
   auto session_id = create_session(library, service, kTestViSession);
-  const std::vector<double> doubles = {0.2, -2.3, 4.5};
+  const auto doubles = std::vector<double>{0.2, -2.3, 4.5};
+  const auto fake_structs = std::vector<FakeCustomStruct>{
+      create_custom_struct(1, 3.2),
+      create_custom_struct(0, 10.0),
+      create_custom_struct(100, 0.)};
```
mgumble, line 1029: observation: ive never seen "0." obviously it is compiling as a double, but I would expected 0.0

mgumble (approved):

source/tests/unit/ni_fake_service_tests.cpp
astarche: Maybe that's too cute. 

From [cppreference ](https://en.cppreference.com/w/c/language/floating_constant):

> For decimal floating constants, the exponent part is optional. If it is omitted, the period is not optional, and either the whole-number or the fraction must be present.

astarche: I'm going to keep it to keep the PRs moving.

PR #542, opened 2022-02-15
Add missing attribute and related enum values for NR.

reckenro:

Adds missing `PDSCH_PRESENT_IN_SSB_RESOURCE_BLOCK` attribute and related enum values to NR.

API cleanliness.
Fixes [AB#1799692](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1799692)

image


generated/nirfmxnr/nirfmxnr.proto
```diff
@@ -350,6 +350,7 @@ enum NiRFmxNRAttribute {
   NIRFMXNR_ATTRIBUTE_SSB_PERIODICITY = 9437332;
   NIRFMXNR_ATTRIBUTE_SSB_ACTIVE_BLOCKS = 9437333;
   NIRFMXNR_ATTRIBUTE_GRID_START = 9437334;
+  NIRFMXNR_ATTRIBUTE_PDSCH_PRESENT_IN_SSB_RESOURCE_BLOCK = 9437335;
```
astarche, line 353: These are in the `obsolete` header/json files for NR. I would think that they don't need to be included, right?

astarche (changes requested):
await: aren't these obsolete?

generated/nirfmxnr/nirfmxnr.proto
reckenro: Ah, oops. I was still going off the header and it's not in the obsolete section there and it's present in the CHM. But, I couldn't find it in any examples so it seems reasonable to leave it out. I'll abandon this PR and the one in scrapigen.

reckenro: It is weird it's not in the obsolete section of the [header file](https://p4.natinst.com/browser/perforce/RfProtocols/RFmx/NR/CSupport/trunk/21.5/source/library/mx/niRFmxNR.h#line_141) though. 

I didn't find a reference to it in any examples though so I think it's still the right call to leave it out of the proto, but I made a task in the RFmx sync up story to make sure since there is a discrepancy: AB#1848496

astarche: Yeah, doesn't hurt to double check, but it does show up obsolete in 2 places, which is pretty convincing to me. It's also easier to add (or even workaround) than remove, not that a spare attr is the end of the world. You could also check the help and/or .NET API.

PR #548, opened 2022-02-16
Check untyped defs in validate examples

astarche:

Pass `--check-untyped-defs` to `mypy` in `validate_examples`.

Add `validate_examples` for `*rfmx*` and `*xnet*` to github workflow.

Add ability to run `validate_examples` without a server address and skip the run-time check.

Ensure that we catch issues like #540 in services that are under development.
* Note that shipping drivers don't all pass `mypy` for various reasons, but are less likely to be broken for API changes.

Ran and passed `python source\codegen\validate_examples.py -p *rfmx*`.

Manual testing of code with bogus references similar to the fixed cases in #540 and ensure they pass.

Tested validate_examples in failed workflow.

### DIFF
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -1,0 +1,39 @@
+name: Validate Python Code
+
+on:
+  push:
+    branches:
+      - main
+      - 'releases/**'
+  pull_request:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repo
+      uses: actions/checkout@v2
+
+    - name: Setup python3
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Install build dependencies
+      run: python -m pip install -r python_build_requirements.txt
+
+    - name: Validate python style with black
+      run: |
+        black source/codegen examples --check
+
+    - name: Install poetry for validate_examples.py
+      uses: abatilo/actions-poetry@v2.0.0
+      with:
+        poetry-version: 1.1.5
+
+    - name: Validate RFmx and XNET examples
+      run: |
+        python source/codegen/validate_examples.py -p *rfmx*
+        python source/codegen/validate_examples.py -p *xnet*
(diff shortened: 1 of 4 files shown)


source/codegen/validate_examples.py
```diff
@@ -40,18 +40,19 @@ def validate_examples(driver_glob_expression: str, ip_address: str, device_name:
...
 
             print(f" -> Running mypy")
-            system(f"poetry run mypy {dir}")
+            system(f"poetry run mypy {dir} --check-untyped-defs")
 
-            for file in dir.glob("*.py"):
-                print(f" -> Running example: {file.name}")
-                PORT = 31763
-                system(rf"poetry run python {file} {ip_address} {PORT} {device_name}")
+            if ip_address:
```
mgumble, line 51: We should add a argument to run simulation too. makes sense to not run them for build right now, but in general a workflow that can easily test simulated (by not passing in the ip_address).

astarche: We run simulated by not passing in a `device_name` (the examples will use simulated when there's no 3rd argument). Running on simulated still requires that there be a server running (with drivers installed if you want the example to pass), so you still need the `ip_address`. It just means that the server doesn't need to have any physical hardware.

mgumble (approved):

.github/workflows/validate_python.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    - name: Install poetry for validate_examples.py
+      uses: abatilo/actions-poetry@v2.0.0
+      with:
+        poetry-version: 1.1.5
+
+    - name: Validate RFmx and XNET examples
+      run: |
+        python source/codegen/validate_examples.py -p *rfmx*
+        python source/codegen/validate_examples.py -p *xnet*
```
DavidCurtiss, line 39: We probably oughta put the parameter in quotes, in case some file is added to the root directory that matches the pattern.

DavidCurtiss (approved):

.github/workflows/validate_python.yml
```diff
@@ -0,0 +1,39 @@
+name: Validate Python Code
```
SparkingSpork, line 1: Nitpick: Consider making this a separate [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) file and running it as part of the build(s) for PRs instead of separately. That way we don't run the full build(s) multiple times when there's invalid syntax and a user fixes it.

SparkingSpork (approved):

reckenro (approved):

PR #557, opened 2022-02-22
Fix type annotation in validate_examples for python <3.9

astarche:

Change `list` to `List` in type annotation in `validate_examples`.

Update `validate_python.yml` to use python `3.7.x` instead of `3.x`.

Allows us to use earlier versions of python, in particular the NI standard `3.7`.

Ran workflow with and without my change on my fork. Succeeding with fix and failing without.


.github/workflows/validate_python.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - name: Setup python3
       uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: '3.7.x'
```
astarche, line 17: Note: I had a surprisingly hard time getting my local build to use python `3.7` even with pyenv set up (I believe) correctly. I think this is something about the `FindPython` module in cmake (with pyenv? on windows?). Anyway, I confirmed this worked.

SparkingSpork: Yeah, I'm not sure how CMake searches for python by default. It's probably not aware of the pyenv directories. Maybe try `Python_ROOT_DIR` as a command line arg for CMake on your local machine and give it the pyenv path or similar?

SparkingSpork (approved):

mgumble (approved):

reckenro (approved):

.github/workflows/validate_python.yml
astarche: Yeah, that would probably work. But `pyenv local/global` set up your PATH so `python` is pointed to the pyenv dir, right? So it seemed more like `cmake` is "trying to hard".

PR #560, opened 2022-02-22
Add LTE session tests (cross-driver and session management).

reckenro:

Adds some tests covering cross driver sessions in LTE and general session management coverage we want per personality.

Fixes [AB#1804662](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1804662)

LTE system-level tests pass locally.

### DIFF
--- a/source/tests/system/nirfmxlte_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxlte_driver_api_tests.cpp
@@ -1,13 +1,14 @@
 #include <gtest/gtest.h>
 
 #include "device_server.h"
-#include "niRFmxLTE.h"
 #include "nirfmxlte/nirfmxlte_client.h"
+#include "nirfsa/nirfsa_client.h"
 #include "rfmx_macros.h"
 
 using namespace ::testing;
 using namespace nirfmxlte_grpc;
 namespace client = nirfmxlte_grpc::experimental::client;
+namespace nirfsa_client = nirfsa_grpc::experimental::client;
 
 namespace ni {
 namespace tests {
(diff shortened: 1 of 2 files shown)


reckenro:
Passing LTE tests:

image

source/tests/system/nirfmxlte_driver_api_tests.cpp
```diff
@@ -1,13 +1,14 @@
 #include <gtest/gtest.h>
 
 #include "device_server.h"
-#include "niRFmxLTE.h"
```
reckenro, line 4: I was going to explore this generally for all personalities but thought I'd go ahead and remove it in this PR for LTE. 

These headers shouldn't need to be included in the system level tests, right? The clients won't have access to them and we should just try to solely rely on the proto generated files for constants and things like that.

DavidCurtiss: Agreed. I just tried out removing it in my copy, where I'm adding the example-based tests, and I got it to work. I had to define float64, int32, and int64, and I had to replace some `_VAL_` constants. The latter is important, and is a good reason to stop including the C header.

DavidCurtiss (approved):

source/tests/system/nirfmxlte_driver_api_tests.cpp
```diff
@@ -54,12 +55,30 @@ InitializeResponse init(const client::StubPtr& stub, const std::string& model)
...
 {
   auto init_response = init(stub(), PXI_5663E);
   auto session = init_response.instrument();
-  // TODO: use EXPECT_SUCCESS here when LTE has get_error and get_error_string
-  EXPECT_RESPONSE_SUCCESS(init_response);
+  EXPECT_SUCCESS(session, init_response);
+
+  auto close_response = client::close(stub(), session, 0);
+
+  EXPECT_RESPONSE_SUCCESS(close_response);
```
reckenro, line 71: Note: In both tests, I'm using EXPECT_RESPONSE_SUCCESS for the `close_response` to match what's done for the similar tests in other personalities. 

The reasoning to use this over EXPECT_SUCCESS is captured in an [Instr driver test](https://github.com/ni/grpc-device/blob/85c6286076e8c42aa411eb40c67ee96710b4c053/source/tests/system/nirfmxinstr_driver_api_tests.cpp#L131): ""Don't use EXPECT_STATUS because that checks the error text, which can report stale errors from previous tests".

source/tests/system/nirfmxlte_driver_api_tests.cpp
DavidCurtiss: Agreed. I just tried out removing it in my copy, where I'm adding the example-based tests, and I got it to work. I had to define float64, int32, and int64, and I had to replace some `_VAL_` constants. The latter is important, and is a good reason to stop including the C header.

reckenro: Sounds good. I'll go through the other RFmx driver api tests and remove the headers in a separate PR.

astarche: We also shouldn't need service headers, if you catch those. I think the only "legitimate" use of the driver headers is for expected error codes. But I haven't seen an instance of that in a while and it's not worth the cost that David mentions.

astarche (approved):

PR #566, opened 2022-02-23
Add LTE examples

DavidCurtiss:

Adds python examples for LTE

[AB#1858415](https://ni.visualstudio.com/DevCentral/_workitems/edit/1858415)
[AB#1858416](https://ni.visualstudio.com/DevCentral/_workitems/edit/1858416)

```
(.venv) Z:\dev\grpc-device\examples\nirfmxlte>python acp-single-carrier.py
Reference level (dBm)        : 22.5

Carrier Absolute Power (dBm) : 18.58660888671875

-----------Offset Channel Measurements-----------
Offset : 0
Lower Relative Power (dB)    : -39.762245178222656
Upper Relative Power (dB)    : -39.762245178222656
Lower Absolute Power (dBm)   : -21.175636291503906
Upper Absolute Power (dBm)   : -21.175636291503906
------------------------------------------
Offset : 1
Lower Relative Power (dB)    : -39.50843048095703
Upper Relative Power (dB)    : -39.50843048095703
Lower Absolute Power (dBm)   : -20.92182159423828
Upper Absolute Power (dBm)   : -20.92182159423828
------------------------------------------
Offset : 2
Lower Relative Power (dB)    : -35.93755340576172
Upper Relative Power (dB)    : -35.93755340576172
Lower Absolute Power (dBm)   : -17.35094451904297
Upper Absolute Power (dBm)   : -17.35094451904297
------------------------------------------

(.venv) Z:\dev\grpc-device\examples\nirfmxlte>python ul-modacc-single-carrier.py
------------------Measurement------------------
Mean RMS Composite EVM (% or dB)     : 24.47988128392011
Max Peak Composite EVM (% or dB)     : 59.352272748947144
Peak Composite EVM Slot Index        : 0
Peak Composite EVM Symbol Index      : 6
Peak Composite EVM Subcarrier Index  : 310
Mean Frequency Error (Hz)            : 147.35012817382812
Mean IQ Origin Offset (dBc)          : -5.3845953941345215
Mean IQ Gain Imbalance (dB)          : 1.833428144454956
Mean IQ Quadrature Error (deg)       : 1.153283715248108
In Band Emission Margin (dB)         : -6.15078726092953
```

### DIFF
--- a/examples/nirfmxlte/acp-single-carrier.py
+++ b/examples/nirfmxlte/acp-single-carrier.py
@@ -1,0 +1,335 @@
+# Steps:
+# 1. Open a new RFmx Session.
+# 2. Configure Frequency Reference.
+# 3. Configure basic signal properties(Center Frequency, RF Attenuation and External Attenuation).
+# 4. Configure Trigger Type and Trigger Parameters.
+# 5. Configure Carrier Bandwidth.
+# 6. Configure Reference Level.
+# 7. Configure Duplex Mode.
+# 8. Configure Link Direction.
+# 9. Select ACP measurement and enable Traces.
+# 10. Configure Measurement Method.
+# 11. Configure Averaging Parameters for ACP measurement.
+# 12. Configure Sweep Time Parameters.
+# 13. Configure Noise Compensation Parameter.
+# 14. Initiate the Measurement.
+# 15. Fetch ACP Measurements and Traces.
+# 16. Close RFmx Session.
+#
+# The gRPC API is built from the C API. RFmx LTE documentation is installed with the driver at:
+# C:\Program Files (x86)\National Instruments\RFmx\LTE\Documentation\ltecvi.chm
+#
+# Getting Started:
+#
+# To run this example, install "RFmx LTE" on the server machine.
+# Link: https://www.ni.com/en-us/support/downloads/software-products/download.rfmx-lte.html
+#
+# For instructions on how to use protoc to generate gRPC client interfaces, see our "Creating a gRPC Client" wiki page.
+# Link: https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+#
+# Refer to the NI-RFmxLTE gRPC Wiki for the latest C Function Reference:
+# Link: https://github.com/ni/grpc-device/wiki/NI-RFmxLTE-C-Function-Reference
+#
+# Running from command line:
+#
+# Server machine's IP address, port number, and physical channel name can be passed as separate command line arguments.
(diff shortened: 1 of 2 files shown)


astarche (approved):

examples/nirfmxlte/acp-single-carrier.py
```diff
@@ -0,0 +1,335 @@
...
+        acp_fetch_absolute_powers_trace_response = raise_if_error(
+            client.ACPFetchAbsolutePowersTrace(
+                nirfmxlte_types.ACPFetchAbsolutePowersTraceRequest(
+                    instrument=instr, selector_string="", timeout=10.0, trace_index=i
+                )
+            )
+        )
+        absolute_powers_trace_results.append(
+            (
+                acp_fetch_absolute_powers_trace_response.x0,
```
mgumble, line 289: Why would we want to pull out the values of each result into a different list?

Similarly should we print these?

examples/nirfmxlte/ul-modacc-single-carrier.py
```diff
@@ -0,0 +1,289 @@
...
+    )
+
+    mod_acc_fetch_composite_evm_response = raise_if_error(
+        client.ModAccFetchCompositeEVM(
+            nirfmxlte_types.ModAccFetchCompositeEVMRequest(
+                instrument=instr, selector_string="", timeout=10.000000
+            )
+        )
+    )
+    mean_rms_composite_evm = mod_acc_fetch_composite_evm_response.mean_rms_composite_evm
```
mgumble, line 222: I feel like it would be cleaner to simply inline this when we print, it would at least reduce the line length and distance between these variables are actually used

mgumble (approved):

examples/nirfmxlte/acp-single-carrier.py
DavidCurtiss: I only printed the things that the C examples print. I decided to gather these other results into a list just so that the user can see that the data is there if they want it, even though we're not printing it. I'm fine with changing that if you disagree.

examples/nirfmxlte/ul-modacc-single-carrier.py
DavidCurtiss: This is similar to the other one: we don't actually print several of these extracted values; I'm just showing that they can be extracted from the response. In that respect, it makes more sense to me for them to be extracted next to where the response is received. But if we don't care about extracting those extra unprinted values, then I'll make this change, too.

mgumble: All of the ones from line 222-251 are being printed. I don't have a strong opinion either way though

PR #567, opened 2022-02-23
Create top-level ci.yml workflow

astarche:

Create top-level ci.yml workflow that calls other workflows.

* Simplifies `build_cmake` workflow and reduces tendency to glom steps there.
* All artifacts are associated with a single build.
   * Hopefully this makes it easier to automate release creation. At least it makes manual creation easier.
* Gives a nice build graph and top-level specification.

Various trial PRs on my fork including one creating artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,3 +26,5 @@ jobs:
      - build_nilrt
      - create_client_artifacts
     uses: ./.github/workflows/trigger_azdo_ci.yml
+    secrets:
+      AZDO_PIPELINE_TRIGGERS: ${{secrets.AZDO_PIPELINE_TRIGGERS}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: CI
+on:
+  push:
+    branches:
+      - main
+      - 'releases/**'
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  cancel_builds:
+    uses: ./.github/workflows/cancel_builds.yml
+  validate_python:
+    uses: ./.github/workflows/validate_python.yml
+  build_desktop:
+    needs: validate_python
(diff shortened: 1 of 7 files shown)


.github/workflows/trigger_azdo_ci.yml
```diff
@@ -0,0 +1,41 @@
+name: Trigger AzDO CIs
```
astarche, line 1: @tkrebes I moved this trigger stuff into a separate workflow. I think this will be a nice change, but I don't know that I can test it pre-submit. But, at least I can double-check if the azdo pipelines run once it goes in. Or I'm open to other ideas.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
+name: CI
```
astarche, line 1: Example run with artifacts (and a build failure for  a typo. oops!): https://github.com/astarche/grpc-device/actions/runs/1889272848
Example successful PR run: https://github.com/astarche/grpc-device/actions/runs/1866276664

DavidCurtiss (approved):

.github/workflows/trigger_azdo_ci.yml
astarche: Oops. Looking at the build results, maybe this doesn't work even on PR builds. I'll try to figure that out.

SparkingSpork (approved):
Very nice! I like being able to see everything on a single Action page and I think this will make changes/additions clearer in the future (e.g., adding an installer stage).

.github/workflows/trigger_azdo_ci.yml
astarche: The failure was because I missed a condition in one of the trigger actions, which caused it to try to trigger the workflow. This failed because PR builds do not have access to secrets. I think that's fixed now. And we'll have to see how it works once the change is in main.

mgumble (approved):

tkrebes (approved):

PR #570, opened 2022-02-28
[Draft] XNET gRPC APIs

ni-pkhare:

TODO: Include high-level description of the changes in this pull request.

TODO: Justify why this contribution should be part of the project.

TODO: Detail what testing has been done to ensure this submission meets requirements.


source/codegen/metadata/nixnet/config.py
```diff
@@ -0,0 +1,51 @@
...
+        'nxDatabaseRef_t[]': 'repeated nidevice_grpc.Session',
+        'u32': 'uint32',
+        'u64': 'uint64',
+        'u8': 'uint32',
+        'f64': 'double',
+        'char[]': 'string',
+        'const char[]': 'string',
+        'nxStatus_t': 'int32',
+        'void[]': 'bytes',
+        'nxTimestamp1ns_t': 'uint64',
```
doshirohan, line 26: This (and probably line 19 above) is causing linux build to fail right now. Timestamps are defined as `u64` in `nixnet.h`. `u64` is [defined as](https://p4.natinst.com/browser/perforce/build/exports/ni/xnet/xnetGlobals/official/export/21.5/21.5.0f186/includes/nixnet.h#line_3157) `unsigned int64` on windows and `unsigned long long int` on linux.

source/codegen/metadata/nixnet/config.py
```diff
@@ -0,0 +1,51 @@
...
+config = {
+    'api_version': '21.0.0',
+    'c_header': 'nixnet.h',
+    'c_function_prefix': 'nx',
+    'service_class_prefix': 'NiXnet',
+    'java_package': 'com.ni.grpc.xnet',
+    'csharp_namespace': 'NationalInstruments.Grpc.NiXnet',
+    'namespace_component': 'nixnet',
+    'close_function': 'Clear',
+    'code_readiness': 'Release',
```
doshirohan, line 11: This should be `NextRelease` until we are ready to release the driver.

source/codegen/metadata/nixnet/config.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+config = {
+    'api_version': '21.0.0',
+    'c_header': 'nixnet.h',
+    'c_function_prefix': 'nx',
+    'service_class_prefix': 'NiXnet',
+    'java_package': 'com.ni.grpc.xnet',
```
doshirohan, line 7: Looks like we don't have consistency between drivers in how these namespaces are defined. Some of them have "ni" repeated in namespace (`com.ni.grpc.nidaqmx`) vs. some omit `ni` from last component. Same for other language namespaces as well. Let's change `xnet` to `nixnet` here to keep it consistent with C# namespace below.

source/codegen/metadata/nixnet/config.py
```diff
@@ -0,0 +1,51 @@
...
+            '64bit': {
+                'name': 'nixnet.dll',
+                'type': 'cdll'
+            }
+        }
+    },
+    'metadata_version': '0.1',
+    'module_name': 'nixnet',
+    'session_class_description': 'An NI-XNET session.',
+    'session_handle_parameter_name': 'vi'
```
doshirohan, line 50: We don't use this anywhere in codegen yet, but let's keep this variable name to `sessionRef` to match with `nxSessionRef_t` parameters. If we start using this for something, we will probably need to add support `nxDatabaseRef_t` as well here, but for now let's just assume `nxSessionRef_t` is our "main" session and add corresponding name here.

source/codegen/metadata/nixnet/functions_addon.py
```diff
@@ -0,0 +1,2 @@
+functions_override_metadata = {
```
doshirohan, line 1: Add a header comment similar to what you have for `enums_addon.py`. You can copy text from some of the other drivers that have it.

doshirohan (approved):

source/codegen/metadata/nixnet/config.py
astarche: Oy, I think there are some force casts for that throughout the codebase. Probably special cases for type == int64/uint64. We might be able to change those to use the grpc_type. Or maybe be more explicit about requiring a force/reinterpret cast in the metadata.

source/codegen/metadata/nixnet/config.py
astarche: I'd change both to `session`. We have to set fields by name in grpc, so these show up in every message and it's nice if it's short. Also `ref`/`handle` are sort of filler and don't map as well/consistently to grpc.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -0,0 +1,1539 @@
...
+                'type': 'u32'
+            }
+        ],
+        'returns': 'nxStatus_t'
+    },
+    'Clear': {
+        'parameters': [
+            {
+                'direction': 'in',
+                'name': 'sessionRef',
```
astarche, line 22: Same as my comment above. I *think* the best name for all of these is to use `session` for the name of everything that is an input/output of `nxSessionRef_t`.

You can use `grpc_name` for that and leave `name` the same (either way is fine, `grpc_name` was more important for RFmx, so that's my current recommendation).

This doesn't need to be in this change. But I'd recommend figuring it out soon. Since all existing code would need to update for it.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -0,0 +1,1539 @@
...
+                'direction': 'in',
+                'name': 'dbFilepath',
+                'type': 'const char[]'
+            }
+        ],
+        'returns': 'nxStatus_t'
+    },
+    'DbSetProperty': {
+        'cname': 'nxdbSetProperty',
+        'codegen_method': 'no',
```
astarche, line 748: How are you tracking the `codegen_method: no` stuff? Maybe a work item and a comment in scrapigen? (you might already have this, I can't remember).

source/codegen/metadata/nixnet/functions.py
```diff
@@ -0,0 +1,1539 @@
...
+            {
+                'direction': 'out',
+                'enum': 'PropertyValue',
+                'name': 'propertyValue',
+                'type': 'void[]'
+            }
+        ],
+        'returns': 'nxStatus_t'
+    },
+    'GetPropertySize': {
```
astarche, line 871: We discussed using this as a size mechanism for `GetProperty` so that the user would not have to call it directly or pass it in. Is that something you're planning to do/tracking?

It's fine as-is for now. I think I even recommended that you add them like this first. So that's great! Just checking on what you're thinking.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -0,0 +1,1539 @@
...
+            }
+        ],
+        'returns': 'nxStatus_t'
+    },
+    'SystemOpen': {
+        'init_method': True,
+        'parameters': [
+            {
+                'direction': 'out',
+                'name': 'systemRef',
```
astarche, line 1327: I guess these should be `system`. Maybe `session` would make sense too.

source/custom/nixnet_service.custom.cpp
```diff
@@ -0,0 +1,4 @@
+#include <nixnet/nixnet_service.h>
```
astarche, line 1: FYI: You actually don't need to include this and it can make a big difference in compile time for bigger drivers (probably doesn't matter for xnet).

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  DB_PROPERTIES_FRM_CLUSTER_REF = 84017153;
+  DB_PROPERTIES_FRM_COMMENT = 50462722;
+  DB_PROPERTIES_FRM_CONFIG_STATUS = 131081;
+  DB_PROPERTIES_FRM_DEFAULT_PAYLOAD = 167903237;
+  DB_PROPERTIES_FRM_FLEX_RAY_BASE_CYCLE = 131104;
+  DB_PROPERTIES_FRM_FLEX_RAY_CH_ASSIGN = 131105;
+  DB_PROPERTIES_FRM_FLEX_RAY_CYCLE_REP = 131106;
+  DB_PROPERTIES_FRM_FLEX_RAY_IN_CYC_REP_CH_ASSIGNS = 134348850;
+  DB_PROPERTIES_FRM_FLEX_RAY_IN_CYC_REP_ENABLED = 33685552;
+  DB_PROPERTIES_FRM_FLEX_RAY_IN_CYC_REP_I_DS = 134348849;
```
astarche, line 163: Is this supposed to be `REP_I_DS`.

Any naming issue can be handled in a separate change. Just track it somewhere.

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  DB_PROPERTIES_CLST_LIN_SCHEDULES = 100728944;
+  DB_PROPERTIES_CLST_LIN_TICK = 16842865;
+  DB_PROPERTIES_CLST_NAME = 50397189;
+  DB_PROPERTIES_CLST_PDU_REFS = 100728840;
+  DB_PROPERTIES_CLST_PD_US_REQD = 33619978;
+  DB_PROPERTIES_CLST_PROTOCOL = 65542;
+  DB_PROPERTIES_CLST_SIG_REFS = 100728839;
+  DB_PROPERTIES_FRM_APPLICATION_PROTOCOL = 131172;
+  DB_PROPERTIES_FRM_CAN_EXT_ID = 33685520;
+  DB_PROPERTIES_FRM_CA_NIO_MODE = 131174;
```
astarche, line 151: I think these are `CAN_IO`. The input was `CANio`, which is a problem for the name converter. So keep an eye out for any  names like that. It looks that's xnet's way to deal with back-to-back acronyms, at least in this case. If it looks like there are a bunch of them, we can consider automating the detection of that (@DavidCurtiss wrote it for RFmx). The volume is way lower for xnet. And the look relatively well behaved. Although the abbreviated names are pretty cryptic.

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  DB_PROPERTIES_PDU_MUX_STATIC_SIG_REFS = 101187594;
+  DB_PROPERTIES_PDU_MUX_SUBFRAME_REFS = 101187595;
+  DB_PROPERTIES_PDU_NAME = 50855937;
+  DB_PROPERTIES_PDU_PAYLOAD_LEN = 524291;
+  DB_PROPERTIES_PDU_SIG_REFS = 101187589;
+}
+
+enum DBPropertyValue {
+  DB_PROPERTY_VALUE_UNSPECIFIED = 0;
+  DB_PROPERTY_VALUE_CA_NIO_MODE_CAN = 1;
```
astarche, line 261: More `CAN_IO`.

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  DB_PROPERTY_VALUE_PROTOCOL_UNKNOWN = 4;
+  DB_PROPERTY_VALUE_PROTOCOL_CAN = 5;
+  DB_PROPERTY_VALUE_PROTOCOL_FLEX_RAY = 6;
+  DB_PROPERTY_VALUE_PROTOCOL_LIN = 7;
+  DB_PROPERTY_VALUE_PROTOCOL_ETHERNET = 8;
+  DB_PROPERTY_VALUE_APP_PROTOCOL_NONE = 9;
+  DB_PROPERTY_VALUE_APP_PROTOCOL_J1939 = 10;
+  DB_PROPERTY_VALUE_FRM_FLEX_RAY_CH_ASSIGN_A = 11;
+  DB_PROPERTY_VALUE_FRM_FLEX_RAY_CH_ASSIGN_B = 12;
+  DB_PROPERTY_VALUE_FRM_FLEX_RAY_CH_ASSIGN_AAND_B = 13;
```
astarche, line 273: `A_AND_B`

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  DB_PROPERTY_VALUE_FRM_FLEX_RAY_CH_ASSIGN_NONE = 14;
+  DB_PROPERTY_VALUE_CLST_FLEX_RAY_SAMP_CLK_PER_P_0125US = 15;
```
astarche, line 275: I'm not sure what this is supposed to be, but `P_0125US` looks a bit suspicious to me.

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  PROPERTIES_SESSION_INTF_FLEX_RAY_SYNC_CH_B_ODD = 135266398;
+  PROPERTIES_SESSION_INTF_FLEX_RAY_TERM = 1048663;
+  PROPERTIES_SESSION_INTF_FLEX_RAY_WAKEUP_CH = 1048661;
+  PROPERTIES_SESSION_INTF_FLEX_RAY_WAKEUP_PTRN = 1048662;
+  PROPERTIES_SESSION_INTF_NAME = 51380243;
+  PROPERTIES_SESSION_INTF_LIN_BREAK_DELIMITER_LENGTH = 1048821;
+  PROPERTIES_SESSION_INTF_LIN_BREAK_LENGTH = 1048688;
+  PROPERTIES_SESSION_INTF_LIN_CHECKSUM_TO_IN_STRM = 34603137;
+  PROPERTIES_SESSION_INTF_LIN_DIAG_P2MIN = 17825911;
+  PROPERTIES_SESSION_INTF_LIN_DIAG_S_TMIN = 17825910;
```
astarche, line 457: `DIAG_S_TMIN` looks suspicious.

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  PROPERTY_VALUE_EPT_RX_FILTER_USE_FLAGS_PRIORITY = 8;
+  PROPERTY_VALUE_EPT_RX_FILTER_USE_FLAGS_DESTINATION_MAC = 9;
+  PROPERTY_VALUE_PHASE_DEVELOPMENT = 10;
+  PROPERTY_VALUE_PHASE_ALPHA = 11;
+  PROPERTY_VALUE_PHASE_BETA = 12;
+  PROPERTY_VALUE_PHASE_RELEASE = 13;
+  PROPERTY_VALUE_DEV_FORM_PXI = 14;
+  PROPERTY_VALUE_DEV_FORM_PCI = 15;
+  PROPERTY_VALUE_DEV_FORM_C_SERIES = 16;
+  PROPERTY_VALUE_DEV_FORM_PX_IE = 17;
```
astarche, line 552: Is this `PXIE`?

generated/nixnet/nixnet.proto
```diff
@@ -0,0 +1,1239 @@
...
+  PROPERTY_VALUE_DEV_FORM_USB = 18;
+  PROPERTY_VALUE_DEV_FORM_PC_IE = 19;
```
astarche, line 554: `PCIE`?

astarche (approved):

source/codegen/metadata/nixnet/config.py
astarche: Reading through the xnet API, I see a lot of `Ref` param names. And also a lot of different prefixes (interface, db, db_object, etc). If someone disagrees with me on this, I'd love to head a counter-argument.

SparkingSpork (approved):

reckenro (approved):

source/codegen/metadata/nixnet/functions.py
doshirohan: I have created tasks on AzDO for tracking remaining API based on different categories (new size mechanism, void return types etc.). We were also planning to update scrapigen to automatically tag these APIs as codegen-no through a hardcoded list of disabled APIs and work our way through till that list is empty.

source/codegen/metadata/nixnet/config.py
doshirohan: @astarche is this change to [service_helpers.mako](https://github.com/ni/grpc-device/pull/570/files#diff-18770917bb2bf21638be75898bfff78d285c52a49af4e2ba348424e57b70f8e0) below is what you had in mind regarding reinterpret cast?

astarche: Yeah, that's fine. Ideally we want to avoid using driver-specific types in the codegen. Often a good way to do that is to key the logic on the grpc_type. Or to add additional info to the metadata. But, in this case, the existing code is obviously hardocding a bunch of driver types, so adding another is OK. You could also consider a helper method that explains what that is for (different typedef'd integer values with the same guaranteed size but different `long` qualifier counts).

PR #571, opened 2022-02-28
Fix several python linter issues in source/codegen

DavidCurtiss:

Fixes issues reported by the ni-python-styleguide tool, except for adding missing docstrings and non-trivial fixes to existing docstrings.

Stepping toward using the ni-python-styleguide

```
> ni-python-styleguide lint source\codegen\*.py | grep -v "Missing docstring\|doc line too long"
source\codegen\metadata_mutation.py:238:1: D415 First line should end with a period, question mark, or exclamation point
source\codegen\metadata_mutation.py:275:1: D205 1 blank line required between summary line and description
source\codegen\metadata_mutation.py:275:1: D212 Multi-line docstring summary should start at the first line
source\codegen\metadata_mutation.py:275:1: D415 First line should end with a period, question mark, or exclamation point
```

### DIFF
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -899,12 +899,18 @@ def get_enum_value_cpp_type(enum: dict) -> str:
 
 
 def is_regular_string_arg(parameter: dict) -> bool:
-    """Exclude byte arrays."""
+    """Return whether the parameter is a "regular" string.
+
+    This excludes byte arrays.
+    """
     return is_string_arg(parameter) and not parameter["grpc_type"] == "bytes"
 
 
 def is_regular_byte_array_arg(parameter: dict) -> bool:
-    """Exclude strings and byte arrays that are mapped to enums."""
+    """Return whether the parameter is a "regular" byte array.
+
+    This excludes byte arrays that are mapped to enums.
+    """
     return parameter["grpc_type"] == "bytes" and not is_enum(parameter)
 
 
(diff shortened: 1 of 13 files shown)


DavidCurtiss:
Need to figure out which way I should fix these relative import complaints. Will reopen when I have a fix for that.

source/codegen/common_helpers.py
```diff
@@ -893,22 +899,22 @@ def get_enum_value_cpp_type(enum: dict) -> str:
 
 
 def is_regular_string_arg(parameter: dict) -> bool:
-    """Excludes: byte arrays"""
+    """Exclude byte arrays."""
```
reckenro, line 902: I think this should be something like
```
"""Return whether the parameter is a "regular" string argument.

This excludes byte arrays."""
```

source/codegen/common_helpers.py
```diff
@@ -893,22 +899,22 @@ def get_enum_value_cpp_type(enum: dict) -> str:
...
     return is_string_arg(parameter) and not parameter["grpc_type"] == "bytes"
 
 
 def is_regular_byte_array_arg(parameter: dict) -> bool:
-    """Excludes: strings and byte arrays that are mapped to enums"""
+    """Exclude strings and byte arrays that are mapped to enums."""
```
reckenro, line 907: Same here, I think it should be something like
I think this should be something like
```
"""Return whether the parameter is a "regular" byte array.

This excludes byte arrays that are mapped to enums."""

This excludes byte arrays.
"""
```

reckenro (approved):

mgumble (approved):

astarche (approved):

PR #572, opened 2022-02-28
Add session API support for XNET

doshirohan:

XNET sessions have two special considerations compared to other drivers we have added to `grpc-device` repo so far.
1. XNET driver owns more than one type of session references (`nxSessionRef_t`, `nxDatabaseRef_t`)
2. It has "accessor" methods for reference type which are like `init` methods, but sessions returned by these APIs are existing sessions and we shouldn't register cleanup lambda for such in session repository

For supporting multiple sessions per driver/service, I have changed `resource_handle_type` metadata in `config.py` to be able to pass in array of handles instead of just single string. Appropriate codegen changes are made in all other places to use correct session repository instance based on type of handle used. We already sort of supported multiple session repositories through `cross_driver_session`. Codegen changes in some places are to merge `CrossDriverSessionDependency` list generated for `cross_driver_session` with the one generated from `session_handle_type` to generate different params, arguments, local variables etc.

For supporting session accessor method (`DbFindObject`) that shouldn't close the reference uses existing `cross_driver_session` tag so that returned session reference uses `add_dependent_session` mechanism for cleanup.

*Note for reviewers:* There are inconsistent usages of `session_repository` and `resource_repository` while naming default repository in service cpp files. I have tried to (I had to) partially correct the naming which has caused changes to generated code for all the drivers. These changes should be harmless and can be skimmed through.

This PR adds support for following APIs for xnet
1. Session APIs using default session repository and corresponding close APIs, session type: `nxSessionRef_t`
   - nxCreateSession, nxCreateSessionByRef  => nxClear 
   - nxSystemOpen => nxSystemClose
 2. Session API with non-default session repository and corresponding close API, session type: `nxDatabaseRef_t`
    - nxdbOpenDatabase => nxdbCloseDatabase
    - nxdbCreateObject => nxdbDeleteObject
3. Session reference accessor methods
    - nxdbFindObject
 
- Manually verified generated server handles for above API to check if session are accessed via correct session repository
- Added fake non-ivi unit test for verifying additional session types are added to corresponding session repository.


source/codegen/service_helpers.py
```diff
@@ -422,6 +422,24 @@ class CrossDriverSessionDependency(NamedTuple):
...
 
+def get_driver_shared_resource_repository_ptr_deps(
+    driver_config: dict,
+) -> List[CrossDriverSessionDependency]:
+    resource_repository_deps = [
+        _create_cross_driver_session_dependency(resource_handle_type)
+        for resource_handle_type in get_resource_handle_types(driver_config)
+    ]
+
+    resource_repository_deps[0] = CrossDriverSessionDependency(
```
doshirohan, line 433: note: I couldn't find a better way to customize field names and alias of "default" repository.

source/codegen/service_helpers.py
```diff
@@ -422,6 +422,24 @@ class CrossDriverSessionDependency(NamedTuple):
...
+        resource_repository_deps[0].resource_handle_type,
+        "ResourceRepositorySharedPtr",
+        resource_repository_deps[0].resource_repository_type,
+        "session_repository_",
```
doshirohan, line 437: note: I haven't changed this to `resource_repository_` as this will essentially cause every function from every driver to have single line change. That's too much noise for this PR. I can send follow up PR once this is done to update field name for default repository.

generated/nixnet/nixnet_service.cpp
```diff
@@ -359,8 +361,9 @@ namespace nixnet_grpc {
     }
     try {
       auto database_ref_grpc_session = request->database_ref();
-      nxDatabaseRef_t database_ref = session_repository_->access_session(database_ref_grpc_session.id(), database_ref_grpc_session.name());
+      nxDatabaseRef_t database_ref = nx_database_ref_t_resource_repository_->access_session(database_ref_grpc_session.id(), database_ref_grpc_session.name());
       u32 close_all_refs = request->close_all_refs();
```
doshirohan, line 365: If `close_all_refs` is true, this library call sets reference counter for this db reference to 0, but we will have dangling entries in `nx_database_ref_t_resource_repository_`. Does it make sense to custom implement `DbCloseDatabase` and when `close_all_refs` is true, iterate over `nx_database_ref_t_resource_repository_` find all entries which have value of `database_ref ` and purge them? If we don't do this, `nx_database_ref_t_resource_repository_` could potentially keep growing in size.

astarche:
@doshirohan Overall this looks really nice. I ran out of energy for tracking it and I like to look closely at the codgen. But the metadata and the output look good; and the overall approach of the implementation looks good. So I expect it will be good!

generated/nixnet/nixnet_service.cpp
astarche: I think we should leave it as-is. We can consider adding that as a feature in the future. I would expect `close_all_refs` to be an infrequently used feature and the consequence is minor. And there are 2 "workarounds" by using grpc-device-style APIs: use named sessions so there's only one grpc-device session or call reset_server to reset everything.

Is `close_all_refs` used in xnet examples? Or recommended anywhere? My understanding is that those "force close" APIs are always discouraged because they circumvent ref counting. But maybe there are some APIs where they are more accepted?

source/codegen/metadata/nixnet/functions.py
```diff
@@ -374,6 +374,7 @@
     },
     'DbCloseDatabase': {
         'cname': 'nxdbCloseDatabase',
+        'custom_close_method': True,
```
astarche, line 377: Note: as long as you are generating functions from scrapigen, you will have to back-port these changes to the extraction tool.

source/codegen/templates/service_registrar.cpp.mako
```diff
@@ -29,15 +29,19 @@ namespace ${namespace} {
 namespace {
 struct LibraryAndService {
   LibraryAndService(
-    const ${resource_repository_type}& resource_repository,
+% for resource_repository_dep in resource_repository_deps:
+    const ${resource_repository_dep.resource_repository_type}& ${resource_repository_dep.local_name},
+% endfor
 % for cross_driver_dep in cross_driver_session_deps:
```
astarche, line 35: Nit: it seems like a lot of these back-to-back session_deps for loops could be combined with just anew helper method that gets all of them. It's OK how it is though.

source/codegen/metadata/nixnet/functions.py
astarche: ALSO: I don't see `custom_close_method` being used. Why was that added?

source/codegen/metadata/nixnet/functions.py
```diff
@@ -389,7 +390,8 @@
         'returns': 'nxStatus_t'
     },
     'DbCreateObject': {
-        'codegen_method': 'no',
+        'init_method': True,
+        'custom_close': "", # we don't need to close reference from this method
```
astarche, line 394: await: We would ideally like all of these to close when the database is closed right? We actually have a mechanism for that with `add_dependent_session`.

```python
db = stub.open_database(...)
table = stub.create_object(db, "table", ...)
row = stub.create_object(table, "row", ...)
stub.close_database(db)
```

Do we want that last line to clean up table and row? If so, I think we can do that with `add_dependent_session`. Each db_object would be a dependent session of the parent_object_ref and it would cascade down.

In the current implementation, is there any way to directly close the sessions with `custom_close=""`? They would be  cleaned up with `reset_server` (that's good!) but there's no way to clean them up individually.

source/codegen/service_helpers.py
```diff
@@ -401,19 +401,19 @@ def get_bitfield_value_to_name_mapping(parameter: dict, enums: dict) -> Dict[int
     }
 
 
-def get_resource_handle_type(config: dict) -> str:
-    return config.get("resource_handle_type", "ViSession")
+def get_resource_handle_types(config: dict) -> List[str]:
```
astarche, line 404: await: you need some test coverage for the new codegen. Ideally that would be fake tests and system level tests. But either/or can be OK. In this case "just fake tests" seems like a good option for now. My PR #574 is an OK example of some basic test coverage. We probably don't want to add a lot of "expected failure/noop" system level cases like that, but it's something to consider. And if you have a test system set up you can write tests that can run on the real hw.

Is it possible to test the database without an xnet device?

astarche (changes requested):
See inline `await` Re add_dependent_session and fake tests. I'm OK with a short-term workaround/whatever for dependent sessions as long as you feel that leaves it in a sane state and you have a plan to fix it.

source/codegen/metadata/nixnet/functions.py
doshirohan: Last line does cleanup table and row from driver side, but I had missed removing them from session repository. `DbCreatedObject` has equivalent `DbDeleteObject` that can be used as `custom_close` here. I have fixed function metadata to do that.

source/codegen/service_helpers.py
doshirohan: I have added non-ivi fake tests for validating initialize/cleanup when there are multiple session repositories owned by driver. We will probably add more system level tests later once we have enough API coverage ready.

source/codegen/templates/service_registrar.cpp.mako
doshirohan: I have merged lists for `cross_driver_session_deps` and `resource_repository_deps` using helper methods and replaced all these back to back loops with single loop.

source/codegen/metadata/nixnet/functions.py
doshirohan: We will be back-porting any changes from here to scrapigen. That will be done once this PR is finalized. 

`custom_close_method` tag is used in `service_helpers.mako`. We can mark only one function as `close_function` in `config.py`. This tag is used for tagging any other APIs which are close methods to generate `remove_session` call. This particular close function is used by `DbOpenDatabase`.

generated/nixnet/nixnet_service.cpp
doshirohan: `close_all_refs` is used by all `nxDbCloseDatabase` in examples.

There is this note in documentation about when `close_all_refs` should be used
> Another option is that only one thread executes nxdbCloseDatabase once, using CloseAllRefs = 1, which closes access for all other threads. This may be convenient when, for example, the main program needs to stop all running threads and be sure the database is closed properly, even if some threads could not execute nxdbCloseDatabase.

But I agree that consequence does seem minor here, so we could leave that for future.

generated/nixnet/nixnet_service.cpp
```diff
@@ -370,6 +373,39 @@ namespace nixnet_grpc {
...
+      auto object_name = request->object_name().c_str();
+
+      auto init_lambda = [&] () {
+        nxDatabaseRef_t db_object_ref;
+        auto status = library_->DbCreateObject(parent_object_ref, object_class, object_name, &db_object_ref);
+        return std::make_tuple(status, db_object_ref);
+      };
+      uint32_t session_id = 0;
+      const std::string& grpc_device_session_name = request->session_name();
+      auto cleanup_lambda = [&] (nxDatabaseRef_t id) { library_->DbDeleteObject(id); };
```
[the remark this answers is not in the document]
astarche: So, I think this code could/should use `add_dependent_session` the same way that `Find` does below. That would ensure that the map is correctly cleaned up when the database is closed AND would avoid deleting all of the objects on `reset`. If that implementation doesn't remove db objs from the map after Delete, that's OK for now at least. You can add a follow up work item to do some extra testing and consider fixing that.

astarche (approved):
Everything looks good except for the `DbCreateObject`/`DbDeleteObject` close method behavior. If you can't easily incorporate my feedback, I'd prefer to get this change submitted and create additional work items to track the issues.

generated/nixnet/nixnet_service.cpp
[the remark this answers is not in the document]
doshirohan: I have updated `CreateObject` to use `cross-driver-session` and removed `custom_close` on that to avoid deleting object during server reset.

`DeleteObject` still has `custom_close_method` set. That will take care of removing the session from server's map when user invokes `DeleteObject` directly. Does this make sense or am I missing something related to sessions added by `add_dependent_session` can not be cleaned up by normal close methods? I assumed since both initiating session repo and dependent session repos are same (nxDatabaseRef_t), this should work.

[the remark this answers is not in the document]
astarche: That sounds and looks correct to me. The test case would be:
1. OpenDatabase
2. CreateDatabaseObj
3. DeleteDatabaseObj
4. CloseDatabase

Step (4) will attempt to close all dependent sessions, including the one that no longer exists. Does it handle that case OK? If not can you change the logic to tolerate the session being there? I don't think that is tested anywhere else, because it's not valid for cross-driver sessions (the driver docs say specifically: don't do it, bad things will happen)

I think it's OK if that's not addressed in this PR. Should be in your acceptance table somewhere.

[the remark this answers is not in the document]
doshirohan: Makes sense to test this scenario out. Added test case for this in acceptance table.

source/codegen/service_helpers.py
mgumble: Sadly you get noise either way :(

mgumble (approved):

source/codegen/metadata/nixnet/functions.py
```diff
@@ -390,8 +390,8 @@
         'returns': 'nxStatus_t'
     },
```
mgumble, line 391: What manual changes? We have been keeping the two in sync which obviously doesn't allow for manual changes

doshirohan: Ah, I should've probably been more verbose on that commit message. There were some changes to metadata that weren't backported from here to scrapigen earlier. We are going to do that soon, but for now I had to manually make those changes after copying the export from scrapigen.

reckenro (approved):

source/codegen/metadata/nixnet/functions.py
mgumble: That makes sense, yeah I saw that commit come through my email and it was a red flag

PR #574, opened 2022-03-01
Add nxsocket service with support for non-status return values

astarche:

Add basic functions of `NiXnetSocket` API to grpc-device for the `nxsocket.h` header.

Add support for non-status return values to handle various methods in `nxsocket.h` with non-status return values.

Note: init is hardcoded to fail because we can't yet construct an `nxIpStackRef_t`.

Incremental progress on Xnet Socket feature. Unblocks other work by providing a shell of a service to work with.

Ran and pased all (1) Xnet test and a cross-section of driver tests.


source/codegen/metadata/nixnetsocket/functions.py
```diff
@@ -0,0 +1,84 @@
...
+                'name': 'type',
+                'type': 'int32_t'
+            },
+            {
+                'direction': 'in',
+                'name': 'prototcol',
+                'type': 'int32_t'
+            },
+            {
+                'direction': 'out',
```
astarche, line 75: My strategy for return values is to add a special parameter for it with a `return_value` tag. This gives us a place to hang metadata and doesn't require a lot of special case code to workaround.

SparkingSpork (approved):

source/codegen/common_helpers.py
```diff
@@ -958,3 +958,8 @@ def get_grpc_field_names_for_param_names(params: List[dict], names: List[str]) -
     of names.
     """
     return [get_grpc_field_name(get_param_with_name(params, name)) for name in names]
+
+
+def is_return_value(parameter: dict) -> bool:
+    """Returns true if parameters is marked as a return_value."""
```
reckenro, line 964: Minor:
```suggestion
    """Returns true if parameter is marked as a return_value."""
```

reckenro (approved):

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -1622,6 +1622,72 @@ TEST_F(NiFakeNonIviServiceTests, SessionAlreadyInCrossDriverSessionRepository_Ge
...
+{
+  constexpr auto FAILED_GET = -1;
+  EXPECT_CALL(library_, GetStringAsReturnedValue(_))
+      .WillOnce(Return(nullptr));
+  ::grpc::ServerContext context;
+  GetStringAsReturnedValueRequest request;
+  GetStringAsReturnedValueResponse response;
+  service_.GetStringAsReturnedValue(&context, &request, &response);
+
+  EXPECT_THAT(response.string_out(), IsEmpty());
```
mgumble, line 1687: One of these is backwards, I think normally it is expected, actual

mgumble (approved):

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
astarche: I do try to do `expected, actual`, since you're right it's the NI standard (at least it is in ASW). But it gets reversed when you have to do `EXPECT_THAT` because the matcher is always on the right so `actual` goes on the left. So it's:

> Expect that [actual] [is whatever]

i.e.,
> Expect that [actual] is empty.
> Expect that [actual] starts with [expected].
> Expect that [actual] container equals [expected].

It may be possible to have a hardcore rule to put [expected] in the left-hand-side in `EXPECT_THAT` statements, but  ... it would be pretty weird because you'd have to invert a lot of the asserts. And it reads weird as a sentence:

> Expect that [expected] is a superstring of [actual]

Possibly for that reason, gtest examples tend to do `actual, expected` in `EXPECT_EQ` as well. ([reference](http://google.github.io/googletest/primer.html))

doshirohan (approved):

PR #575, opened 2022-03-01
Add SockAddr type and converters to xnetsocket service

astarche:

Add support for the `nxsockaddr` union type as a `oneof` type in `grpc-device`.

Our existing codegen does not support this type structure. To avoid adding a lot of complexity to codegen, I've bolstered the `supports_standard_copy_convert` concept and implemented a `SockAddrHolder` struct that does the heavy lifting.

Fixes [AB#1871124](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1871124).

Required throughout `nxsocket.h`. 

Ran and passed all unit tests and all `Xnet` tests.


source/codegen/metadata_validation.py
```diff
@@ -218,17 +219,6 @@ def _validate_function(function_name: str, metadata: dict):
                             raise Exception(
                                 f"Failed to validate callback_param with name {callback_param['name']}"
                             )
-                if parameter.get("pointer", False):
```
astarche, line 221: A lot of xnet inputs are passed by pointer. I could add `supports_standard_copy_convert` to this list, but there already a lot of unintuitive exceptions. Also, I don't expect someone to add this tag on accident. It might not do what they want, but this validation doesn't help explain what the gap is.

source/codegen/service_helpers.py
```diff
@@ -107,11 +107,7 @@ def _create_standard_arg(parameter):
         return f"CallbackRouter::handle_callback, "
     elif "callback_token" in parameter:
         return f"handler->token(), "
-    elif (
-        not is_output
```
astarche, line 111: Note: All of these negatives can be captured as the positive `is_size_param_passed_by_ptr`.

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -0,0 +1,117 @@
...
+TEST(XnetConvertersTests, IPv4GrpcSockAddr_ConvertFromGrpc_CreatesIPv4NXSockAddr)
+{
+  constexpr auto PORT = 1234;
+  constexpr auto ADDRESS = 0xAABBCCDD;
+  auto grpc_sock_addr = SockAddr{};
+  grpc_sock_addr.mutable_ipv4()->set_port(PORT);
+  grpc_sock_addr.mutable_ipv4()->set_addr(ADDRESS);
+
+  auto converted_addr = convert_from_grpc<nxsockaddr>(grpc_sock_addr);
+  auto addr_ptr = static_cast<nxsockaddr*>(converted_addr);
```
astarche, line 24: Note: `static_cast` tests the implicit conversion. `reinterpret_cast` is what the driver would do given the address family.

source/codegen/metadata/nixnetsocket/functions.py
```diff
@@ -1,10 +1,35 @@
...
+            },
+        ],
+        'returns': 'int32_t'
+    },
     'Close': {
         'cname': 'nxclose',
         'parameters': [
             {
                 'direction': 'in',
-                'grpc_type': 'nidevice_grpc.Session',
```
reckenro, line 7: Huh, was this deletion on purpose? Also, I'm surprised it didn't seem to change the generated proto or service.

SparkingSpork (approved):

reckenro (approved):

source/codegen/metadata/nixnetsocket/functions.py
reckenro: Oh nevermind, I see in the config.py we have `nxSOCKET` mapped to `nidevice_grpc.Session` so this is taken care of.

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -0,0 +1,117 @@
...
+  const auto TRUNCATED_ADDRESS = std::vector<char>{0x0, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8, 0x9, 0xA, 0xB, 0xC, 0xD, 0xE, 0xF};
+  constexpr auto SCOPE_ID = 8888;
+  auto grpc_sock_addr = create_addr_ipv6(PORT, FLOW_INFO, ADDRESS, SCOPE_ID);
+
+  auto converted_addr = convert_from_grpc<nxsockaddr>(grpc_sock_addr);
+
+  EXPECT_IPV6_ADDR(converted_addr, PORT, FLOW_INFO, TRUNCATED_ADDRESS, SCOPE_ID);
+}
+
+TEST(XnetConvertersTests, UnsetGrpcSockAddr_ConvertFromGrpc_CreatesZeroedOutSockAddr)
```
mgumble, line 104: Maybe I am wrong, but to me "zeroed out" would have elements with the value of zero. Rather than a vector with size of 0

mgumble (approved):
Sorry for the delay on this review!

source/tests/unit/xnet_converters_tests.cpp
astarche: Hm, yeah. "Zeroed out" in this context is more in the implementation domain. The struct is literally all zeros:

```c++
std::memset(&addr, 0, sizeof(addr));
```

And the asserts are what that looks like at an API level (empty/unspecified).

I think it's reasonable to be verbose here. "CreatesEmptySockAddrWithUnspecifiedAddressFamily".

PR #579, opened 2022-03-02
Fix misspelled param names in SpecAn.

reckenro:

Fixes two misspelled param names for SpecAn ("measurerment" => "measurement").

SpecAn has the misspellings in their API but we can fix it on our end in the proto file.

Build succeeds.


reckenro:
Note: Filed AB#1871307 to RFmx to fix it on the driver side of things.

astarche (approved):

PR #583, opened 2022-03-03
Add get_last_error support for RFmx init methods

### LINKED ISSUE #494: RFSG and SpecAn GetError tests fail (order dependent)
`ErrorFromDriver_GetErrorMessage_ReturnsUserErrorMessage`

Both of these tests can fail depending on what tests they run with.

Both of these are passing `GetError` a session id from a request that *failed to init the session*. So there's some weirdness in how we're handling that.

astarche:

Change RFmx `init_method`s to `get_last_error` on failed init.

Normally, `GetError` in RFmx gets the last error for the given session. But in the case of failed Init, there is no session to check. The fallback logic uses thread-local storage to get the latest error on the calling thread. This does not work well for `grpc-device` because messages are handled on a thread pool.

Calling `GetError` immediately after a failed init and adding it to the response ensures that we get the correct error message.

Fixes #494 

Fixes [AB#1872435](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1872435)

### Alternatives considered

1. Do Nothing. Reason Rejected: Failed Init can be frustrating to debug without the message.
2. Try to implement a global `get_last_error`. Reason Rejected: This doesn't require API changes but is not easier to implement and can be incorrect with multiple clients.
3. Implement a better error/status reporting strategy like an error service or rich status messages. Reason Rejected: this is feature-level and potentially a major behavior breaking change. The `grpc-device` approach to error messages is not perfect, but it's working OK and we'll stick with the overall approach for now.

Ran and passed all new tests and RFmx tests.

### DIFF
--- a/generated/nirfmxinstr/nirfmxinstr.proto
+++ b/generated/nirfmxinstr/nirfmxinstr.proto
@@ -1054,6 +1054,7 @@ message InitializeResponse {
   int32 status = 1;
   nidevice_grpc.Session instrument = 2;
   int32 is_new_session = 3;
+  string error_message = 4;
 }
 
 message InitializeFromNIRFSASessionRequest {
@@ -1064,6 +1065,7 @@ message InitializeFromNIRFSASessionRequest {
 message InitializeFromNIRFSASessionResponse {
   int32 status = 1;
   nidevice_grpc.Session instrument = 2;
+  string error_message = 3;
 }
 
 message InitializeFromNIRFSASessionArrayRequest {
@@ -1074,6 +1076,7 @@ message InitializeFromNIRFSASessionArrayRequest {
 message InitializeFromNIRFSASessionArrayResponse {
   int32 status = 1;
   nidevice_grpc.Session instrument = 2;
+  string error_message = 3;
 }
 
 message IsSelfCalibrateValidRequest {
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -1055,7 +1055,7 @@ def is_get_last_error_output_param(parameter: dict) -> bool:
     return "get_last_error" in parameter
 
 
-def get_driver_api_params(parameters: List[dict]) -> bool:
+def get_driver_api_params(parameters: List[dict]) -> List[dict]:
     """Return all parameters that are passed as parameters to the driver API.
 
     Excludes:
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -1048,3 +1048,18 @@ def get_grpc_field_names_for_param_names(params: List[dict], names: List[str]) -
 def is_return_value(parameter: dict) -> bool:
     """Whether the parameter is marked as a return_value."""
     return parameter.get("return_value", False)
+
+
+def is_get_last_error_output_param(parameter: dict) -> bool:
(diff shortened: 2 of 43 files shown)


generated/nirfmxinstr/nirfmxinstr.proto
```diff
@@ -1054,6 +1054,7 @@ message InitializeResponse {
   int32 status = 1;
   nidevice_grpc.Session instrument = 2;
   int32 is_new_session = 3;
+  string error_message = 4;
```
reckenro, line 1057: Do we not need it for GetNIRFSASession and GetNIRFSASessionArray? I guess since those are adding a dependent session and have an input of an RFmx session they should be fine to look up the error message on failure like normal, right?

source/codegen/common_helpers.py
```diff
@@ -1048,3 +1048,18 @@ def get_grpc_field_names_for_param_names(params: List[dict], names: List[str]) -
...
     """Whether the parameter is marked as a return_value."""
     return parameter.get("return_value", False)
+
+
+def is_get_last_error_output_param(parameter: dict) -> bool:
+    """Return True if pararameter is a get_last_error parameter."""
+    return "get_last_error" in parameter
+
+
+def get_driver_api_params(parameters: List[dict]) -> bool:
```
reckenro, line 1058: The return value type hint should be `List[dict]` I believe.

reckenro (approved):

generated/nirfmxinstr/nirfmxinstr.proto
astarche: Yeah, any errors for those should be associated with the initiating RFmx session.

source/codegen/metadata/nirfmxlte/config.py
```diff
@@ -1,16 +1,16 @@
 # -*- coding: utf-8 -*-
 config = {
```
astarche, line 2: Oops. I'd rather not reformat these. But unlike the functions.py files, they're not in strict "pretty print" format in the first place. So maybe it's OK. I'm going to try to fix it.

source/tests/system/nirfmxspecan_session_tests.cpp
```diff
@@ -107,16 +97,31 @@ TEST_F(NiRFmxSpecAnSessionTest, InitializedSession_CloseSession_ClosesDriverSess
...
   rfmxspecan::InitializeResponse init_response;
   call_initialize(kRFmxSpecAnTestInvalidRsrc, "", "", &init_response);
   EXPECT_EQ(kInvalidRsrc, init_response.status());
 
   nidevice_grpc::Session session = init_response.instrument();
-  expect_error_string(session, init_response.status(), kRFmxSpecAnErrorResourceNotFoundMessage);
+  EXPECT_EQ(std::string(kRFmxSpecAnErrorResourceNotFoundMessage), init_response.error_message());
+}
+
+TEST_F(NiRFmxSpecAnSessionTest, InitWithErrorFromDriver_ReInitSuccessfully_ErrorMessageIsEmpty)
```
mgumble, line 115: nit: reinit is probably 1 word not two :)

mgumble (approved):

PR #592, opened 2022-03-09
[Draft] Get/Set APIs XNET

ni-pkhare:

Adds GetProperty API in `custom.cpp` file and the proto message definitions of it in `custom_proto.mako` file.

The response message contains **oneof** which lists different datatypes. Since the datatype of value returned by property can be an array, **repeated** fields are used, but **oneof** does not allow repeated keyword inside it, hence a different message is created for array types and the main **oneof** containes those message fields.
 
There are also maps created for getting the datatype of value a property would return. These maps are used inside the wrapper API and based on the datatype, different switch-cases are created for setting the appropriate response field. 

Following codegen changes are made:
1. In enums.py, for Property, DBProperty, SubProperty, an additional tag called `'type'` is added which basically tells the datatype of value it returns.
2. The function is tagged with `codegen_method: CustomCodeButNoProtoMessage`. I have introduced this new codegen method so that `rpc GetProperty(GetPropertyRequest) returns (GetPropertyResponse);` gets generated in the proto file but the **request** and **response** message are not because we are custom implementing it in `custom_proto.mako`
3. An additional logic for generating `std::map<std::uint32_t, Type> property_type_map_` inside `nixnet_service.h` has been added. Here the 2nd parameter `Type` is used which is basically an `enum`. There is also an additional logic written to generate this enum in the same file so that it can be used.


ni-pkhare:
@astarche I have added you to this PR to get some initial feedback for this API as the other ones would be similar to this one only.

source/codegen/metadata/nixnet/custom_proto.mako
```diff
@@ -0,0 +1,31 @@
+message u32Array {
```
astarche, line 1: Messages need to be `PascalCased`: `U32Array`, `RefArray`.

source/custom/nixnet_service.custom.cpp
```diff
@@ -1,2 +1,168 @@
...
+      }
+      case ref_array_: {
+        std::vector<u32> property_value(property_size, 0);
+        status = library_->GetProperty(session_ref, property_id, property_size, property_value.data());
+        if (!status_ok(status)) {
+          response->set_status(status);
+          return ::grpc::Status::OK;
+        }
+        auto response_iterator = response->mutable_ref_array()->mutable_ref()->begin();
+        for (auto it = property_value.begin(); it != property_value.end(); it++, response_iterator++) {
```
astarche, line 154: This is not correct. Since this is a session, you need to `access_session` to get the grpc session from the db_ref.

Also: is it possible that the db_ref is not yet tracked by grpc-device? @doshirohan 's change handles that for `FindDBRef` by basically making it an `init_method`. The only challenge there is that we need to provide an array of session_names so that the user can name the sessions.

You might want to defer this to a separate task and make sure you know how this is used.

source/custom/nixnet_service.custom.cpp
```diff
@@ -1,2 +1,168 @@
...
+        break;
+      }
+      case ref_array_: {
+        std::vector<u32> property_value(property_size, 0);
+        status = library_->GetProperty(session_ref, property_id, property_size, property_value.data());
+        if (!status_ok(status)) {
+          response->set_status(status);
+          return ::grpc::Status::OK;
+        }
+        auto response_iterator = response->mutable_ref_array()->mutable_ref()->begin();
```
astarche, line 153: I think `std::transform` is a better way to handle this type of copy operation. There are quite a few examples in the repo.

source/codegen/service_helpers.py
```diff
@@ -246,7 +246,9 @@ def _create_param(parameter, expand_varargs=True, repeated_parameters=None):
             return "..."
     elif common_helpers.is_array(type):
         array_size = _get_array_param_size(parameter)
-        return f"{type[:-2]} {name}[{array_size}]"
+        if type[:-2] == "void":
```
astarche, line 249: This is a pretty confusing change, I at least understand what it's doing. Is there a reason that `void[]` doesn't work? I know that's unconventional, but it does match how grpc-device usually deals with buffer types.

At least put a comment here on what it's doing and why.

source/codegen/metadata_validation.py
```diff
@@ -87,7 +87,7 @@ class RULES:
         "returns": str,
         Optional("cname"): str,
         Optional("codegen_method"): And(
-            str, lambda s: s in ("public", "private", "CustomCode", "no", "python-only")
+            str, lambda s: s in ("public", "private", "CustomCode", "no", "python-only", "CustomCodeButNoProtoMessage")
```
astarche, line 90: Consider: would this be better as a separate tag? I think we've repeatedly run into problems where we wanted to mix-and-match `codegen_method` features.

I would also say `CustomCodeCustomProtoMessage`. You're using both "Custom" and "No" to mean "not generated but hardcoded". "Custom" is a better name for that and it's better to be consistent.

source/codegen/service_helpers.py
```diff
@@ -284,6 +286,15 @@ def get_output_lookup_values(enum_data):
         out_value_format += f"{{{formated_value}, {i + 1}}},"
     return out_value_format
 
+def get_id_type_value(enum_data):
```
astarche, line 289: All of these new method names are EXTREMELY generic. It's very difficult to tell from the names that this is codegen for a very specific feature that is only used to generate a few fields in codegen (i.e., it's not even used in the service cpp generator).

First we need a name for your feature and then we need to use it everywhere.

`type_mapped_enum`?
`enum_oneof_selector`?
`type_selector_enum`?

Also, some of these names are obviously insufficient. Like this one should mention that it's a map/lookup. Something like `generate_enum_oneof_selector_map`

source/codegen/service_helpers.py
```diff
@@ -419,6 +430,26 @@ def should_generate_mappings(enum_name: str) -> bool:
...
 
+def get_enums_to_map_type(enums: dict) ->List[str]:
+    """Get a list of the enums used by functions, for which mappings should be generated."""
+    list_of_enums: List[str] = []
+    for enum_name in enums.keys():
+        if "generate-mapping-type" in enums[enum_name]:
+            list_of_enums.append(enum_name)
+    return list_of_enums
+
+def get_type_from_enum(enums: dict) -> str:
```
astarche, line 441: `generate_type_tag_enum_for_<your_name_for_this_feature>`. This is being used to select the field of a union. This is similar to "tagged" or "discriminated" unions, though in this case the "tag" is not part of the union.

source/codegen/service_helpers.py
```diff
@@ -419,6 +430,26 @@ def should_generate_mappings(enum_name: str) -> bool:
...
+    """return a comma seperated string of different data types used in enums value type field"""
+    s = ""
+    distinct_type = set()
+    for enum_name in enums.keys():
+        if "generate-mapping-type" in enums[enum_name]:
+            for i, value in enumerate(enums[enum_name]["values"]):
+                type = value["type"]
+                if type not in distinct_type:
```
astarche, line 449: Instead of this block, can you just build a set the call:

```python
str.join(", ", sorted(distinct_type)
```

source/codegen/templates/service.h.mako
```diff
@@ -58,6 +60,10 @@ struct ${service_class_prefix}FeatureToggles
 % endfor
 };
 
+% if type_from_enum != "":  
+enum Type { ${type_from_enum} };
```
astarche, line 64: This can't just be called Type!!!

source/custom/nixnet_service.custom.cpp
```diff
@@ -1,2 +1,168 @@
...
+    }
+
+    u32 property_size{};
+    auto status = library_->GetPropertySize(session_ref, property_id, &property_size);
+    if (!status_ok(status)) {
+      response->set_status(status);
+      return ::grpc::Status::OK;
+    }
+
+    switch (property_type_map_[property_id]) {
```
astarche, line 52: You'll need to make sure and test this thoroughly. There's not codesharing that helps validate this logic is correct.

astarche:
I think the overall structure looks good. Thanks for sending out the Draft review!

generated/nixnet/nixnet_service.h
```diff
@@ -91,6 +94,9 @@ class NiXnetService final : public NiXnet::Service {
...
..._array_},{33685539, boolean_},{33685540, boolean_},{33685541, boolean_},{131110, u32_},{131075, u32_},{131152, u32_},{84017217, ref_},{33685568, boolean_},{100794434, ref_array_},{100794435, ref_array_},{50462724, string_},{134348897, u32_array_},{134348899, u32_array_},{100794464, ref_array_},{131079, u32_},{100794376, ref_array_},{33685605, boolean_},{196609, u32_},{50528258, string_},{196617, u32_},{196611, u32_},{16973828, f64_},{84082693, ref_},{16973830, f64_},{16973831, f64_},{33751088, boolean_},{33751089, boolean_},{196658, u32_},{84082739, ref_},{50528264, string_},{50528272, string_},{196626, u32_},{84082705, ref_},{16973843, f64_},{16973844, f64_},{196629, u32_},{50528278, string_},{262153, u32_},{100925441, ref_array_},{84148226, ref_},{262147, u32_},{50593796, string_},{50593799, string_},{84148229, ref_},{84213761, ref_},{50659333, string_},{327689, u32_},{33882128, boolean_},{327700, u32_},{84213777, ref_},{327699, u32_},{327698, u32_},{100990979, ref_array_},{100990980, ref_array_},{327715, u32_},{327717, u32_},{327714, u32_},{33882144, boolean_},{17104934, f64_},{327713, u32_},{17104935, f64_},{327716, u32_},{50659330, string_},{151322665, u64_},{327720, u32_}, };
```
astarche, line 97: Did you already discuss "properties we don't know about" with Harsha and/or Xnet? Are we OK with not exposing a way to access them?

source/codegen/metadata/nixnet/custom_proto.mako
ni-pkhare: Done.

source/custom/nixnet_service.custom.cpp
ni-pkhare: I have added a logic for both ref and ref_array type propertyids, could you pl have a look at it?

source/custom/nixnet_service.custom.cpp
ni-pkhare: Yeah, changed that and other places. Now using std::transform function

source/codegen/metadata_validation.py
ni-pkhare: Done.

source/codegen/templates/client.cpp.mako
```diff
@@ -45,6 +47,7 @@ namespace ${namespace} {
   stub_param = f"const {stub_ptr_alias}& stub"
   is_streaming = common_helpers.has_streaming_response(f)
   client_params = client_helpers.get_client_parameters(f, enums)
+  include_in_client = functions[function].get('include_in_client', True)
```
reckenro, line 50: `include_in_client` isn't used.

generated/nixnet/nixnet_service.h
```diff
@@ -31,6 +31,8 @@ struct NiXnetFeatureToggles
   bool is_enabled;
 };
 
+enum TypeIdentifier { db_ref_array_, string_, boolean_, u64_, u32_, db_ref_, f64_, u8_array_, u32_array_, parent_ref_, structure_, i32_, string_array_, dev_ref_array_, intf_ref_array_, dev_ref_,  };
```
reckenro, line 34: I don't see any references to the types {parent_ref_, structure_, dev_ref_array_, intf_ref_array_, dev_ref_} in the custom service implementations. I was curious if we have follow up work to handle these property types.

reckenro (approved):
My main concern is about handling some of the types I mentioned in xnet_service.h's `TypeIdentifier` that I didn't see in any of the switch statements in the custom service implementations.

If you end up addressing it in this PR can you reset me? Otherwise I'm fine if it's tracked and handled in a follow up change.

generated/nixnet/nixnet_service.h
ni-pkhare: There will be a follow up PR for adressing all these typeidentifiers that you have mentioned. This PR handlee support for most commonly used types. Those will be added in next PR.

ni-pkhare: Apart from structures, the others will most probably be changed to db_ref or db_ref_ in enums.py and hence won't be autogenerated here as well. If not then they will be addressed in next PR else I'll make that change in the same PR. I have put out an email to the XNET expert to clarify if these can be used as db_ref or db_ref_array or not.

mgumble (approved):

astarche (approved):

source/codegen/service_helpers.py
ni-pkhare: Done

source/codegen/templates/client.cpp.mako
ni-pkhare: Done

source/codegen/templates/client.h.mako
```diff
@@ -50,6 +51,7 @@ using namespace nidevice_grpc::experimental::client;
   stub_param = f"const {stub_ptr_alias}& stub"
   is_streaming = common_helpers.has_streaming_response(f)
   client_params = client_helpers.get_client_parameters(f, enums)
+  include_in_client = functions[function].get('include_in_client', True)
```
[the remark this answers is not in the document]
ni-pkhare: Done

source/codegen/common_helpers.py
```diff
@@ -415,6 +415,16 @@ def pascal_to_snake(pascal_string):
...
+    functions_for_proto = {"public", "CustomCode", "CustomCodeCustomProtoMessage"}
+    return [
+        name
+        for name, function in functions.items()
+        if function.get("codegen_method", "public") in functions_for_proto
+    ]
+
+
+def filter_proto_rpc_functions_for_message(functions):
     """Return function metadata only for functions to include for generating proto rpc methods."""
```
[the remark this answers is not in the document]
ni-pkhare: Done

PR #595, opened 2022-03-09
Initial implementation of nx functions

danestull:

Add several of the nx socket methods.

This begins adding the needed socket functions for xnet socket api functionality. 

All builds and tests pass
Screen Shot 2022-03-09 at 3 00 29 PM

### DIFF
--- a/source/codegen/metadata/nixnetsocket/functions.py
+++ b/source/codegen/metadata/nixnetsocket/functions.py
@@ -25,33 +25,6 @@ functions = {
         ],
         'returns': 'int32_t'
     },
-    'Accept': {
-        'cname': 'nxaccept',
-        'parameters': [
-            {
-                'direction': 'in',
-                'name': 'socket',
-                'type': 'nxSOCKET'
-            },
-            {
-                'direction': 'in',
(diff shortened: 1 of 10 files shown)


source/codegen/metadata/nixnetsocket/functions.py
```diff
@@ -25,6 +25,177 @@
...
+            },
+            {
+                'direction': 'in',
+                'name': 'flags',
+                'type': 'int32_t'
+            },
+        ],
+        'returns': 'int32_t'
+    },
+    'RecV': {
```
astarche, line 151: Why `RecV` instead of `Recv`?

astarche, line 151: Why `RecV` instead of `Recv`?

astarche (approved):

reckenro (approved):

source/codegen/metadata/nixnetsocket/functions.py
danestull: For some reason I thought `Rec` was short for receive and the `v` was something different and not that the whole `recv` was shorthand for Receive. I've updated it now

PR #597, opened 2022-03-09
Add get and set sockopt methods for xnetsocket service.

reckenro:

Includes the getsockopt and setsockopt methods to the xnetsocket gRPC service.

Fixes [AB#1872848](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1872848).

TODO: Still figuring out how to get system tests running.

### DIFF
--- a/generated/nixnetsocket/nixnetsocket.proto
+++ b/generated/nixnetsocket/nixnetsocket.proto
@@ -64,7 +64,7 @@ message SockAddr {
 }
 
 message SockOptData {
-  int32 opt = 1;
+  SocketOptions opt = 1;
   oneof data {
     int32 data_int32 = 2;
     bool data_bool = 3;
--- a/generated/nixnetsocket/nixnetsocket.proto
+++ b/generated/nixnetsocket/nixnetsocket.proto
@@ -21,6 +21,7 @@ service NiXnetSocket {
   rpc Close(CloseRequest) returns (CloseResponse);
   rpc GetLastErrorNum(GetLastErrorNumRequest) returns (GetLastErrorNumResponse);
   rpc GetLastErrorStr(GetLastErrorStrRequest) returns (GetLastErrorStrResponse);
+  rpc GetSocketOption(GetSocketOptionRequest) returns (GetSocketOptionResponse);
   rpc IsSet(IsSetRequest) returns (IsSetResponse);
   rpc Select(SelectRequest) returns (SelectResponse);
   rpc SetSocketOption(SetSocketOptionRequest) returns (SetSocketOptionResponse);
@@ -29,6 +30,18 @@ service NiXnetSocket {
 
 enum NiXnetSocketAttribute {
   NIXNETSOCKET_ATTRIBUTE_UNSPECIFIED = 0;
+}
+
+enum SocketOptions {
+  SOCKET_OPTIONS_UNSPECIFIED = 0;
+  SOCKET_OPTIONS_RX_DATA = 257;
+  SOCKET_OPTIONS_RCV_BUF = 258;
+  SOCKET_OPTIONS_SND_BUF = 259;
+  SOCKET_OPTIONS_NON_BLOCK = 260;
+  SOCKET_OPTIONS_BIND_TO_DEVICE = 261;
+  SOCKET_OPTIONS_ERROR = 262;
+  SOCKET_OPTIONS_LINGER = 263;
+  SOCKET_OPTIONS_REUSE_ADDR = 264;
 }
 
 message SockAddrIPv4 {
@@ -51,7 +64,7 @@ message SockAddr {
 }
 
 message SockOptData {
-  int32 opt = 1;
+  SocketOptions opt = 1;
   oneof data {
     int32 data_int32 = 2;
     bool data_bool = 3;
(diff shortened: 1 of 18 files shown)


reckenro:
I'm getting a build version of these changes up for draft review while I switch over to testing NR/LTE.

generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -21,15 +21,29 @@ service NiXnetSocket {
...
   rpc Select(SelectRequest) returns (SelectResponse);
+  rpc SetSocketOption(SetSocketOptionRequest) returns (SetSocketOptionResponse);
   rpc Socket(SocketRequest) returns (SocketResponse);
 }
 
 enum NiXnetSocketAttribute {
   NIXNETSOCKET_ATTRIBUTE_UNSPECIFIED = 0;
 }
 
+enum SocketOptions {
```
reckenro, line 35: TODO: This isn't the exhaustive list of options that can be used in Get or SetSocketOption. It seems like it can at least also be used for some TCP options like TCP_NO_DELAY (see [here in SocketTests.cpp](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/xnet/xnetCored/nixntipstack_tests/source/nixntipstackt/SocketTests.cpp&version=GC72eb6a7f4204d6ae570dfcd207c804c1d7817ca5&line=3665&lineEnd=3666&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)).

source/custom/nixnetsocket_service.custom.cpp
```diff
@@ -1,2 +1,94 @@
...
+
+// Returns true if it's safe to use outputs of a method with the given status.
+inline bool status_ok(int32 status)
+{
+  return status >= 0;
+}
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+::grpc::Status NiXnetSocketService::GetSocketOption(::grpc::ServerContext* context, const GetSocketOptionRequest* request, GetSocketOptionResponse* response)
```
reckenro, line 21: This was getting to complicated to codegen in a straightforward way, so I opted to custom implement it. We might be able to codegen this instead but I'd rather get the service implementation reviewed first before attempting to get it generating.

source/custom/nixnetsocket_service.custom.cpp
```diff
@@ -1,2 +1,94 @@
...
+      case nixnetsocket_grpc::GetSocketOptionRequest::OptNameEnumCase::OPT_NAME_ENUM_NOT_SET: {
+        return ::grpc::Status(::grpc::INVALID_ARGUMENT, "The value for opt_name was not specified or out of range");
+        break;
+      }
+    }
+
+    response->mutable_optval()->set_opt(request->opt_name());
+    nxsocklen_t optlen{};
+    int32_t status;
+    switch (request->opt_name()) {
```
reckenro, line 49: This is ugly, but seemed like the most straightforward way to allocate the void* passed to the library method and set it back the output response. I tried for a bit with `convert_to_grpc` but wasn't sure how to pass the reference to the data to be filled by GetSocketOption in the first place without doing something like what's seen here.

source/custom/xnetsocket_converters.h
```diff
@@ -149,5 +149,72 @@ inline TimeValHolder convert_from_grpc(const pb_::Duration& input)
...
+  // size() method is used to simplify codegen calculating the size of the
+  // selected data.
+  nxsocklen_t size() const
+  {
+    switch (opt_data.data_case()) {
+      case SockOptData::DataCase::kDataInt32:
+        return sizeof(int32_t);
+        break;
+      case SockOptData::DataCase::kDataString:
+        return opt_data.data_string().size();
```
reckenro, line 202: TODO: Check if this needs to be based off c string length instead.

source/tests/system/nixnetsocket_driver_api_tests.cpp
```diff
@@ -136,6 +136,27 @@ TEST_F(NiXnetDriverApiTests, InvalidSocket_Select_ReturnsAndSetsExpectedErrors)
   EXPECT_XNET_ERROR(SOCKET_COULD_NOT_BE_FOUND_ERROR, select_last_error);
   EXPECT_THAT(SOCKET_COULD_NOT_BE_FOUND_MESSAGE, select_last_error_str.error());
 }
+
+TEST_F(NiXnetDriverApiTests, Socket_GetSocketOptionMulticastTTL_ResponseIsReasonable)
```
reckenro, line 140: These tests fail on EXPECT_SUCCESS with -1 response. Not sure if we're at the point to be able to test these or not, need to work on them and cross reference with SocketTests.cpp.

source/custom/nixnetsocket_service.custom.cpp
```diff
@@ -1,2 +1,94 @@
...
+        status = library_->GetSocketOption(socket, level, opt_name, &data_bool, &optlen);
+        response->set_status(status);
+        if (status_ok(status)) {
+          response->mutable_optval()->set_data_bool(data_bool);
+        }
+        break;
+      }
+      case SocketOptions::SOCKET_OPTIONS_BIND_TO_DEVICE:
+      case SocketOptions::SOCKET_OPTIONS_ERROR: {
+        std::string data_string(256 - 1, '\0');  // Guessing here that max is 256... don't know how to figure out actual max size for string options.
```
reckenro, line 75: TODO: Figure out what size to allocate for the string socket options... I guessed 256 up front but that was arbitrary based on past max values for error strings.

generated/nixnetsocket/nixnetsocket.proto
astarche: OK, that is at least nxTCP_NODELAY. I only see one such option in the header. It we *can't* get an exhaustive list: that's a problem for this implementation.

reckenro:
Abandoning this and splitting into SetSockOpt PR (first) and later GetSockOpt.

PR #600, opened 2022-03-10
Add SetSockOpt to xnet socket service.

reckenro:

Add SetSockOpt to xnetsocket service.

- Adds new metadata tag "proto_only" for parameters we wish to only expose in the proto file (doesn't map directly to param for API). This allows the client to pass in a SockOptData message for the data field to the SetSockOpt method. Then the other inputs to the driver API can be derived from it which simplifies things.

Fixes [AB#1872848](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1872848)

Added four new conversion tests for the SockOptDataHolder and all XnetConvertersTests pass.


source/codegen/metadata/nixnetsocket/custom_proto.mako
```diff
@@ -16,3 +16,27 @@ message SockAddr {
     SockAddrIPv6 ipv6 = 2;
   }
 }
+
+enum OptName {
```
reckenro, line 20: The majority of these came from the "Socket options" #define section in the header. Looking at SocketTests.cpp, however, showed a few more being used like TCP_NODELAY and IP_ and IPv6_ options. I only included the ones I saw in the example tests but we need to iron out this list and determine if it is exhaustive or not.

If it ends up not being exhaustive, I can keep it but make the SockOptData `opt_name` field  be a oneof to either input a known option or pass in a raw value for the option. Either way should work fine with the conversion for SetSockOpt since the client's setting the data type (bool, int, or string).

If it's not exhaustive though, for GetSockOpt I think I'll need an extra param indicating the type (since it can't be derived from opt_name) or seperate out into different methods for each type.

reckenro, line 20: I planned to have this enum generated and in enums.py but it wasn't picked up as a function enum during proto generation since it wasn't marked on a parameter anywhere in functions.py. I planned to wait until I add GetSockOpt which will include the enum on a param to move it over to enums.py.

source/custom/xnetsocket_converters.h
```diff
@@ -149,5 +149,68 @@ inline TimeValHolder convert_from_grpc(const pb_::Duration& input)
...
+        return sizeof(bool);
+      default:
+        return 0;
+        break;
+    }
+  }
+};
+
+template <typename TSockOptData>
+inline SockOptDataHolder convert_from_grpc(const SockOptData& input)
```
astarche, line 210: nit: change this to SockOptDataInputConverter

source/custom/xnetsocket_converters.h
```diff
@@ -149,5 +149,68 @@ inline TimeValHolder convert_from_grpc(const pb_::Duration& input)
 {
   return TimeValHolder(input);
 }
+
+struct SockOptDataHolder {
+  SockOptData opt_data;
+  int32_t data_int;
```
astarche, line 155: Can you change this to a `union` like `SocAddrHolder` (now `SockAddrInputConverter`)? Or make an argument for why not? You would need a switch case in the constructor because, like with `oneof`s, you need to only assign one value. But it makes it clearer that only one case is a "live" value.

generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -49,6 +50,30 @@ message SockAddr {
...
+  OPT_NAME_SO_SND_BUF = 259;
+  OPT_NAME_SO_NON_BLOCK = 260;
+  OPT_NAME_SO_BIND_TO_DEVICE = 261;
+  OPT_NAME_SO_ERROR = 262;
+  OPT_NAME_SO_LINGER = 263;
+  OPT_NAME_SO_REUSE_ADDR = 264;
+}
+
+message SockOptData {
+  OptName opt_name = 1;
```
astarche, line 69: Check for understanding: The reason for having opt_name in the struct is to make it easier to implement the converter right? So we have all the data we need from a single grpc type? I think that's OK. it seems like mildly-negative-to-neutral from a client perspective.

I think we talked about this a little bit speculating that maybe doing this allows us to pair the valid value with the correct type in the type system. Like a tagged union in typescript. But I don't think that `protobuf` gives us a way to do that, at least not using enums. So the behavior is the same as if you separated the name from the data.

astarche (approved):

source/codegen/metadata/nixnetsocket/custom_proto.mako
reckenro: In new update it's in enums.py and generating.

source/custom/xnetsocket_converters.h
reckenro: Done.

generated/nixnetsocket/nixnetsocket.proto
reckenro: Actually, I don't know if it made it easier to implement the converter. I thought it was nice to have them coupled but I don't think that's true. it makes more sense having SockOptData just hold the oneof data and have the user use OptName enum for the optname input.

Yeah, I think that was the initial reason I had them together but it dosn't make sense in the current state. Now SockOptData just has the oneof data field and the user inputs a separate field on the request for the opt name.

source/custom/xnetsocket_converters.h
reckenro: I tried but I think it was having trouble having a string in a union. I tried messing with it a little with a destructor in the case the string is set to destruct it but I think it came down to the compiler not liking the string as one of the data members.

astarche: Ah, makes sense. Yeah, leave it then. I think `std::variant` would be best but we can't yet use that on nilrt.

PR #605, opened 2022-03-15
[Owners Review] Added CAN example for XNET

kothariaman:
Implemented CAN examples for XNET. Ported and replicated them from C to python

Existing C examples are ported to Python using gRPC API's through the gRPC server

Locally tested the python version and matched the output with the C API.
It reports a grpc error when server is not running
image


examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
+
```
ni-pkhare, line 1: Add a comment stating what this example actually does and some additional information on getting started. you can see what other have done.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+            value_buffer = value_Buffer
+        ))
+    
+    if response.status == SUCCESS: 
+        print("Signals sent:")
+        print("Signal 1: ", value_Buffer[0])
+        print("Signal 2: ", value_Buffer[1],end="\n\n")
+        i += 1
+        if i > 10 :
+            i = 0
```
ni-pkhare, line 77: This can be simplified to `i = i+1 if i <= 10 else 0`

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+SUCCESS = 0
+
+# Display Error Function
+def displayerror_andexit(Status, Source):
+    status_string = [None] * 1024
+    nixnet_types.StatustoString(Status,status_string.__sizeof__(),Source) 
+    print("\n\nERROR at {}!\n{}\n", Source, status_string)
+    print(Status)
+    print("\nExecution stopped.\nPress any key to quit\n")
+
```
ni-pkhare, line 25: Not sure, if this blank line is required. Probably we can remove it and it would still look readable. Ill leave that to you.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+DATABASE = "NIXNET_example"
+CLUSTER = "CAN_Cluster"
+SIGNAL_LIST = "CANEventSignal1,CANEventSignal2"
+NUM_SIGNALS = 2
+SUCCESS = 0
+
+# Display Error Function
+def displayerror_andexit(Status, Source):
+    status_string = [None] * 1024
+    nixnet_types.StatustoString(Status,status_string.__sizeof__(),Source)
```
ni-pkhare, line 21: Spaces after `,` while passing arguments to `StatusToString`. Something like `nixnet_types.StatustoString(Status, status_string.__sizeof__(), Source)`

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+# Parameters
+INTERFACE = "CAN1"
+DATABASE = "NIXNET_example"
+CLUSTER = "CAN_Cluster"
+SIGNAL_LIST = "CANEventSignal1,CANEventSignal2"
+NUM_SIGNALS = 2
+SUCCESS = 0
+
+# Display Error Function
+def displayerror_andexit(Status, Source):
```
ni-pkhare, line 19: Shouldn't the function name be `display_error_and_exit` ? Is this name used in the C Example? Anyways we can still change it to make it more clear and readable.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+
+# Parameters
+INTERFACE = "CAN1"
+DATABASE = "NIXNET_example"
+CLUSTER = "CAN_Cluster"
+SIGNAL_LIST = "CANEventSignal1,CANEventSignal2"
+NUM_SIGNALS = 2
+SUCCESS = 0
+
+# Display Error Function
```
ni-pkhare, line 18: Since this a local function, the description of this function should come inside the body. Something like this.

```
def display_error_and_exit(Status, Source)::
"""Display error message and exit."""
... rest of the code
```

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+    # Update the signal data
+    response = client.WriteSignalSinglePoint(
+        nixnet_types.WriteSignalSinglePointRequest(
+            session_ref = m_SessionRef,
+            value_buffer = value_Buffer
+        ))
+    
+    if response.status == SUCCESS: 
+        print("Signals sent:")
+        print("Signal 1: ", value_Buffer[0])
```
ni-pkhare, line 73: Here we can use `f` strings. Something like this `print(f"Signal 1: {value_buffer[0]}")`

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
...
+        print("Signal 2: ", value_Buffer[1],end="\n\n")
```
ni-pkhare, line 74: Similarly here. `print(f"Signal 2: {value_Buffer[1]\n}")`

ni-pkhare:
The python build failures can be corrected by installing ni-python-style-guide and black. I dont remember how to do that. You could check it on wiki of grpc-device channel in Teams.

ni-pkhare (approved):

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,87 @@
+
+import grpc
+import nixnet_pb2 as nixnet_types
+import nixnet_pb2_grpc as grpc_nixnet
+import getch
+
+SERVER_ADDRESS = "localhost"
+SERVER_PORT = "31763"
+
```
ni-pkhare, line 9: We should also add a logic to take Server_adress and server_port as command line arguments from the clients. See other examples.

kothariaman: Added this logic

doshirohan: `pip install ni-python-styleguide` and `pip install black`. 

To execute them:

> ni-python-styleguide --exclude .\source\codegen\metadata\ lint .\source\codegen\ .\examples\

This will show you any errors flagged by ni styleguide

> black source/codegen examples

This command will report any issues flagged by black and fix it as well.

Here's the wiki that has some relevant information
[grpc-device (Teams) -> wiki](https://teams.microsoft.com/l/entity/com.microsoft.teamspace.tab.wiki/tab::7426a53d-3098-4c17-9bb0-5ba77dc51c7f?context=%7B%22subEntityId%22%3A%22%7B%5C%22pageId%5C%22%3A7%2C%5C%22origin%5C%22%3A2%7D%22%2C%22channelId%22%3A%2219%3A658a48f1bcd240dc9c56f5eb35761898%40thread.tacv2%22%7D&tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e)

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,119 @@
+r"""
+ This example writes a signal value when a keyboard character is pressed.
+ This is used to demonstrate a signal single point output session. 
+ This example uses hardcoded signal names that use the NIXNET_example database. 
+
+The gRPC API is built from the C API. NI-XNET documentation is installed with the driver at:
+  C:\Program Files (x86)\IVI Foundation\IVI\Drivers\niXNET\Documentation\NIXNETCref.chm
+Getting Started:
+
+To run this example, install "NI-DCPower Driver" on the server machine:
```
ni-pkhare, line 10: Everywhere it says NI-DCPOWER, pl change it accordingly.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,119 @@
...
+  https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+
+Refer to the NI DCPOWER gRPC Wiki for the latest C Function Reference:
+  https://github.com/ni/grpc-device/wiki/NI-XNET-C-Function-Reference
+
+ Running from command line:
+
+Server machine's IP address, port number, and resource name can be passed as separate command line
+arguments.
+  > python measure-record.py <server_address> <port_number>
```
ni-pkhare, line 24: Change it to use XNET example, right now this is the dcpower example

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
+r""" Write Signal Data.
```
doshirohan, line 1: Change file name to follow conventions for python.

Maybe `can-signal-single-point-output.py`?

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+
+ This example writes a signal value when a keyboard character is pressed.
+ This is used to demonstrate a signal single point output session. 
+ This example uses hardcoded signal names that use the NIXNET_example database. 
+
+The gRPC API is built from the C API. NI-XNET documentation is installed with the driver at:
+  C:\Program Files (x86)\IVI Foundation\IVI\Drivers\niXNET\Documentation\NIXNETCref.chm
```
doshirohan, line 8: Documentation is installed at
> C:\Users\Public\Documents\National Instruments\NI-XNET\Documentation\NI-XNET Manual.chm

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+  https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+
+Refer to the NI XNET gRPC Wiki for the latest C Function Reference:
+  https://github.com/ni/grpc-device/wiki/NI-XNET-C-Function-Reference
+
+ Running from command line:
+
+Server machine's IP address, port number, and resource name can be passed as separate command line
+arguments.
+  > python CAN_SIGNAL_SINGLE_POINT_OUTPUT.py <server_address> <port_number>
```
doshirohan, line 25: Update command once you rename the file.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+For instructions on how to use protoc to generate gRPC client interfaces, see our "Creating a gRPC
+Client" wiki page:
+  https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+
+Refer to the NI XNET gRPC Wiki for the latest C Function Reference:
+  https://github.com/ni/grpc-device/wiki/NI-XNET-C-Function-Reference
+
+ Running from command line:
+
+Server machine's IP address, port number, and resource name can be passed as separate command line
```
doshirohan, line 23: There is no "resource name" parameter in this case. Fix this line.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+SIGNAL_LIST = "CANEventSignal1,CANEventSignal2"
+NUM_SIGNALS = 2
+SUCCESS = 0
+
+# Display Error Function
+def display_error_andexit(status, source):
+    """Display error message and exit."""
+    response = client.StatusToString(nixnet_types.StatusToStringRequest(status_id=status))
+    print("\n\nERROR at {}!\n{}\n", source, response.status_description)
+    print("\nExecution stopped.\nPress any key to quit\n")
```
doshirohan, line 59: Doesn't look like this line is going to wait here for user input. So this message doesn't make sense. Make these changes to keep consistency with other gRPC examples we ship out of this repo.

- I think instead of printing error and exiting, you can `raise Exception(response.status_description)` here. That will also take care of printing call stack, so users would know which line resulted in error.

- Also, put main body of this file in try-except-finally block. `except` can catch RPC errors and print RPC specific error message. `finally` block can be used for closing session if  `session_reference.id` is not zero (id != 0 makes sure that error wasn't during session initialization and we should close the sesssion).

- Rename this method to check_for_error and check for `status != SUCCESS` here to raise exception conditionally. Callers below can call this API unconditionally after each gRPC calls and print messages you have in success scenarios after the `check_for_error`. Those won't be executed in case of errors because of `raise Exception` here. 

Look at other examples to refer to this pattern.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+    client.Clear(nixnet_types.ClearRequest(session_ref=session_reference))
+    exit(1)
+
+
+i = 0
+session_reference = 0
+value_buffer = [0] * NUM_SIGNALS
+
+# Create the communication channel for the remote host and create connections to the NI-DCPower and
```
doshirohan, line 68: ~NI-DCPower~ => NI-XNET

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+    """Display error message and exit."""
+    response = client.StatusToString(nixnet_types.StatusToStringRequest(status_id=status))
+    print("\n\nERROR at {}!\n{}\n", source, response.status_description)
+    print("\nExecution stopped.\nPress any key to quit\n")
+    client.Clear(nixnet_types.ClearRequest(session_ref=session_reference))
+    exit(1)
+
+
+i = 0
+session_reference = 0
```
doshirohan, line 65: You don't really need to declare `session_reference` variable in advance.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+# Create the communication channel for the remote host and create connections to the NI-DCPower and
+# session services.
+channel = grpc.insecure_channel(f"{SERVER_ADDRESS}:{SERVER_PORT}")
+client = grpc_nixnet.NiXnetStub(channel)
+
+# Display parameters that will be used for the example.
+print("Interface: " + INTERFACE, "Database: " + DATABASE, "Signal List: " + SIGNAL_LIST, sep="\n")
+
+# Create an XNET session in SignalOutSinglePoint mode
+CreateSession_Response = client.CreateSession(
```
doshirohan, line 77: Change variable names to follow lower_snake_case convention.

`create_session_response`

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+    display_error_andexit(CreateSession_Response.status, "CreateSession")
+
+print("\nPress any key to transmit new signal values or q to quit\n")
+
+while not (getch.getch()).decode("UTF-8") == "q":
+    value_buffer[0] = float(i)
+    value_buffer[1] = float(i * 10)
+
+    # Update the signal data
+    response = client.WriteSignalSinglePoint(
```
doshirohan, line 101: Every gRPC call is going to have a response. Let's name this based on which API this is for. `write_signal_response`. Same for the `Clear` call below as well.

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+        nixnet_types.WriteSignalSinglePointRequest(
+            session_ref=session_reference, value_buffer=value_buffer
+        )
+    )
+
+    if response.status == SUCCESS:
+        print("Signals sent:")
```
doshirohan, line 108: Do you want to add newline after each of these print statements or does it look good as is?

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
```diff
@@ -0,0 +1,120 @@
...
+        print(f"Signal 1: {value_buffer[0]}")
+        print(f"Signal 2: {value_buffer[1]}", end="\n\n")
+        i = i + 1 if i < 10 else 0
+    else:
+        display_error_andexit(response.status, "WriteSignalSinglePoint")
+
+# Clears the X-NET Session
```
doshirohan, line 115: Clears => Clear
X-NET => XNET (it doesn't refer to network of x-men ;P )

doshirohan (approved):

examples/nixnet/CAN_SIGNAL_SINGLE_POINT_OUTPUT.py
doshirohan: Put screenshot of output in your PR description. It would easier for Harsha as well to visualize the output.

kothariaman: It look good as it is currently. I will update the screenshot in PR description.

Akshata440 (approved):

astarche:
Both `black` and `ni-python-styleguide` are in `python_build_requirements.txt`. One way to get those is to active the venv that is created in `build \venv`. When you are editing in grpc-device with vscode, you should select that as your venv, which will give you those tools in your command prompt etc. The wiki does have instructions for how to configure black in vscode.

`validate_examples.py` will install a lot of stuff in the poetry project that it creates for validating examples. It actually doesn't include `ni-python-styleguide`, but it could.

I don't think we've defined a best practice for how to run ni-python-styleguide during development. It's not quite as straightforward as configuring black in your IDE. You should be able to enable it via `flake8` ide integration using our special ruleset, but I don't have a step-by-step for how to do that. But *you can* run it from the command line when you are in a venv with in installed, which is generally what I have been doing. FYI @DavidCurtiss

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
...
+SERVER_PORT = "31763"
+
+# Read in cmd args
+if len(sys.argv) >= 2:
+    SERVER_ADDRESS = sys.argv[1]
+if len(sys.argv) >= 3:
+    SERVER_PORT = sys.argv[2]
+
+# Parameters
+INTERFACE = "CAN1"
```
astarche, line 46: Consider; We usually allow the device name to be passed in as a command line argument as well. Does that work in this case? Or are there just too many names that need to match your MAX config?

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Update the signal data
+        write_signal_response = client.WriteSignalSinglePoint(
+            nixnet_types.WriteSignalSinglePointRequest(
+                session_ref=session_reference, value_buffer=value_buffer
+            )
+        )
+        check_for_error(write_signal_response.status)
+
+        print("Signals sent:")
```
astarche, line 104: Check for understanding: you say in the PR that the output is the same as the C API. But this is only printing data from the write buffer. So all that validates (for now) is that write didn't return an error?

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
...
+    if rpc_error.code() == grpc.StatusCode.UNAVAILABLE:
+        error_message = f"Failed to connect to server on {SERVER_ADDRESS}:{SERVER_PORT}"
+    elif rpc_error.code() == grpc.StatusCode.UNIMPLEMENTED:
+        error_message = (
+            "The operation is not implemented or is not supported/enabled in this service"
+        )
+    print(f"{error_message}")
+
+finally:
+    if "session_reference" in vars() and session_reference.id != 0:
```
astarche, line 120: I'm pretty sure that this can all just be `if session_reference:`.

Also make sure that you include some error cases in your testing (and include that in the PR notes).

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
...
+    print("\nPress any key to transmit new signal values or q to quit\n")
+
+    while not (getch.getch()).decode("UTF-8") == "q":
+        value_buffer[0] = float(i)
+        value_buffer[1] = float(i * 10)
+
+        # Update the signal data
+        write_signal_response = client.WriteSignalSinglePoint(
+            nixnet_types.WriteSignalSinglePointRequest(
+                session_ref=session_reference, value_buffer=value_buffer
```
astarche, line 99: nit: I don't like this long name `session_reference` and it's going to be all over the place. What about just `session`?

Also, @doshirohan have y'all considered changing `session_ref` to `session`? I think that came up in a previous PR.

astarche (approved):

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
...
+Server machine's IP address and port number can be passed as separate command line
+arguments.
+  > python can-signal-single-point-output.py <server_address> <port_number>
+If they are not passed in as command line arguments, then by default the server address will be
+"localhost:31763"
+"""
+
+import sys
+
+import getch
```
astarche, line 31: Consider: Do we really need an external dependency for `q`-to-quit? There are several examples that use `ctrl+c` and maybe one that uses esc. Will one of those work for you instead? Can you require <enter> for a newline instead of a character? Can you just run 10 times without user input?

Also, if you add this, you need to add it to validate_examples as a dependency to install.

astarche:
The current failure is from `validate_examples` which you can run locally. It is complaining about methods that are not defined (are you using something that isn't in the main proto yet?). And also about a missing dependency (which I mentioned in a comment; you should remove. but, if you do need an external dep, you can add it to validate_examples in the `poetry add` statement).

examples/nixnet/can-signal-single-point-output.py
kothariaman: Yeah, we have implemented the same as passing the device name as interface in command line argument.

examples/nixnet/can-signal-single-point-output.py
kothariaman: No, the function `check_for_error` raises a exception whenever there is a error and the print statements will be executed only if the status doesn't indicate a error.

examples/nixnet/can-signal-single-point-output.py
kothariaman: Sure

examples/nixnet/can-signal-single-point-output.py
kothariaman: Yeah, so we are going ahead with running this 10 times without user input and printing output. Removing the `getch` dependency.

examples/nixnet/can-signal-single-point-output.py
kothariaman: Renaming this to session.

examples/nixnet/can-signal-single-point-output.py
astarche: That's what I was thinking, sorry if my message was not clear.

That sounds fine.

kothariaman: @astarche  I have reset your vote. Could you please review and approve the changes once again?

astarche (approved):

examples/nixnet/can-signal-single-point-output.py
doshirohan: I just noticed this comment. @astarche we do have a task to address your earlier PR comment to fix session reference param names (and also enum names that you had called out on other PR). We will be fixing up examples as well during that change.

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
+r""" Write Signal Data.
```
doshirohan, line 1: There seems to be an empty `PortExamplesC-Python` file above. I couldn't add a comment there, so adding it here. Make sure remove that file before you complete the PR.

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -0,0 +1,125 @@
...
+
+ This example writes a signal value when a keyboard character is pressed.
```
doshirohan, line 3: >  writes a signal value when a keyboard character

This behavior has changed. Make sure to update the comment.

reckenro (approved):

PR #606, opened 2022-03-15
XNET StatusToString, DbGetDBCAttributeSize, DbGetDBCAttribute API and interface name sanitization support

ni-pkhare:

This PR adds support for `StatusToString` , `DbGetDBCAttributeSize`, `DbGetDBCAttribute` API of XNET.
Alongs side, I have added sanitized name support in client files

# Why should this Pull Request be merged?

For StatusToString API, have made some codegen changes to not generate `auto status =` sentence before library call since the function returns `void`.
Hence now the library call should look like `library_->StatusToString(...)` instead of 
`auto status = library_->StatusToString(...)`

`Note: After making the sanitization changes to the client files, now for some more drivers there are changes in client files which makes them consistent with service files.`

I have custom implementation of DbGetDBCAttribute API because it was similar to Get/Set APIs where size of output buffer had to be calculated using DBGetDBCAttributeSize API


astarche (changes requested):
await: keep the name as-is. Make sure that the client wrapper code is using `sanitize_names`.

generated/nixnet/nixnet.proto
```diff
@@ -918,7 +919,7 @@ message CreateSessionResponse {
 message CreateSessionByRefRequest {
   string session_name = 1;
   repeated nidevice_grpc.Session array_of_database_ref = 2;
-  string interface = 3;
```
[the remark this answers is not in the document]
doshirohan: @asumit the issue you ran into was while trying to compile a C application right? From what I remember C client wasn't compiling because `interface` is keyword in C (which means users won't be use xnet-gRPC in C). Is that correct?

source/codegen/templates/service_helpers.mako
```diff
@@ -690,15 +690,17 @@ ${initialize_standard_input_param(function_name, parameter)}
 
 
 ## Handles populating the response message after calling the driver API.
-<%def name="populate_response(output_parameters, indent_level=0, init_method=False)">\
+<%def name="populate_response(function_data, output_parameters, indent_level=0, init_method=False)">\
 <%
   get_last_error_outputs = service_helpers.get_last_error_output_params(output_parameters)
   normal_outputs = [p for p in output_parameters if not p in get_last_error_outputs]
+  return_type = service_helpers.get_function_return_type(function_data)
```
doshirohan, line 697: Instead of making this change here, you could add `'status_expression' : '0'` attribute on `StatusToString` API. That way in metadata we are tagging the function to always succeed and status variable will be generated (with value of `status_expression`) for this API as well. Then you don't need this additional check of setting `response->status` differently here.

doshirohan (approved):

source/codegen/templates/service_helpers.mako
ni-pkhare: yeah you are right. I was not aware of this. I have made the change.

generated/nixnet/nixnet.proto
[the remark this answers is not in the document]
ni-pkhare: I have added support to use sanitize_names in client files as well. Besides XNET, there was 1 API in Scope which was also using a reserved keyword "event" which got corrected as part of sanitization support.

astarche (approved):

source/codegen/metadata/nixnet/functions.py
```diff
@@ -507,6 +508,7 @@
             },
             {
                 'direction': 'in',
+                'include_in_proto': False,
```
reckenro, line 511: [Minor] For accuracy, can you update `attributeText`'s direction to 'out' on line 516. 

I think after you update it you'll have to either add size mechanism (you could use passed-in with value 'attributeTextSize') or add a size suppression to functions_addon.py like what's done for xnet sockets: https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nixnetsocket/functions_addon.py.

source/codegen/client_helpers.py
```diff
@@ -6,6 +6,7 @@
 from typing import List, Tuple
 
 import common_helpers
+import metadata_mutation
```
reckenro, line 9: I don't think you need this or the private `_get_sanitized_client_parameters` method calling sanitize_names. It should happen before any of the files (including client files) are generated. The cppName updates should suffice.

reckenro (approved):

source/codegen/metadata/nixnet/custom_proto.mako
```diff
@@ -145,4 +145,19 @@ message DbGetDatabaseListResponse {
     string alias_buffer = 2;
     string file_path_buffer = 3;
     uint32 number_of_databases = 4;
+}
+
+message DbGetDBCAttributeRequest {
```
doshirohan, line 150: Why do we need to implement these messages in custom proto? Looks like we should be able to auto generate these.

doshirohan (approved):

mgumble (approved):

source/codegen/metadata/nixnet/functions.py
ni-pkhare: Done. Thanks for noticing this.

source/codegen/metadata/nixnet/custom_proto.mako
ni-pkhare: Done using CustomCode codegen_method for this API

source/codegen/client_helpers.py
ni-pkhare: Yeah, that makes sense, now I have made the necessory changes. Thanks. Now the client files are consistent with the service files.

source/codegen/templates/client_helpers.mako
```diff
@@ -9,7 +9,8 @@ from client_helpers import ParamMechanism
 <%def name="build_request(client_params)">\
 %   for param in client_params:
 <%
-  field_name = common_helpers.get_grpc_field_name_from_str(param.name)
+  field_name = common_helpers.get_grpc_field_name_from_str(param.cppName)
```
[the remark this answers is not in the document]
ni-pkhare: Right now, I am leaving it as is, but will surely come and revisit this.

PR #609, opened 2022-03-16
Support for frame input/output APIs in XNET

doshirohan:

XNET frame related APIs accept void* parameter which could contain different types of frames. CAN/LIN/FlexRay/J1939 frames share same `struct` in C with different sized payloads. Ethernet frames use a separate struct.

- I have modified request/response messages for these APIs to pass "oneof" different types of frames instead of dealing with void buffer. Based on type of frame appropriate oneof field would be set.
- For APIs that return frames, I have added additional "proto_only" field for user to specify type of frame they are expecting so that server handle can figure out how to interpret buffer returned by C API. For APIs with input frame parameter, this information is implicit based on which oneof user has set for frame parameter.
- Custom converters have been implemented to iterate over frames using iterators provided by C library.

- This PR adds framework for read/write frame type of APIs along with codegen for one API of each type. Once this framework is submitted, I will update metadata for other APIs as well and submit the generated code.

- Executed one of examples which deals with conversion of CAN frames to and from Signals which exercises both above APIs
- Above tests conversion for LIN, FlexRay and J1939 frames as well since they use same format and share same conversion method.
- I haven't tested ENET frames conversion yet. I will be doing this once we have the hardware setup ready (might be done after completing this PR).

Note: There is issue with how different buffer sizes are calculated in server implementation. "len" mechanism calculates buffer size in terms of "number of elements", but XNET APIs expect it in number of bytes. With current implementation, we resize the buffers correctly but pass incorrect buffer_size parameter to C library. This is tracked as separate task since it affects other XNET APIs as well and will be addressed in separate PR.


generated/nixnet/nixnet_service.cpp
```diff
@@ -166,6 +167,78 @@ namespace nixnet_grpc {
...
+  ::grpc::Status NiXnetService::ConvertSignalsToFramesSinglePoint(::grpc::ServerContext* context, const ConvertSignalsToFramesSinglePointRequest* request, ConvertSignalsToFramesSinglePointResponse* response)
+  {
+    if (context->IsCancelled()) {
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      auto session_ref_grpc_session = request->session_ref();
+      nxSessionRef_t session_ref = session_repository_->access_session(session_ref_grpc_session.id(), session_ref_grpc_session.name());
+      auto value_buffer = const_cast<f64*>(request->value_buffer().data());
+      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size());
```
doshirohan, line 210: note: This calculation is incorrect. XNET expects number of bytes in `size_of_value_buffer` parameter, but we are passing in number of f64 elements here.

generated/nixnet/nixnet_service.cpp
```diff
@@ -166,6 +167,78 @@ namespace nixnet_grpc {
...
+  //---------------------------------------------------------------------
+  ::grpc::Status NiXnetService::ConvertFramesToSignalsSinglePoint(::grpc::ServerContext* context, const ConvertFramesToSignalsSinglePointRequest* request, ConvertFramesToSignalsSinglePointResponse* response)
+  {
+    if (context->IsCancelled()) {
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      auto session_ref_grpc_session = request->session_ref();
+      nxSessionRef_t session_ref = session_repository_->access_session(session_ref_grpc_session.id(), session_ref_grpc_session.name());
+      auto frame_buffer = convert_from_grpc<u8>(request->frame_buffer());
```
doshirohan, line 180: note: `frame_buffer` is of type `FrameBuffer` and uses same concept as `SockAddrInputConverter` for implicit conversion to `u8*` during library call and for the `size` calculation below.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -141,16 +140,17 @@
             {
                 'direction': 'in',
                 'name': 'frameBuffer',
-                'size': {
-                    'mechanism': 'len',
-                    'value': 'numberOfBytesForFrames'
-                },
-                'type': 'void[]'
+                'type': 'u8',
```
doshirohan, line 143: note: Changed type from `void[]` to `u8` because `validate_metadata` expects size mechanism to be present on input arrays. This was easier way to get around that.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -222,6 +224,14 @@
             {
                 'direction': 'out',
                 'name': 'numberOfBytesReturned',
+                'type': 'u32',
+                'include_in_proto': False
+            },
+            {
```
doshirohan, line 230: note: Additional `frame_type` parameter required only when API returns a frame. This helps server identify how to interpret the `void*` buffer returned by C API.

source/custom/nixnet_service.custom.cpp
```diff
@@ -803,4 +805,68 @@ ::grpc::Status NiXnetService::DbGetDatabaseList(::grpc::ServerContext* context,
   }
 }
 
-}  // namespace nixnet_grpc
\ No newline at end of file
+// template <>
```
doshirohan, line 808: @astarche I couldn't find a way to make this template instantiation. Since this method is specific to XNET, I guess this should be fine right?

source/custom/nixnet_converters.h
```diff
@@ -0,0 +1,171 @@
...
+namespace pb_ = ::google::protobuf;
+
+namespace nixnet_grpc {
+
+struct FrameHolder {
+  FrameHolder(const pb_::RepeatedPtrField<FrameBuffer>& input)
+  {
+    // all frames in repeated field must be of same type
+    switch (input[0].frame_case()) {
+      case FrameBuffer::kCan:
```
astarche, line 16: Why is it better to have a separate frame type for every one of these? Could the types just be:

```proto
enum FrameType { Frame, EnetFrame }
```

Both the underlying type and the proto type are the same. And it doesn't look like we validate that the actual *data* is a frame (could we do that by looking at the type?).

I think this implementation is OK, though. It makes a little more sense to me with just 2 frame types, but I'm not sure I understand all the considerations. 

Pros/Cons: 
* Using only 2 frame types would make it "just work" with a new frame type.
* Having a type struct for each subtype allows us to change them independently in grpc (this would be an important consideration in normal proto design. But, in this case, I'm not sure if the underlying C API representation supports that or not).
* In this implementation, is it possible for a user to request "LIN" data and then have us return an array of "LIN" frames that are actually "CAN" frames? Are we implying type-checking that we're not actually doing?

astarche (approved):
Looks really good overall!

reckenro (approved):

generated/nixnet/nixnet.proto
```diff
@@ -823,154 +825,191 @@ enum WriteState {
...
+  int32 status = 1;
+  string alias_buffer = 2;
+  string file_path_buffer = 3;
+  uint32 number_of_databases = 4;
+}
+
+message Frame {
+  uint64 timestamp = 1;
+  uint32 identifier = 2;
+  uint32 type = 3;
```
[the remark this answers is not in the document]
doshirohan: No, they are not same. There are multiple sub-types of each frame like data frame, bus error, remote etc. Type here points to that.

source/custom/nixnet_converters.h
doshirohan: I had initially thought of using Frame for LIN/CAN/FlexRay and J1939, since underlying data structure is same (difference is mostly in payload lengths, and some connotation about how `PayloadLength` gets calculated. But these differences are hidden away by macros from xnet driver.

Main reason for using separate types is mostly to do with discoverability in request/response messages. I couldn't find a single name good enough to imply that use this field to pass in either of those 4 frame types. Other name `EnetFrame` stands on its own and implies the type very well. I will discuss this with Waldemar post this PR to see if there is single name that domain users would understand. Maybe just `Frame` could be good enough too.

> In this implementation, is it possible for a user to request "LIN" data and then have us return an array of "LIN" frames that are actually "CAN" frames? Are we implying type-checking that we're not actually doing?

Yes, this is possible. We don't do any type checking of any sort on server side. I am not sure if this is completely possible either. There are few FrameType constants of each type in header file (and also few more frame types which are common for all frames). But if we could merge the frames into single field as you suggested, that would solve this problem too.

astarche: > I will discuss this with Waldemar post this PR to see if there is single name that domain users would understand. Maybe just Frame could be good enough too.

That sounds good. Whatever you decide is OK with me (including the current implementation). And you don't need to block this PR on the decision if you don't want to.

mgumble (approved):

source/custom/nixnet_converters.h
doshirohan: I discussed this with Waldemar and he mentioned having separate fields for each frame type (even though underlying struct type is same) is better. There is no single name we could use to refer to all those 4 types of frames.

Regarding validation if the frames being returned are of same type that user has requested, Waldemar said that is not required. Type of frame depends on how session is initialized. In C API, users can initialize LIN session and pass CAN frame to write API, and that doesn't throw any error either. Users might see weird output though. If users try to pass J1939 frame for LIN session, that might throw error because J1939 supports longer payload than LIN. As long as we are reporting such errors from driver back to users (which we do), this should be sufficient.

PR #610, opened 2022-03-16
Add Enums for nxSocket

danestull:

Adds defined enums from nxSocket for use.

Adding enums helps to complete the XNET use case for socket functions and helps make usage of several functions easier. 

Builds passed, updated test class to use some of the enums for the socket methods and all ran tests passed.

### DIFF
--- a/generated/nixnetsocket/nixnetsocket.proto
+++ b/generated/nixnetsocket/nixnetsocket.proto
@@ -44,11 +44,45 @@ enum NiXnetSocketAttribute {
   NIXNETSOCKET_ATTRIBUTE_UNSPECIFIED = 0;
 }
 
+enum AddressFamily {
+  ADDRESS_FAMILY_UNSPECIFIED = 0;
+  ADDRESS_FAMILY_INET = 2;
+  ADDRESS_FAMILY_INET6 = 10;
+}
+
+enum IPProtocol {
+  IP_PROTOCOL_IP = 0;
+  IP_PROTOCOL_TCP = 6;
+  IP_PROTOCOL_UDP = 8;
+  IP_PROTOCOL_IPV6 = 12;
+}
+
+enum SocketProtocolType {
+  SOCKET_PROTOCOL_TYPE_UNSPECIFIED = 0;
+  SOCKET_PROTOCOL_TYPE_STREAM = 1;
+  SOCKET_PROTOCOL_TYPE_DGRAM = 2;
+}
+
+enum SocketOptionLevel {
+  SOCKET_OPTION_LEVEL_UNSPECIFIED = 0;
+  SOCKET_OPTION_LEVEL_SOCKET = 13;
+}
+
 enum OptName {
+  option allow_alias = true;
   OPT_NAME_UNSPECIFIED = 0;
   OPT_NAME_TCP_NODELAY = 1;
-  OPT_NAME_IP_MULTICAST_TTL  = 6;
+  OPT_NAME_IP_ADD_MEMBERSHIP = 3;
+  OPT_NAME_IP_DROP_MEMBERSHIP = 4;
+  OPT_NAME_IP_MULTICAST_IF = 5;
+  OPT_NAME_IP_MULTICAST_TTL = 6;
+  OPT_NAME_IPV6_JOIN_GROUP = 12;
+  OPT_NAME_IPV6_LEAVE_GROUP = 13;
+  OPT_NAME_IPV6_ADD_MEMBERSHIP = 12;
+  OPT_NAME_IPV6_DROP_MEMBERSHIP = 13;
+  OPT_NAME_IPV6_MULTICAST_IF = 14;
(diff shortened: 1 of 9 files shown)


source/codegen/metadata/nixnetsocket/enums.py
```diff
@@ -1,18 +1,144 @@
...
+            {
+                'name': 'IP_MULTICAST_IF',
+                'value': 5
+            },
+            {
+                'name': 'IP_MULTICAST_TTL',
                 'value': 6
             },
+            {
+                'name': 'IPV6_JOIN_GROUP',
```
reckenro, line 123: Thanks for rounding these OptNames out, I think this is all of them from the header and have an email out to see if this is this is all the optnames or not.

I do think we should go ahead and add `IPV6_ADD_MEMBERSHIP` and `IPV6_DROP_MEMBERSHIP` with 12 and 13, respectively. It's duplicated with JOIN/LEAVE_GROUP but I think the codegen will properly add the allow_alias = true to allow the duplicate values in the proto Enum.

source/codegen/metadata/nixnetsocket/enums.py
```diff
@@ -1,18 +1,144 @@
...
+                'name': 'DOWN',
+                'value': 0
+            },
+            {
+                'name': 'UP',
+                'value': 1
+            }
+        ]
+    },
+     'IPStackInffoStringFormat': {
```
reckenro, line 15: Typo: 
```suggestion
     'IPStackInfoStringFormat': {
```

source/codegen/metadata/nixnetsocket/enums.py
```diff
@@ -1,18 +1,144 @@
...
+            {
+                'name': 'AF_INET6',
+                'value': 10
+            }
+        ]
+    },
+    'ProtocolFamilies': {
+        'values': [
+            {
+                'name': 'PF_UNSPEC',
```
reckenro, line 47: I wonder if it's better for the '..._Unspec' enum values where it's 0 if it's better to just leave it off and have the codegen generate the UNSPECIFIED 0 enum value.

source/codegen/metadata/nixnetsocket/enums.py
```diff
@@ -1,18 +1,144 @@
...
+            {
+                'name': 'TEXT',
+                'value': 1
+            }
+        ]
+    },
+    'AddressFamilies': {
+        'values': [
+            {
+                'name': 'AF_UNSPEC',
```
reckenro, line 31: I think we should strip prefixes like 'AF_' from enum value names when their covered by the Enum's name (AddressFamilies covers AF_, ProtocolFamilies covers PF_, etc.)

source/codegen/metadata/nixnetsocket/enums.py
```diff
@@ -1,18 +1,144 @@
...
+                'name': 'JSON',
+                'value': 0
+            },
+            {
+                'name': 'TEXT',
+                'value': 1
+            }
+        ]
+    },
+    'AddressFamilies': {
```
reckenro, line 28: I _think_ in general we should shoot for Enum names be singular. So this becomes AddressFamily. Except for Flag enums where more than one might be used in conjunction. @astarche , does this sound right?

source/codegen/metadata/nixnetsocket/enums.py
astarche: I like that idea. 

`UNSPECIFIED` is a widely accepted standard/practice. We'll stick with weird driver names pretty far, but when it's this close, I prefer UNSPECIFIED.

source/codegen/metadata/nixnetsocket/enums.py
danestull: I added these duplicate values and added the needed suppression to get allow_alias and build working

source/codegen/metadata/nixnetsocket/enums.py
danestull: I've striped these off

generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -42,11 +42,45 @@ enum NiXnetSocketAttribute {
...
+  SOCKET_OPTION_LEVEL_UNSPECIFIED = 0;
+  SOCKET_OPTION_LEVEL_SOCKET = 13;
+}
+
 enum OptName {
+  option allow_alias = true;
   OPT_NAME_UNSPECIFIED = 0;
   OPT_NAME_TCP_NODELAY = 1;
-  OPT_NAME_IP_MULTICAST_TTL  = 6;
+  OPT_NAME_IP_ADD_MEMBERSHIP = 3;
```
reckenro, line 73: Note: We'll need to handle these new OptName values in the conversion code but I don't think that should be part of this change. Added AB#1889111 to address that.

reckenro (approved):

astarche (approved):

mgumble (commented):
Looks good!

PR #614, opened 2022-03-17
Introduce len-in-bytes size mechanism for XNET APIs that require size in number of bytes

doshirohan:

This PR introduces new size mechanism `len-in-bytes` for input parameters. Existing `len` mechanism computes length of input array and passes that as size parameter to library API. Some of XNET APIs expect "number of bytes" to be passed to library calls.  So this new mechanism multiplies size of array by type of array to compute number of bytes.

I have tagged appropriate APIs from XNET with new mechanism.

Size for these APIs was passed incorrectly while calling the C API resulting in error. This new mechanism fixes this issue for XNET APIs.

- Executed python examples which exercise some of these APIs. Earlier we had observed errors during driver calls for these, now these APIs seem to return expected value.


astarche:
@doshirohan Have y'all been using `validate_examples.py` to run/lint examples? You don't need to, but it might be helpful. And the non-runtime validation will run on the PR build.

generated/nixnet/nixnet_service.cpp
```diff
@@ -1145,7 +1145,7 @@ namespace nixnet_grpc {
       auto session_ref_grpc_session = request->session_ref();
       nxSessionRef_t session_ref = session_repository_->access_session(session_ref_grpc_session.id(), session_ref_grpc_session.name());
       auto value_buffer = const_cast<f64*>(request->value_buffer().data());
-      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size());
+      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size()) * sizeof(f64);
```
astarche, line 1148: Shouldn't this go inside the `static_cast`? I think this introduces a warning because `sizeof` is a `size_t` and loses precision.

source/codegen/templates/service_helpers.mako
```diff
@@ -491,8 +491,15 @@ ${initialize_standard_input_param(function_name, parameter)}
 % else:
 <%
   size_field_name = size_sources[-1]
+  size_from_param_name = parameter["determine_size_from"][0] if isinstance(parameter["determine_size_from"], list) else parameter["determine_size_from"]
```
astarche, line 494: Consider: Would it make more sense to assert that determine_size_from is not a list when `len-in-bytes` is set? Linked size params don't necessarily have the same type/size. Unless you need that to work for some case in xnet ...

source/codegen/templates/service_helpers.mako
```diff
@@ -491,8 +491,15 @@ ${initialize_standard_input_param(function_name, parameter)}
...
+  size_field_param = common_helpers.get_param_with_name(parameters, size_from_param_name)
```
astarche, line 495: I wouldn't call this the "size_field_param". More like the `linked_array_param`?

astarche (approved):

mgumble (approved):

source/codegen/templates/service_helpers.mako
doshirohan: `determine_size_from` is always initialized to list in `metadata_mutation`. I have added check for len of that shouldn't be greater than 1 and raise exception.

I don't understand `determine_size_from` mechanism properly and couldn't find any usage of that tag in `functions.py` to validate my changes against. @astarche can you confirm if this logic is what you had in mind?

generated/nixnet/nixnet_service.cpp
doshirohan: Moved `sizeof` inside `static_cast`

source/codegen/templates/service_helpers.mako
doshirohan: I had named this based on `size_field_name` above. I have renamed it to `linked_array_param` now.

doshirohan:
> @doshirohan Have y'all been using `validate_examples.py` to run/lint examples? You don't need to, but it might be helpful. And the non-runtime validation will run on the PR build.

@astarche I am aware of `validate_examples.py` but we haven't used it yet. @kothariaman once you have examples in good shape you can use [validate_examples.py](https://github.com/ni/grpc-device/blob/main/source/codegen/validate_examples.py) to run all of them in single go. This will probably take care of compiling proto file as well making it easier to validate examples against latest proto from your branch.

source/codegen/templates/service_helpers.mako
astarche: yeah, that looks good. Sorry for the confusion. `determine_size_from` is populated in a transform. It's sort of like a back-pointer to the size mechanism field (which is from array->size param). It can have multiple elements for "linked arrays" in which case there is logic to make sure that they are all the same length. My assumption was that that logic *did not* work for len-in-bytes (because element size may differ). So, the assert you added should help us catch any case where we try to use linked len-in-bytes and at least make sure that someone works through the codegen.

PR #616, opened 2022-03-18
Add loopback system tests for xnet socket

astarche:

Adds loopback system tests for nxet socket.

Makes several fixes and additions to enable this:
* Fix in/out size param in `getsockname`.
* Fix param directions in `recv` methods.
* Make `accept` an `init_method` because it returns a socket.
* Add `inet_pton` and `inet_aton` methods for converting address strings to `addr` struct byte formats.

Add test coverage for xnet socket and enable more testing.

Fixes [AB#1868705](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1868705).

Ran and passed all Xnet tests and all unit tests.


generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -82,15 +84,30 @@ message VirtualInterface {
...
+  oneof addr {
+    IPv4Addr ipv4 = 1;
+    IPv6Addr ipv6 = 2;
+  }
+}
+
 message SockAddrIPv4 {
   uint32 port = 1;
-  uint32 addr = 2;
+  IPv4Addr addr = 2;
```
astarche, line 104: I was originally "collapsing" the xnet addr structs (`nxin_addr` and `nxin6_addr`) into the containing `sockaddr` structs (which also have port/etc). But, with the `aton`/`pton` APIs, we now have methods that populate these directly. So I updated sockaddr to have the additional struct level as well.

I'm also considering changing the names of these types, maybe to be closer to the nxsocket ones. I.e., InAddr, In6Addr, SockAddrIn, etc. I don't love those names (and am a little unsure on the capitalization) but I'm not convinced my names are an improvement. And the C API is out best touchstone. (I'm planning to save this for a future PR, if any, but would welcome feedback.

generated/nixnetsocket/nixnetsocket_service.cpp
```diff
@@ -55,12 +55,22 @@ namespace nixnetsocket_grpc {
     try {
       auto socket_grpc_session = request->socket();
       nxSOCKET socket = session_repository_->access_session(socket_grpc_session.id(), socket_grpc_session.name());
+
       auto addr = allocate_output_storage<nxsockaddr, SockAddr>();
       nxsocklen_t addrlen {};
-      auto status = library_->Accept(socket, &addr, &addrlen);
+      auto init_lambda = [&] () {
+        auto socket_out = library_->Accept(socket, &addr, &addrlen);
```
astarche, line 62: Annoying subtle change. When you `Accept`, you pass in a socket (the listening socket) and get out a new socket. So `Accept` needs to be an `init_method`.

I also wanted to use `socket` as the name of both the input and the output. That "just works" in grpc because of the request/response split, but we need a different `cpp_local_name` for one or the other in the implementation. So there are a few codegen changes for that.

generated/nixnetsocket/nixnetsocket_service.cpp
```diff
@@ -297,7 +370,7 @@ namespace nixnetsocket_grpc {
       auto socket_grpc_session = request->socket();
       nxSOCKET socket = session_repository_->access_session(socket_grpc_session.id(), socket_grpc_session.name());
       auto addr = allocate_output_storage<nxsockaddr, SockAddr>();
-      nxsocklen_t addrlen {};
+      auto addrlen = static_cast<nxsocklen_t>(sizeof(addr.storage));
```
astarche, line 373: Annoying subtle change: `addrlen` is an in/out parameter. If we don't initialize it, the driver will error out because it thinks we didn't give it space for an address. I'm solving this by enabling "hardcoded" on output parameters. We're already setting up the max required storage size, we just need to tell the driver that.

source/codegen/templates/service_helpers.mako
```diff
@@ -786,69 +790,69 @@ ${initialize_response_buffer(parameter_name=parameter_name, parameter=parameter)
 ${copy_to_response_with_transform(source_buffer=raw_response_field, parameter_name=parameter_name, transform_x=convert_x_to_enum, size=common_helpers.get_size_expression(parameter))}\
 %       elif parameter['type'] == 'ViReal64':
         if(${parameter_name} == (int)${parameter_name}) {
-          response->set_${parameter_name}(static_cast<${namespace_prefix}${parameter["enum"]}>(static_cast<int>(${parameter_name})));
```
astarche, line 789: This method wasn't distinguishing "parameter_name" and "field_name", so I updated to make the "socket" in, "socket" out thing work in accept.

source/custom/xnetsocket_converters.h
```diff
@@ -24,27 +24,22 @@ using ResourceRepositorySharedPtr_ = std::shared_ptr<nidevice_grpc::SessionResou
 struct SockAddrInputConverter {
   SockAddrInputConverter(const SockAddr& input)
   {
-    // memset zero ensures:
-    // (a) any unset bytes in the ipv6 addr are zero.
-    // (b) If neither oneof case is selected, sa_familly will be nxAF_UNSPEC.
-    std::memset(&addr, 0, sizeof(addr));
```
astarche, line 30: Thanks to @reckenro for the tip to use braced initialization for these instead of memset. I've updated that throughout. With matching `{}` in the member declarations.

source/tests/system/nixnetsocket_driver_api_tests.cpp
```diff
@@ -27,29 +28,38 @@ constexpr auto INVALID_SOCKET_ERROR = -13008;
...
+constexpr auto IPV4_LOCALHOST_ADDRESS_STR = "127.0.0.1";
+constexpr auto IPV6_LOCALHOST_ADDRESS_STR = "::1";
+const auto IPV6_LOCALHOST_ADDR = "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\x1"s;
+const auto IPV6_ZERO_ADDR = "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"s;
 
 constexpr auto SCHEMA = "file:///NIXNET_Documentation/xnetIpStackSchema-03.json";
 const auto SIMPLE_INTERFACE_CONFIG = R"(      
         {
-          "address": "inherit",
+          "address": "generated",
```
astarche, line 44: I tweaked this to be exactly what the nxsocket C API tests use. I don't think most of the changes matter, but I think it's better to have fewer variables to worry about.

generated/nixnetsocket/nixnetsocket.proto
reckenro: Yeah... I think changing the names to be closer to the C API is probably the safe bet. At worst, the same and at best the client is familiar with the naming schemes from XNET C API or socket programming in general.

reckenro (approved):

source/tests/system/nixnetsocket_driver_api_tests.cpp
```diff
@@ -185,13 +208,26 @@ class NiXnetSocketNoHardwareTests : public NiXnetSocketDriverApiTests {
 
 #define EXPECT_SUCCESS(response)               \
   if (1) {                                     \
-    EXPECT_EQ(0, (response).status());         \
+    EXPECT_LE(0, (response).status());         \
```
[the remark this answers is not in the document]
astarche: I should add a comment. Basically this is to say that all positive statuses are considered SUCCESS. There are several methods that return a size on success. Some return "0 or 1". It's kind of all over the place.

I guess an alternative would be a variant of EXPECT_SUCCESS that takes in an expected status (like expect failure). Or EXPECT_SUCCESS_ZERO_ONLY or something.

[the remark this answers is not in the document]
DavidCurtiss: Maybe add a separate EXPECT_NO_ERROR macro that does the LE check?? Or else you might explicitly EXPECT_WARNING like we did in the RFmx tests.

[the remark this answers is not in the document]
astarche: I like EXPECT_NO_ERROR. I think EXPECT_SUCCESS_WITH_STATUS is probably the next best, since a lot of these are explicitly not warnings.

mgumble (approved):

PR #617, opened 2022-03-18
Auto generate remaining Frame APIs for XNET

doshirohan:

Earlier PR added convert* helper methods for copying Frames to and from gRPC along with 1 API of each type. This PR auto generated remaining frame APIs from XNET.

Adds additional API support for XNET.

- In previous PR I had tested Convert* APIs that dealt with reading and writing Frames which did not require hardware. With that we are confident of frame conversions happening properly. The specific APIs from this PR need hardware and will be tested using python examples ones they are ready.


generated/nixnet/nixnet_service.cpp
```diff
@@ -974,6 +1043,63 @@ namespace nixnet_grpc {
...
+  //---------------------------------------------------------------------
+  ::grpc::Status NiXnetService::ReadFrame(::grpc::ServerContext* context, const ReadFrameRequest* request, ReadFrameResponse* response)
+  {
+    if (context->IsCancelled()) {
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      auto session_ref_grpc_session = request->session_ref();
+      nxSessionRef_t session_ref = session_repository_->access_session(session_ref_grpc_session.id(), session_ref_grpc_session.name());
+      u32 size_of_buffer = request->size_of_buffer();
```
ni-pkhare, line 1056: I think essentially this is wrong. firstly we are taking size_of_buffer as an input parameter (which we should not). Secondly this API needs information from client regarding how many frames they want to see in output (Probably LV has also exposed this additional parameter for LV clients) because the documentation states that "**Standard CAN and LIN frames are always 24 bytes in length. To read a specific number of standard CAN and/or LIN frames, multiply that number by 24.**" So we would essentially require this information to allocate the necessory buffer size for output. But here we would also need to be carefull of CAN FD and FlexRay frames because each frame might vary in length.

Same would be applicable to WriteFrame also probably.

astarche: I think your proposal is basically "it would be better to calculate `size_of_buffer` from `frame_type` and `<some_additional_parameter>`. And that `<some_additional_parameter>` is required to deal with variable length frame types?

I agree in general that it is better to not require `size_of_buffer `if possible. That's a good observation. It *might* be better to replace `size_of_buffer` with `<some_additional_parameter>` depending on what the additional parameter was. What exactly parameter are you proposing? Or are you proposing that `size_of_buffer` is essentially optional? I agree with you that it is worth putting effort into improving this.

It will always be OK to me to submit functional code that is harder to use, **especially since we don't yet have any system tests for xnet**. So, for that reason, I would recommend continuing with this change. That will allow testing in the current state, which will help determine how best to improve the API (and what mistakes we have made).

> If we expose the additional parameter to client (which we should to get the correct results), the request message format would also change and essentially this and WriteFrame would become "CustomCodeCustomProto" API.

**I disagree** with this, and I am honestly a little frustrated by it. This is the first set of Xnet changes that uses converters correctly instead of `custom_code`. And I **strongly** prefer converters! There are 2 tools that you can use to accomplish what you want:
1. `'proto_only': True,`, which was recently added by @reckenro for `SetSockOpt`.
2. `hardcoded_value` to hardcode a library param to an expression.

So, for example, you could add `<some_additional_parameter>` with `proto_only`. You could set `size_of_buffer` to be `"include_in_proto": False` and `"hardcoded_value": "determine_size_from_request(request->frame_type(), request->some_additional_parameter())"`. And you could implement `determine_size_from` in `xnet_converters.h`.

You can also see from recent xnet socket changes that we have some up with a lot of ways to use converters and other metadata tricks to support different API types without custom_code. Most of these have been easy to add and can significantly improve testability.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -1025,6 +1037,14 @@
             {
                 'direction': 'out',
                 'name': 'numberOfBytesReturned',
+                'type': 'u32',
+                'include_in_proto': False
+            },
+            {
+                'direction': 'in',
```
astarche, line 1044: Are you working on adding these to scrapigen?

astarche (approved):

generated/nixnet/nixnet_service.cpp
doshirohan: We do have a task to fix APIs that accept size_of_buffer in bytes from user to instead expose that parameter in different way that client in any language can pass in. We have about 8 APIs affected by this and most of them either return Frame, Signal or Timestamp. I think only Frame ones might be difficult to convert since frames could be of variable length, but other two are straight forward.

I have been trying to figure out how to use converters for other APIs which accept/return custom message instead of void* and `SetSockOpt` seems like a good reference for that. Use of converters and hardcoded_value in that is exactly what I have been wanting to play around with to codegen those. We will probably try to give that a shot too.

source/codegen/metadata/nixnet/functions.py
doshirohan: Yes, all of the changes from here will be backported to scrapigen.

PR #618, opened 2022-03-18
[minor] Codegen proto message for DbGetDatabaseList instead of custom_proto

doshirohan:

Request/Response messages for DbGetDatabaseList API were hardcoded in `custom_proto.mako`. There is no customization required for these, so auto generated them through codegen instead.

Moving these messages out of `custom_proto.mako` is better for long term maintenance.

- Verified no diff in generated messages

### DIFF
--- a/source/custom/nixnet_service.custom.cpp
+++ b/source/custom/nixnet_service.custom.cpp
@@ -1056,7 +1056,7 @@ u32 GetStateSize(u32 state_id)
     response->set_status(status);
     if (status_ok(status)) {
       response->set_alias_buffer(alias_buffer.c_str());
-      response->set_file_path_buffer(file_path_buffer.c_str());
+      response->set_filepath_buffer(file_path_buffer.c_str());
       response->set_number_of_databases(number_of_databases);
     }
     return ::grpc::Status::OK;
--- a/source/custom/nixnet_service.custom.cpp
+++ b/source/custom/nixnet_service.custom.cpp
@@ -99,8 +99,11 @@ u32 GetStateSize(u32 state_id)
   return state_size;
 }
 
+
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


source/custom/nixnet_service.custom.cpp
```diff
@@ -1056,7 +1056,7 @@ ::grpc::Status NiXnetService::DbGetDatabaseList(::grpc::ServerContext* context,
     response->set_status(status);
     if (status_ok(status)) {
       response->set_alias_buffer(alias_buffer.c_str());
-      response->set_file_path_buffer(file_path_buffer.c_str());
+      response->set_filepath_buffer(file_path_buffer.c_str());
```
astarche, line 1059: This is custom_code for the `DbGetDatabaseListSizes` thing, right? it would be nice to add comments on why these are customcode (just noticing now that I forgot why). A lot of them are pretty obviously custom, but some of these look very similar to generated code.

astarche (approved):

source/custom/nixnet_service.custom.cpp
doshirohan: Updated documentation for all of the custom implementations.

source/custom/nixnet_service.custom.cpp
```diff
@@ -204,6 +207,10 @@ ::grpc::Status NiXnetService::ReadState(::grpc::ServerContext* context, const Re
 
 //---------------------------------------------------------------------
 //---------------------------------------------------------------------
+// WriteState API has an input of type void * called StateValue. There are different possible values for 
+// StateValue based on Based on the value of StateID passed in by user. We have defined complex message
```
ni-pkhare, line 211: "based on" is mentioned twice here, can we correct this?

doshirohan: Fixed the comment.

reckenro (approved):

PR #619, opened 2022-03-21
Autogen XNET 3 ReadSignal APIs

ni-pkhare:

This PR adds support for 3 ReadSignal APIs which takes in some additional parameters from users (not to be passed to C API) but used for calculating buffersize of output arrays.

For achieving this, I added helper functions in `xnet_converters.h` and using those as `hard_coded` values to calculate buffer size.

For ReadSignalXY API, there is 1 input parameter which has to be passed as a pointer to the C API, for this I tagged that paramter with `pointer:True` in metadata and a new tag called `input_passed_by_ptr` was introduced to handle it in service.cpp mako file.


source/codegen/service_helpers.py
```diff
@@ -111,6 +111,8 @@ def _create_standard_arg(parameter):
         return f"&{parameter_name}_copy, "
     elif not is_array and is_output:
         return f"&{parameter_name}, "
+    elif not is_array and not is_output and "pointer" in parameter:
+        return f"&{parameter_name}, "
```
ni-pkhare, line 115: @astarche this is the place where I tried to put a logic so that `&` is introduced in service implementation. But this is causing `&` to be introduced in some other drivers as well.

astarche: Ah, I see. Some `pointer` fields are pointers on the stack that are passed by value:

```c++
auto reserved = nullptr;
library_->Calll(reserved);
```

And some, like this (and `is_size_param_passed_by_ptr`) we want to pass a pointer to a value on the stack.

```c++
auto timestamp = request->timestamp();
library_->Calll(&reserved);
```

Yeah, I like your proposal of adding a new flag. I would suggest that you still set "pointer" to true and add an additional flag like "input_passed_by_pointer" or "pass_address_to_driver" or something like that. Then you don't need to change the simple code that handles "pointer" for the library. And the new complexity/logic can go in this method and people reading it have some chance to understand how the decision making works.

And ... since we mentioned in the discussion. To implement this with an InputConverter. It would need to implement *implicit conversion to the pointer type*, NOT override the `&` operator (that was for OutputConverters). It would be pretty simple to implement though, if you prefer that option. That's how I would implement it, but everything is an InputConverter in nxsocket...

ni-pkhare: I have implemented the 1st option. That was quite easy to implement.

source/custom/nixnet_service.custom.cpp
```diff
@@ -1218,6 +1218,22 @@ void convert_to_grpc(const void* input, nixnet_grpc::FrameBuffer* output, u32 fr
...
+u32 determine_size_of_output_buffer(u32 size_of_value_buffer, u32 size_of_timestamp_buffer)
+{
+  if(size_of_value_buffer == 0U) {
+    return size_of_timestamp_buffer;
+  }
+  if(size_of_timestamp_buffer == 0U) {
+    return size_of_value_buffer;
+  }
+  return std::min(size_of_timestamp_buffer, size_of_value_buffer);
+}
```
ni-pkhare, line 1236: For ReadSignalXY API, if both sizeofvaluebuffer and sizeoftimestampbuffer are non zero values then, minimum has to be taken for both valuebuffer and timestampbuffer as per the documentation.

generated/nixnet/nixnet_service.cpp
```diff
@@ -169,7 +169,7 @@ namespace nixnet_grpc {
       auto session_ref_grpc_session = request->session_ref();
       nxSessionRef_t session_ref = session_repository_->access_session(session_ref_grpc_session.id(), session_ref_grpc_session.name());
       u8* value_buffer = (u8*)request->value_buffer().c_str();
-      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size());
+      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size() * sizeof(u8));
```
ni-pkhare, line 172: note: This change to multiply buffer size with size of data type for params expecting "len-in-bytes" was introduced in earlier PR. The autogenerated change was introduced in this PR because of bad merge in previous PR.
Same is mentioned in Rohan's [PR](https://github.com/ni/grpc-device/pull/622)

source/codegen/metadata/nixnet/functions.py
```diff
@@ -166,28 +172,32 @@
                 'direction': 'out',
                 'name': 'valueBuffer',
                 'size': {
-                    'mechanism': 'passed-in',
-                    'value': 'sizeOfValueBuffer'
+                    'mechanism': 'custom-code',
```
doshirohan, line 175: This is actually `passed-in` now :)

Same for `timeStampBuffer` param below, and in other APIs.

source/codegen/metadata/nixnet/functions.py
```diff
@@ -147,6 +147,12 @@
                 'name': 'sessionRef',
                 'type': 'nxSessionRef_t'
             },
+            {
```
doshirohan, line 150: Would it work if we put this next to `sizeOfValueBuffer` parameter, since that's the param it is replacing? That way there won't much change in signature from user perspective.

Same for other APIs as well.

generated/nixnet/nixnet.proto
```diff
@@ -1490,15 +1489,29 @@ message ReadSignalWaveformRequest {
     TimeOut timeout = 2;
     double timeout_raw = 3;
   }
-  uint32 size_of_value_buffer = 4;
+  uint32 number_of_samples = 4;
```
astarche, line 1492: nit: I might be misunderstanding this. "number_of_samples" is "samples_per_signal" right? Can we use that terminology instead (if that's accurate)? I'm concerned that a client might think that this is "total_number_of_samples" which is "number_of_samples * samples_per_signal".

ni-pkhare: Yes, that makes much more sense, I'll change it everywhere.

astarche (approved):

astarche:
Overall approach looks good. Trusting you and @doshirohan on making sure the math is right. And of course will need to test it to make sure it works!

mgumble (approved):

reckenro (approved):

PR #620, opened 2022-03-21
[minor] Rename `from` in xnet API since it is reserved keyword in python

doshirohan:

`ConvertTimestamp100nsTo1ns` and `ConvertTimestamp1nsTo100ns` APIs in XNET have parameter named `from`. As `from` is reserved keyword in python, we can't directly set this param on request object. There is a workaround to use `setattr` that is reported [here](https://github.com/protocolbuffers/protobuf/issues/6815). The workaround (as well as parameter names which are reserved keywords) are not easy to discover and cumbersome to use. This is how request would look like with the workaround:
```
req_object = nixnet_types.ConvertTimestamp1nsTo100nsRequest()
setattr(req_object, "from", Time)
```
Since python is one of the important use case for `grpc-device`, I am renaming these parameters to make it easy to use in python.

- For both above APIs suffixed `from` parameter with timestamp type that it represents.
- Renamed `to` parameter in above APIs as well to maintain consistency even though that does not cause any issue.

- This change makes it easy to use above APIs in python client

- Python API seems to be properly generated after the change


astarche (approved):

reckenro (approved):

doshirohan:
@kothariaman I have renamed the `from` param for convert timestamp APIs in this PR. Once you pull in latest changes, make sure to replace `setattr` call in your example with usual param setter pattern we have followed.

PR #621, opened 2022-03-21
Add new Optname Enum conversions

danestull:

Handles new Optname Enum value conversions.

Allows for users to use the remaining enums for SockOpt methods that were not implemented.

Passed all builds and added several new tests for new conversion mechanisms added which passed. 
Screen Shot 2022-03-21 at 1 30 04 PM


source/codegen/metadata/nixnetsocket/custom_proto.mako
```diff
@@ -54,11 +54,23 @@ message Linger {
   int32 l_linger = 2;
 }
 
+message IPMReq {
+  uint32 imr_multiaddr = 1;
```
astarche, line 58: await: I introduced message types for nxin_addr and nxin6_addr (IPv4Addr and IPv6Addr; though I may change the names later) in #616 , I like that both (a) for consistency with the C API, (b) to provide meaning to disorienting data, and (c) to make it easier to copy to/from request/response objects. I'll admit it's not the killerest feature though. Can you check out that PR. And then update to use those and/or make a counter-argument?

danestull: Yeah I will want to update to have these messages use those new addr types you introduced for consistency with the API that makes sense.

source/custom/xnetsocket_converters.h
```diff
@@ -327,6 +346,10 @@ struct SockOptDataInputConverter {
         break;
       case SockOptData::DataCase::kDataLinger:
         return sizeof(data_linger);
+      case SockOptData::DataCase::kDataIpmreq:
```
mgumble, line 349: nit: I think we would want this to be kDataIpMreq

mgumble (approved):

source/custom/xnetsocket_converters.h
```diff
@@ -412,6 +452,23 @@ struct SockOptDataOutputConverter {
...
+        break;
+      }
+      case OptName::OPT_NAME_IPV6_ADD_MEMBERSHIP:
+      case OptName::OPT_NAME_IPV6_DROP_MEMBERSHIP: {
+        // Reinterpret unsigned char to char.
+        auto multiaddr_out = reinterpret_cast<const char*>(data_ipv6mreq.ipv6mr_multiaddr.addr);
+        auto multiaddr_size = sizeof(data_ipv6mreq.ipv6mr_multiaddr.addr);
+        output.mutable_data_ipv6mreq()->set_ipv6mr_multiaddr(
+            {&multiaddr_out[0],
+             &multiaddr_out[multiaddr_size]});
```
reckenro, line 468: I'm not sure if I have the syntax right but are the arguments supposed to be `(pointer_to_size, size)`? If that's the case I think the second argument should change to just be `multiaddr_size`.

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -435,6 +469,47 @@ TEST(XnetConvertersTests, LingerSockOptData_ConvertToGrpc_ConvertsToSockOptDataW
...
+      ipv6mreq_pointer->ipv6mr_multiaddr.addr,
+      IPV6MR_MULTIADDR.data(),
+      sizeof(ipv6mreq_pointer->ipv6mr_multiaddr.addr));
+  ipv6mreq_pointer->ipv6mr_interface = IPV6MR_INTERFACE;
+
+  auto grpc_data = SockOptData{};
+  convert_to_grpc(storage, &grpc_data);
+
+  EXPECT_EQ(SockOptData::DataCase::kDataIpv6Mreq, grpc_data.data_case());
+  EXPECT_THAT(grpc_data.data_ipv6mreq().ipv6mr_multiaddr(), ElementsAreArray(IPV6MR_MULTIADDR.data(), IPV6MR_MULTIADDR.size()));
```
reckenro, line 509: Can you add an assert here matching my other comment? I think you can assert that the multiaddr().size() on the grpc data equals the `IPV6MR_MULTIADDR.size()`.

reckenro (approved):

source/codegen/metadata/nixnetsocket/custom_proto.mako
danestull: I've updated everything to use the new IPv4 and IPv6 Addrs

source/custom/xnetsocket_converters.h
danestull: Updated names so that we get this name for the Datacase

source/custom/xnetsocket_converters.h
danestull: Alex's change introduced `copy_ipv6_addr_to_output` helper that sets the IPv6 output message with the storage data. This is what we use now in all outputs dealing with IPv6Addrs.

source/tests/unit/xnet_converters_tests.cpp
danestull: I went ahead and still added an assert here for ensuring the sizes are the same before comparing the elements.

source/custom/xnetsocket_converters.h
astarche: From the [std::string reference](https://en.cppreference.com/w/cpp/string/basic_string/basic_string): I think @reckenro  is referring to (4). This incantation might work as (6) `[first, last)`, but I'm not sure. Pointers should satisfy the contract of `InputIt`, but I Am Not A Compiler. What we definitely **DO NOT WANT** is (5) (`char*`) because that will only copy up to the first null byte. Our test case should catch that because of the `0x0` byte at the beginning. But you can double-check that by changing this to use only pass in the `char*` and see if you can get it to fail (`EXPECTED "string data" ACTUAL ""`)

astarche (approved):

source/custom/xnetsocket_converters.h
reckenro: Oh I see, nevermind. I missed the curly braces and thought those were args to `set_ipv6mr_multiaddr`. Yeah, for strings it probably falls into case (6) where it's `[first, last)` so this works out.

PR #622, opened 2022-03-22
Replace string_array property type with string for xnet

doshirohan:

Some of the properties for XNET are specified as 1DString in documentation. Behind the scenes these properties are still represented as string with comma separated values. Earlier we were tagging them as `string_array` in proto messages and in metadata, but still handling them as scalar strings. This PR changes that behavior to tag those properties as scalar strings, as we don't return `repeated` element for these properties.

Other alternative was to expose these properties as `repeated string` in proto and handle conversion between repeated string and comma separated lists in server implementation. This will require us to modify any other APIs that accept these "list" values to accept `repeated string` instead. This is definitely feasible, but not really required since many NI drivers return list of strings as comma separated values and we handle them as string in gRPC.

This PR brings consistency between field name in the message for these string properties and underlying data type. Earlier the field name was `string_array`, but data type was just `string`.

Executed python example that read string array property and ensured that the value is correct


generated/nixnet/nixnet_service.cpp
```diff
@@ -169,7 +169,7 @@ namespace nixnet_grpc {
       auto session_ref_grpc_session = request->session_ref();
       nxSessionRef_t session_ref = session_repository_->access_session(session_ref_grpc_session.id(), session_ref_grpc_session.name());
       u8* value_buffer = (u8*)request->value_buffer().c_str();
-      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size());
+      u32 size_of_value_buffer = static_cast<u32>(request->value_buffer().size() * sizeof(u8));
```
ni-pkhare, line 172: Isnt sizeof(u8) 1 byte only? but yeah this is also correct.

ni-pkhare (approved):

generated/nixnet/nixnet_service.cpp
doshirohan: Yes, that's correct. But this is auto-generated code and I had tagged the API in metadata to call out that this parameter expects length in bytes. This auto generated change came in this PR because of merge conflicts. I will add a note here so that owners won't get confused about the change.

doshirohan, line 172: note: This change to multiply buffer size with size of data type for params expecting "len-in-bytes" was introduced in earlier PR. The autogenerated change was introduced in this PR because of bad merge in previous PR.

mgumble (approved):

reckenro (approved):

PR #625, opened 2022-03-23
Update XNET Frame input APIs to remove buffer size parameters

doshirohan:

XNET APIs used for reading (or converting to) frames used to pass buffer_size parameter to indicate number of bytes that should be allocated for frame buffer while calling C API. gRPC clients being aware of C buffer sizes is not correct, so we are replacing those parameters with alternatives that could be used for calculating size in bytes on server side.

For APIs returning frames, we have modified API signature to accept `num_of_frames` clients are expecting and `max_payload_per_frame`. We can use these two parameters to calculate maximum buffer size required while calling the C API. These buffers are then resized appropriately using `number_of_bytes_returned` value returned by C API.

LV API uses similar approach where they either accept number of frames from client or query number of available frames in the queue from XNET. Based on frame type of current session, they query max frame size from XNET. And then use these two to compute buffer size. In case of LV implementation some of these values are cached with the session (apparently for performance reasons). Since we can not cache values in gRPC server, and we expect users to be aware of these values anyway as they needed these in C application for allocating buffer, we are requesting these values to be passed in by user.

This change simplifies frame APIs in XNET where client need not be aware of server implementation and data types their size on server side.

Executed Write/Read APIs for frames for Ethernet to check we get same frame back.


generated/nixnet/nixnet.proto
```diff
@@ -1103,10 +1103,11 @@ message ConnectTerminalsResponse {
 message ConvertByteArrayToFramesSinglePointRequest {
   nidevice_grpc.Session session_ref = 1;
   bytes value_buffer = 2;
-  uint32 size_of_buffer = 3;
+  uint32 num_of_frames = 3;
```
astarche, line 1106: nit: we used `number_of_signals` for #619 so I would use `number_of_frames` here.

Also, for some reason `num_frames` sounds fine/normal to me but `num_of_frames` sounds weird.

source/custom/nixnet_converters.h
```diff
@@ -8,6 +8,8 @@ namespace pb_ = ::google::protobuf;
 
 namespace nixnet_grpc {
 
+#define EnetFrameHeaderLength sizeof(nxFrameEnet_t) - 1  // last byte in nxFrameEnet_t is u8 FrameData[1]
```
astarche, line 11: nit: prefer `constexpr` and `UPPER_SNAKE_CASE` for constants C++.

```cpp
constexpr auto ENET_FRAME_HEADER_LENGTH = sizeof(nxFrameEnet_t) -1; // last byte in nxFrameEnet_t is u8 FrameData[1]
```

you could also include the static cast here if you always need a `u16`:

```cpp
constexpr auto ENET_FRAME_HEADER_LENGTH = static_cast<u16>(sizeof(nxFrameEnet_t) -1); // last byte in nxFrameEnet_t is u8 FrameData[1]
```

(I think the current code is a compile warning).

astarche (approved):
Looks nice!

reckenro (approved):

generated/nixnet/nixnet.proto
doshirohan: Updated to `number_of_frames` for consistency.

source/custom/nixnet_converters.h
doshirohan: Updated to use `constexpr` with `static_cast`

PR #626, opened 2022-03-24
Adding enums for Frame APIs

Akshata440:

XNET Frame and EnetFrame types have enums associated with `frame type` and `flags` fields of the frame. This PR introduces these enums and updates custom implementation of frame input and output convert APIs to use enum and raw values appropriately.

For `flags` fields in frame we are using concept equivalent to `bitfield_as_enum_array` and appropriately compute flag value by or-ing all values user has sent for request and splitting values into array for response in the `convert` APIs.

We had added `FrameType` enum earlier for specifying which type of frame user wants to read (this would point to can/lin/flexray/enet etc.). Since enum associated with "type" field of frame is also called `FrameType` and has different values like data frame etc., earlier enum has been renamed to `Protocol` after discussion with Waldemar.

Enum tagging makes it easier to use these APIs.

TODO:


source/custom/nixnet_service.custom.cpp
```diff
@@ -1187,15 +1187,39 @@ void convert_to_grpc(std::vector<u8>& input, google::protobuf::RepeatedPtrField<
...
+      if(enet_flags_enum_omap_it != enetflags_output_map.end()) {
+        enet_frame->add_flags(EnetFlags::ENET_FLAGS_ENET_FLAGS_NETWORK_SYNCED);
+      }
+    }
+    if(nxEnetFrame->Flags & 0x00010000) {
+      auto enet_flags_enum_omap_it = enetflags_output_map.find(0x00010000);
+      if(enet_flags_enum_omap_it != enetflags_output_map.end()) {
+        enet_frame->add_flags(EnetFlags::ENET_FLAGS_ENET_FLAGS_ERROR);
+      }
+    }
```
doshirohan, line 1222: You need to `flags_raw` here.

source/custom/nixnet_service.custom.cpp
```diff
@@ -1204,30 +1228,45 @@ void convert_to_grpc(const void* input, nixnet_grpc::FrameBuffer* output, u32 fr
...
+    if(nxFrame->Flags & 0x04) {
+     frame->add_flags(FrameFlags::FRAME_FLAGS_FRAME_FLAGS_FLEX_RAY_PREAMBLE);
+    }
+    if(nxFrame->Flags & 0x10) {
+      frame->add_flags(FrameFlags::FRAME_FLAGS_FRAME_FLAGS_FLEX_RAY_CH_A);
+    }
+    if(nxFrame->Flags & 0x20) {
+      frame->add_flags(FrameFlags::FRAME_FLAGS_FRAME_FLAGS_FLEX_RAY_CH_B);
+    }
     frame->set_info(nxFrame->Info);
```
doshirohan, line 1251: You need to set `flags_raw` here.

source/custom/nixnet_service.custom.cpp
```diff
@@ -1187,15 +1187,39 @@ void convert_to_grpc(std::vector<u8>& input, google::protobuf::RepeatedPtrField<
   }
 }
 
-void convert_to_grpc(const void* input, nixnet_grpc::FrameBuffer* output, u32 frame_type)
+void convert_to_grpc(const void* input, nixnet_grpc::FrameBufferResponse* output, u32 protocol, std::map<std::int32_t, std::int32_t> enetflags_output_map)
 {
-  if (frame_type == nixnet_grpc::FrameType::FRAME_TYPE_ENET) {
-    nixnet_grpc::EnetFrame* enet_frame = new nixnet_grpc::EnetFrame();
+  if (protocol == nixnet_grpc::Protocol::PROTOCOL_ENET) {
```
doshirohan, line 1192: Since these `if/else` blocks are much bigger now, it would be helpful to extract those into private helper methods `convert_frame_to_grpc` and `convert_enet_frame_to_grpc`

doshirohan (approved):

source/codegen/metadata/nixnet/custom_proto.mako
```diff
@@ -238,29 +238,66 @@ message DbSetPropertyResponse {
...
   uint32 identifier = 2;
-  uint32 type = 3;
-  uint32 flags = 4;
-  uint32 info = 5;
-  bytes payload = 6;
+  oneof type_enum{
+     FrameType type = 3;
+     uint32 type_raw = 4;
+  }
+  repeated FrameFlags flags = 5;
```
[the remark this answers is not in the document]
astarche: I actually think it would be better to have `flags_raw` as a separate field not in a `oneof`. Since this is a logical OR we can simply OR together all flags from the FrameFlags array and the value of flags_raw. Of course most users will set one or the other and will work fine.

I think `bitfield`s are a very good candidate for the `_raw` concept. Some users will want to use existing code that `|` flags together and it's more difficult to "fake `_raw` as you can often do with typecasts on enums. Also, bitfields are often designed for extensibility, so there is a likelihood of "values we don't know about" that users might want to set through raw.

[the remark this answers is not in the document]
astarche: I think I'm also going to add codegen for bitfield inputs for nxsocket as well. fyi.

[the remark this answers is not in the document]
astarche: I think it's fine to add this later if implementation cost is a problem. You can check with Harsha. I believe that it would be easy to add later, especially with this being all custom_code. I think having a second non-oneof `_raw` field should be the "standard" implementation and that's what I'll do for codegen. But it's OK with me to have a functional subset of the "standard" implementation.

source/custom/nixnet_service.custom.cpp
```diff
@@ -1187,15 +1187,39 @@ void convert_to_grpc(std::vector<u8>& input, google::protobuf::RepeatedPtrField<
...
+    nixnet_grpc::EnetFrameResponse* enet_frame = new nixnet_grpc::EnetFrameResponse();
     nxFrameEnet_t* nxEnetFrame = (nxFrameEnet_t*)input;
-    enet_frame->set_type(nxEnetFrame->Type);
+    enet_frame->set_type(static_cast<nixnet_grpc::EnetFrameType>(nxEnetFrame->Type));
+    enet_frame->set_type_raw(nxEnetFrame->Type);
     enet_frame->set_device_timestamp(nxEnetFrame->DeviceTimestamp);
     enet_frame->set_network_timestamp(nxEnetFrame->NetworkTimestamp);
-    enet_frame->set_flags(nxEnetFrame->Flags);
+    if(nxEnetFrame->Flags & 0x80000000) {
```
astarche, line 1199: Note: this is really impossible to review. This is a perfect example of why both codegen and tests are important. Please add it to your retrospective.

Normally, if you were writing this without codegen, you would need a way to implement this with a loop. I believe in this case it would be iterating over the mapping (for unmapped it would be iterating over a list of flags) and checking for each value and then adding it. The "unrolled" implementation makes sense when it comes from a loop in the codegenerator.

astarche (approved):

reckenro (approved):

source/custom/nixnet_service.custom.cpp
mgumble: I agree some unit tests here would be really helpful

mgumble (approved):

source/custom/nixnet_service.custom.cpp
Akshata440: Yes, I will be adding unit tests in subsequent PRs

source/custom/nixnet_converters.h
```diff
@@ -105,7 +105,15 @@ struct FrameHolder {
       current_frame->Type = grpc_frame.enet().type();
       current_frame->DeviceTimestamp = grpc_frame.enet().device_timestamp();
       current_frame->NetworkTimestamp = grpc_frame.enet().network_timestamp();
-      current_frame->Flags = grpc_frame.enet().flags();
+      current_frame->Flags = 0;
+      for(int i=0; i<grpc_frame.enet().flags_size(); i++)
+      {
+        auto enet_flags_enum_imap_it = enetflags_input_map.find(grpc_frame.enet().flags()[i]);
+        if(enet_flags_enum_imap_it == enetflags_input_map.end()) {
+          return throw std::invalid_argument("The value for destination_mapped was not specified or out of range.");
```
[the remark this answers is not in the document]
Akshata440: Done

source/codegen/metadata/nixnet/custom_proto.mako
```diff
@@ -238,29 +238,66 @@ message DbSetPropertyResponse {
...
-  uint32 flags = 4;
-  bytes frame_data = 5;
+message EnetFrameRequest {
+  oneof type_enum{
+    EnetFrameType type = 1;
+    uint32 type_raw = 2;
+  }
+  uint64 device_timestamp = 3;
+  uint64 network_timestamp = 4;
+  repeated EnetFlags flags = 5;
```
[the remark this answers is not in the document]
Akshata440: Done

source/custom/nixnet_service.custom.cpp
Akshata440: Done

source/custom/nixnet_service.custom.cpp
Akshata440: Done

source/custom/nixnet_service.custom.cpp
Akshata440: Extracted into helpers

source/custom/nixnet_converters.h
```diff
@@ -105,7 +105,15 @@ struct FrameHolder {
       current_frame->Type = grpc_frame.enet().type();
```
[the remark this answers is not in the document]
Akshata440: Done

source/codegen/metadata/nixnet/custom_proto.mako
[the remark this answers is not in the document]
Akshata440: Will be adding this implementation in later PRs

PR #629, opened 2022-03-24
Add tests for AddrInfo related methods.

reckenro:

Adds tests for the AddrInfo functions.

Fixes [AB#1895026](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1895026)

We wanted to get the functions in as soon as possible so this is a follow up to add tests for them.

Added both conversion and system-level tests, all of which pass.


reckenro:
image

generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -186,7 +195,7 @@ message AddrInfo {
 }
 
 message AddrInfoHint {
-  GetAddrInfoFlags flags = 1;
+  int32 flags = 1;
```
reckenro, line 198: This field and flags on `AddrInfo` either needed to be an `int32` or `repeated GetAddrInfoFlags` to allow for a combination of flags. I'm honestly not sure which is better but I went with int32 since it was simpler and allows the client to do bitwise operations on the flags / construct them how they would for the C API.

If, however, the client just wants to add/remove flags and leave the bitwise combination to us we could use repeated Flags enum and do the translations to and from an int in the conversion code.

generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -297,7 +306,10 @@ message GetNameInfoRequest {
   SockAddr addr = 2;
   int32 host_len = 3;
   int32 serv_len = 4;
-  int32 flags = 5;
+  oneof flags_enum {
```
reckenro, line 309: I updated this to use the `GetNameInfoFlags` enum. I only saw these used one at a time (no combination) in [GetAddrNameInfoTests](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Fxnet%2FxnetCored%2Fnixntipstack_tests%2Fsource%2Fnixntipstackt%2FGetAddrNameInfoTests.cpp&_a=contents&version=GBmain) so I kept them as just the enum.

For this one, I could potentially mark it in functions.py as a 'bitfield_as_enum_array`. I think currently we only support that for output parameters but I could do some work to get it generating for input params as well.

source/codegen/metadata/nixnetsocket/functions.py
```diff
@@ -163,12 +163,28 @@
...
                 'name': 'node',
-                'type': 'const char[]'
+                'type': 'const char[]',
+                'proto_only': True
+            },
+            {
+                'direction': 'in',
+                'name': 'node_api',
+                'type': 'const char[]',
+                'hardcoded_value': 'request->node() == "" ? nullptr : node',
```
reckenro, line 173: I'm not crazy about this but something like this was needed to get the driver-api tests to pass. They would fail when passing an empty string instead of nullptr (the C API tests pass nullptr). This seemed like the easiest way to generate the necessary conditional. I could also look into something like a 'nullptr-if-empty' tag for strings and have it generate this but that might be overkill if it's not widely used.

generated/nixnetsocket/nixnetsocket.proto
astarche: I created AB#1896770 that's related to this. I'm leaning towards exposing both and |-ing it all together. See discussion in #626 ( 'note: FrameFlags is used as "bitfield_as_enum_array" here. ').

Note that the codgen for `bitfield_as_enum_array` is only for outputs. That PR is adding a custom code variant "inspired by" bitfield_as_enum_array. We might be somewhere in between (or have some of each) with the different bitfields on nxsocket.

source/codegen/metadata/nixnetsocket/functions.py
astarche: I think the `optional` tag is logically a good match for this, but we'd need new code. That's another case where we treat "empty-as-null", but onyl with linked arrays. I thought I ran into this with other methods but they all accepted `""` to mean the same thing as `nullptr`. I guess that's not the case here?

I think this is OK though.

astarche (approved):

generated/nixnetsocket/nixnetsocket.proto
reckenro: That sounds good. I did that for AddrInfoHint and AddrInfo, so now they have both `repeated GetAddrInfoFlags flags = 1` and `int32 flags_raw = 2` fields. Conversion code was updated to | the fields for input (AddrInfoHint). For output, I set `flags_raw` to equal the flags gotten from the driver API and then added a helper `convert_to_addr_info_flags` to take the int flags and fill out the repeated `flags` field based on it. Updated tests to test both `flags` and `flags_raw` and added new test to directly test the new helper.

source/codegen/metadata/nixnetsocket/functions.py
reckenro: Yeah, so it was complaining in the `NumericPortTranslation_GetAddrInfo_ReturnsExpectedAddrInfo` test I added before adding this with the `node` argument as an empty string. I think this is because some hint flags were set for the test, specifically `GET_ADDR_INFO_FLAGS_NUMERICHOST`. I've also seen in all other cases that it seems to accept an empty string as `nullptr`.

And yeah, I think some version of `optional` would make sense. But I'm going to leave it as-is for now. If more instances like this crop up, though, we should reconsider.

generated/nixnetsocket/nixnetsocket.proto
reckenro: @astarche , for this one since it's on a request message (in functions.py) I'm assuming we'll address it with AB#1896770 which will add support for `bitfield_as_enum_array` on input params. Is how I updated the field in this PR okay for now or do you want me to make any further updates before we get to that task?

source/custom/xnetsocket_converters.h
```diff
@@ -563,6 +567,17 @@ inline AddrInfoHintInputConverter convert_from_grpc(const AddrInfoHint& input)
   return AddrInfoHintInputConverter(input);
 }
 
+inline void convert_to_addr_info_flags(int32_t flags, pb_::RepeatedField<int32_t>& get_addr_info_flags)
+{
+  int flag_to_check = GetAddrInfoFlags::GET_ADDR_INFO_FLAGS_BYPASS_CACHE;
+  while (flag_to_check != 0) {
+    if (flags & flag_to_check) {
+      get_addr_info_flags.Add(flag_to_check);
```
astarche, line 575: I'm thinking we should check `GetAddrInfoFlags_IsValid` here as well. In some scenario where there is an extra "flag we don't know about" I think it's better that we only include it in `_raw` instead of forcing it in here.

reckenro: Agreed. Technically, I don't think we would've hit that scenario since we're starting from the max value of the `GetAddrInfoFlags` enum and bit shifting down through them and only adding them (any ones in raw we don't know about wouldn't be found iterating through the `GetAddrInfoFlags`). I did add it for sanity's sake, though. I guess in general part of this is trusting that `GetAddrInfoFlags` is a proper Flags enum (each element is a power of 2) so bit shifting through them is valid but I think that's a fair assumption to make.

Also, I found protoc generates a `_MAX` constant for enums so I replaced `flag_to_check` with `GetAddrInfoFlags_MAX` instead so it should be future-proof if more flags get added (unlikely).

astarche: Gotcha, I missed that. Still a good update though.

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -599,6 +599,103 @@ TEST(XnetConvertersTests, IPv4AddrOutputConverter_ConvertToGrpc_ConvertsToAddres
...
+  EXPECT_EQ(GetAddrInfoFlags::GET_ADDR_INFO_FLAGS_ALL, first_addr_info.flags(0));
+  EXPECT_EQ(AddressFamily::ADDRESS_FAMILY_INET, first_addr_info.family());
+  EXPECT_EQ(SocketProtocolType::SOCKET_PROTOCOL_TYPE_STREAM, first_addr_info.sock_type());
+  EXPECT_EQ(IPProtocol::IP_PROTOCOL_TCP, first_addr_info.protocol());
+  EXPECT_STREQ("AddrHostName", first_addr_info.canon_name().c_str());
+  EXPECT_EQ(ADDR, first_addr_info.addr().ipv4().addr().addr());
+  EXPECT_EQ(PORT, first_addr_info.addr().ipv4().port());
+  auto next_addr_info = grpc_output[1];
+  EXPECT_EQ(nxAI_PASSIVE | nxAI_ADDRCONFIG, next_addr_info.flags_raw());
+  EXPECT_EQ(2, next_addr_info.flags().size());
```
mgumble, line 675: nit; this is a lot digest maybe pull it into a helper function since we are checking 2 addresses

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -599,6 +599,103 @@ TEST(XnetConvertersTests, IPv4AddrOutputConverter_ConvertToGrpc_ConvertsToAddres
...
+
+TEST(XnetConvertersTests, AddrInfoFlagsAsInt_ConvertToRepeatedFlagsEnum_AllFlagsPresent)
+{
+  int32_t flags = nxAI_BYPASS_CACHE | nxAI_V4MAPPED | nxAI_PASSIVE;
+
+  AddrInfo grpc_addr_info{};
+  convert_to_addr_info_flags(flags, *(grpc_addr_info.mutable_flags()));
+
+  EXPECT_EQ(3, grpc_addr_info.flags().size());
+  EXPECT_EQ(GetAddrInfoFlags::GET_ADDR_INFO_FLAGS_BYPASS_CACHE, grpc_addr_info.flags(0));
```
mgumble, line 694: Question: I may have missed it but how are these flags ordered?

mgumble (approved):

source/tests/unit/xnet_converters_tests.cpp
reckenro: They're in descending order right now so you might get (2048, 4, 1). I can change the conversion code to have it in ascending if it's useful. I'm not sure if there's an order of importance to the flags.

reckenro: I added AB#1900318 to the API review story to come back to this in general for any repeated flag enum fields to evaluate if the order's important and update accordingly.

source/tests/system/nixnetsocket_driver_api_tests.cpp
```diff
@@ -237,6 +240,26 @@ class NiXnetSocketNoHardwareTests : public NiXnetSocketDriverApiTests {
     EXPECT_XNET_ERROR(GENERIC_NXSOCKET_ERROR, INVALID_ARGUMENT_ERROR, INVALID_ARGUMENT_MESSAGE, (response)); \
   }
 
+#define ARRAY_ARG(...) \
+  {                    \
+    __VA_ARGS__        \
+  }
+
+#define EXPECT_ADDR_EQ(addr_info, expected_family, expected_addr)                                                      \
```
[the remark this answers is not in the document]
reckenro: Ah, thanks. Moved the Actual to the end.

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -599,6 +599,103 @@ TEST(XnetConvertersTests, IPv4AddrOutputConverter_ConvertToGrpc_ConvertsToAddres
   EXPECT_EQ(ADDRESS, grpc_addr.addr());
 }
 
+TEST(XnetConvertersTests, AddrInfoHintInputConverter_ConvertFromGrpc_ConvertsToAddrInfoHint)
+{
+  AddrInfoHint hints{};
+  constexpr auto EXPECTED_FLAGS = nxAI_PASSIVE | nxAI_V4MAPPED;
+  hints.set_family(AddressFamily::ADDRESS_FAMILY_INET6);
```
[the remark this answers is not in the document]
reckenro: Done.

source/tests/unit/xnet_converters_tests.cpp
reckenro: Made an `EXPECT_ADDR_INFO` helper that makes this section easier to absorb.

PR #634, opened 2022-03-29
Updated message description and comments

kothariaman:
Updated comments and message description post feedback given by team after demo.

Comments changes for better meaningful information.

Locally tested the python version.

### DIFF
--- a/examples/nixnet/can-signal-single-point-output.py
+++ b/examples/nixnet/can-signal-single-point-output.py
@@ -49,7 +49,6 @@ DATABASE = "NIXNET_example"
 CLUSTER = "CAN_Cluster"
 SIGNAL_LIST = "CANEventSignal1,CANEventSignal2"
 NUM_SIGNALS = 2
-SUCCESS = 0
 
 
 def check_for_error(status):
@@ -89,7 +88,7 @@ try:
     session = create_session_response.session
     print("Session Created Successfully.\n")
 
-    print("Writing 10 values to CAN Port.\n")
+    print("Writing 10 values to CAN Interface.\n")
 
     while i < 10:
         value_buffer[0] = float(i)
--- a/examples/nixnet/can-signal-single-point-output.py
+++ b/examples/nixnet/can-signal-single-point-output.py
@@ -87,8 +87,9 @@ try:
     check_for_error(create_session_response.status)
 
     session = create_session_response.session
-
     print("Session Created Successfully.\n")
+
+    print("Writing 10 values to CAN Port.\n")
 
     while i < 10:
         value_buffer[0] = float(i)
--- a/examples/nixnet/ethernet-frame-stream-input-reader.py
+++ b/examples/nixnet/ethernet-frame-stream-input-reader.py
@@ -36,7 +36,8 @@ import nixnet_pb2_grpc as grpc_nixnet
 
 CHOOSE_MONITOR_OR_ENDPOINT_TEXT = (
     "\nPress 'm' followed by enter to configure the input stream session to use the monitor path"
-    " to monitor all network traffic. :"
+    " to monitor all network traffic else press any other key followed by enter to use the"
+    " endpoint path which filters traffic based on VLAN ID and priority. :"
 )
 FRAME_CHECK_SEQUENCE_SIZE = 4
 MAX_ENET_FRAME_SIZE = 1518
--- a/examples/nixnet/ethernet-frame-stream-input-reader.py
+++ b/examples/nixnet/ethernet-frame-stream-input-reader.py
@@ -36,8 +36,7 @@ import nixnet_pb2_grpc as grpc_nixnet
 
 CHOOSE_MONITOR_OR_ENDPOINT_TEXT = (
     "\nPress 'm' followed by enter to configure the input stream session to use the monitor path"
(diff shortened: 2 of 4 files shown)


examples/nixnet/flexray-signal-single-point-input.py
```diff
@@ -100,8 +100,9 @@ def check_for_error(status):
     )
     check_for_error(set_property_response.status)
 
+    # To get more meaningful data use this with output example
```
harsha-bhushan, line 103: Avoid using 'meaningful'. Rephrase this to:
If no values are being written on the FlexRay port, the signals read would contain the default value of 0.0

examples/nixnet/flexray-signal-single-point-input.py
```diff
@@ -100,8 +100,9 @@ def check_for_error(status):
...
     while i < 10:
-        # Update the signal data
+        # Update the signal data, the first read normally return the default payload values.
```
harsha-bhushan, line 105: Revert this. I think the comment on 103 (rephrased statement) would convey this too.

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -87,9 +87,10 @@ def check_for_error(status):
     check_for_error(create_session_response.status)
 
     session = create_session_response.session
-
     print("Session Created Successfully.\n")
 
+    print("Writing 10 values to CAN Port.\n")
```
doshirohan, line 92: I think this should be `CAN interface`

examples/nixnet/can-signal-single-point-output.py
```diff
@@ -87,9 +87,10 @@ def check_for_error(status):
     check_for_error(create_session_response.status)
```
doshirohan, line 87: Unused variable `SUCCESS` on line 52 above. GitHub doesn't let me add comment there, so adding it here.

examples/nixnet/ethernet-frame-stream-input-reader.py
```diff
@@ -36,8 +36,7 @@
 
 CHOOSE_MONITOR_OR_ENDPOINT_TEXT = (
     "\nPress 'm' followed by enter to configure the input stream session to use the monitor path"
-    " to monitor all network traffic else just the enter key to use the endpoint path which filters"
-    " traffic based on VLAN ID and priority. :"
+    " to monitor all network traffic. :"
```
doshirohan, line 39: You previous message was fine, it just needed rewording for clarity.

> "\nPress 'm' followed by enter to configure the input stream session to use the monitor path"
    " to monitor all network traffic else press enter key to use the endpoint path which filters"
    " traffic based on VLAN ID and priority. :"
    
The change is in `press enter key`. I assume just pressing `enter` would configure it in filtered mode.


> Press 'm' followed by enter

Is this the behavior? Do you user need to press just `m` or `m + enter` to configure as monitor.

examples/nixnet/lin-signal-single-point-output.py
```diff
@@ -114,7 +114,7 @@ def check_for_error(status):
         )
         check_for_error(write_state_response.status)
 
-    print("Values are incremented from 0 to 10 in this example.\n")
+    print("Writing 10 values to LIN Port.\n")
```
doshirohan, line 117: This should be `LIN interface`

doshirohan (approved):

examples/nixnet/ethernet-frame-stream-input-reader.py
kothariaman: We need to press 'm' followed by enter because otherwise reading just a character will require importing getch and that anyways we ain't doing.

doshirohan: Sure, that makes sense. And for non-monitor mode just `enter` works, right?

kothariaman: No, for the else case it doesn't work with just `enter` and that's why i changed the comment.

doshirohan: Okay. In that case the second half of comment should read
> else press any other key followed by enter to use the endpoint path which filters

examples/nixnet/lin-signal-single-point-output.py
harsha-bhushan: yes. interface is the 'word' to use in all examples.

examples/nixnet/flexray-signal-single-point-input.py
kothariaman: Done

examples/nixnet/flexray-signal-single-point-input.py
kothariaman: Done

examples/nixnet/can-signal-single-point-output.py
kothariaman: Done

examples/nixnet/can-signal-single-point-output.py
kothariaman: Done

examples/nixnet/lin-signal-single-point-output.py
kothariaman: Done.

mgumble (approved):

PR #635, opened 2022-03-29
Users/dstull/sock opt string length

danestull:

Fetches Size of string socket options.

This allows us to know how large the string value for string sock opts are before we allocate them, therefore allowing us to be more efficient. This is done by calling the GetSockOpt with a nullptr for the value reference first and this is done in the new size() method of the converter. 

Created new test `StringSockOptData_GetSize_FetchesDataFromLibrary'
All builds and tests pass
Screen Shot 2022-03-29 at 9 45 28 AM


source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -601,6 +608,25 @@ TEST(XnetConvertersTests, IPv4AddrOutputConverter_ConvertToGrpc_ConvertsToAddres
...
 }
 
+TEST(XnetConvertersTests, StringSockOptData_GetSize_FetchesSizeFromLibrary)
+{
+  constexpr auto DEFAULT_SOCK_OPT_STRING_SIZE = 255;
+  auto ACTUAL_SOCK_OPT_STRING_SIZE = 12;
+  NiXnetSocketMockLibrary library;
+  auto storage = allocate_output_storage<void*, SockOptData>(&library, OptName::OPT_NAME_SO_BIND_TO_DEVICE);
+  EXPECT_EQ(DEFAULT_SOCK_OPT_STRING_SIZE, storage.string_length);
+  // Expect call GetSockOpt(socket, level, opt_name, nullptr, &string_length)
```
mgumble, line 618: not sure if this comment provides a lot, it is obvious we expect the call from line 623

mgumble (approved):

reckenro (approved):

generated/nixnetsocket/nixnetsocket.proto
```diff
@@ -148,7 +148,7 @@ message GatewayAddress {
 
 message VirtualInterface {
   string xnet_interface_name = 1;
-  string vlane_name = 2;
+  string vlan_name = 2;
```
astarche, line 151: fyi: I already made this change.

source/tests/unit/xnet_converters_tests.cpp
```diff
@@ -601,6 +608,25 @@ TEST(XnetConvertersTests, IPv4AddrOutputConverter_ConvertToGrpc_ConvertsToAddres
   EXPECT_EQ(ADDRESS, grpc_addr.addr());
 }
 
+TEST(XnetConvertersTests, StringSockOptData_GetSize_FetchesSizeFromLibrary)
+{
+  constexpr auto DEFAULT_SOCK_OPT_STRING_SIZE = 255;
+  auto ACTUAL_SOCK_OPT_STRING_SIZE = 12;
```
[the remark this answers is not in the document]
danestull: I've updated these both to just `const` not sure the exact reason constexpr was chose for several other instances, perhaps just so its set at compile time? But I see no problem with just having these both be consts

source/tests/unit/xnet_converters_tests.cpp
danestull: removed

generated/nixnetsocket/nixnetsocket.proto
danestull: Mose had mentioned you've made some updates so I'll be sure to handle any merge conflicts / remove this from mine

astarche (approved):

source/custom/xnetsocket_converters.h
```diff
@@ -548,8 +549,7 @@ struct SockOptDataOutputConverter {
         output.set_data_bool(data_int == 0 ? false : true);
         break;
       }
-      case OptName::OPT_NAME_SO_BIND_TO_DEVICE:
-      case OptName::OPT_NAME_SO_ERROR: {
+      case OptName::OPT_NAME_SO_BIND_TO_DEVICE: {
```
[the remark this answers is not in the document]
danestull: Updated these after merging should be all consistent!

PR #638, opened 2022-03-30
Change interface to interface_name in xnet CreateSession API

astarche:

Change `interface` to `interface_name` in xnet CreateSession API.

`interface` is a non-standard reserved keyword in MSVC (not C/C++). This can cause failures in some msvc setups when windows.h is included.

Ran and passed static type checking with `python .\source\codegen\validate_examples.py -p *xnet*`.

### DIFF
--- a/generated/nixnet/nixnet_service.cpp
+++ b/generated/nixnet/nixnet_service.cpp
@@ -359,7 +359,7 @@ namespace nixnet_grpc {
       auto database_name = request->database_name().c_str();
       auto cluster_name = request->cluster_name().c_str();
       auto list = request->list().c_str();
-      auto interface_parameter = request->interface().c_str();
+      auto interface_name = request->interface_name().c_str();
       u32 mode;
       switch (request->mode_enum_case()) {
         case nixnet_grpc::CreateSessionRequest::ModeEnumCase::kMode: {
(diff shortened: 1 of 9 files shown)


generated/nixnet/nixnet_service.cpp
```diff
@@ -359,7 +359,7 @@ namespace nixnet_grpc {
       auto database_name = request->database_name().c_str();
       auto cluster_name = request->cluster_name().c_str();
       auto list = request->list().c_str();
-      auto interface_parameter = request->interface().c_str();
+      auto interface_name = request->interface_name().c_str();
```
astarche, line 362: I actually can't figure out why this code doesn't compile with errors on xnet.

I *did* check that it has errors if you include xnet stuff from a test cpp file. So, my guess is either:
1. Somehow we are disabling non-standard CXX extensions for the server exe but not the test exe.
2. Something about the includes/ordering in the server suppresses this error: i.e.: not including windows.h, doing some kind of `undef` workaround, etc. 

At this point, I think we've had enough trouble with this issue that it is just worth implementing the workaround. However, if you disagree, let me know. I'm also pretty confident that it *is* possible to workaround and *should be* valid `protobuf` to use `interface` as a field name. Just more trouble than it's worth.

doshirohan: Interestingly I ran into same error yesterday when I was doing some changes to includes of custom service file for XNET. After trying out lot of different things late night yesterday, I was able to fix it with some ordering of includes but couldn't figure out either what exactly the workaround was. I agree, its just too much trouble and one that users might run into too. So renaming does seem like a good idea here.

doshirohan (approved):

reckenro (approved):

mgumble (approved):

PR #641, opened 2022-03-31
Update type for properties that return device and interface references

doshirohan:

Earlier `GetProperty` API used to return drivers handles for properties reading device and interface references as u32. This causes issue when users want to call `GetProperty` subsequently to read a property of interface and wants to pass-in session object for that interface handle.

There are two ways to fix this. We could either create separate repositories for device and interface handles, and return session object from these repositories OR we could add device and interface session handles to existing `session_repository_` itself as dependent session. Second option makes more sense since these handles don't have equivalent "close" APIs, so it makes sense to manage lifetime of these handles based on `session_ref` used for querying these handles.

We are still exposing different fields in request and response messages for device and interface references to make them more discoverable, but on server side lookup will always happen in `session_repository_`.

This addresses bug where gRPC clients wouldn't be able to query device and interface properties.

Tested by calling `GetProperty` APIs for reading device and interface properties.


source/custom/nixnet_converters.h
```diff
@@ -3,6 +3,7 @@
 #include <nixnet.h>
 #include <nixnet.pb.h>
 
+#include "server/converters.h"
```
doshirohan, line 6: note: This include is required for correctness, to ensure that template instantiations for convert methods here works.

source/custom/nixnet_service.custom.cpp
```diff
@@ -431,46 +442,114 @@ ::grpc::Status NiXnetService::GetProperty(::grpc::ServerContext* context, const
...
           nxDatabaseRef_t property_value;
           status = library_->GetProperty(session, property_id, property_size, &property_value);
           return std::make_tuple(status, property_value);
         };
         uint32_t session_id = 0;
-        status = nx_database_ref_t_resource_repository_->add_dependent_session("", init_lambda, initiating_session_id, session_id);
-        if (!status_ok(status)) {
-          response->set_status(status);
-          return ::grpc::Status::OK;
+        auto cleanup_lambda = [&](nxDatabaseRef_t id) {};
```
doshirohan, line 451: note: add_dependent_session here was incorrect. `initiating_session_id` is actually from `session_repository_`, but created session is db session and should be added to db resource repository. Calling add_dependent_session on either of the repositories would end up in incorrect behavior.

source/custom/nixnet_service.custom.cpp
```diff
@@ -431,46 +442,114 @@ ::grpc::Status NiXnetService::GetProperty(::grpc::ServerContext* context, const
...
+        auto initiating_session_id = session_repository_->access_session_id(session_grpc_session.id(), session_grpc_session.name());
+        auto init_lambda = [&]() {
+          nxSessionRef_t property_value;
+          status = library_->GetProperty(session, property_id, property_size, &property_value);
+          return std::make_tuple(status, property_value);
+        };
+        uint32_t session_id = 0;
+        // We are adding it to session_repository_ and not to "device repository", because devices don't have a close API,
+        // so it makes sense to tie their lifetime with sessions lifetime.
+        status = session_repository_->add_dependent_session("", init_lambda, initiating_session_id, session_id);
```
doshirohan, line 494: note: We can use add_dependent_session here and below because both sessions belong to same repository

source/custom/nixnet_service.custom.cpp
astarche: This doesn't sound right to me, or I am not following your logic.

Note that all sessions are in the global `session_repository_`. We use the `*_resource_repository_` to resolve the driver-specific handle. When you register a `dependent_session`, the `initiating_session` will always be the ID in the `session_repository_` because we need to be able to link their lifetimes even if they have different resource types. 

***Remember that `dependent_sessions` were specifically added for cross-driver sessions. So it certainly should work to use them with different repositories.***

The question "should this be a dependent session or not?" should be about: "do we need to call close on this object?" and "would it be valid to remove this object from the map when some initiating object is closed?" But none of your comments seem to be about that question.

I thought that the DB objects did not have a "close" method and, therefore, we didn't have a "normal" way to clean them up. And also, their lifetimes are specifically linked to the owning object (a hierarchy leading up to the database itself). If we call `add_session` for objects without a close method, we are effectively leaking them in the map (until reset server is called).

It's also not clear either from your PR description or the "testing done" that you found a specific behavioral problem with this code or verified that it was fixed. And there is nothing about ensuring that didn't break "clean up of database objects".

source/custom/nixnet_service.custom.cpp
```diff
@@ -431,46 +442,114 @@ ::grpc::Status NiXnetService::GetProperty(::grpc::ServerContext* context, const
...
+        if (status_ok(status)) {
+          response->mutable_dev_ref()->set_id(session_id);
+        }
+        break;
+      }
+      case dev_ref_array_: {
+        auto initiating_session_id = session_repository_->access_session_id(session_grpc_session.id(), session_grpc_session.name());
```
astarche, line 501: You should be able to put all of these in helper methods and call them in a test.

source/custom/nixnet_service.custom.cpp
astarche: await: If you're not just reverting this change, I would like to see a follow up with more details/tests/etc.

doshirohan: Looks like I had misunderstood `add_dependent_session` concept. Looking through the code I thought the entry will be added to session repository that you called `add_dependent_session` on. I also assumed that generated session_id are unique per-repository, didn't realize that they are unique per server instance and everything has entry in `session_repository_`. I have reverted this change.

source/custom/nixnet_service.custom.cpp
doshirohan: Yep, I have some other cleanup to do here in all get/set properties APIs. I will refactor them into helpers for all of these as part of that.

astarche (approved):

PR #643, opened 2022-04-01
Add system tests for XNET LIN

ni-pkhare:

Adds some system tests for XNET LIN

For loopback tests where first we are writing signals and then reading signals, 2 XNET modules must be connect physically, so added 1 message for it to be printed in case of failure of tests.

Ran SystemTestsRunner.exe


source/tests/system/nixnet_lin_driver_api_tests.cpp
```diff
@@ -0,0 +1,167 @@
...
+using namespace ::testing;
+using namespace ::nixnet_utilities;
+using nlohmann::json;
+
+namespace ni {
+namespace tests {
+namespace system {
+namespace {
+
+typedef pb::uint8 u8;
```
astarche, line 28: nit: prefer type aliases to `typedef` in C++11+. In this case:

```c++
using u8 = pb::uint8;
```

source/tests/system/nixnet_lin_driver_api_tests.cpp
```diff
@@ -0,0 +1,167 @@
...
+  constexpr auto NUM_FRAMES = 2;
+  constexpr auto NUM_SIGNALS = 2;
+  std::vector<nixnet_grpc::FrameBufferRequest> frames;
+  auto session = EXPECT_SUCCESS(client::create_session(stub(), "NIXNET_exampleLDF", "Cluster", "SlaveSignal1_U16,MasterSignal1_U16", "", CREATE_SESSION_MODE_SIGNAL_CONVERSION_SINGLE_POINT)).session();
+  nixnet_grpc::FrameRequest* frame = new nixnet_grpc::FrameRequest();
+  frame->set_timestamp(0);
+  frame->add_flags(FrameFlags::FRAME_FLAGS_UNSPECIFIED);
+  frame->set_identifier(4);
+  frame->set_type(FrameType::FRAME_TYPE_LIN_DATA);
+  frame->set_payload("\0\1\2\3\4\5\6\7");
```
astarche, line 75: Does this actually set the whole string as a payload? I think the `\0` in a char* constant will end up null terminating it.

In other cases I have used the `""s` suffix from `std::string_literals`. (A `std::string` can contain null, but a C string can't).

source/tests/system/nixnet_lin_driver_api_tests.cpp
```diff
@@ -0,0 +1,167 @@
...
+  frame2->set_payload("\0\1\2\3\4\5\6\7");
+  frames.push_back(nixnet_grpc::FrameBufferRequest());
+  frames.back().set_allocated_lin(frame2);
+
+  auto convert_frames_to_signals_single_point_response = EXPECT_SUCCESS(client::convert_frames_to_signals_single_point(stub(), session, NUM_SIGNALS, frames));
+  std::vector<f64> value_buffer_copy(convert_frames_to_signals_single_point_response.value_buffer().begin(), convert_frames_to_signals_single_point_response.value_buffer().end());
+  auto convert_signals_to_frames_single_point_response = EXPECT_SUCCESS(client::convert_signals_to_frames_single_point(stub(), session, value_buffer_copy, NUM_FRAMES, 8, Protocol::PROTOCOL_LIN));
+  EXPECT_SUCCESS(client::clear(stub(), session));
+
+  EXPECT_EQ(2, convert_frames_to_signals_single_point_response.timestamp_buffer().size());
```
astarche, line 92: Have you been able to look at the frames in the debugger at least? This should be the same data that is populated in the frames above, right?

source/tests/system/nixnet_utilities.h
```diff
@@ -0,0 +1,195 @@
...
+namespace nixnet_utilities {
+namespace {
+
+namespace client = nixnet_grpc::experimental::client;
+namespace pb = google::protobuf;
+using namespace nixnet_grpc;
+
+typedef pb::uint32 u32;
+
+GetPropertyResponse get_property(const client::StubPtr& stub, const nidevice_grpc::Session& session_ref, const client::simple_variant<nixnet_grpc::Property, pb::uint32>& property_id)
```
astarche, line 28: Are these actually called?

source/tests/system/nixnet_utilities.h
```diff
@@ -0,0 +1,195 @@
...
+    return response;                 \
+  })(response_)
+
+#define EXPECT_XNET_ERROR(error, response) \
+  if (1) {                                 \
+    EXPECT_EQ(error, (response).status()); \
+  }
+
+namespace nixnet_utilities {
+namespace {
```
astarche, line 20: nit: I don't think anonymous namespaces work in headers because they get pulled into the including translation unit. In this case, Normally they would be used in cpp files to avoid defining stuff in shared/global namespaces and potentially violating the One Definition Rule.

Here are the relevant core guidelines:
* https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed
* https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed2

You can use a `detail` namespace for a similar purpose, that's what I'd use for a shared header-only library. I also use the `_` prefix or postfix in headers that are only used internally (like you're doing below).

In this case, I don't see where the methods are called, so I'm not sure what the intent is.

astarche (approved):
Looks good!

source/tests/system/nixnet_lin_driver_api_tests.cpp
doshirohan: Yep, me and Sumit were debugging similar scenario earlier. This sets payload to empty string. Another `protoc` generated workaround we found was to specify size of string explicitly. So below snippet should work too.

`frame->set_payload("\0\1\2\3\4\5\6\7", 8);`

source/tests/system/nixnet_utilities.h
astarche: Speaking of the ODR, you also probably need to inline these methods. At least before they get included twice.

source/tests/system/nixnet_lin_driver_api_tests.cpp
ni-pkhare: Yes, they were same.

source/tests/system/nixnet_utilities.h
ni-pkhare: Right now, only few functions from them are being used but these all were added because they were being used in some other tests which will be added soon. If ill find any function not being used then ill remove them in subsequent PRs

source/tests/system/nixnet_utilities.h
```diff
@@ -0,0 +1,202 @@
...
+}
+
+inline void _set_property_value(SetPropertyRequest& request, const double& value)
+{
+  request.set_f64_scalar(value);
+}
+
+inline void _set_property_value(SetPropertyRequest& request, const std::vector<std::string>& value)
+{
+  // TODO: set_string_array expects a single string?!?
```
reckenro, line 82: Is this tracked somewhere?

reckenro (approved):

mgumble (approved):

SparkingSpork (approved):

source/tests/system/nixnet_lin_driver_api_tests.cpp
```diff
@@ -0,0 +1,187 @@
...
+  EXPECT_EQ(2, convert_frames_to_signals_single_point_response.value_buffer().size());
+  EXPECT_EQ(2, convert_signals_to_frames_single_point_response.buffer().size());
+  assert_lin_frames_are_equal(frame1, &lin_1);
+  assert_lin_frames_are_equal(frame2, &lin_2);
+  EXPECT_SUCCESS(client::clear(stub(), session));
+  frames.clear();
+}
+
+TEST_F(NiXnetLINDriverApiTests, FrameStreamInputFromExample_FetchData_DataLooksReasonable)
```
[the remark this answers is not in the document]
ni-pkhare: We planned to just check the status here and as part of some other test we will check the frame.

source/tests/system/nixnet_lin_driver_api_tests.cpp
```diff
@@ -0,0 +1,187 @@
...
+  EXPECT_SUCCESS(client::clear(stub(), output_session));
+  EXPECT_SUCCESS(client::clear(stub(), input_session));
+
+  EXPECT_EQ(output_value_vtr[0], read_signal_single_point_response.value_buffer()[0]) << "LIN1 and LIN2 must be connected together.";
+  EXPECT_EQ(output_value_vtr[1], read_signal_single_point_response.value_buffer()[1]) << "LIN1 and LIN2 must be connected together.";
+}
+
+TEST_F(NiXnetLINDriverApiTests, WriteSignalXYFromExample_ReadSignalXYFromExample_SignalMatches)
+{
+  constexpr auto NUM_SAMP = 3;
```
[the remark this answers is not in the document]
ni-pkhare: Done

source/tests/system/nixnet_lin_driver_api_tests.cpp
```diff
@@ -0,0 +1,187 @@
...
+};
+
+TEST_F(NiXnetLINDriverApiTests, ConvertFramesToFromSignalsFromExample_FetchData_DataLooksReasonable)
+{
+  constexpr auto NUM_FRAMES = 2;
+  constexpr auto NUM_SIGNALS = 2;
+  std::vector<nixnet_grpc::FrameBufferRequest> frames;
+  auto session = EXPECT_SUCCESS(client::create_session(stub(), "NIXNET_exampleLDF", "Cluster", "SlaveSignal1_U16,MasterSignal1_U16", "", CREATE_SESSION_MODE_SIGNAL_CONVERSION_SINGLE_POINT)).session();
+  nixnet_grpc::FrameRequest* frame1 = new nixnet_grpc::FrameRequest();
+  frame1->set_timestamp(0);
```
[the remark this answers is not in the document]
ni-pkhare: Done

source/tests/system/nixnet_utilities.h
ni-pkhare: yesh, recently we clubbed string_array case with string itself, so not we dont need to deal with it differently. Hence I have removed that part.

PR #648, opened 2022-04-05
Add get_last_error mechanism to IVI init methods.

reckenro:

Updates IVI based drivers to have an error_message field appended to initialization response messages. It's populated using the same `get_last_error` codegen pattern used for RFmx.

Fixes [AB#1873996](https://ni.visualstudio.com/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1873996)
When an initialization method fails, we should tack on the error message to the response. We did this with RFmx to avoid some problems with the previous pattern of getting the error message after the initialization response was returned.

Ran all *SessionTest* tests and they all pass.
Ran all updated examples and they still pass. Also, for each one I updated the initialization to fail (invalid resource) and it printed out the failure as expected.

## Breaking Change

Marking this as a breaking change. It's not API breaking as the client can continue to use their pre-existing code unaware of the new field on the end of initialization responses. It is breaking in the behavioral sense, though, since there is a new pattern to detect errors for initialization calls as outlined in the updated examples.

### DIFF
--- a/examples/nidcpower/measure-record.py
+++ b/examples/nidcpower/measure-record.py
@@ -68,6 +68,14 @@ def check_for_error(vi, status):
         raise Exception(error_message_response.error_message)
 
 
+def check_for_initialization_error(response):
+    """Raise an exception if an error was returned from Initialize."""
+    if response.status < 0:
+        raise RuntimeError(f"Error: {response.error_message or response.status}")
+    if response.status > 0:
+        sys.stderr.write(f"Warning: {response.error_message or response.status}\n")
+
+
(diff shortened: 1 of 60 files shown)


reckenro:
Updated:
- DC Power
- DMM
- Digital Pattern
- FGEN
- RFSA
- RFSG
- Scope
- Switch

Not Updated:
- Sync - No `GetError` method similar to the other drivers to use for get_last_error.
- Tclk - no initialize methods

mgumble (approved):

astarche (approved):

Issue #658, opened 2022-04-12
Ethernet frame is missing last four bytes in the payload


Akshata440:
![image](https://user-images.githubusercontent.com/47028785/162894522-0ed1a1dd-f011-41ab-9454-b3bfb7492049.png)
Using XNET `WriteFrame` API, I am writing an ethernet frame. After reading the same frame using `ReadFrame` API, the last four bytes of payload are zeroed out.

PR #654, opened 2022-04-08
Add system test for XNET - Ethernet

Akshata440:
### What does this Pull Request accomplish?
Adding System test for Ethernet

### Why should this Pull Request be merged?

Added a loopback test where I am writing an ethernet frame and then reading back the same frame.

### What testing has been done?

Ran the SystemTestRunner.exe on a machine with Xnet hardware. Test passed successfully.

PR #659, opened 2022-04-12
Fixed payload size issue

Akshata440:
Changes made to fix the issue #658 

`releases\1.5` branch will be fast-forwarded to include this fix and minor/patch release will be published with this bug fix as it fixes showstopper bug for ENET frames in XNET.

While calling `WriteFrame` API, we need to pass in buffer with last 4 bytes of buffer reserved for adding FCS (frame checksum). in gRPC service we take payload data as input from user and allocate buffer and copy payload that user has sent in. Earlier we had missed allocating extra 4 bytes for the FCS while allocating buffer for copying data. Hence, 4 bytes of the payload were not being written correctly. I have added a new constant called `ETHERNET_FRAME_FCS_SIZE` which is now used for calculating payload sizes while reading and writing frames.

Ran system tests in #654 to confirm the whole payload is read correctly.
Before : 
image

After : 
image

### DIFF
--- a/source/custom/nixnet_converters.h
+++ b/source/custom/nixnet_converters.h
@@ -16,6 +16,7 @@ void SetSessionInfoResponse(const u32& input, nixnet_grpc::SessionInfoResponse* 
 }  // namespace converters
 
 constexpr auto ENET_FRAME_HEADER_LENGTH = static_cast<u16>(sizeof(nxFrameEnet_t) - 1);  // last byte in nxFrameEnet_t is u8 FrameData[1]
+constexpr auto ENET_FRAME_FCS_SIZE = 4;
 
 struct FrameHolder {
   FrameHolder(const pb_::RepeatedPtrField<FrameBufferRequest>& input, std::map<std::int32_t, std::int32_t> enetflags_input_map)
@@ -103,7 +104,7 @@ struct FrameHolder {
         throw std::invalid_argument("All FrameBufferRequest instances in repeated field should have same oneof set for the frame");
       }
 
-      size_t frame_size = ENET_FRAME_HEADER_LENGTH + grpc_frame.enet().frame_data().length();
+      size_t frame_size = ENET_FRAME_HEADER_LENGTH + ENET_FRAME_FCS_SIZE + grpc_frame.enet().frame_data().length();
       frame_data.resize(frame_size, 0);
       nxFrameEnet_t* current_frame = (nxFrameEnet_t*)frame_data.data();
       // The Length field in ENET write frame is big-endian. Typecast to u16 before doing the conversion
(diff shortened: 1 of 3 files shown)


source/custom/nixnet_converters.h
```diff
@@ -103,7 +104,7 @@ struct FrameHolder {
         throw std::invalid_argument("All FrameBufferRequest instances in repeated field should have same oneof set for the frame");
       }
 
-      size_t frame_size = ENET_FRAME_HEADER_LENGTH + grpc_frame.enet().frame_data().length();
+      size_t frame_size = ENET_FRAME_HEADER_LENGTH + ENET_FRAME_FCS_SIZE + grpc_frame.enet().frame_data().length();
```
doshirohan, line 107: Can you call the helper to compute size instead? That way the logic will be single place.

```
get_frame_buffer_size(
  number_of_frames=1,
  max_payload_per_frame=grpc_frame.enet().frame_data().length(),
  protocol=Protocol::ENET`
);
```

doshirohan (approved):

source/custom/nixnet_converters.h
Akshata440: Done

source/custom/nixnet_converters.h
```diff
@@ -16,6 +16,9 @@ void SetSessionInfoResponse(const u32& input, nixnet_grpc::SessionInfoResponse*
 }  // namespace converters
 
 constexpr auto ENET_FRAME_HEADER_LENGTH = static_cast<u16>(sizeof(nxFrameEnet_t) - 1);  // last byte in nxFrameEnet_t is u8 FrameData[1]
+constexpr auto ENET_FRAME_FCS_SIZE = 4;
+
+u32 get_frame_buffer_size(int32 number_of_frames, u32 max_payload_size, u32 protocol);
```
Akshata440, line 21: Note : Moved this helper declaration here since I needed to declare this before using it on line 109 below.

mgumble:
Well this is a major bummer, releasing XNET with this bug. @asumit  do you think we just didn't do enough manual testing to catch this before creating the release? 

@Akshata440 You mentioned we would fast forward the release branch, but we need to update the artifacts too and create yet another release I guess with a minor version bump?
@astarche Thoughts on versioning?

mgumble (approved):

astarche (approved):

astarche:
The version should be `1.5.1`. The idea of fast-forwarding and using the same branch sounds right to me as well.

reckenro (approved):

doshirohan:
> Well this is a major bummer, releasing XNET with this bug. @asumit do you think we just didn't do enough manual testing to catch this before creating the release?
> @Akshata440 You mentioned we would fast forward the release branch, but we need to update the artifacts too and create yet another release I guess with a minor version bump?


We did test the API manually and we were in process of automating that via test when we encountered the issue. During testing, we had used XNET C API for generating frames and were reading those gRPC API in loopback test to validate the data. This had passed, though this issue had turned out to be in write frame itself which we couldn't catch with that manual test.

We are planning to create 1.5.1 patch release for this once the PR goes in.

@Akshata440  @mgumble @astarche @reckenro @asumit

PR #667, opened 2022-06-02
Address missing ni-scope enums.

reckenro:

- Fixes ni-scope missing enums (deleted or referenced by appropriate function/attribute)
- Fixed examples with new enum values. Also, fixed the examples to use GetErrorMessage instead of ErrorMessage.

Fixes [AB#1991317](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1991317)

Build passes. Ran validate_examples.py on niscope to find and fix changed/removed enums.
To ensure old client with new server is able to still run installed and ran the new server with these updates on a lvtest machine, on dev machine built old client protos and ran calls to Init and Configure Clock. These calls passed and operated the same when connecting to the new server on the lvtest machine and when connecting to an old server on localhost.


reckenro:
### Decision process for missing enums

- 5900AccessoryInAScopeSessionValues (Removed)
	- In https://p4.natinst.com/browser/perforce/build/exports/ni/nisc/niScopeHeaders/official/export/22.5/22.5.0d4/includes/niScopeHeaders/niScopeInternalAttributes.h?revision_at=65207436#line_409
	- Internal attributes header and I don't think we're exposing that attribute
	- Also mixed bag of strings and doubles which we don't support. Could maybe separated into two enums.
- AccessoryTypes (Removed)
	- Also in that internal attributes header. Don't think it's public nor relating to any attributes we're exposing
- AgcAverageControl (Removed)
	- I think it used to be used by obsolete attribute NISCOPE_ATTR_DDC_AGC_AVERAGE_CONTROL
- AoutParallelOutputSource (Removed)
	- Used by obsolete attribute 
	NISCOPE_ATTR_DDC_AOUT_PARALLEL_OUTPUT_SOURCE
- BoutParallelOutputSource (Removed)
	- Used by obsolete attribute 
	NISCOPE_ATTR_DDC_BOUT_PARALLEL_OUTPUT_SOURCE
	
- AttributeRetrievalModes (Removed)
	- I think this is just used for niScope_GetAttributeWithOptionsVi* which we don't have included
	
- CalEndActions (Removed)
	- Used in niScope_CalEnd which we don't have included

- CalibrationAccessorySources (Removed)
	- Used in niScope_CalSetAccessorySource which we don't have included

- CalibrationOptions (Removed)
	- Used in niScope_CalSelftCalibrate which we don't have included

- ClockingTerminalValues (Added)
	- Mark generate-mappings and reference from ExportSignal and 
	ConfigureClock
	

- CoefficientType  (Removed)
	- Obsolete, not referenced by anything

- ConfigurationListStepTriggerSources (Removed)
	- Internal, referenced by attribute we don't include: 
	NISCOPE_ATTR_CONFIGURATION_LIST_STEP_TRIGGER_SOURCE
	
- ControlBandwidthLimitBandpassFilters (Removed)
	- Internal, referenced by attribute we don't include: 
	NISCOPE_ATTR_CONTROL_BANDWIDTH_LIMIT_BANDPASS_FILTER

- CoordinateConverterInput (Removed)
	- Think it would be used by obsolete attribute 
	NISCOPE_ATTR_DDC_COORDINATE_CONVERTER_INPUT that we don't include
	
- DDCMuxModes (Removed)
	- Would be used by NISCOPE_ATTR_DDC_MUX_MODE but we don't include it (internal)

- DataJustificationMode (Removed)
	- Would be used by obsolete 
	NISCOPE_ATTR_FETCH_DATA_JUSTIFICATION but we don't include it
	
- DeviceStatuses (Removed)
	- Internal and not referenced by anything

- DiscriminatorFIRInputSource (Removed)
	- Would be used by obsolete/not included 
	NISCOPE_ATTR_DDC_DISCRIMINATOR_FIR_INPUT_SOURCE
	
- DiscriminatorFIRSymmetry (Removed)
	- Would be used by obsolete/not included 
	NISCOPE_ATTR_DDC_PFIR_SYMMETRY and 
	NISCOPE_ATTR_DDC_DISCRIMINATOR_FIR_SYMMETRY
	
- DiscriminatorFIRSymmetryType (Removed)
	- Would be used by obsolete/not included 
	NISCOPE_ATTR_DDC_PFIR_SYMMETRY_TYPE and 
	NISCOPE_ATTR_DDC_DISCRIMINATOR_FIR_SYMMETRY_TYPE
	
- FetchRecordModes (Removed)
	- Would be used by internal/not included attribute NISCOPE_ATTR_FETCH_RECORD_MODE 

- FetchWaitModes (Removed)
	- Would be used by internal/not included attribute NISCOPE_ATTR_FETCH_WAIT_MODE 

- HorzConfigModes (Removed)
	- Would be used by internal/not included attribute NISCOPE_ATTR_HORZ_CONFIG_MODE 

- InputImpedance (Added)
	- Mark generate-mappings and reference from NISCOPE_ATTR_INPUT_IMPEDANCE

- InternalExportSignalValues (Removed)
	- Internal and not referenced by anything

- InternalExternalCalibrationRoutingTerminal (Removed)
	- Internal and not referenced by anything

- Interpolation (Removed)
	- I think this is relating to IVISCOPE_ATTR_INTERPOLATION which isn't included

- LinearizationCalMethods (Removed)
	- Would be used in some internal/not included linearization cal attributes

- LvfpgaCoercionValues (Removed)
	- Related to private fpga prototype interface attributes

- MaxInputFrequency (Added)
	- Same as InputImpedance. Mark generate-mappings and reference from NISCOPE_ATTR

- MaxTime (Removed)
	- We could force-include but it exceeds an integer limit and we don't have a good way of having an mapped enum that's not referenced. How would user get the corresponding enum values?

- MeasFilterType  (Added)
	- Use it instead of FilterType enum (deleted) and reference this from attribute 
	NISCOPE_ATTR_MEAS_FILTER_TYPE
	
- MeasPercentageMethod (Added)
	- Use over PercentageMethod enum (deleted) and reference from attribute 
	NISCOPE_ATTR_MEAS_PERCENTAGE_METHOD
	
- MruSignalNames (Removed)
	- Internal defines used in Signal Name Query in RFSA, I think it's safe to remove them here

- OptionsForErrorHandling (Removed)
	- Related to error handling related to niScopeCal.h which we're not referencing/including

- P2PStreamRelativeTo (Added)
	- Should be referenced by 
	NISCOPE_ATTR_P2P_STREAM_RELATIVE_TO
	
- PllCommitActions (Removed)
	- Would be used by internal/not included attribute NISCOPE_ATTR_PLL_COMMIT_ACTION

- Privileges (Removed)
	- Would be used by niScope_RequestPrivelege which we don't include right now.

- ProgFIRFilterRealComplex (Removed)
	- Covered by DataProcessingMode enum (same values and names). Not sure what this would be used by.
- ProgFIRFilterSymmetry (Removed)
	- Would be used by obsolete/not included 
	NISCOPE_ATTR_DDC_PFIR_SYMMETRY
	
- ProgFIRFilterSymmetryType  (Removed)
	- Would be used by obsolete/not included 
	NISCOPE_ATTR_DDC_PFIR_SYMMETRY_TYPE
	
- QInputToCoordConverter (Removed)
	- I think this would be referenced by obsolete/not included NISCOPE_ATTR_DDC_COORDINATE_CONVERTER_INPUT 

- RecordCountModes (Removed)
	- Internal record count modes. Don't know where it'd be used from.
	
- ReferenceParameterForRoutingAndStoringAndFetchingIntRef (Removed)
	- Referenced from niScopeCal.h but don't think we have a use for it.

- ResamplerFilterMode (Removed)
	- I think this would be used in relation to some obsolete/not included attributes 
	NISCOPE_ATTR_DDC_RESAMPLER_***
	
- RouteInternalReferenceOptions (Removed)
	- Used by niScope_CalRouteInternalReference which we aren't including

- SampClkTimepaceSrc (Added)
	- Generate mappings and link from 
	NISCOPE_ATTR_SAMP_CLK_TIMEBASE_SRC

- SampleCountModes (Removed)
	- Similar to RecordCountModes, internal and not sure where they'd be used from in our service

- SampleMode (Added)
	- Should be referenced by 
	NISCOPE_ATTR_SAMPLE_MODE
	
- SessionState (Removed)
	- No idea, can't find references to the values anywhere

- StreamingPositionType (Removed)
	- No idea, can't find references anywhere

- SyncoutClkSelect (Removed)
	- I think it's related to obsolete/not included attribute
	NISCOPE_ATTR_DDC_SYNCOUT_CLK_SELECT
	
- TVTriggerSignalFormat (Removed)
	- Covered by VideoSignalFormat

- TimeHistogram (Removed)
	- Covered by 
	ClearableMeasurement
	
- TimingNcoFreqOffsetBits (Removed)
	- Related to NCO obsolete/not included attributes	

- TriggerSource (Added)
	- Generate mappings and link from 
	NISCOPE_ATTR_TRIGGER_SOURCE

- VideoFormats (Removed)
	- Internal/private that's not referenced anywhere

- VoltageHistogram (Removed)
	- Covered by 
	ClearableMeasurement

mgumble (approved):

generated/niscope/niscope.proto
```diff
@@ -999,10 +1089,19 @@ message ConfigureChanCharacteristicsResponse {
 
 message ConfigureClockRequest {
   nidevice_grpc.Session vi = 1;
-  string input_clock_source = 2;
-  string output_clock_source = 3;
-  string clock_sync_pulse_source = 4;
-  bool master_enabled = 5;
+  oneof input_clock_source_enum {
```
astarche, line 1092: This is a more significant breaking change than I think we've made in the past. Because the types of the slots have changed in an incompatible way (not just a rename or a type change). So, for example, slot 2 was a string and is now an enum. Slot 5 was a bool and is now a string. If only the name changes (or type changes to a compatible type, like int -> enum), then pre-existing clients still work. But source updates will be required if they re-protoc. But with this type of change, pre-existing clients will not work (random behavior, crash, etc).

Have you talked with @harsha-bhushan about whether we're OK with that? For sure, we would need to highlight it more in our release notes.

I think it's possible to avoid this, but it might be a bit tricky in the codegen. You could add a flag to enum fields for a `compatibility_mode`, and then give them indexes at the end:

```proto
message ConfigureClockRequest {
  nidevice_grpc.Session vi = 1;
  oneof input_clock_source_enum {
    ClockingTerminalValues input_clock_source_mapped = 6;
    string input_clock_source_raw = 2;
  }
  oneof output_clock_source_enum {
    ClockingTerminalValues output_clock_source_mapped = 7;
    string output_clock_source_raw = 3;
  }
  oneof clock_sync_pulse_source_enum {
    ClockingTerminalValues clock_sync_pulse_source_mapped = 8;
    string clock_sync_pulse_source_raw = 4;
  }
  bool master_enabled = 5;
}
```

at least .... I think you can do that.

astarche: There's a little more detail on `oneof` compatibility here: https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility_issues . That would only apply if you took my alternative above of keeping the numbering. I think adding a field to a `oneof` is safe, but you should test the case where an old client didn't set the field. That shouldn't cause insane behavior, but it might change behavior (from "default value" to "one of case -> none").

harsha-bhushan: My recommendation - make it as less breaking as possible.
Between codegen & backward compatibility for existing clients, we should prefer the latter. Client source breaking on re-protoc is still OK to an extent but unexpected behavior/ breaking compatibility without re-protoc is not OK.

danestull: I was able to get the correct less breaking behavior of maintaining the field numbers using the 'grpc_raw_field_number" and "grpc_mapped_field_number"

astarche: Looks great! That fix is much nicer than what I had in mind.

Did you test with an old client and confirm the behavior was OK? Can you do that and update the `Testing Done` to describe what you did?

astarche (approved):

generated/niscope/niscope.proto
danestull: I was able to get this testing done though it took a little more time than expected. I put a new version of the server (containing these changes) on an lvtest machine and connected to it and tested running a simple init configure clock with the new builds on my dev machine client that all worked as expected. On my dev machine I then reverted back to main and rebuilt everything. I ran locally with old client and server to verify just the string passing behavior and that passed as expected. Then I used the old client running simple init/configure clock to connect to the new server on the lvtest machine and that passed successfully so I think we have avoided the massively breaking changes!

PR #671, opened 2022-06-21
Create script to check if Linux RT feed needs updating.

reckenro:

Fixes AB#2009217

Creates a script to detect if the Linux RT Feed needs updating. The script will be supplied with a list of files that have changed and determine if any of them 

Stepping stone to be able to plug this into a GitHub Action to run on the releases branch so the developer creating a release knows if they need to update the Linux RT feed to point to the new release or not.

Ran the script against several inputs to make sure it's working as intended:

```
PS ...\grpc-device> "source/custom/niscope_service.custom.cpp" | python source/codegen/validate-linux-rt.py C:\dev\reckenro-git\grpc-device\source\codegen\metadata\
Linux RT Feed needs updating.
PS ...\grpc-device> "source/custom/rfsg_service.custom.cpp" | python source/codegen/validate-linux-rt.py ...\grpc-device\source\codegen\metadata\
Linux RT Feed does not need updating.
PS ...\grpc-device> git diff --name-only 46e1b9edd846505c21a9ee34bcdfc1689cc13cc9 HEAD | python source/codegen/validate-linux-rt.py ...\grpc-device\source\codegen\metadata\
Linux RT Feed needs updating.
PS ...\grpc-device>
```

### DIFF
--- a/source/codegen/validate-linux-rt.py
+++ b/source/codegen/validate-linux-rt.py
@@ -1,0 +1,69 @@
+"""Script for validating if a set of changed files warrants updating the Linux RT Feed.
+Expects output of git diff --name-only <last-release-commit> <current-commit> to be piped in so it can be read from sys.stdin"""
+
+import argparse
+import fnmatch
+import os
+import sys
+from template_helpers import load_metadata
+from typing import List
+
+
+def _check_core_files_affecting_rt(changed_files: List[str]) -> bool:
+    files_affecting_linux_rt = [
+        "source/config/*",
+        "source/custom/ivi_errors.h",
+        "source/protobuf/*",
+        "source/server/*",
+    ]
+
+    for pattern in files_affecting_linux_rt:
+        if fnmatch.filter(changed_files, pattern):
+            return True
+    return False
+
+
+def _check_driver_files_affecting_rt(driver_name: str, changed_files: List[str]) -> bool:
+    files_affecting_linux_rt = [
+        f"generated/{driver_name}/*",
+        f"examples/{driver_name}/*",
+        f"source/custom/{driver_name}*",
+    ]
+
+    for pattern in files_affecting_linux_rt:
+        if fnmatch.filter(changed_files, pattern):
+            return True
+    return False
+
+
+def _check_all(metadata_dir: str, changed_files: List[str]) -> bool:
+
+    if _check_core_files_affecting_rt(changed_files):
+        return True
+    rt_supported_drivers = [
+        driver
+        for driver in os.listdir(metadata_dir)
+        if load_metadata(f"{metadata_dir}/{driver}/")["config"]["linux_rt_support"]
+    ]
+    for driver in rt_supported_drivers:
+        if _check_driver_files_affecting_rt(driver, changed_files):
+            return True
+    return False
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Check if any file has changed that affects the Linux RT Feed."
+    )
+    parser.add_argument(
+        "metadata",
+        help="The path to the directory containing all of the metadata.",
+    )
+    args = parser.parse_args()
+    changed_files = list()
+    for line in sys.stdin:
+        changed_files.append(line.strip())
+    if _check_all(args.metadata, changed_files):
+        print("Linux RT Feed needs updating.")
+    else:
+        print("Linux RT Feed does not need updating.")


source/codegen/validate-linux-rt.py
```diff
@@ -0,0 +1,73 @@
...
+    ]
+
+    for pattern in files_affecting_linux_rt:
+        if fnmatch.filter(changed_files, pattern):
+            return True
+    return False
+
+
+def _check_all(metadata_dir: str, changed_files: List[str]) -> bool:
+
```
mgumble, line 44: nit: extra space

mgumble (approved):

source/codegen/validate-linux-rt.py
```diff
@@ -0,0 +1,73 @@
...
+import os
+import sys
+from typing import List
+
+from template_helpers import load_metadata
+
+
+def _check_core_files_affecting_rt(changed_files: List[str]) -> bool:
+    files_affecting_linux_rt = [
+        "source/config/*",
```
astarche, line 18: It seems like this is a little too easy to break by adding a new source folder. Those will silently default to not requiring a linux RT update. But shouldn't the opposite be the default? Or should it return an error for unexpected folders?

astarche: Or is there a common profile of a change that doesn't affect RT that we specifically care about? Like adding a non-RT driver? Or a bugfix release that happens to not touch any code on RT?

Is this feature more about notifying *why* RT changed in a release than *whether* it changed?

astarche (approved):
I have a lot of questions on how this is supposed to work. But if it meets your requirements, it works for me.

source/codegen/validate-linux-rt.py
reckenro: So the thought on this feature is to let whoever's making the grpc-device release know if they should update the Linux RT Feed after making the release. Essentially, letting them know whether they should go update the version of grpc-device Linux RT is pulling in: https://github.com/ni/meta-nilrt/blob/nilrt/master/hardknott/recipes-ni/ni-grpc-device/ni-grpc-device_git.bb

So this script will run based on a diff between the commit the draft release would be made from and the last release and try to tell the developer whether they will need to update the Linux RT Feed based on that.

With that all being said, this approach might not be the best. Right now it's working on a whitelist of RT supported driver updates and this set of files updating. But maybe it should be a blacklist of non-RT supported driver changes. So if there's any change other than one to a driver that doesn't support RT, we flag it as "needing to update Linux RT".

source/codegen/validate-linux-rt.py
reckenro: Removed.

source/codegen/validate-linux-rt.py
```diff
@@ -0,0 +1,73 @@
...
+        f"source/custom/{driver_name}*",
+    ]
+
+    for pattern in files_affecting_linux_rt:
+        if fnmatch.filter(changed_files, pattern):
+            return True
+    return False
+
+
+def _check_all(metadata_dir: str, changed_files: List[str]) -> bool:
```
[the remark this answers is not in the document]
reckenro: Renamed to `_needs_linux_rt_feed_update`

reckenro:
Talked with @astarche  offline about this and changed the approach. I'll reset both of you. The new approach is more akin to "non-RT driver changes" only or "non-RT driver changes and CMakeLists.txt change" only indicating updating the Linux RT Feed is likely not needed. Any changes outside those two scenarios probably warrants updating the Linux RT Feed.

source/codegen/validate-linux-rt.py
```diff
@@ -0,0 +1,73 @@
...
+import argparse
+import fnmatch
+import os
+import sys
+from typing import List
+
+from template_helpers import load_metadata
+
+
+def _check_core_files_affecting_rt(changed_files: List[str]) -> bool:
```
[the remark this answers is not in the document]
reckenro: I'll mark this resolved. Changed it to what we talked about offline to make the recommendation more sane.

source/codegen/validate-linux-rt.py
reckenro: We talked about this offline and I updated the approach to account for files we don't know about being added. The new approach is trying to see if the set of changes are likely just touching non-RT drivers to determine if the Linux RT feed should be updated or not.

source/codegen/validate-linux-rt.py
```diff
@@ -0,0 +1,66 @@
...
+    non_rt_driver_changes = set()
+
+    for pattern in files_affecting_linux_rt:
+        non_rt_driver_changes |= set(fnmatch.filter(changed_files, pattern))
+    return non_rt_driver_changes
+
+
+def _need_linux_rt_feed_update(metadata_dir: str, changed_files: Set[str]) -> bool:
+    codegen_changes = _get_codegen_changes(changed_files)
+    remaining_changes = changed_files - codegen_changes
```
reckenro, line 35: Remaining question I thought of is if we can additionally ignore "source/tests/*" changes like we can codegen changes... I think that's reasonable and avoids trying to figure out if a test change is related to a non-rt driver or not. @astarche , thoughts on that?

reckenro: Actually, I went ahead and made this change. I think it's reasonable to discard test changes. If there's only test changes plus non-RT driver changes then it's probably still safe to say the set of changes isn't affecting RT.

astarche (approved):

source/codegen/validate-linux-rt.py
```diff
@@ -0,0 +1,72 @@
...
+        driver
+        for driver in os.listdir(metadata_dir)
+        if not load_metadata(f"{metadata_dir}/{driver}/")["config"]["linux_rt_support"]
+    ]
+    non_rt_driver_changes = set()
+    for non_rt_driver in non_rt_drivers:
+        non_rt_driver_changes |= _get_non_rt_driver_changes(non_rt_driver, remaining_changes)
+    remaining_changes = remaining_changes - non_rt_driver_changes
+    if non_rt_driver_changes:
+        if len(remaining_changes) == 1 and next(iter(remaining_changes)) == "source/CMakeLists.txt":
```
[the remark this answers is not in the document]
reckenro: Ah, yes that works. Moved that check up to the main if statement so it's `if non_rt_driver_changes and remaining_changes == {"source/CMakeLists.txt"}`

mgumble (approved):

PR #672, opened 2022-06-23
Trigger script to validate linux rt feed after a new tag is pushed.

reckenro:

Fixes [AB#2009227](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2009227).

Adds a validate_linux_rt GitHub Action to run whenever a new tag is pushed (i.e., a new release is published). It will get the files that have changed between the last release and the new one and call validate_linux_rt.py to see if the Linux RT dependency on grpc-device needs to be updated to the new release version or not.

This will give a spot to check for developers creating releases in grpc-device to see if they need to update the Linux RT Feed post release.

In my fork, I pushed up some tags to test out the different scenarios
- Patch to previous release (tag v1.4.1 pushed up)
  - Job failed early: https://github.com/reckenro/grpc-device/actions/runs/2586323611
- Pre-release tag pushed up (tag v1.5.2-rc1)
  - Workflow didn't trigger
- New release tag pushed up (tag v1.5.2)
  - Workflow ran to completion: https://github.com/reckenro/grpc-device/runs/7121479474?check_suite_focus=true
  - Workflow created an issue in my fork to update Linux RT Feed: https://github.com/reckenro/grpc-device/issues/9

### DIFF
--- a/.github/workflows/validate_linux_rt.yml
+++ b/.github/workflows/validate_linux_rt.yml
@@ -39,6 +39,7 @@ jobs:
     - name: Fail if new tag is not latest tag
       if: ${{ github.ref_name != steps.latesttag.outputs.tag}}
       uses: actions/github-script@v3
+      with:
         script: |
             core.setFailed('The new tag is not the latest tag. Checking if the linux rt feed needs updating doesn't make sense on a patch to a previous release.')
 
--- a/.github/workflows/validate_linux_rt.yml
+++ b/.github/workflows/validate_linux_rt.yml
@@ -29,6 +29,6 @@ jobs:
           git describe --abbrev=0 --tags ${{github.ref_name}}^
         )"
 
-    - name: Run validate_linux_rt.py
+    - name: Update Linux RT Feed to ${{ github.ref_name }}?
       run: |
         git diff --name-only ${{ steps.previoustag.outputs.tag}}..${{ github.ref_name }} | python source/codegen/validate-linux-rt.py source/codegen/metadata
--- a/.github/workflows/validate_linux_rt.yml
+++ b/.github/workflows/validate_linux_rt.yml
@@ -26,7 +26,7 @@ jobs:
       id: previoustag
       run: >-
         echo "::set-output name=tag::$(
-          git describe --abbrev=0 ${{github.ref_name}}^
+          git describe --abbrev=0 --tags ${{github.ref_name}}^
         )"
 
     - name: Run validate_linux_rt.py
--- a/.github/workflows/validate_linux_rt.yml
+++ b/.github/workflows/validate_linux_rt.yml
@@ -25,4 +25,4 @@ jobs:
 
     - name: Run validate_linux_rt.py
       run: |
-        git diff --name-only ${{ steps.previoustag.outputs.tag}}..${{github.ref_name}} | python source/codegen/validate-linux-rt.py
+        git diff --name-only v1.5.1..${{github.ref_name}} | python source/codegen/validate-linux-rt.py
--- a/.github/workflows/validate_linux_rt.yml
+++ b/.github/workflows/validate_linux_rt.yml
@@ -28,4 +28,4 @@ jobs:
 
     - name: Run validate_linux_rt.py
       run: |
-        git diff --name-only v1.5.1..${{github.ref_name}} | python source/codegen/validate-linux-rt.py
+        git diff --name-only v1.5.1..${{github.ref_name}} | python source/codegen/validate-linux-rt.py source/codegen/metadata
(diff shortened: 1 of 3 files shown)


source/codegen/validate-linux-rt.py
```diff
@@ -36,9 +37,12 @@ def _get_non_rt_driver_changes(driver_name: str, changed_files: Set[str]) -> Set
 
 def _need_linux_rt_feed_update(metadata_dir: str, changed_files: Set[str]) -> bool:
     codegen_changes = _get_codegen_changes(changed_files)
+    _print_changed_files("Codegen changes not affecting RT:", codegen_changes)
```
reckenro, line 40: [Note] I went back and added these print statements to give more information in the GitHub action output (look under the section "Update Linux RT's grpc-device dependency to v1.5.2?" in https://github.com/reckenro/grpc-device/runs/7024760828?check_suite_focus=true.

.github/workflows/validate_linux_rt.yml
```diff
@@ -0,0 +1,34 @@
+name: Validate Linux RT Feed
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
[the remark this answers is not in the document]
mgumble: Wont this also happen when a PR is tagged with breaking change or if we used tags for other things?

.github/workflows/validate_linux_rt.yml
```diff
@@ -0,0 +1,34 @@
...
+        python-version: '3.x'
+
+    - name: Install build dependencies
+      run: python -m pip install -r python_build_requirements.txt
+
+    - name: Get previous tag
+      id: previoustag
+      run: >-
+        echo "::set-output name=tag::$(
+          git describe --abbrev=0 --tags ${{github.ref_name}}^
```
astarche, line 29: The last tagged commit would not necessarily be an ancestor of the current commit. I.e., there could be cherry-picks into the previous release branch. Would this catch that? I think that's relatively common.

I expect it's possible to get all `v` tags in the repo and sort them by semver. Not sure if that's the best way to do it. Sorting by date would work in most cases, but potentially not patch releases.

.github/workflows/validate_linux_rt.yml
```diff
@@ -0,0 +1,34 @@
...
+        )"
+
+    - name: Update Linux RT's grpc-device dependency to ${{ github.ref_name }}?
```
astarche, line 32: Is this aspirational? We're not actually updating here right? Maybe "Test update"

astarche (changes requested):
await: would like to see the follow up to review comments.

.github/workflows/validate_linux_rt.yml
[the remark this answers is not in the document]
SparkingSpork: This should only trigger on tags which match the filter `v*.*.*` are pushed. It would trigger if you used that format of tags for other things but if tags are only used when creating a new release this should be fine.

.github/workflows/validate_linux_rt.yml
```diff
@@ -0,0 +1,34 @@
+name: Validate Linux RT Feed
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  validate_linux_rt:
+    name: Validate Linux RT Feed
```
SparkingSpork, line 9: Maybe `Validate Linux RT Feed Package Version` so it's clearer what you're actually checking.

.github/workflows/validate_linux_rt.yml
SparkingSpork: Is someone going to be monitoring this script? Could we have this action make an AzDO work item in the backlog automatically?

.github/workflows/validate_linux_rt.yml
SparkingSpork: Same for the name of the workflow.

SparkingSpork (changes requested):

.github/workflows/validate_linux_rt.yml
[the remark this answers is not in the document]
reckenro: Yeah, also in the recent update I changed the filter to 'v[0-9]+.[0-9]+.[0-9]+' to restrict it a bit more. Realized we probably only want to trigger this check for release tags pushed up and not pre-releases (v1.5.1-rc0 or v1.5.1-internal).

.github/workflows/validate_linux_rt.yml
reckenro: Yeah, I updated these to sort by semver like the step I have in the create_release action.

.github/workflows/validate_linux_rt.yml
reckenro: Yeah, I like that better. Updated both the workflow and job name to the name you suggested.

astarche (approved):

PR #673, opened 2022-06-23
Remove mgumble from ownersfile

mgumble:

Leaving NI, while this is open source I will lose my 2FA not sure how that would work out

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,10 +1,10 @@
 # Default code owner for the repository
-* @astarche
+* @astarche @reckenro
 
 # Owners for the build files
-/CMakeLists.txt      @SparkingSpork @astarche
-/.github/**/*.yml    @SparkingSpork @astarche
-/scripts/            @SparkingSpork @astarche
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,14 +1,14 @@
 # Default code owner for the repository
-* @astarche @mgumble
```
astarche, line 2: @reckenro can we add you to these spots Mose is vacating? @DavidCurtiss (or anyone else) would you like to be added to any of these sections?

This can be addressed after this PR. Not like I can stop Mose leaving with open PR comments.

astarche (approved):

CODEOWNERS
mgumble: Any interest @reckenro  @DavidCurtiss  @dmondrik ?

DavidCurtiss: I've enjoyed not being responsible for reviewing everything. :) But I'm willing to step in if I'm the best person for it.

PR #674, opened 2022-06-29
Add GitHub action to make draft release at end of releases/* branch CIs.

reckenro:

- Creates new 'create_release' step on the CI to run when the CI is triggered from a releases/* branch
- New Create Release GitHub action creates a draft release with a proper title, tag, artifacts, and a template body.

Fixes [AB#2009191](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2009191)

Makes the release process more streamlined by eliminating a lot of the manual steps previously needed to create a release.

Created a releases/test branch on my fork with these changes, which after the CI ran created this draft release in my fork:

https://github.com/reckenro/grpc-device/releases/tag/untagged-f9b263420efe9d7cbd42

image


.github/workflows/create_release.yml
```diff
@@ -0,0 +1,90 @@
...
+    - uses: vimtor/action-zip@v1
+      with:
+        files: downloads/ni-grpc-device-server-windows-x64
+        dest: downloads/artifacts/ni-grpc-device-server-windows-x64.zip
+
+    - name: "Get previous tag"
+      id: previoustag
+      run: >-
+        echo "::set-output name=tag::$(
+          git tag --list --sort=-version:refname | grep '^\(v[0-9]\+.[0-9]\+.[0-9]\+\)$' | head -1
```
reckenro, line 58: Note: I'll use this for the linux RT validation action too. `git tag` looks at all tags in the repo, not just those accessible from the current commit so it addresses that. Sorting by the reverse version should get the tags in latest release to first release order. The grep is to just get actual release tags... this avoids it returning any internal or rc suffixes we want to avoid.

astarche: Does the regex match [rcs like this one](https://github.com/ni/grpc-device/releases/tag/v1.1.0-rc1)?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,90 @@
...
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - uses: actions/download-artifact@v3
+      with:
+        name: ni-grpc-device-client
+        path: downloads/ni-grpc-device-client
+
+    - uses: vimtor/action-zip@v1
```
astarche, line 24: Download artifact unzips the artifact. So for windows we have to re-zip it. Is that right?

.github/workflows/ci.yml
```diff
@@ -20,6 +20,13 @@ jobs:
     uses: ./.github/workflows/build_nilrt.yml
   create_client_artifacts:
     uses: ./.github/workflows/create_client_artifacts.yml
+  create_release:
+    if: ${{ startsWith(github.ref, 'refs/heads/releases') }}
```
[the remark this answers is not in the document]
SparkingSpork: And what will happen if you have a draft open and this runs again? Will it update the draft or make a new one?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,90 @@
...
+          Breaking Changes:
+          * List out any breaking changes introduced since the last release.
+
+          This release contains the ni-grpc-device-server application for Windows, Linux Desktop, and Linux RT.
+
+          It also contains client bundles that contain the proto files needed to build a gRPC client as well as several example clients.
+
+          Documentation on using this release can be found in our [README](https://github.com/ni/grpc-device#downloading-a-release) and [wiki](https://github.com/ni/grpc-device/wiki).
+        name: "NI gRPC Device Server ${{ steps.semvers.outputs.patch }}"
+        tag: ${{ steps.semvers.outputs.v_patch }}
```
SparkingSpork, line 89: Is this creating a new tag as well? I'd be leery of letting the CI create tags in this manner per Alex's concerns about multiple pushes to the release branch.

SparkingSpork (changes requested):
Ditto to Alex's concerns.

.github/workflows/ci.yml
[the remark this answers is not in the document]
reckenro: If there's a merge or cherry-pick after the release branch is created. It'll run the step again as part of the CI and create a new draft release similar to how it would on a new release branch.

I added to the if statement to avoid the create_release step running on PRs to a releases/* branch  though by adding `&& github.event_name == 'push'`. I'm testing that on a PR from this branch to a releases/* branch in my fork to make sure that it doesn't get kicked off.

It will make a new draft release currently. I could switch it to have an "allowUpdates" on the actual create_release action which I think will replace a draft release with the same proposed tag if it exists. If you look at my release drafts on my fork (https://github.com/reckenro/grpc-device/releases) right now I have a bunch of drafts triggered by different iterations of this action that would create the tag v1.5.2 if they were published.

.github/workflows/create_release.yml
reckenro: No, it'll just get the tags "v*" with the three numbers separated by '.'s. I was thinking that this script would just assume you're making a normal release so it should get the last normal release tag (last "v*" tag without any hyphen suffix) to avoid any pre-releases.

.github/workflows/create_release.yml
reckenro: Yes, or at least downloading the windows artifacts through these actions was just downloading the files/folders so we need to zip them up to include in the artifacts. The linux ones weren't zipped and instead are just single tar.gz files so they're fine to directly add to the artifacts. 

I'm not sure if they're zipped in the first place when the artifacts are uploaded or if it's just a directory. It might just be the directory and then when you click a link to one of them (say from a CI run) it zips them up and downloads that. Because from there, the linux ones are zipped too.

In any case, zipping them up in the workflow seems correct and produces artifacts in the draft release matching the form of the artifacts in our previous releases.

.github/workflows/create_release.yml
astarche: gotcha. that sounds reasonable. So, if you want to make `rc`s you need to change the name in the draft or make a new one or whatever. That sounds fine.

astarche (approved):

PR #675, opened 2022-06-29
Add Windows system testing to CI/PR builds

DavidCurtiss:

Add Windows system testing to CI/PR builds

Better test coverage

PR


.github/workflows/build_nilrt.yml
```diff
@@ -163,7 +163,6 @@ jobs:
 
     - name: Upload Tests Artifact
       uses: actions/upload-artifact@v2
-      if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/releases') }}
```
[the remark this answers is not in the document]
DavidCurtiss: Reverted that change. NILRT testing didn't work out.

.github/workflows/run_ubuntu_system_tests.yml
```diff
@@ -0,0 +1,30 @@
...
+        name: ni-grpc-device-tests-linux-glibc2_27-x64
+
+    - run: ls -R
+
+    - name: Untar Test Build Files
+      run: >-
+        tar -xvzf ni-grpc-device-tests-linux-glibc2_27-x64.tar.gz
+
+    - name: Run System Tests
+      run: ./SystemTestsRunner --gtest_filter=-NiRFSA*:NiRFSG*:NiDCPower*CalSelfCalibrate*
```
DavidCurtiss, line 29: These tests don't work on Linux. We should fix them, if possible, but in a separate change.

source/tests/system/nidcpower_session_tests.cpp
```diff
@@ -156,7 +158,7 @@ TEST_F(NiDCPowerSessionTest, InvalidSession_CloseSession_ReturnsInvalidSessionEr
   EXPECT_TRUE(status.ok());
   EXPECT_EQ(kInvalidDCPowerSession, response.status());
   std::string error_message = get_error_message(response.status());
-  EXPECT_STREQ(kInvalidDCPowerSessionMessage, error_message.c_str());
+  EXPECT_THAT(error_message.c_str(), HasSubstr(kInvalidDCPowerSessionMessage));
```
reckenro, line 161: Curious: Was this failing on Windows, Ubuntu, or both?

source/tests/system/nidmm_session_tests.cpp
```diff
@@ -140,7 +142,7 @@ TEST_F(NiDmmSessionTest, InvalidSession_CloseSession_ReturnsInvalidSesssionError
   EXPECT_TRUE(status.ok());
   EXPECT_EQ(kInvalidDmmSession, response.status());
   std::string error_message = get_error_message(response.status());
-  EXPECT_STREQ(kInvalidDmmSessionMessage, error_message.c_str());
+  EXPECT_THAT(error_message.c_str(), HasSubstr(kInvalidDmmSessionMessage));
```
reckenro, line 145: Same as DC Power, curious where this was failing.

reckenro (approved):

source/tests/system/nidcpower_session_tests.cpp
DavidCurtiss: On Ubuntu, IIRC.

SparkingSpork (approved):

.github/workflows/run_ubuntu_system_tests.yml
```diff
@@ -0,0 +1,30 @@
...
+      timeout-minutes: 27
```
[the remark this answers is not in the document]
DavidCurtiss: My thoughts were: AWS probably charges by the hour (though I didn't check), so let's keep it under 30 minutes so we can potentially get two runs in before the hour's up. In actuality, the (filtered) Ubuntu tests took about half a minute, and the (unfiltered) Windows tests took 18 minutes. So we've got ample room to grow before we hit the timeout.

astarche (approved):

source/tests/system/nirfmxwlan_driver_api_tests.cpp
```diff
@@ -185,8 +185,8 @@ TEST_F(NiRFmxWLANDriverApiTests, OFDMModAccTXPCompositeFromExample_FetchData_Dat
   const auto txp_fetch_measurement_response = client::txp_fetch_measurement(stub(), session, "", 10.0);
   EXPECT_WARNING(RISING_EDGE_DETECTION_FAILED_WARNING, RISING_EDGE_DETECTION_FAILED_WARNING_STR, session, txp_fetch_measurement_response);
 
-  EXPECT_LT(0.0, ofdm_mod_acc_fetch_composite_rmsevm_response.composite_rms_evm_mean());
-  EXPECT_LT(0.0, ofdm_mod_acc_fetch_composite_rmsevm_response.composite_data_rms_evm_mean());
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_composite_rmsevm_response.composite_rms_evm_mean());
+  EXPECT_NE(0.0, ofdm_mod_acc_fetch_composite_rmsevm_response.composite_data_rms_evm_mean());
```
dmondrik, line 189: What caused a previously passing test for less-than to now fail? Was it just an invalid assumption with a ticking time bomb?

dmondrik (approved):

source/tests/system/nirfmxwlan_driver_api_tests.cpp
DavidCurtiss: When we wrote the RFmx tests we didn't plan ahead of time what the expected output values _ought to_ be, we just looked at the actual value returned by the simulated device, decided it was reasonable, and wrote the test to watch for similar values. Sometimes, like in this case, the value started out less than or greater than zero during our initial testing, but further testing showed it to be sometimes the opposite. Mostly, though, we're just trying to verify that we can query through gRPC to get _some_ sort of value without it erroring. The _actual_ values should be tested by the tests for the actual driver.

PR #683, opened 2022-07-09
Add ProductId to device enumeration

epetersoni:

Adds `ProductId` to the `DeviceProperties` returned with an `EnumerateDevicesResponse`. This is desired for clients that want to easily determine the actual HW products that match the enumerated devices.

Simple change that just adds further information available via the already-in-use SystemConfigurationAPI to the enumerate device response.

Since the `DeviceProperties` data structure is only adding a new field, there's no need for existing clients or servers to update unless they want to take advantage of the new information.

* Added a unit test and verified that all unit tests pass.
* All integration tests pass.
* Added to an existing system test and it still passes.

### DIFF
--- a/source/protobuf/session.proto
+++ b/source/protobuf/session.proto
@@ -39,6 +39,7 @@ message DeviceProperties {
   string model = 2;
   string vendor = 3;
   string serial_number = 4;
+  bytes product_id = 5;
 }
 
 message EnumerateDevicesRequest {}
--- a/source/protobuf/session.proto
+++ b/source/protobuf/session.proto
@@ -39,7 +39,7 @@ message DeviceProperties {
   string model = 2;
   string vendor = 3;
   string serial_number = 4;
-  bytes product_id = 5;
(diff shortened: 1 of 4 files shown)


source/protobuf/session.proto
```diff
@@ -39,6 +39,7 @@ message DeviceProperties {
   string model = 2;
   string vendor = 3;
   string serial_number = 4;
+  bytes product_id = 5;
```
epetersoni, line 42: I had this as a `string` at first, but that caused errors about the values containing invalid UTF-8 data when serializing a protocol buffer when I ran the system tests.

The error says to "Use the `bytes` type if you intend to send raw bytes." so I switched to that, but I'm not certain this is what we actually want.

epetersoni:
Need to change the product ID to use unsigned int type instead of string.

reckenro (approved):

reckenro:
@epetersoni , Alex is out so you can submit this whenever you're ready.

Issue #685, opened 2022-07-22
Session name conflicts between drivers are reported with error code 1, which is ambiguous


bkeryan:
If the session name passed to a session creation RPC (like nidaqmx CreateTask or RF/MI Initialize RPCs) is used by a different driver, the grpc-device server returns error code 1 as a driver API error. However, looking up this error code using the driver API's error message function causes the example programs to displays "Warning: 1" or an inappropriate error message like "Command requires GPIB Controller to be Controller-In-Charge."

Steps to reproduce:
1. Change `examples/nidcpower/measure-record.py` to set `SESSION_NAME = "my task"`
2. Run `examples/nidcpower/measure-record.py`, then while it is running, run `examples/nidaqmx/analog-input.py`:
```
(.venv) PS Z:\grpc-device\examples\nidaqmx> python .\analog-input.py
Traceback (most recent call last):
  File ".\analog-input.py", line 77, in <module>
    raise_if_error(response)
  File ".\analog-input.py", line 63, in raise_if_error
    raise Exception(f"Error: {response.error_string}")
Exception: Error: Command requires GPIB Controller to be Controller-In-Charge.
```
3. Restart the grpc-device server.
4. Add `time.sleep(60.0)` to the nidaqmx example.
5. Run the nidaqmx example, then while it is running, run the nidcpower example:
```
(.venv) PS Z:\grpc-device\examples\nidcpower> python .\measure-record.py
Warning: 1
Traceback (most recent call last):
  File ".\measure-record.py", line 107, in <module>
    check_for_error(vi, configure_measure_when.status)
  File ".\measure-record.py", line 68, in check_for_error
    raise Exception(error_message_response.error_message)
Exception: IVI: The session handle is not valid.
```
It didn't create a NI-DCPower session, but the example program kept going because it reported error code 1 as a warning.
![image](https://user-images.githubusercontent.com/17914399/180476743-fa18d76f-c0bf-44a6-891e-25b5ed147457.png)

[AB#2085853](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2085853)

bkeryan:
Note that NI software already uses error code 1 for two different error messages:
![image](https://user-images.githubusercontent.com/17914399/180478188-5113cda1-78ac-425e-9729-bb39f24d811f.png)

In my opinion, the grpc-device server should report this error in one of two ways:
[...]

PR #687, opened 2022-07-27
Updates behavior when new session name is in use in different resource repository.

reckenro:

- Created a new `SessionException` struct in the shared library
- Throw a SessionException from the session resource repository when trying to add a session that's already in a different resource repository instead of returning a warning status of 1.
- Update codegen so all "Init" service handlers catch the new `SessionException` and return a gRPC Status of `INVALID_ARGUMENT` with the exception's message attached.
- Update the session resource repository unit test to now expect the `SessionException` to be thrown when trying to add a session where the session name is already in use in a different resource repository.

Addresses #685 

If a proposed session name is already used by a different driver, grpc-device was returning a warning status 1 on the return message's status field. This both allows programs making the call to keep running since it's treated as a warning and if the client does check the status code it can be ambiguous depending on the driver.

This change addresses this by instead immediately returning a gRPC Status of `INVALID_ARGUMENT` with a message explaining that the session name is already in use.

Updated a unit test testing the old behavior to now ensure the new SessionException is thrown if the session name is used in a different resource repository.

Ran the examples like they were specified in the Issue.

Now, running the daqmx example while dcpower is running using "my task" as the session name produces:
```console
PS ~\grpc-device\examples\nidaqmx> python .\analog-input.py
The session name "my task" is being used by a different driver.
```

And similarly, running the dcpower example while the daqmx example is running (with `time.wait(60.0)`) stops the program with the same message:
```console
PS ~\grpc-device\examples\nidcpower> python .\measure-record.py
The session name "my task" is being used by a different driver.
```

### DIFF
--- a/source/server/session_resource_repository.h
+++ b/source/server/session_resource_repository.h
@@ -114,7 +114,7 @@ int SessionResourceRepository<TResourceHandle>::add_session(
 
   // session_name resolves to a session in a different resource repository.
   if (!resource_map_->contains(session_from_repository)) {
-    throw nidevice_grpc::SessionException("The session name \"" + session_name + "\" is already being used in a different driver's service. Please use a different name.");
+    throw nidevice_grpc::SessionException("The session name \"" + session_name + "\" is being used by a different driver.");
   }
 
   session_id = session_from_repository;
--- a/source/server/session_resource_repository.h
+++ b/source/server/session_resource_repository.h
@@ -6,6 +6,7 @@
 #include <unordered_map>
 
 #include "session_repository.h"
+#include "shared_library.h"
 
 namespace nidevice_grpc {
 
@@ -112,11 +113,8 @@ int SessionResourceRepository<TResourceHandle>::add_session(
   }
 
   // session_name resolves to a session in a different resource repository.
-  // Report error.
   if (!resource_map_->contains(session_from_repository)) {
-    session_id = 0;
-    const int kError_InvalidArg = 42;
-    return kError_InvalidArg;
+    throw nidevice_grpc::SessionException("The session name \"" + session_name + "\" is already being used in a different driver's service. Please use a different name.");
   }
 
   session_id = session_from_repository;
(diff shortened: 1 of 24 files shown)


source/server/session_resource_repository.h
```diff
@@ -112,11 +113,8 @@ int SessionResourceRepository<TResourceHandle>::add_session(
   }
 
   // session_name resolves to a session in a different resource repository.
-  // Report error.
   if (!resource_map_->contains(session_from_repository)) {
-    session_id = 0;
-    const int kError_InvalidArg = 1;
-    return kError_InvalidArg;
+    throw nidevice_grpc::SessionException("The session name \"" + session_name + "\" is already being used in a different driver's service. Please use a different name.");
```
reckenro, line 117: I'm still ruminating on this message. It doesn't sound quite right because I think technically it could re-use a session from a different driver using the same resource type... I may be wrong about that though.

Also, I'm not sure if we should make a suggestion at the end of the message or not.

astarche: Ideas:

`The session name "my_cool_sesh" is in use and refers to an incompatible driver session resource`

`The session name "hw_that_does_stuff" is being used by a different driver`

You're correct that you would not hit this for drivers with compatible session types. I.e., all IVI drivers can share sessions (at least at this level, we'll see what happens when it gets to the API) and same with RFmx. I don't think we need to be too precise about that. But yeah "in a different driver's service" is too specific/wrong and a little fluffy anyway. But I think "by a different driver" is clear enough even if it's not 100% accurate.

astarche (commented):
Nice fix!

astarche:
@reckenro I added the breaking change tag here. The old behavior was pretty wrong and a bit of an edge/probably-a-mistake case anyway. But (I think) it's possible that someone could have worked around the old behavior in a way that would now break.

source/server/session_resource_repository.h
reckenro: Thanks, I went with the latter. It's short and accurate enough to know what's going on.

astarche (approved):

PR #689, opened 2022-08-03
[Owners] Add RFmx SpecAn Restricted service for internal functions

madrasm:

Adds a new nirfmxspecan_restricted service to support calling internal driver functions from internal products like Instrument Studio.

Required for committed feature to support grpc with RFmx SpecAn clients like Instrument Studio.

Added and ran system tests for all the functions in the new service.


madrasm:
@epetersoni @cumitche Can one of you please peer review this PR? I don't see how to add you as reviewers to this PR - may be I lack some permissions. Thanks!

madrasm:
@Wedgetable and @achanyou - FYI. Please feel free to peer review this as Curtis is OOO.

source/codegen/metadata/nirfmxspecan_restricted/config.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+config = {
+    "api_version": "21.0.0",
+    "restricted_driver_service": True,
```
madrasm, line 4: NOTE: New key to support functions exported from driver DLL but not in public header.

source/codegen/metadata/nirfmxspecan_restricted/functions.py
```diff
@@ -0,0 +1,98 @@
...
+            {
+                'direction': 'in',
+                'name': 'selectorStringOutSize',
+                'type': 'int32'
+            },
+            {
+                'direction': 'out',
+                'name': 'selectorStringOut',
+                'size': {
+                    'mechanism': 'passed-in',
```
madrasm, line 24: NOTE: This function does not support the IVI dance.

achanyou, line 24: How do we know when to use 'passed-in' vs. 'ivi-dance' for the mechanism?

madrasm: I found it the hard way when my test failed. I guess we just have to know the API - did not help that this one is private and not documented. Having said that, most APIs support ivi-dance and we should first use that and see if it works.

CMakeLists.txt
```diff
@@ -506,6 +506,7 @@ if(WIN32)
         "source/tests/system/nirfmxnr_session_tests.cpp"
         "source/tests/system/nirfmxspecan_driver_api_tests.cpp"
         "source/tests/system/nirfmxspecan_session_tests.cpp"
+        "source/tests/system/nirfmxspecan_restricted_driver_api_tests.cpp"
```
Wedgetable, line 509: Minor: Could you alphabetize this new entry?

madrasm: Fixed in commit 3f1229d.

source/codegen/metadata/nirfmxspecan_restricted/CHANGES.md
```diff
@@ -0,0 +1,6 @@
+# Changes in metadata 
+
+## config.py
+
+The following metadata was added:
```
achanyou, line 5: Is this here because it is a change from the base configuration? In which case, should this also mention the new "restricted_driver_service" tag? Or is this just left over from the component the file was copied from?

source/codegen/metadata/nirfmxspecan_restricted/config_addon.py
```diff
@@ -0,0 +1,5 @@
+# We need to maintain the version here since it needs to be updated by the build process on GitHub
+config_additional_config = {
+    'module_version': '1.3.3.dev0',
+    'latest_runtime_version_tested_against': '21.3.0',
```
achanyou, line 4: Is this supposed to be the driver version we tested against? Is the version accurate? Or is there a convention that we should we be testing against a specific older version?

source/codegen/metadata/nirfmxspecan_restricted/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+from .attributes import attributes
+from .enums import enums
+from .config import config
+
+metadata = {
+    "functions": functions,
+    "attributes": attributes,
+    "enums": enums,
+    "config": config
+}
```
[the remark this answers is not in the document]
madrasm: One of the checks that runs on the PR validates the python code and it passed. I think this is ok.

source/codegen/metadata/nirfmxspecan_restricted/config.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+config = {
+    "api_version": "21.0.0",
```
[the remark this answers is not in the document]
madrasm: I copied this from the nirfmxspecan/config.py file to keep them in sync.

source/codegen/metadata/nirfmxspecan_restricted/config_addon.py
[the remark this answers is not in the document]
madrasm: No idea. I just copied these values from the corresponding nirfmxspecan files.

achanyou (approved):

source/codegen/metadata/nirfmxspecan_restricted/config_addon.py
madrasm: I copied these over from the corresponding nirfmxspecan files. On my machine, I ran my tests with the latest 22.8 dev builds of the specan driver. But I am not sure what version the automated test pipeline is setup to use.

source/codegen/metadata/nirfmxspecan_restricted/CHANGES.md
madrasm: I think so. I copied the corresponding nirfmxspecan file. I added restricted_driver_service in commit 
c22188b.

source/codegen/metadata/nirfmxspecan_restricted/config.py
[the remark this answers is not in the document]
Wedgetable: @astarche or @reckenro, do either of you have guidance on this `api_version` and the `latest_runtime_version_tested_against` question below?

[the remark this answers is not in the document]
reckenro: Yeah, so I think this is a remnant of us starting from nimi-python metadata. We copied it over as a starting point but never cleaned up what we weren't using. nimi-python uses api_version, module_version, and latest_runtime_version_tested_against fields for different parts of their build but we haven't yet.

As for guidance, for this PR I think it's fine to match nirfmxspecan's values for these since they're not used. I think we should do a metadata cleanup as a whole for grpc-device to decide what values we can clear out. We have some metadata related features coming up on our squad so I think I'll bring this back up when we get to those.

source/codegen/metadata/nirfmxspecan_restricted/config_addon.py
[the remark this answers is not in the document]
Wedgetable: (See Ryan's reply [here](https://github.com/ni/grpc-device/pull/689#discussion_r937014132) - this value is currently unused so this is fine.)

Wedgetable (approved):

SparkingSpork (approved):

source/codegen/metadata/nirfmxspecan_restricted/config.py
```diff
@@ -0,0 +1,58 @@
...
+        "int8": "int32",
+        "uInt16": "uint32",
+        "uInt32": "uint32",
+        "uInt64": "uint64",
+        "uInt8": "uint32",
+        "uInt8[]": "bytes",
+        "NIComplexSingle": "nidevice_grpc.NIComplexNumberF32",
+        "NIComplexDouble": "nidevice_grpc.NIComplexNumber",
+    },
+    "driver_name": "NI-RFMXSPECAN-RESTRICTED",
```
reckenro, line 30: I'm thinking that we should add 'code_readiness': 'Prototype', or maybe a new type of "restricted" so this doesn't get included by default in our client artifacts (proto and examples). @astarche , thoughts on this? I think then the clients (like InstrumentStudio) that do want to use this restricted service can enable it on the server through the config file.

madrasm: I tried with code_readiness: Prototype and my tests started failing. I debugged this and it is because my tests are using DeviceServer in tests/system/device_server.cpp and this class only supports NextRelease and up. Should I make DeviceServer work with Prototype (instead of NextRelease)?

astarche: I like the idea to distinguish internal APIs with `code_readiness`. I think our current set of readiness states doesn't represent this well .

I think we should **avoid** having the "released" state of internal APIs be "prototype". I think it would also be helpful to have the "pre-release" state of internal apis be **different** from the pre-release state of (future) public APIs. But I don't think that's an absolute requirement.

I think this would be good:

```
InternalNextRelease, // Dev state for internal APIs
NextRelease,
Internal, // release state for internal APIs
Release
```

I like "internal", "restricted" or "unstable" as the qualifier for internal APIs. Any of those are fine.

I'd be OK with using `Prototype` instead of `NextReleaseInternal` for the dev state. But `Prototype` has a different meaning in our other apps, so I don't like that as much.

We borrowed the terms for readiness from ASW, which is not a perfect fit for this problem. If either of y'all want to consider a more significant rewording, here's an idea:

```
Development, // Dev state for Internal APIs and APIs that are not approaching release
Unstable, // Release state for Internal APIs
Hardening, // Dev state for Public APIs as they approach release
Release
```

Also note that `NextRelease` and `Internal` APIs may have the same behavior. That would make sense to a client. Using a pre-release driver API should be more-or-less similar to using an internal/unstable API that we don't intend to harden.

@madrasm Yes. Whatever we choose as the "dev state of internal APIs" should be the lowest readiness enabled by `device_server`. We've never used the "readiness" states that are "lower" than what `device_server` enables. Those lower "readiness" states are a bit dubious. The reason we would use them is to stage a change that has the potential to destabilize the system test suite (but that we still want to test with integration tests). That would not be a driver readiness, but possible the readiness of some internal feature. We don't need to worry about that too much, but whatever change we make we can document `readiness_enabled_by_device_server - 1 => a_dev_feature_that_we_want_to_disable_in_auto_tests`. In ASW, that readiness is `prototype`, but if it's `incomplete` or `tbd` that's OK too.

madrasm: Fixed in commit [b63bc2e](https://github.com/ni/grpc-device/pull/689/commits/b63bc2e2ccc3599d68327e9a5f4c8fcc41e95dc0). Went with 'Restricted' in the new code readiness fields to match with the nomenclature in the new files, etc.

source/codegen/metadata/nirfmxspecan_restricted/config.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+config = {
+    "code_readiness": "RestrictedNextRelease",
+    "api_version": "21.0.0",
+    "restricted_driver_service": True,
```
reckenro, line 5: I think you can remove this `"restricted_driver_service": True,` now and update your new `is_restricted_driver_service` helper to look at `code_readiness` instead.

reckenro (approved):

source/codegen/metadata/nirfmxspecan_restricted/config.py
madrasm: Fixed in commit [e63fe2a](https://github.com/ni/grpc-device/pull/689/commits/e63fe2a19661c5774909bfb32d2f2d940c609cbe).

source/codegen/metadata/nirfmxspecan_restricted/CHANGES.md
astarche: @madrasm : TL;DR it doesn't make since to have a "CHANGES.md" when there's no original source for it to have changed from. This file should be deleted.

@reckenro @DavidCurtiss

The original purpose of these files was to track any divergence from the hapigen metadata for MI drivers. I'm not sure that they were useful for that purpose (i.e., compared to looking at git history or writing a diff script) and I never used them when I worked on non-MI drivers. . We also have a different solution for tracking breaking_changes (pr tags).

The big test of the original intent of these drivers will be in David's feature to enable continuous updates. Have those notes been helpful? Maybe someone has discovered some other value for them?

If not, can we delete them? Or all of the non-hapigen ones?

astarche (approved):
Looks good!

source/codegen/metadata/nirfmxspecan_restricted/CHANGES.md
DavidCurtiss: I didn't realize they were here. Either way, we can delete them. If we decide they'll be helpful, we can get them from the history.

DavidCurtiss (approved):

source/codegen/metadata/nirfmxspecan_restricted/CHANGES.md
madrasm: Deleted the changes.md I added in commit [a5c4b90](https://github.com/ni/grpc-device/pull/689/commits/a5c4b905d7062c775a6b492a7571544fab9620ba).

reckenro: Yeah, I think it was semi-useful when we were adding new MI drivers early on to kind of see what kind of changes the other driver made in one place. But as of now, I'd just look at the git history or diff against nimi-python versions of the files if I was curious. So from my perspective we should be fine deleting them all.

reckenro: I'll make a quick PR to clear out the rest.

PR #209, opened 2021-05-20
Fixed nisync's ReadMultipleTriggerTimeStamp timeout behavior

tkrebes:
### What does this Pull Request accomplish?

Allows ReadMultipleTriggerTimeStamp to set timestamps_read when the operation times out.

This build on #203

### Why should this Pull Request be merged?

If the timeout duration is reached before the number time stamps read equals timestamps_to_read, ReadMultipleTriggerTimeStamp will return the number of read time stamps along and a timeout error. This normal operating behavior.

### What testing has been done?

Ran `.\SystemTestsRunner.exe --gtest_filter=*TimeStamp*` on Windows
```
Note: Google Test filter = *TimeStamp*
[==========] Running 7 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 7 tests from NiSyncDriver6683Test
[ RUN      ] NiSyncDriver6683Test.EnableTimeStampTrigger_ReturnsSuccess
[       OK ] NiSyncDriver6683Test.EnableTimeStampTrigger_ReturnsSuccess (1846 ms)
[ RUN      ] NiSyncDriver6683Test.EnableTimeStampTriggerWithInvalidTerminal_ReturnsError
[       OK ] NiSyncDriver6683Test.EnableTimeStampTriggerWithInvalidTerminal_ReturnsError (9 ms)
[ RUN      ] NiSyncDriver6683Test.DisableTimeStampTriggerNotReserved_ReturnsError
[       OK ] NiSyncDriver6683Test.DisableTimeStampTriggerNotReserved_ReturnsError (10 ms)
[ RUN      ] NiSyncDriver6683Test.GivenNoTrigger_ReadTriggerTimeStamp_ReturnsTimeoutError
[       OK ] NiSyncDriver6683Test.GivenNoTrigger_ReadTriggerTimeStamp_ReturnsTimeoutError (129 ms)
[ RUN      ] NiSyncDriver6683Test.GivenNoTrigger_ReadMultipleTriggerTimeStamp_ReturnsTimeoutError
[       OK ] NiSyncDriver6683Test.GivenNoTrigger_ReadMultipleTriggerTimeStamp_ReturnsTimeoutError (127 ms)
[ RUN      ] NiSyncDriver6683Test.GivenOneTrigger_ReadMultipleTriggerTimeStamp_ReturnsOneTimeStampAndTimeoutError
[...]

PR #692, opened 2022-08-04
API errors now return information via gRPC metadata (not the response)

dmondrik:

API errors are now returned as part of gRPC metadata rather than as part of the response. This is more natural for gRPC.
- The approach relies on custom methods that are per resource type, since in at least some cases the message comes dynamically from the resource handle (session).
  - Ideally the names of those methods could be simple and rely on parameter overloading to resolve the difference. However, XNET has 2 resources types that are both 32-bit integers. Therefore the simplest solution is to include the resource type name as part of the method name.
- The methods should be hand maintained because many of them require a fallback process. In MI for example, it returns the last error code for the session as well as the description. However, that may not be the status code we wanted to look up (that can happen if 2 threads are called in parallel and both fail with different status codes.)
  - In that case, such drivers have a different method of looking up a static error message for a given error code. The name and signature of that fallback method varies greatly.

[Task 2103382](https://ni.visualstudio.com/DevCentral/_workitems/edit/2103382): Change generated code to return UNKNOWN on error status.

I built locally. I have not run any tests manually but I am expecting they will run as part of this PR.


source/server/converters.h
```diff
@@ -317,6 +318,19 @@ typename TypeToStorageType<TDriverType, TGrpcType>::StorageType allocate_output_
 }
 
 }  // namespace converters
+
+const int kMaxGrpcErrorDescriptionSize = 4096;
+
+inline ::grpc::Status ApiErrorAndDescriptionToStatus(google::protobuf::int32 status, std::string& description)
+{
+    converters::trim_trailing_nulls(description);
+    // TODO: May need full-fledged json escape logic.
```
dmondrik, line 327: Note: I have since created a task for this. It shouldn't block submission.

bkeryan, line 327: It also ought to convert the string encoding to UTF-8, but that looks like a much larger issue than just error reporting. 

The string encoding may vary by API. Most are ANSI. DAQmx has a separate UTF-8 DLL which this server is not using.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,4 +1,12 @@
 #include <nidaqmx/nidaqmx_service.h>
 
 namespace nidaqmx_grpc {
+
+::grpc::Status NiDAQmxService::ConvertApiErrorStatusForTaskHandle(google::protobuf::int32 status, TaskHandle task)
+{
+    std::string description(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
+    library_->GetErrorString(status, description.data(), nidevice_grpc::kMaxGrpcErrorDescriptionSize);
```
bkeryan, line 8: `DAQmxGetErrorString` returns a static error message. For dynamic errors, call `DAQmxGetExtendedErrorInfo`.

source/custom/nidcpower_service.custom.cpp
```diff
@@ -55,4 +55,18 @@ ::grpc::Status NiDCPowerService::MeasureMultiple(::grpc::ServerContext* context,
...
+::grpc::Status NiDCPowerService::ConvertApiErrorStatusForViSession(google::protobuf::int32 status, ViSession vi)
+{
+    const ViInt32 buffer_size = 4096;
+    ViStatus error_code {};
+    std::string description(buffer_size, '\0');
+    // Try first to get the most recent error with a dynamic message.
+    library_->GetError(vi, &error_code, buffer_size, description.data());
+    if (error_code != status) {
+        // Since another thread has changed the status, fall back to the static message lookup.
+        library_->ErrorMessage(vi, status, description.data());
```
bkeryan, line 67: This function (and many of the similar MI error_message functions) assumes the buffer size is >= 256. That assumption is currently true, but consider documenting it.

source/custom/nifgen_service.custom.cpp
```diff
@@ -37,4 +37,17 @@ ::grpc::Status NiFgenService::CreateAdvancedArbSequence(::grpc::ServerContext* c
...
 
+::grpc::Status NiFgenService::ConvertApiErrorStatusForViSession(google::protobuf::int32 status, ViSession vi)
+{
+    ViStatus error_code {};
+    std::string description(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
+    // Try first to get the most recent error with a dynamic message.
+    library_->GetError(vi, &error_code, nidevice_grpc::kMaxGrpcErrorDescriptionSize, description.data());
+    if (error_code != status) {
+        // Since another thread has changed the status, fall back to the static message lookup.
+        library_->ErrorHandler(vi, status, description.data());
```
bkeryan, line 48: NI-FGEN has two similar functions with almost the same signature: `niFgen_ErrorHandler` and `niFgen_error_message`. 
- In `niFgenPAL.cpp`, `ErrorHandler` is a wrapper that calls `GetErrorInfo` for some errors and `error_message` for others, and I think `error_message` is the static error lookup. 
- In `niFgenMunichAPI.cpp`, `error_message` calls `ErrorHandler` ¯\\\_(ツ)\_/¯

Does this matter? Not much, but I think `niFgen_error_message` is closer to what the comment says this does.

source/codegen/templates/service_helpers.mako
```diff
@@ -719,30 +713,41 @@ ${initialize_hardcoded_parameter(parameter)}
...
 ## Handles populating the response message after calling the driver API.
-<%def name="populate_response(function_data, output_parameters, indent_level=0, init_method=False)">\
+<%def name="populate_response(function_data, parameters, indent_level=0, init_method=False)">\
 <%
+  output_parameters = [p for p in parameters if common_helpers.is_output_parameter(p)]
   get_last_error_outputs = service_helpers.get_last_error_output_params(output_parameters)
   normal_outputs = [p for p in output_parameters if not p in get_last_error_outputs]
 %>\
 <%block filter="common_helpers.indent(indent_level)">\
+${populate_error_check(function_data, parameters, indent_level)}
```
DavidCurtiss, line 723: You shouldn't pass down the passed-in indent_level here. That would basically double it. I'd have to try it out (or dig around a while longer) to figure out the proper value, but my gut tells me you should leave it off in this case (so that it's 0).

source/server/converters.h
```diff
@@ -317,6 +318,19 @@ typename TypeToStorageType<TDriverType, TGrpcType>::StorageType allocate_output_
...
+    for (size_t index = 0; (index = description.find('"', index)) != std::string::npos; ++index) {
+        description.replace(index, 1, "\\\"");
```
DavidCurtiss, line 329: I'm pretty sure you'll have an infinite loop here (until you run out of memory). If I'm reading it right, when you find a `"` character, you replace it with `\"` and then you still only increment `index` by 1, *which means you'll immediately find the `"` character you just [re]added and escape it again*!

(Also, instead of using `replace(index, 1, "\\\"")` you could use `insert(index, '\\')`, if you want.)

source/custom/nidcpower_service.custom.cpp
DavidCurtiss, line 67: Bug: Since `trim_trailing_nulls` (used in `ApiErrorDescriptionToStatus`) only "erase[s] from the first non-null character (going backwards) to the end," you can run into scenarios like this:
* Line 64 fills in `description` with 47 characters
* Line 67 overwrites `description` with only 33 characters
* As a result, after the `'\0'` at character 34, characters 35-47 will still have the original data from line 64, and `trim_trailing_nulls` will only trim the string down to 47 characters.

You need to either update `trim_trailing_nulls` accordingly, or null out the description buffer again before line 67.

DavidCurtiss (changes requested):

source/server/converters.h
bkeryan: If you don't want to add a dependency on a full-fledged JSON parser, consider the `encodeAndStrcpy` function from jsonz.

source/server/converters.h
```diff
@@ -317,6 +318,19 @@ typename TypeToStorageType<TDriverType, TGrpcType>::StorageType allocate_output_
...
+    }
+    std::string errorMessage = "{ \"code\":\"" + std::to_string(status) + "\", \"" + description + "\" }";
```
bkeryan, line 331: Google documents that this is a human-readable message:
https://cloud.google.com/apis/design/errors#error_model

There is a separate `repeated Any details` for additional details, but I don't know if we have access to that in all programming languages.

bkeryan: status.proto: https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto
C# RpcException provides a `Status` property: https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Core.Api/RpcException.cs#L77
Python provides an `rpc_status.from_call` function: https://github.com/grpc/grpc/blob/master/examples/python/errors/client.py#L36

dmondrik: I am aware of that, and while I saw that on their cloud.google.com site, I didn't see that documentation anywhere else. There are numerous problems with it, however:
- The C++ gRPC::Status object `error_details` says it's another serialized gRPC::Status object, such as an inner exception, with the same restrictions.
- The `error_details` don't appear to be propagated to the client exception

Since we're using UNKNOWN, I think we're staying as true to the spirit of the documentation as possible. That field does contain a human-readable message.

source/custom/nidcpower_service.custom.cpp
dmondrik: Good point, I'll make sure to clear the buffer in each case before the second read.

source/custom/nidaqmx_service.custom.cpp
dmondrik: Thanks, I guess as long as it's on the same thread, that will always correctly return the last error for that thread?

source/server/converters.h
jasonmreding: It's confusing because of the naming, but the Status message listed in that link from Google describing a richer error model than what gRPC provides natively is different from this Status object. What is described in that link is a convention based approach to providing richer error reporting than what you get by default with just an error string and one of the 17 predefined error codes. In particular, it defines several message types which can be included as part of the error to provide more detailed information.

While I had difficulty finding documentation about how this is accomplished, I finally pieced together that this information is passed as part of the metadata in the response trailers using the string "grpc-status-details-bin" as the metadata key. Since this is convention based and not part of gRPC proper, it isn't supported by all languages:

https://www.grpc.io/docs/guides/error/

> This richer error model is already supported in the C++, Go, Java, Python, and Ruby libraries, and at least the grpc-web and Node.js libraries have open issues requesting it. 

However, I have no idea what those libraries look like in each of these languages. There is also an open issue to support this for C#. In the meantime, I did find these examples of how it can be done from C#:

https://github.com/chwarr/grpc-dotnet-google-rpc-status
https://github.com/nano-byte/grpc-rich-error

Whether it would be better for NI to adopt this convention in some NI standard way is something I've been mulling over for Measurement SDK. Not all of the message classes defined by Google work well for us, but we could easily publish our own message types that do. Also, C# and LabVIEW don't currently have libraries available that make accessing the detailed error information easy. LabVIEW currently doesn't even support gRPC metadata at all.

For now, maybe the best thing is to concede to encoding driver error message/code into a json string that is returned as the error message reported natively by gRPC. That's probably human readable enough to be useful for most users. We can always decide to return more detailed error information later using the convention described above which may duplicate some of the error information included in the standard error message.

source/custom/nifgen_service.custom.cpp
dmondrik: Lovely. Thanks for being thorough.

source/codegen/templates/service_helpers.mako
```diff
@@ -719,30 +713,41 @@ ${initialize_hardcoded_parameter(parameter)}
...
 <%
-  get_last_error_output_name = common_helpers.get_grpc_field_name(get_last_error_output)
-  get_last_error_method_name = get_last_error_output["get_last_error"]
+  session = 0
+  handle_type = resource_handle_types[0]
+  for parameter in input_parameters:
+    if parameter['type'] in resource_handle_types:
+      session = common_helpers.get_cpp_local_name(parameter)
+      handle_type = parameter['type']
+      break
```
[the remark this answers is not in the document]
dmondrik: I actually don't think that's desirable. Some methods have no session handle parameters. And in fact this would incorrectly assert for all Init methods, because this is looking only at the `in` parameters. In that case we just pick something and that should be good enough.

CMakeLists.txt
```diff
@@ -358,6 +358,7 @@ add_executable(IntegrationTestsRunner
     "${service_output_dir}/nifake/nifake_service.cpp"
     "${service_output_dir}/nifake_non_ivi/nifake_non_ivi_client.cpp"
     "${service_output_dir}/nifake_non_ivi/nifake_non_ivi_service.cpp"
+    "${custom_dir}/nifake_service.custom.cpp"
     "${custom_dir}/nifake_non_ivi_service.custom.cpp"
```
SparkingSpork, line 362: Your approach I think results in a smaller diff, but we generally try to keep the files in alphabetical order here. 
```suggestion
    "${custom_dir}/nifake_non_ivi_service.custom.cpp"
    "${custom_dir}/nifake_service.custom.cpp"
```

source/server/converters.h
dmondrik: That sounds good. I'll add to the comment. It won't happen in this PR because I have a separate task for it and I'll want to take time to add good test cases for it.

SparkingSpork (approved):
Approving based on CMakeLists changes and deferring to others on the code generation/API changes.

source/server/converters.h
dmondrik: David I will increment by 2. While your suggestion to use `insert` is tantalizing, that takes an iterator instead of an index. :(

astarche: `grpc-device` already uses `nlohmann::json` to parse the config file.

source/server/converters.h
bkeryan: Thanks for the correction @jasonmreding. I mixed up the hand-written `Status` and the code-generated `Status`.

In Python, on the client, you call `rpc_status.from_call(call)` to unpack the call's `grpc-status-details-bin` response trailer into a `status_pb2.Status` (the codegen from `status.proto`). It also validates that the error code and message match.
https://github.com/grpc/grpc/blob/master/src/python/grpcio_status/grpc_status/rpc_status.py#L33

> For now, maybe the best thing is to concede to encoding driver error message/code into a json string that is returned as the error message reported natively by gRPC. That's probably human readable enough to be useful for most users.

JSON allows adding fields without breaking compatibility, but switching from JSON to "richer error model" would be a breaking change.

Other ways to represent the error code without JSON:
- "Error -50150: The software has entered an unknown state..." (English-specific)
- "-50150: The software has entered an unknown state..." (language-agnostic, but more cryptic)

jasonmreding: > JSON allows adding fields without breaking compatibility, but switching from JSON to "richer error model" would be a breaking change.

When I said switch I meant also use the richer error model. When using the richer error model, you still need to report error information through the standard gRPC mechanism. Presumably we would continue to report errors through the standard mechanism as is done in this PR (not break it). I'm not actually sure how the standard mechanism encodes that data over http. I'm guessing it's either part of the response body or trailing headers. If it's headers, I'm not sure what the metadata key is that is used.

> Other ways to represent the error code without JSON:

My concern with that format, is it makes it very difficult to trap/handle specific errors programmatically. With JSON, you can at least get at the error code in a standard way even though it's more work.

jasonmreding: Looks like standard error info is sent as trailing headers using "grpc-status" and "grpc-message".

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses

dmondrik: Note that the Python example you linked to is experimental, which means it's subject to change. And the only example I remember seeing was getting the metadata by expanding both parameters at the call site, which meant it wouldn't throw like we want it to. Or at least that's what it looked like to me.

Jason also the C# examples that do have richer fields are only defined for certain error types. There's no obvious way to extend it for others such as UNKNOWN.

source/server/converters.h
DavidCurtiss: "I will increment by 2" -- careful there. You shouldn't have a "+= 2" anywhere here. You just need an extra ++index after doing the `replace`, and when that's combined with the for loop's ++index it'll add up to 2.

dmondrik: Meaning, I'll change `++index` in the for loop to `index += 2` because it only ever loops around when it does a replace.

source/custom/nidcpower_service.custom.cpp
dmondrik: Added `static_assert` wherever the function signature indicated that the buffer length was assumed, such as `char message[256]`.

source/server/converters.h
bkeryan: BTW, I don't expect this to be addressed in this PR.

@astarche @reckenro Is the string encoding already on your radar or should I file a bug/issue?

source/codegen/templates/service.h.mako
```diff
@@ -95,6 +96,9 @@ private:
 % for resource_handle_type in resource_repository_deps:
   ${resource_repository_deps[resource_handle_type].resource_repository_alias} ${resource_repository_deps[resource_handle_type].field_name};
 % endfor
+% for resource_handle_type in resource_handle_types:
+  ::grpc::Status ConvertApiErrorStatusFor${resource_handle_type}(google::protobuf::int32 status, ${resource_handle_type} ${config["session_handle_parameter_name"]});
```
reckenro, line 100: Nit: Personally, the method name looks a little weird for RFmx to me: `ConvertApiErrorStatusForniRFmxInstrHandle`. I'm fine with it as is but consider capitalizing the `resource_handle_type` with something like `resource_handle_type[0].upper() + resource_handle_type[1:]`.

source/custom/niscope_service.custom.cpp
```diff
@@ -564,4 +564,18 @@ ::grpc::Status NiScopeService::ReadMeasurement(::grpc::ServerContext* context, c
   }
 }
 
+::grpc::Status NiScopeService::ConvertApiErrorStatusForViSession(google::protobuf::int32 status, ViSession vi)
```
reckenro, line 567: I'm marking it here because I know scope has a lot of custom service handlers. But we'll need to go through and update all of them across the _service.custom.cpp files so their status checks call the new ConvertApiErrorStatusForViSession.

This can be done in this change or in a separate one.

reckenro:
Looks like there's four integration tests expecting errors that'll need to be updated to fit the new expectations:

`NiFakeNonIviServiceTests.InitWithReturnedSessionFailsInit_AccessSession_SessionIsNotInMapAndStatusValueIsReturned`
`NiFakeNonIviServiceTests.GetStringAsReturnedValueReturnsNull_ReturnsErrorAndEmptryString`
`NiFakeNonIviServiceTests.InitWithError_CallsGetLatestErrorAndReturnsMessage`
`NiFakeServiceTests.NiFakeService_FunctionCallErrors_ResponseValuesNotSet`

I'm expecting a good bit of System tests to start failing too that expect and check errors (run after a successful build_desktop action is run).

source/server/converters.h
jasonmreding: > Jason also the C# examples that do have richer fields are only defined for certain error types. There's no obvious way to extend it for others such as UNKNOWN.

I'm not sure what examples you are referring to. I'm not aware of any conventions that say this extended error model should only be used with certain grpc status codes. To my knowledge, it would still be up to us to decide on a convention for which grpc status code to return in a given situation. My thought is that most things would get mapped to UNKNOWN. I'm not sure if there is value in sometimes reporting something like INVALID_ARGUMENT vs. just always reporting UNKNOWN. The client receiving the error ultimately has to do something:

``` CSHARP
if (errorDetails is MessageType1) { /* display custom error info for message1 */}
else if (errorDetails is MessageType2) {/* display custom error info for message2 */}
else {/* generic error display fallback */}
```
so maybe the status codes could help minimize the number of checks for different types of error info? I think the more likely answer is we'll have to do better about documenting which types of extended error information can be returned from a given rpc, basically the equivalent of documenting which types of exceptions can be thrown from a function.

source/server/converters.h
reckenro: Good catch. I don't remember it being brought up before or tracked anywhere so I'd say filing a bug/issue sounds appropriate.

source/server/converters.h
bkeryan, line 333: Does this produce an invalid JSON object? I don't see an object key for `description`

```
{ "code":"-50150", "An unexpected error occurred..." }
```

source/server/converters.h
bkeryan: > `grpc-device` already uses `nlohmann::json` to parse the config file.

Great, using that library to produce this JSON string is literally a one-liner.

https://godbolt.org/z/vT1hajGnj
```
#include <iostream>
#include <nlohmann/json.hpp>

std::string createStatusJson(int code, std::string message)
{
    return to_string(nlohmann::json{{"code", code}, {"message", message}});
}
```

source/server/converters.h
bkeryan: > When I said switch I meant also use the richer error model. When using the richer error model, you still need to report error information through the standard gRPC mechanism. Presumably we would continue to report errors through the standard mechanism as is done in this PR (not break it).

Okay, that makes sense.

> I'm not sure what examples you are referring to

I linked to a Python example that uses an experimental API.

> I'm not aware of any conventions that say this extended error model should only be used with certain grpc status codes. 

I'm not aware of any either. 

One of the error detail messages that Google provides is `DebugInfo`, which adds a server-side stack trace. This seems applicable to all errors.

> And the only example I remember seeing was getting the metadata by expanding both parameters at the call site, which meant it wouldn't throw like we want it to. Or at least that's what it looked like to me.

gRPC attaches the response trailer metadata to the exception. I'm not sure how this works in Python, but it's pretty clear in C#.

There is a 3rd party C# library implementing the richer error model: https://github.com/nano-byte/grpc-rich-error

> The client receiving the error ultimately has to do something:
> 
> ```
> if (errorDetails is MessageType1) { /* display custom error info for message1 */}
> else if (errorDetails is MessageType2) {/* display custom error info for message2 */}
> else {/* generic error display fallback */}
> ```

`details` is a `repeated Any`, so you can include multiple detail objects.

To extend the example from nano-byte's C# library:
```
try
{
    // ...
}
catch (RpcException ex)
{
    var info = ex.GetDetail<ErrorInfo>();
    if (info != null)
    {
        // ...
    }
    var debugInfo = ex.GetDetail<DebugInfo>();
    if (debugInfo != null)
    {
        // print stack trace...
    }
}
```

bkeryan, line 333: Representing the error code as a number rather than a string would make it easier to parse. Error codes should not be more than 32 bits, so there is no problem with JSON implementations that use floating point internally limiting to 2^53.

source/server/converters.h
bkeryan: https://github.com/ni/grpc-device/issues/693

source/server/converters.h
dmondrik: Sigh. `quality(spec) > quality(implementation)` 😬

dmondrik: Thanks, With keys requiring quotes I went overboard.

astarche (approved):

source/codegen/templates/service.h.mako
dmondrik: Good suggestion

source/custom/niscope_service.custom.cpp
dmondrik: Done in this PR. @reckenro if you don't mind, please take another look at those.

source/custom/nidcpower_service.custom.cpp
```diff
@@ -43,16 +44,33 @@ ::grpc::Status NiDCPowerService::MeasureMultiple(::grpc::ServerContext* context,
       ViReal64* current_measurements = response->mutable_current_measurements()->mutable_data();
 
       auto status = library_->MeasureMultiple(vi, channel_name, voltage_measurements, current_measurements);
+      if (status < VI_SUCCESS) {
```
reckenro, line 47: Can you do this check instead of the CheckStatus on line 40? The CheckStatus will throw a `DriverErrorException` on warnings which will now cause the UNKNOWN gRPC status.

Or you could also change CheckStatus to do this check instead of `status != 0` and then wrap both library_ calls in it.

dmondrik: I definitely don't want a warning to return UNKNOWN. I need to evaluate each of these more closely.

source/custom/nisync_service.custom.cpp
```diff
@@ -26,14 +26,23 @@ ::grpc::Status NiSyncService::ReadMultipleTriggerTimeStamp(::grpc::ServerContext
     ViUInt32 timestamps_read {};
     auto status = library_->ReadMultipleTriggerTimeStamp(vi, terminal, timestamps_to_read, timeout, time_seconds_buffer, time_nanoseconds_buffer, time_fractional_nanoseconds_buffer, detected_edge_buffer, &timestamps_read);
     response->set_status(status);
-    if (status == VI_SUCCESS || status == NISYNC_ERROR_DRIVER_TIMEOUT) {
-      response->set_timestamps_read(timestamps_read);
+    if (status < VI_SUCCESS) {
```
reckenro, line 29: Hmm, this one's weird. It seems in https://github.com/ni/grpc-device/pull/209 this was added to still set the timestamps_read field even when the error is `NISYNC_ERROR_DRIVER_TIMEOUT`. I'm wondering if we should keep that behavior by checking `status < VI_SUCCESS && status != NISYNC_ERROR_DRIVER_TIMEOUT`.

astarche (approved):

source/custom/nixnet_service.custom.cpp
```diff
@@ -1083,12 +1121,13 @@ ::grpc::Status NiXnetService::DbGetDatabaseList(::grpc::ServerContext* context,
     u32 number_of_databases{};
 
     status = library_->DbGetDatabaseList(ip_address, size_of_alias_buffer, const_cast<char*>(alias_buffer.c_str()), size_of_file_path_buffer, const_cast<char*>(file_path_buffer.c_str()), &number_of_databases);
-    response->set_status(status);
-    if (status_ok(status)) {
-      response->set_alias_buffer(alias_buffer.c_str());
-      response->set_filepath_buffer(file_path_buffer.c_str());
-      response->set_number_of_databases(number_of_databases);
+    if (!status_ok(status)) {
```
reckenro, line 1124: Update the status check on line 1114 for `library_->DbGetDatabaseListSizes`.

source/custom/nisync_service.custom.cpp
bkeryan: FYI, DAQmx read functions can also return (partial) samples along with a timeout error:

https://www.ni.com/docs/en-US/bundle/ni-daqmx-c-api-ref/page/daqmxcfunc/daqmxreadanalogf64.html

> A value of 0 indicates to try once to read the requested samples. If all the requested samples are read, the function is successful. Otherwise, the function returns a timeout error and returns the samples that were actually read.

source/server/converters.h
```diff
@@ -317,6 +318,21 @@ typename TypeToStorageType<TDriverType, TGrpcType>::StorageType allocate_output_
 }
 
 }  // namespace converters
+
+const int kMaxGrpcErrorDescriptionSize = 4096;
```
bkeryan, line 322: I wasn't aware of this before, but https://cloud.google.com/apis/design/errors#error_model has a specific guideline about error size:

> Note: When using gRPC, errors are included in the headers, and total headers in responses are limited to 8 KB (8,192 bytes). Ensure that errors do not exceed 1-2 KB in size.

source/custom/nidaqmx_service.custom.cpp
bkeryan: Yes, `DAQmxGetExtendedErrorInfo` returns the last error for the current thread.

DavidCurtiss (approved):

dmondrik:
> Looks like there's four integration tests (run as part of build_desktop part of CI) expecting errors that'll need to be updated to fit the new expectations:
> 
> `NiFakeNonIviServiceTests.InitWithReturnedSessionFailsInit_AccessSession_SessionIsNotInMapAndStatusValueIsReturned` `NiFakeNonIviServiceTests.GetStringAsReturnedValueReturnsNull_ReturnsErrorAndEmptryString` `NiFakeNonIviServiceTests.InitWithError_CallsGetLatestErrorAndReturnsMessage` `NiFakeServiceTests.NiFakeService_FunctionCallErrors_ResponseValuesNotSet`
> 
> I'm expecting a good bit of System tests to start failing too that expect and check errors (run after a successful build_desktop action is run).

I'll be waiting for that now that I've fixed up the unit tests.

source/custom/niscope_service.custom.cpp
```diff
@@ -45,18 +46,18 @@ ::grpc::Status NiScopeService::Fetch(::grpc::ServerContext* context, const Fetch
     ViReal64* waveform = response->mutable_waveform()->mutable_data();
     std::vector<niScope_wfmInfo> waveform_info(num_waveforms, niScope_wfmInfo());
     auto status = library_->Fetch(vi, channel_list, timeout, num_samples, waveform, waveform_info.data());
-    response->set_status(status);
-    if (status == 0) {
-      convert_to_grpc(waveform_info, response->mutable_wfm_info());
+    if (status < VI_SUCCESS) {
```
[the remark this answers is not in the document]
dmondrik: @reckenro I think I got it fixed up correctly now.

source/custom/nisync_service.custom.cpp
dmondrik: That's going to be fun.

source/custom/nixnet_service.custom.cpp
dmondrik: Argh!!!

source/server/converters.h
dmondrik: What are the situations in which it will vary? I tested local servers for both C# and Python and 4K came through just fine. I know NI has error messages longer than 1KB but do we have any longer than 2KB?

source/custom/nisync_service.custom.cpp
astarche: I guess that's probably a bug.

Would it work to incorporate that into `status_ok`? That's implemented per-driver to decide what status codes are errors vs warnings (where "warnings" we copy outputs but "errors" we don't).

So basically the question is "is it valid to treat a timeout error as a warning throughout the daq/sync API?"

Or are there other entrypoints (potentially) where timeouts need to be handled as errors? If so, we could add an optional `errors_to_handle_as_warnings: List[int]` to function metadata.

dmondrik: @astarche that (adding the check to `status_ok`) sounds right to me except that it's usually an inline method in the generated file instead of being a member method that I could implement in the custom file. In the case of this driver (nisync) the generated check does not cover this same error code. How should we pull that off?

astarche: First: for the sync method in this diff. This is `custom` code, so I think it makes sense to take @reckenro 's suggestion: restore the original condition and avoid a regression here.

I'm thinking more of the issue with DAQ that Brad pointed out and maybe a general fix.

The `status_ok` methods codegen from an expression in config.py like [this one for DAQ](https://github.com/ni/grpc-device/blob/cca4d8a11a1cbec001c6267137c3e5e2b2d0e9ec/source/codegen/metadata/nidaqmx/config.py#L135). So, if the expression can be changed to something like `'status_ok': 'status >= 0 || status == DAQmx_Timeout', then that's an easy fix.

Otherwise, we'd need to change [the codegen for the callers of `status_ok` ](https://github.com/ni/grpc-device/blob/cca4d8a11a1cbec001c6267137c3e5e2b2d0e9ec/source/codegen/templates/service_helpers.mako#L730) so they can conditionally check additional error codes based on some new field in the `function_data`.

source/server/converters.h
bkeryan: I suspect that this guideline errs on the side of caution, in order to leave room for other response trailers. Also, I think the "richer error model" sends the status code and message twice: once as separate `grpc-status` and `grpc-message` response trailers, and again as fields in the `google.rpc.Status` message that is sent in the `grpc-status-details-bin` response trailer.

> I know NI has error messages longer than 1KB but do we have any longer than 2KB?

Static error messages: I don't know.

Dynamic error messages:
- RF/MI: as we discussed previously, these drivers may or may not truncate the dynamic part of the error messages to 1024 bytes, depending on which implementation of `GetError` is called. 
- DAQmx: I don't think the DAQmx C API has a hard limit, but nidaqmx-python truncates error messages to 2 KB: https://github.com/ni/nidaqmx-python/blob/master/nidaqmx/errors.py#L182

source/custom/nisync_service.custom.cpp
bkeryan: > So basically the question is "is it valid to treat a timeout error as a warning throughout the daq/sync API?"

No, I don't think this is valid. 

Not all timeout errors correspond to a user-specified timeout parameter. Some timeout errors indicate that a hardware operation didn't complete within a fixed time. For example, when writing to an EEPROM or flash ROM, the driver may poll the status register to determine when the write cycle is complete. It would be inappropriate to hide timeout errors from `DAQmxSelfCal` or `DAQmxCloseExtCal`.

If the driver has separate error codes for user-specified timeouts and hardware polling timeouts, you might be able to get by with it, but I don't think this is always the case.

Let's look at DAQmx's timeout error codes:
- DAQmxErrorSamplesNotYetAvailable: used for DAQmx read; can return acquired samples
- DAQmxErrorCannotReadRelativeToRefTrigUntilDone; used for DAQmx read; can't return acquired samples 
- DAQmxErrorSamplesCanNotYetBeWritten: used for DAQmx write
- DAQmxErrorTimeoutExceeded: used for hardware polling and DAQmxAutoConfigureCDAQSyncConnections, which has a user-specified timeout

> Or are there other entrypoints (potentially) where timeouts need to be handled as errors? If so, we could add an optional errors_to_handle_as_warnings: List[int] to function metadata.

I think this is less likely to have unintended consequences.

> I'm thinking more of the issue with DAQ that Brad pointed out and maybe a general fix.
> 
>The status_ok methods codegen from an expression in config.py like [this one for DAQ](https://github.com/ni/grpc-device/blob/cca4d8a11a1cbec001c6267137c3e5e2b2d0e9ec/source/codegen/metadata/nidaqmx/config.py#L135). So, if the expression can be changed to something like 'status_ok': 'status >= 0 || status == DAQmx_Timeout, then that's an easy fix.

I think fixing this for DAQmx requires more thought and should be handled separately.

Do you even need to fix this for DAQmx?  Do the other DAQmx APIs that use exceptions (.NET and Python) return samples in timeout cases?

Do they return any other outputs for read/write errors? nidaqmx-python adds the number of samples read/written to the exception object, and this is not specific to timeout errors. 

If it's important to support these use cases, perhaps it would be better to allow clients to opt out of exception-throwing for specific calls.

astarche: > Do you even need to fix this for DAQmx? For comparison, do the DAQmx APIs that use exceptions (.NET and Python) return samples in timeout cases?

Good question. It looks to me like the python API [just throws for any error from read](https://github.com/ni/nidaqmx-python/blob/1f5586385d4da2cfa922834adc77a1f768aff259/nidaqmx/_task_modules/read_functions.py#L31). So that's a good argument for "do nothing".

dmondrik: I fixed the one issue Ryan brought up above and will leave well enough alone for now for other cases. I'm going to mark this issue resolved.

bkeryan (approved):

source/custom/nirfmxinstr_restricted_service.custom.cpp
```diff
@@ -1,2 +1,12 @@
+#include <nirfmxinstr_restricted/nirfmxinstr_restricted_service.h>
+
 namespace nirfmxinstr_restricted_grpc {
-}  // namespace nirfmxinstr_restricted_grpc
\ No newline at end of file
+
+::grpc::Status NiRFmxInstrRestrictedService::ConvertApiErrorStatusForNiRFmxInstrHandle(google::protobuf::int32 status, niRFmxInstrHandle instrumentHandle)
+{
+    std::string description(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
+    // TODO: How do we get access to a library that can return a useful string?
```
bkeryan, line 8: `nirfmxinstr_library` and `nirfmxinstr_restricted_library` are C++ wrappers for different subsets of the same DLL. Perhaps you could add `GetError` to `nirfmxinstr_restricted`'s metadata and set `'codegen_method': 'private'` to keep it out of `nirfmxinstr_restricted.proto`.

dmondrik: Since typing that, I had thought of pretty much the same thing. After this PR gets submitted, though. I've already had to fetch/merge 3 times and it's getting to be not-fun.

dmondrik: [Task 2107350](https://ni.visualstudio.com/DevCentral/_workitems/edit/2107350): Ensure RF restricted APIs have access to "error message" function.

PR #695, opened 2022-08-08
Add RFmxInstr Restricted service for internal functions

achanyou:

Adds a new nirfmxsinstr_restricted service to support calling internal driver functions from internal products like Instrument Studio.

The python files for this service were manually created.

Required for committed feature to support grpc with RFmx SpecAn clients like Instrument Studio.

Added some automated system tests. As stated in the original work item, many of these functions only work in conjunction with many other calls into the driver. More in-depth testing will be done later as the project progresses.


source/tests/system/nirfmxinstr_restricted_driver_api_tests.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+TEST_F(NiRFmxInstrRestrictedDriverApiTests, GetAttributeAuthor)
+{
+  const auto session = init_session(stub(), PXI_5663E);
+  EXPECT_SUCCESS(session, client::set_attribute_i32(stub(), session, "", NiRFmxInstrAttribute::NIRFMXINSTR_ATTRIBUTE_TUNING_SPEED, 2));
+  initiate_to_enter_committed_state(session);
+  
+  auto get_attribute_author_response = restricted_client::get_attribute_author(restricted_stub(), session, "", NIRFMXINSTR_ATTRIBUTE_TUNING_SPEED);
+  EXPECT_RESPONSE_SUCCESS(get_attribute_author_response);
+  printf("\nattribute author %d\n", get_attribute_author_response.attr_val());
```
astarche, line 153: Leftover `printf`.

generated/nirfmxinstr_restricted/nirfmxinstr_restricted.proto
```diff
@@ -0,0 +1,440 @@
...
+  string info_json = 2;
+}
+
+message GetRFmxVersionRequest {
+  nidevice_grpc.Session instrument = 1;
+}
+
+message GetRFmxVersionResponse {
+  int32 status = 1;
+  string r_fmx_version = 2;
```
astarche, line 79: These should be `rfmx_version` of perhaps just `version`. I think the easiest way to fix this is using the `grpc_name` field in the `functions` metadata.

source/codegen/service_helpers.py
```diff
@@ -426,6 +426,11 @@ def get_driver_service_readiness(config: dict) -> str:
     return to_cpp_readiness(readiness)
 
 
+def is_restricted_driver_service(config):
+    """Whether the driver service is restricted."""
+    return config.get("restricted_driver_service", False)
```
astarche: I think @madrasm had a different representation of this concept? Using `code-readiness`?

astarche (approved):

SparkingSpork (approved):

generated/nirfmxinstr_restricted/nirfmxinstr_restricted.proto
```diff
@@ -0,0 +1,440 @@
...
+  rpc GetLatestConfigurationSnapshot(GetLatestConfigurationSnapshotRequest) returns (GetLatestConfigurationSnapshotResponse);
+  rpc GetSnapshotState(GetSnapshotStateRequest) returns (GetSnapshotStateResponse);
+  rpc DeleteSnapshot(DeleteSnapshotRequest) returns (DeleteSnapshotResponse);
+  rpc ConvertForPowerUnitsUtility(ConvertForPowerUnitsUtilityRequest) returns (ConvertForPowerUnitsUtilityResponse);
+  rpc UnregisterSpecialClientSnapshotInterest(UnregisterSpecialClientSnapshotInterestRequest) returns (UnregisterSpecialClientSnapshotInterestResponse);
+  rpc GetCalibrationPlaneNames(GetCalibrationPlaneNamesRequest) returns (GetCalibrationPlaneNamesResponse);
+  rpc GetCalibrationPlaneEnabled(GetCalibrationPlaneEnabledRequest) returns (GetCalibrationPlaneEnabledResponse);
+  rpc GetExternalAttenuationTableNames(GetExternalAttenuationTableNamesRequest) returns (GetExternalAttenuationTableNamesResponse);
+  rpc GetActiveTableName(GetActiveTableNameRequest) returns (GetActiveTableNameResponse);
+  rpc GetSParameterExternalAttenuationType(GetSParameterExternalAttenuationTypeRequest) returns (GetSParameterExternalAttenuationTypeResponse);
```
Wedgetable, line 50: Please remove this method - this is from the public header and was added to [nirfmxinstr.proto](https://github.com/ni/grpc-device/blob/main/generated/nirfmxinstr/nirfmxinstr.proto#L71).

source/codegen/service_helpers.py
reckenro: Yeah, can you change the config.py to use `"code_readiness": "RestrictedNextRelease"` like what's done in https://github.com/ni/grpc-device/pull/689. You should also be able to re-use that version of `is_restricted_driver_service(config)` instead.

reckenro (approved):

generated/nirfmxinstr_restricted/nirfmxinstr_restricted.proto
achanyou: Fixed in commit 5763b81.

generated/nirfmxinstr_restricted/nirfmxinstr_restricted.proto
achanyou: Done. Thanks for letting me know!

source/codegen/service_helpers.py
achanyou: Looks like this got fixed in the merge.

PR #698, opened 2022-08-08
Update error strings to fix failing system tests

Wedgetable:

Removes IVI and hex code specification from error strings on 3 tests, and checks for the substring instead:

* `NiFgenSessionTest.InitWithErrorFromDriver_ReturnsUserErrorMessage`
* `NiRFSADriverApiTests.ErrorMessage_ReturnsErrorMessage`
* `NiRFSGDriverApiTests.ErrorMessage_ReturnsErrorMessage`

Fixes 3 system tests that are consistently failing.

Ran tests locally on a machine with drivers - the tests now pass.


DavidCurtiss:
When I automated the system tests, I found that at least some of these IVI errors will report the hex code on one platform but not the other. If you hit the same thing, then you can remove the "IVI:" prefix, and change the check from EXPECT_EQ to `EXPECT_THAT(..., HAS_SUBSTR(...))`

If you do this, then you'll also need to `#include <gmock/gmock.h>` if it's not already included, and add `using namespace ::testing;`

For reference, you can see my changes to the dcpower tests [here](https://github.com/ni/grpc-device/pull/675/files#diff-311bfa03f686d483ffc820012e65425a6dc1f6e2abc4ae0c785b1f4165171ba8).

reckenro (approved):

source/tests/system/nifgen_session_tests.cpp
```diff
@@ -149,7 +152,7 @@ TEST_F(NiFgenSessionTest, InitWithErrorFromDriver_ReturnsUserErrorMessage)
   call_init_with_options(kTestInvalidFgenRsrc, "", "", &initialize_response);
 
   EXPECT_EQ(kInvalidFgenRsrc, initialize_response.status());
-  EXPECT_STREQ(kViErrorFgenResourceNotFoundMessage, initialize_response.error_message().c_str());
+  EXPECT_THAT(initialize_response.error_message().c_str(), HasSubstr(kViErrorFgenResourceNotFoundMessage));
```
dmondrik, line 155: Note that we're both changing code in this area. Second to commit gets to merge it.

source/tests/system/nirfsa_driver_api_tests.cpp
```diff
@@ -599,9 +599,9 @@ TEST_F(NiRFSADriverApiTests, ErrorMessage_ReturnsErrorMessage)
   const auto response = client::error_message(stub(), session, IVI_ATTRIBUTE_NOT_SUPPORTED_ERROR);
 
   EXPECT_SUCCESS(session, response);
-  EXPECT_EQ(
-      "IVI: (Hex 0xBFFA0012) Attribute or property not supported.",
-      response.error_message());
+  EXPECT_THAT(
+      response.error_message(),
+      HasSubstr("IVI: Attribute or property not supported."));
```
astarche, line 604: This still has the `IVI: ` prefix, so it's not a `substr` of the original message. Shouldn't it be the same as the one below?

astarche (approved):

source/tests/system/nifgen_session_tests.cpp
Wedgetable: I'm going to wait for this last update to go through CI checks, but I'll PM you before I merge just in case you're about to submit as well.

PR #699, opened 2022-08-11
RFC: prototype of custom trailing metadata for status info

bkeryan:

This shows what it would look like to return the status code as custom trailing metadata.

It shouldn't, it's just an RFC. If we did this for real, it would make sense to pass the ServerContext into the status helper methods and stop encoding the message as JSON.

Manually tested updated example:
```
(.venv) PS Z:\grpc-device\examples\nidcpower> python .\measure-record.py localhost 31763 PXI1Slot2
Error -1074135024: {"code":-1074135024,"message":"IVI: Invalid value for parameter or property.\n\nProperty: Measure When\nRequested Value: Automatically after Source Complete\nPossible Values: On Demand\n\nStatus Code: -200077"}
```
The new part is the "Error xxxx:" at the beginning.

### DIFF
--- a/examples/nidcpower/measure-record.py
+++ b/examples/nidcpower/measure-record.py
@@ -218,6 +218,9 @@ try:
 
 except grpc.RpcError as rpc_error:
     error_message = rpc_error.details()
+    for key, value in rpc_error.trailing_metadata() or []:
+        if key == "nidevice-status-code":
+            error_message = f"Error {int(value)}: {error_message}"
     if rpc_error.code() == grpc.StatusCode.UNAVAILABLE:
         error_message = f"Failed to connect to server on {SERVER_ADDRESS}:{SERVER_PORT}"
     elif rpc_error.code() == grpc.StatusCode.UNIMPLEMENTED:
--- a/generated/nidaqmx/nidaqmx_service.cpp
+++ b/generated/nidaqmx/nidaqmx_service.cpp
@@ -58,6 +58,7 @@ namespace nidaqmx_grpc {
       auto port_list = request->port_list().c_str();
       auto status = library_->AddCDAQSyncConnection(port_list);
       if (!status_ok(status)) {
+        context->AddTrailingMetadata("nidevice-status-code", std::to_string(status));
         return ConvertApiErrorStatusForTaskHandle(status, 0);
       }
       response->set_status(status);
@@ -81,6 +82,7 @@ namespace nidaqmx_grpc {
       auto channel_names = request->channel_names().c_str();
       auto status = library_->AddGlobalChansToTask(task, channel_names);
       if (!status_ok(status)) {
+        context->AddTrailingMetadata("nidevice-status-code", std::to_string(status));
         return ConvertApiErrorStatusForTaskHandle(status, task);
       }
       response->set_status(status);
@@ -107,6 +109,7 @@ namespace nidaqmx_grpc {
       while (true) {
         auto status = library_->AddNetworkDevice(ip_address, device_name, attempt_reservation, timeout, nullptr, 0);
         if (!status_ok(status)) {
+          context->AddTrailingMetadata("nidevice-status-code", std::to_string(status));
           return ConvertApiErrorStatusForTaskHandle(status, 0);
         }
         uInt32 device_name_out_buffer_size = status;
@@ -121,6 +124,7 @@ namespace nidaqmx_grpc {
           continue;
         }
         if (!status_ok(status)) {
+          context->AddTrailingMetadata("nidevice-status-code", std::to_string(status));
           return ConvertApiErrorStatusForTaskHandle(status, 0);
         }
         response->set_status(status);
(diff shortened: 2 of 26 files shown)


dmondrik:
Since mine has now submitted, can you pull it into yours? That way it'll be easier to spot the difference.

dmondrik:
I can see how this would be cleaner. I can also see how we'd get the metadata in C#. This would allow us to remove a lot of json references that I had added. Is there any reason *not to do this*? @ccifra @astarche

bkeryan:
> I can see how this would be cleaner. I can also see how we'd get the metadata in C#. This would allow us to remove a lot of json references that I had added. Is there any reason not to do this? @ccifra @astarche

@dmondrik One potential issue: `grpc-labview` doesn't allow the client to access trailing metadata yet. However, given the level of support for other `.proto` language features like `oneof`, as well as scheduling factors I won't go into here, that may not be a deal breaker.

ccifra:
What do you consider are the advantages of this approach vs putting the status code / message in the Status object itself?  I could see us either putting device error information in the status error_details or the error_message.

dmondrik:
> What do you consider are the advantages of this approach vs putting the status code / message in the Status object itself?

With or without Brad's change, the message stays in the Status object itself. The question is where the numeric code goes.

bkeryan:
> What do you consider are the advantages of this approach vs putting the status code / message in the Status object itself? I could see us either putting device error information in the status error_details or the error_message.

The advantage of this approach is that it allows returning NI-specific status fields without overloading the gRPC-defined status fields.

There are two ways that a gRPC service can return the `Status` object:
- The standard way is to return the status code and message using trailing metadata, as defined by https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
  - `grpc-status` is the status code as decimal text.
  - `grpc-message` is the error message, percent-encoded.
- The "richer error model" additionally returns a [google.rpc.Status](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto) protobuf as trailing metadata `grpc-status-details-bin`.

The standard way only has two fields, and overloading them is problematic.
- `grpc-status` is supposed to be one of these status codes: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md . Returning an NI-specific status code in this field would be ambiguous.
- `grpc-message` is supposed to be a developer-facing, English error message. @dmondrik recently changed grpc-device to return a JSON string instead. When parsing the JSON string, the client needs to handle the case where `grpc-message` actually contains a developer-facing, English error message and not a JSON string.

The "richer error model" adds `repeated google.protobuf.Any details` which can be used to return additional data. 

grpc-dotnet doesn't have an implementation of the "richer error model", but there is a 3rd party implementation that looks reasonable: https://github.com/nano-byte/grpc-rich-error

gRPC Python has an implementation of the "richer error model", but the APIs are marked "experimental".

bkeryan:
What are the advantages of "richer error model" over "custom trailing metadata"? 

"Richer error model" uses protobuf to encode the additional data, so you don't have to document "this field is an ASCII string" vs "this field is a URL-encoded string" vs. "this field is a binary protobuf with this message type", etc.

Custom trailing metadata has two data types: ASCII strings and binary.

astarche:
I don't see a problem with this proposal. Thanks for taking the time to research and describe it. Comparing `custom-tralling-metadata` to `rich-error-model`: `rich-error-model` is a bit cleaner, but depending on external libraries and/or experimental features for a common workflow (that we're already API-rev-ing) seems like a serious problem to me. `json-in-the-message` is nice and intuitive for someone that "didn't read the help" and "didn't ask to deal with http2 headers", but I appreciate that we don't want to depend on an unconventional try/parse for something like this.

Could also do either/both/all-three (not necessarily all at once) as far as I'm concerned. We could make a specific recommendation, but users could choose an alternative if they evaluate the tradeoffs differently.

ccifra:
given the lack of support for the richer error model in Status I think this is a good approach.  In the future we could always augment the Status with additional information if needed.

dmondrik:
@bkeryan I'll take this and run with it.

dmondrik:
I had sent an email on Friday but I'll copy it here for posterity. Despite our initial enthusiasm for this approach, it doesn't seem feasible to me now. Here are 2 issues with testing. The first may be overcome but the second seems like a deal-breaker.

1.	Unit tests invoke the server directly with a ServerContext. There is no API by which I can inspect it on return to see that it added to the trailing metadata. The implementation of that data is private with explicitly declared friend classes, so I can’t just subclass it like it might be possible if the data had a protected access method. Does this imply the context somehow needs to be mocked so we can expect a call with certain parameters?
2.	The system tests that use the experimental client stuff hides the ClientContext, which is what I’d need access to in order to get the metadata that the server call contained. I see no good way other than ripping it all out.

@astarche replied with: Aw bummer. (2) seems especially bad. That’s the sort of use case we should be making better by using grpc status so that it’s auto-converted to language-specific exception types. This allows the failure to bubble up through abstraction layers without needing to pass context/status in/out. We could catch, enrich, and rethrow a different exception from the generated client code. That’s could be reasonable, but we could to that with status-in-the-response as well.

FYI @ccifra

PR #700, opened 2022-08-11
Add unit tests for ApiErrorAndDescriptionToStatus

dmondrik:

Adds unit tests for `ApiErrorAndDescriptionToStatus` which was added in the previous PR.

Getting the previous PR to a sufficiently passing state required modifying tests to *indirectly* ensure its behavior, but this is a more direct verification.

[Task 2103383](https://ni.visualstudio.com/DevCentral/_workitems/edit/2103383): Ensure description strings are properly JSON encoded.

Built locally and ran the unit tests successfully.

### DIFF
--- a/source/tests/unit/converters_tests.cpp
+++ b/source/tests/unit/converters_tests.cpp
@@ -2,12 +2,14 @@
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
 #include <server/converters.h>
 
 #include <array>
 
 using namespace nidevice_grpc::converters;
 using namespace ::testing;
+using nlohmann::json;
 
 namespace ni {
 namespace tests {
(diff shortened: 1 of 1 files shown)


reckenro (approved):

astarche (approved):

source/tests/unit/converters_tests.cpp
```diff
@@ -170,6 +172,41 @@ TEST(ConvertersTests, ArrayAndRawValue_ConvertBitfieldAsEnumArrayInput_ReturnsOr
...
+  nlohmann::json jsonError{{"custom", 1}};
+
+  auto error_message = nidevice_grpc::ApiErrorAndDescriptionToStatus(0, empty_description, jsonError).error_message();
+
+  EXPECT_THAT(error_message, HasSubstr("\"custom\":1"));
+}
+
+TEST(ConvertersTests, QuotedLocalizedDescription_ApiErrorAndDescriptionToStatus_MessageIsEncoded)
+{
+  const auto TEST_MESSAGE = std::string("\"chaîne 😀 localisée\"");
```
bkeryan, line 200: What encoding are you using in this source file?

MSVC has command line options to specify that the source code is UTF-8: https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170

Hex escapes are more portable: `"\cha\xC3\xAEe \xF0\x9F\x98\x80 localis\xC3\xA9e\""`

dmondrik: Readability is important too. I found that specifying the string prefix `u8` will guarantee it, so I'll add that to my next PR.

bkeryan: Based on https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170 I would not be surprised if MSVC interprets the source as Windows-1252 and then converts the `u8` strings from Windows-1252 to UTF-8. :)

`u8` strings also support the `\uNNNN` escape.

dmondrik: So our C++ standard representative tells me passing a u8 literal to std::string will stop compiling as of C++20, so I guess I can't do that. Oh well.

bkeryan: Interesting: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r0.html

> Code surveys have so far revealed little use of u8 literals. Google and Facebook have both reported less than 1000 occurrences in their code bases, approximately half of which occur in test code. Representatives of both organizations have stated that, given the actual size of their code base, this is approximately equivalent to 0.
>
> Searches on Debian code search found uses in only a few packages and, within those packages, a small number of uses (mostly single digit use counts), most of which occurred in tests.

PR #702, opened 2022-08-12
Sort ni-fake function metadata.

reckenro:

- Sorts ni-fake functions.py methods alphabetically
- Fixes misspelling in `ReadDataWithMultpleIviTwistParamSets`
- Removed ni-fake metadata for `'codegen_method': 'no'` functions

We're working on generating the MI drivers' grpc-device metadata through a hapigen plugin. I've started work on that and am testing it with NI-FAKE. I realized that the output from hapigen is sorted which made it hard to diff the metadata and was causing updates in the generated output.

I'm making these manual updates to ni-fake metadata now so that when we update it from hapigen output the metadata diffs will align and we won't have functions moving around in the generated files.

Unit and Integration tests leveraging NI Fake pass.


source/codegen/metadata/nifake/functions.py
```diff
@@ -101,10 +101,10 @@
         ],
         'returns': 'ViStatus'
     },
-    'BoolArrayOutputFunction': {
+    'BoolArrayInputFunction': {
```
reckenro, line 104: [Note] This diff is hard to read, but I just moved the functions around to have them sorted. The best way to see this is looking at the updates in nifake.proto to see the new ordering of service methods.

I also removed the `'codegen_method': 'no'` methods as they weren't being used in generation and I am removing them in the hapigen output. I figured if we want to add them back we'll add them through the hapigen metadata / plugin. I don't think having any of them in grpc-device metadata is useful. It also will help aligning the functions.py metadata diffs when we update ni-fake metadata from hapigen output.

astarche (approved):

PR #703, opened 2022-08-15
Generate some old hand-written methods.

reckenro:

Generate the service handlers for some methods that originally couldn't be handled by the code generator.

Fixes [AB#2106015](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2106015)

While we're updating the source of some of the metadata for MI drivers, we thought it'd make sense to see if there were any methods we could generate now since some time has passed since writing them by hand.

Note: There's still more "CustomCode" methods. The point of this PR was to pick off any low hanging fruit that didn't require changes to the code generator. The code generator might be updated at a later point in time to handle the remaining CustomCode scenarios like unique status checks, calling into other methods to get an array size, etc.

Compared old custom code to new generated code.. Scope and Tclk system tests pass locally.

### DIFF
--- a/source/custom/niscope_service.custom.cpp
+++ b/source/custom/niscope_service.custom.cpp
@@ -546,16 +546,16 @@ void CheckStatus(int status)
 
 ::grpc::Status NiScopeService::ConvertApiErrorStatusForViSession(google::protobuf::int32 status, ViSession vi)
 {
-  ViStatus error_code{};
-  std::string description(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
-  // Try first to get the most recent error with a dynamic message.
-  library_->GetError(vi, &error_code, nidevice_grpc::kMaxGrpcErrorDescriptionSize, &description[0]);
-  if (error_code != status) {
-    // Since another thread has changed the status, fall back to the static message lookup.
-    description.assign(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
-    library_->GetErrorMessage(vi, status, nidevice_grpc::kMaxGrpcErrorDescriptionSize, &description[0]);
-  }
-  return nidevice_grpc::ApiErrorAndDescriptionToStatus(status, description);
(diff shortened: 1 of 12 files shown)


astarche (approved):

source/custom/niscope_service.custom.cpp
```diff
@@ -634,16 +546,16 @@ ::grpc::Status NiScopeService::ReadMeasurement(::grpc::ServerContext* context, c
...
-    std::string description(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
-    // Try first to get the most recent error with a dynamic message.
-    library_->GetError(vi, &error_code, nidevice_grpc::kMaxGrpcErrorDescriptionSize, &description[0]);
-    if (error_code != status) {
-        // Since another thread has changed the status, fall back to the static message lookup.
-        description.assign(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
-        library_->GetErrorMessage(vi, status, nidevice_grpc::kMaxGrpcErrorDescriptionSize, &description[0]);
-    }
-    return nidevice_grpc::ApiErrorAndDescriptionToStatus(status, description);
+  ViStatus error_code{};
```
dmondrik, line 549: Nit: In the generated code there's a space before `{};` so I kept that style in the custom code I wrote. You've removed it here.

dmondrik (changes requested):
It makes sense to keep it simple for the low-hanging fruit.

source/custom/niscope_service.custom.cpp
reckenro: Ah, I have my vs code to auto-format when I save based on our [clang format file](https://github.com/ni/grpc-device/blob/main/.clang-format). I reverted these formatting changes, though. 

We can update the .clang-format file and add something like `SpaceBeforeCpp11BracedList: true` to get the space before braces when auto-formatted. But we can do that some other time.

PR #707, opened 2022-08-19
Mark existing response error message for Init methods as deprecated.

tdunkle:

Updates the `proto_helpers` to automatically mark deprecated any response fields that are `get_last_error` output parameters named "error_message" or "error_num".

Fixes [AB#2103384](https://ni.visualstudio.com/DevCentral/_workitems/edit/2103384/)

We should mark these fields deprecated since the way we report errors has changed (see also: #692 )

grpc-device builds in Debug configuration locally, and the generated proto files mark the intended response fields deprecated.

### DIFF
--- a/source/codegen/proto_helpers.py
+++ b/source/codegen/proto_helpers.py
@@ -135,15 +135,15 @@ def get_message_parameter_definitions(parameters):
                 # type.
                 parameter_type = f'repeated {parameter["enum"]}'
             grpc_field_number = generate_parameter_field_number(parameter, used_indexes)
-            get_last_error = 0
+            is_get_last_error_output_param = False
             if common_helpers.is_get_last_error_output_param(parameter):
-                get_last_error = 1
+                is_get_last_error_output_param = True
             parameter_definitions.append(
                 {
                     "name": parameter_name,
                     "type": parameter_type,
                     "grpc_field_number": grpc_field_number,
-                    "get_last_error": get_last_error,
+                    "is_get_last_error_output_param": is_get_last_error_output_param,
                 }
             )
     return parameter_definitions
--- a/source/codegen/proto_helpers.py
+++ b/source/codegen/proto_helpers.py
@@ -135,13 +135,13 @@ def get_message_parameter_definitions(parameters):
                 # type.
                 parameter_type = f'repeated {parameter["enum"]}'
             grpc_field_number = generate_parameter_field_number(parameter, used_indexes)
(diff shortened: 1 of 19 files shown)


source/codegen/templates/proto_helpers.mako
```diff
@@ -100,7 +100,18 @@ message ${common_helpers.snake_to_pascal(function)}Request {
 %>\
 message ${common_helpers.snake_to_pascal(function)}Response {
 % for parameter in response_parameters:
-  ${parameter["type"]} ${parameter["name"]} = ${parameter["grpc_field_number"]};
+<%
+  parameter_name = parameter["name"]
+%>\
+% if (parameter.get("is_get_last_error_output_param", True)):
+%   if ((parameter_name == "error_message") | (parameter_name == "error_num")):
```
reckenro, line 107: We shouldn't need this check on if the param name is "error_message" or "error_num". I think you just need to modify the first statement to default to False instead if the `is_get_last_error_output_param` isn't found in the parameter dictionary:

So I think it'll produce the same generated output if it's just a single if/else here with `paramater.get("is_get_last_error_output_param", False)`

source/codegen/proto_helpers.py
```diff
@@ -135,11 +135,13 @@ def get_message_parameter_definitions(parameters):
                 # type.
                 parameter_type = f'repeated {parameter["enum"]}'
             grpc_field_number = generate_parameter_field_number(parameter, used_indexes)
+            is_get_last_error_output_param = common_helpers.is_get_last_error_output_param(parameter)
```
tdunkle, line 138: I noticed after posting the PR that our style guidelines don't allow for this line to be longer than 100 characters, so I need to shorten it.

reckenro: Oh yeah. For anything formatting I'd recommend installing Black to automatically reformat it. You can look at https://github.com/ni/grpc-device/wiki/Development-Tips-and-Tricks#black-format-on-save to see how to install and run it from a shell or as a vs-code extension.

source/codegen/templates/proto_helpers.mako
astarche: Yeah, and for the record. `parameter_name in ["error_message", "error_num"]` is preferred in python. And (logical) `or` instead of (bitwise) `|`.

astarche (approved):

source/codegen/proto_helpers.py
tdunkle: I shortened the line by changing the name of the field we write to the parameter definition to `is_get_last_error_output`.

PR #708, opened 2022-08-19
Review feedback for nirfmxinstr_restricted

achanyou:

Implement Curtis' review feedback for the nirfmxinstr_restricted service:

- Alphabetize methods to match existing proto files
- Calculate array_size parameter for LoadConfigurationFromJSON from json_string.

Improves readability and ease-of-use.

The changes build. Existing auto-tests run.


source/codegen/metadata/nirfmxinstr_restricted/functions.py
```diff
@@ -1,86 +1,46 @@
 functions = {
-    'RegisterSpecialClientSnapshotInterest': {
```
achanyou, line 2: Most of the changes in this file are non-functional. I've alphabetized the functions. I have a comment below identifying the one functional change.

source/codegen/metadata/nirfmxinstr_restricted/functions.py
```diff
@@ -957,27 +894,14 @@
...
-                    'value': 'arraySize'
-                },
+                'name': 'filePath',
                 'type': 'char[]'
             }
         ],
         'returns': 'int32'
     },
-    'GetActiveTableName': {
+    'LoadConfigurationsFromJSON': {
```
achanyou, line 903: This is the one functional change in this file, adding a custom implementation for LoadConfigurationsFromJSON to infer the arraySize from the jsonString parameter.

source/codegen/metadata/nirfmxinstr_restricted/functions.py
```diff
@@ -957,27 +894,14 @@
...
+        'codegen_method': 'CustomCode',
```
astarche, line 904: I don't think you need `CustomCode` at the `function` level for this. You use a `custom-code` size mechanism. [Example](https://github.com/ni/grpc-device/blob/d30a1cb0c11b9622e0c4954d3a99d755992a3dde/source/codegen/metadata/nidaqmx/functions.py#L142).

```python

            {
                'direction': 'in',
                'name': 'arraySize',
                'size': {
                    'mechanism': 'custom-code',
                    'value': 'jsonString.length()'
                },
                'type': 'int32'
            }
```

astarche (approved):

reckenro (approved):

source/codegen/metadata/nirfmxinstr_restricted/functions.py
achanyou: I tried it, and it doesn't actually eliminate the arraySize parameter from the grpc request message. I think it is because the 'size' tag doesn't apply to the scalar int32 value.

Let me know if I'm not understanding this the right way around, but I don't think I can add the size tag to the jsonString parameter either to achieve this.

reckenro: I must've glossed over this the first time. I think something like this will get you the generated code you want:

```
'LoadConfigurationsFromJSON': {
        'parameters': [
            {
                'direction': 'in',
                'grpc_name': 'instrument',
                'name': 'instrumentHandle',
                'type': 'niRFmxInstrHandle'
            },
            {
                'direction': 'in',
                'name': 'jsonString',
                'type': 'char[]',
                'size': {
                    'mechanism': 'len',
                    'value': 'arraySize',
                },
            },
            {
                'direction': 'in',
                'name': 'arraySize',
                'type': 'int32'
            }
        ],
        'returns': 'int32'
    },
```

achanyou: Thanks for the example, Ryan! This did the trick.

PR #709, opened 2022-08-23
Updated NI-DCPower example to rely on exceptions

dmondrik:

- Removes checks for errors after each call since they are no longer warranted.
- Adds handling for the UNKNOWN exception code including parsing the fields.

Part of [Task 2103385](https://ni.visualstudio.com/DevCentral/_workitems/edit/2103385): Update examples to handle new errors

Verified that the error handling is correct both for failure to connect (which hasn't changed) as well as invalid resource.
```
>> python measure-record.py localhost 31763 dev1
Device was not recognized. The device is not supported with this driver or version.

Invalid Identifier: dev1
Error status: -1074118656
```


examples/nidcpower/measure-record.py
```diff
@@ -93,7 +77,8 @@ def check_for_initialization_error(response):
         )
     )
     vi = initialize_with_channels_response.vi
-    check_for_initialization_error(initialize_with_channels_response)
+    if initialize_with_channels_response.status > 0:
```
dmondrik, line 80: Note: Since initialization no longer needs to check for errors, just warnings, it felt cleaner to just inline the check. I could keep this as a separate `check_for_warning` helper if desired. I'm not sure how much it might be used elsewhere in the other examples though.

examples/nidcpower/measure-record.py
```diff
@@ -224,9 +201,13 @@ def _on_close(event):
         error_message = (
             "The operation is not implemented or is not supported/enabled in this service"
         )
+    elif rpc_error.code() == grpc.StatusCode.UNKNOWN:
+        # The error details is a JSON string containing both the status code and the description.
```
bkeryan, line 205: Is that guaranteed? Does gRPC itself ever return `StatusCode.UNKNOWN` with a non-JSON message?

bkeryan: > My earlier searching said the server never returned it, but apparently it can, at least in theory:
> [Server side application throws an exception (or does something other than returning a Status code to terminate an RPC)](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

Yeah, it's definitely possible for the client to return `StatusCode.UNKNOWN` when the server returns an unexpected response:
- https://github.com/grpc/grpc/blob/master/src/core/lib/surface/call.cc#L983
- https://github.com/grpc/grpc/blob/master/src/core/lib/transport/status_conversion.cc#L67

Even if `ni_grpc_device_server.exe` never hits any of these cases, a more complicated setup like an HTTP load balancer or proxy server could conceivably hit one of this cases.

Compared to the code you removed and the rest of the error handling code, adding a try/except doesn't seem that bad to me.

```
    elif rpc_error.code() == grpc.StatusCode.UNKNOWN:
        try:
            error_details = json.loads(error_message);
            error_message = f"Failed with status {error_details['code']}\n{error_details['message']}"
        except JSONDecodeError:
            pass
```

> I don't see any other good options. 

Custom trailing metadata or "richer error model" (`google.rpc.Status`) would not have this ambiguity. :)

examples/nidcpower/measure-record.py
bkeryan: Warnings have error messages too. You can use `client.ErrorMessage()` to query what they are.

examples/nidcpower/measure-record.py
```diff
@@ -103,23 +88,20 @@ def check_for_initialization_error(response):
             attribute_value=nidcpower_types.NiDCPowerInt32AttributeValues.NIDCPOWER_INT32_MEASURE_WHEN_VAL_AUTOMATICALLY_AFTER_SOURCE_COMPLETE,
         )
     )
-    check_for_error(vi, configure_measure_when.status)
```
reckenro, line 106: I think we should keep these checks and helper, just change them to check_for_warning. If the status is > 0, we can get the error message and write to `sys.stderr`.

reckenro:
To fix the validate_python step in the build, I'd recommend installing Black to automatically reformat it. You can look at https://github.com/ni/grpc-device/wiki/Development-Tips-and-Tricks#black-format-on-save to see how to install and run it from a shell or as a vs-code extension.

dmondrik:
> To fix the validate_python step in the build, I'd recommend installing Black to automatically reformat it. You can look at https://github.com/ni/grpc-device/wiki/Development-Tips-and-Tricks#black-format-on-save to see how to install and run it from a shell or as a vs-code extension.

I opened it in VS Code and the problem was obvious. The trailing semicolon after the function call exposes my C/C# background.

examples/nidcpower/measure-record.py
dmondrik: Yep, I came to that conclusion eventually. And it's good because in a chat with Brad, he pointed out that the fetch operation can return a warning, so I added a call to it there too.

examples/nidcpower/measure-record.py
```diff
@@ -103,23 +95,20 @@ def check_for_initialization_error(response):
...
 
-    # set the voltage level.
+    # Set the voltage level.
     configure_voltage_level = client.ConfigureVoltageLevel(
         nidcpower_types.ConfigureVoltageLevelRequest(vi=vi, level=VOLTAGE_LEVEL)
     )
-    check_for_error(vi, configure_voltage_level.status)
```
reckenro, line 112: I'm not familiar enough with the API, but should we call the new `check_for_warning` in all the places we had `check_for_error` before? It might be the case that we don't expect warnings from these calls but I'm leaning towards keeping them to notify the user in case they do occur.

dmondrik: It was important to do on each call when that was the only way to discover that an error happened. I think it just litters the examples unnecessarily to check for warnings on each call. I'd prefer to keep it limited to the places where warnings can actually happen. @bkeryan @marcoskirsch @harsha-bhushan thoughts?

danestull (approved):

examples/nidcpower/measure-record.py
bkeryan: > I think it just litters the examples unnecessarily to check for warnings on each call. 

I agree.

> I'd prefer to keep it limited to the places where warnings can actually happen. @bkeryan @marcoskirsch @harsha-bhushan thoughts?

The only NI-DCPower warnings I'm aware of are measurement-related, so I would expect them to be returned from `FetchMultiple` or *maybe* `Initiate`.

tdunkle (approved):

examples/nidcpower/measure-record.py
marcoskirsch: Generally speaking MI APIs don't return warnings other than to indicate things like "an overrange occurred".
In `nimi-python` APIs the warnings get logged. [See here](https://github.com/ni/nimi-python/blob/3e4372d32409594d6968c9a392b3d0078ba7e8bf/generated/nidcpower/nidcpower/errors.py#L112).

bkeryan: > Generally speaking MI APIs don't return warnings other than to indicate things like "an overrange occurred".

@marcoskirsch That's what I meant by "measurement-related". Do you expect these to be returned from any function other than fetch/measure?

marcoskirsch: nope, I was adding a but more info to your correct answer.

dmondrik: I had put it on Initialize knowing that IVI defines 2 warnings for not-supported (ID query and self-test) but it doesn't seem any NI drivers use that. If someone can confirm we don't warn on those anywhere, I'll remove the first warning check.

reckenro (approved):

PR #713, opened 2022-08-30
Make C++ exception be similar to C# RpcException

dmondrik:

The exception that the experimental client currently raises has only the message and nothing else. That makes it hard to glean other information about the error. This change makes it fairly similar to the C# exception, exposing the gRPC status code and trailer metadata as well.

This PR has no functional changes.

I will likely make a follow-on change that returns the gRPC error number in the trailer rather than as JSON.

Built locally and ensured unit tests still pass.

### DIFF
--- a/source/tests/system/nidaqmx_driver_api_tests.cpp
+++ b/source/tests/system/nidaqmx_driver_api_tests.cpp
@@ -675,13 +675,6 @@ class NiDAQmxDriverApiTests : public Test {
     return stub()->CfgWatchdogDOExpirStates(&context, request, &response);
   }
 
-  // void raise_if_error(::grpc::Status status)
-  // {
-  //   if (!status.ok()) {
-  //     throw new std::runtime_error(status.error_message());
-  //   }
-  // }
-
   std::unique_ptr<NiDAQmx::Stub>& stub()
   {
     return nidaqmx_stub_;
--- a/source/tests/system/nidaqmx_driver_api_tests.cpp
+++ b/source/tests/system/nidaqmx_driver_api_tests.cpp
@@ -1,10 +1,10 @@
 #include <gmock/gmock.h>
 #include <google/protobuf/util/time_util.h>
 #include <gtest/gtest.h>  // For EXPECT matchers.
-#include <nlohmann/json.hpp>
 
 #include <algorithm>
 #include <cstring>
+#include <nlohmann/json.hpp>
 #include <random>
 #include <stdexcept>
 #include <vector>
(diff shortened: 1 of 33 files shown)


source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -675,12 +675,12 @@ class NiDAQmxDriverApiTests : public Test {
     return stub()->CfgWatchdogDOExpirStates(&context, request, &response);
   }
 
-  void raise_if_error(::grpc::Status status)
-  {
-    if (!status.ok()) {
-      throw new std::runtime_error(status.error_message());
-    }
-  }
+  // void raise_if_error(::grpc::Status status)
```
reckenro, line 678: Should we just delete this commented out `raise_if_error` function?

dmondrik: Thanks, I knew I forgot something. :)

reckenro (approved):

astarche (approved):

source/tests/utilities/client_helpers.h
```diff
@@ -55,10 +55,34 @@ class simple_variant {
...
+class grpc_driver_error : public std::runtime_error {
+ private:
+  ::grpc::StatusCode code_;
+  std::multimap<std::string, std::string> trailers_;
+
+ public:
+  grpc_driver_error(const std::string& message, ::grpc::StatusCode code, const std::multimap<grpc::string_ref, grpc::string_ref>& trailers)
+      : std::runtime_error(message), code_(code)
+  {
+    for (auto trailer : trailers) {
```
bkeryan, line 67: FYI: `for (const auto& trailer : trailers) {` would avoid a copy, but it's probably not a big deal due to the use of `string_ref`.

source/tests/utilities/client_helpers.h
```diff
@@ -55,10 +55,34 @@ class simple_variant {
...
+      trailers_.emplace(trailer.first.data(), trailer.second.data());
```
bkeryan, line 68: `string_ref::data()` isn't guaranteed to be NUL-terminated. A `string_ref` may refer to a substring.

`std::string(trailer.first.data(), trailer.first.size())` or `std::string(trailer.first.begin(), trailer.first.end())` would work.

source/tests/utilities/client_helpers.h
```diff
@@ -55,10 +55,34 @@ class simple_variant {
...
+    }
+  }
+  ::grpc::StatusCode StatusCode() const
```
bkeryan, line 71: FYI: C# [RpcException](https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Core.Api/RpcException.cs) also includes the `Status` object, but I guess this object has enough info to reconstruct it.

bkeryan (changes requested):

source/tests/utilities/client_helpers.h
dmondrik: Yeah, and I'm trying to limit unnecessary copy operations. It's something we could always add later if needed.

bkeryan (approved):

PR #714, opened 2022-08-30
Update dmm digital fgen scope Examples

danestull:

Update DMM, Digital, FGEN, and SCOPE examples to new error convention.

This updates the examples so they work appropriately with the new convention and cleans up the examples as well. 

Examples still run as expected, when in an error state such as providing an non existent device the examples report errors correctly. 
Screen Shot 2022-08-30 at 2 49 40 PM


examples/niscope/fetch.py
```diff
@@ -56,83 +57,65 @@
...
-            )
-        ).status,
+    niscope_client.ConfigureHorizontalTiming(
+        niscope_types.ConfigureHorizontalTimingRequest(
+            vi=vi,
+            min_sample_rate=50000000,
+            min_num_pts=samples,
+            ref_position=50.0,
+            num_records=1,
+            enforce_realtime=True,
```
dmondrik, line 104: This is existing stuff, but I'm just now realizing how inconsistent the examples are w.r.t. a trailing comma on the last parameter. @reckenro shouldn't `black` be enforcing this one way or the other?

dmondrik (approved):
These look so much better than they were already!

reckenro (approved):

examples/niscope/fetch.py
bkeryan: FYI, clang-format also uses trailing comma as a hint to put each element on a separate line.

danestull: I've updated these so any that are not all on a single line are now the trailing comma style since that was the most used.

astarche (approved):

PR #715, opened 2022-08-31
Return the NI error code as trailing metadata instead of in the Status

dmondrik:

Return the NI status code as trailing metadata instead of in the Status object.

The advantage of this approach is that it allows returning NI-specific status fields without overloading the gRPC-defined status fields. In particular, we can avoid JSON-encoding the combination of fields. (And since we just added that, we can remove this use of JSON.)

This completes the work initiated by Brad in his [partial RFC implementation](https://github.com/ni/grpc-device/pull/699).

The existing architecture for trailing metadata already supports multiple key-value pairs, which suits our needs quite well.

[Task 2133937](https://ni.visualstudio.com/DevCentral/_workitems/edit/2133937): Move error number from status object to trailing metadata

Built locally. Unit tests pass.


source/server/converters.h
```diff
@@ -322,25 +321,21 @@ typename TypeToStorageType<TDriverType, TGrpcType>::StorageType allocate_output_
...
 }
 
-inline ::grpc::Status ApiErrorAndDescriptionToStatus(int32_t status, std::string& description)
-{
-  nlohmann::json jsonError;
-  return ApiErrorAndDescriptionToStatus(status, description, jsonError);
-}
-
-inline ::grpc::Status ApiErrorToStatus(int32_t status)
+template <typename TServerContext>
```
dmondrik, line 332: Note to reviewers: This is templatized so as to allow unit tests to Mock the context.

source/tests/system/nidcpower_session_tests.cpp
```diff
@@ -151,27 +154,36 @@ TEST_F(NiDCPowerSessionTest, InvalidSession_CloseSession_ReturnsInvalidSessionEr
...
+  catch (const nidevice_grpc::experimental::client::grpc_driver_error& ex) {
+    EXPECT_EQ(::grpc::StatusCode::UNKNOWN, ex.StatusCode());
+    const auto& error = ex.Trailers().find("ni-error")->second;
+    EXPECT_EQ(kInvalidDCPowerSession, std::stoi(error));
+    EXPECT_THAT(ex.what(), HasSubstr(kInvalidDCPowerSessionMessage));
+  }
 }
 
-TEST_F(NiDCPowerSessionTest, InitWithErrorFromDriver_ReturnsUserErrorMessage)
+TEST_F(NiDCPowerSessionTest, InitWithErrorFromDriver_ReturnsDriverErrorWithUserErrorMessage)
```
dmondrik, line 174: Note to reviewers: Several of these classes had 99%-identical tests for the 2 conditions. I consolidated them into a single test to make it easier on me.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
```diff
@@ -1695,7 +1693,7 @@ TEST_F(NiFakeNonIviServiceTests, InitWithReturnedSession_AccessSession_ReturnsIn
   EXPECT_EQ(SESSION_HANDLE, accessed_handle);
 }
 
-TEST_F(NiFakeNonIviServiceTests, InitWithReturnedSessionFailsInit_AccessSession_SessionIsNotInMapAndStatusValueIsReturned)
+TEST_F(NiFakeNonIviServiceTests, InitWithReturnedSessionFailsInit_AccessSession_SessionIsNotInMap)
```
dmondrik, line 1696: Note to reviewers: Some of these unit tests are incapable of knowing the status being returned because 1) the concrete google class for the server context defines its methods as non-virtual which we cannot Mock, and 2) that class does not expose a way to check what has been added. 

I chose to remove that portion of a few unit tests because that functionality is thoroughly tested in the system tests.

source/custom/nixnetsocket_service.custom.cpp
```diff
@@ -2,26 +2,26 @@
...
-    // XNET sockets have special behavior that returns both the API error code and the sockets error number.
-    int errorNumber = library_->GetLastErrorNum();
-    nlohmann::json jsonError{{"errorNumber", errorNumber}};
-    return nidevice_grpc::ApiErrorAndDescriptionToStatus(status, description, jsonError);
+  // This implementation assumes this method is always called on the same thread where the error occurred.
+  std::string description(nidevice_grpc::kMaxGrpcErrorDescriptionSize, '\0');
+  library_->GetLastErrorStr(&description[0], nidevice_grpc::kMaxGrpcErrorDescriptionSize);
+  // XNET sockets have special behavior that returns both the API error code and the sockets error number.
+  int errorNumber = library_->GetLastErrorNum();
+  context->AddTrailingMetadata("socket-error", std::to_string(errorNumber));
```
bkeryan, line 12: Let's prefix our custom metadata with `ni-` to ensure that it never conflicts with [standard HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).

(This is more important for headers than for trailers, but I think it's a good practice to follow in both cases.)

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1073,13 +1079,17 @@ TEST_F(NiDAQmxDriverApiTests, AOVoltageChannel_WriteAODataWithOutOfRangeValue_Re
...
-  auto write_data = generate_random_data(AO_MIN, AO_MAX, 100);
-  write_data[80] += AO_MAX;
-  WriteAnalogF64Response write_response;
-  auto status = write_analog_f64(write_data, write_response);
+  try {
+    auto write_data = generate_random_data(AO_MIN, AO_MAX, 100);
+    write_data[80] += AO_MAX;
+    WriteAnalogF64Response write_response;
+    write_analog_f64(write_data, write_response);
+    EXPECT_FALSE(true);
```
bkeryan, line 1087: There's a more direct way to do this:
https://google.github.io/googletest/reference/assertions.html#success-failure

```
FAIL() << "We shouldn't get here.";
```

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1362,21 +1372,27 @@ TEST_F(NiDAQmxDriverApiTests, CalculateReversePolyCoefficientsWithPositiveRevers
 TEST_F(NiDAQmxDriverApiTests, AIVoltageChannel_WaitForValidTimestamp_ReturnsError)
 {
   create_ai_voltage_chan(0.0, 1.0);
```
bkeryan, line 1374: It looks like these tests are organized as arrange/act/assert. You probably ought to keep the spacing, to the extent possible.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -714,16 +725,11 @@ class NiDAQmxDriverApiTests : public Test {
     EXPECT_EQ(::grpc::Status::OK.error_code(), status.error_code());
   }
 
-  void EXPECT_DAQ_ERROR(int32_t expected_error, const std::string& error_message)
+  void EXPECT_DAQ_ERROR(int32_t expected_error, const nidevice_grpc::experimental::client::grpc_driver_error& ex)
```
bkeryan, line 728: This seems applicable to all of the drivers. Have you considered single-sourcing it in a header file?

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -714,16 +725,11 @@ class NiDAQmxDriverApiTests : public Test {
...
-    auto error = json::parse(error_message);
-    EXPECT_EQ(expected_error, error.value("code", 0));
-  }
-
-  void EXPECT_DAQ_ERROR(int32_t expected_error, const ::grpc::Status& status)
-  {
-    EXPECT_EQ(::grpc::StatusCode::UNKNOWN, status.error_code());
-    EXPECT_DAQ_ERROR(expected_error, status.error_message());
+    EXPECT_EQ(::grpc::StatusCode::UNKNOWN, ex.StatusCode());
+    const auto& error = ex.Trailers().find("ni-error")->second;
```
bkeryan, line 731: I'm not a fan of tests that crash if the SUT doesn't return an error, but I guess these checks are all inside catch blocks, so they would only crash if the SUT returned a non-driver error.

Single-sourcing the error macro would make it less tedious to handle this case.

bkeryan (changes requested):

source/tests/system/nidaqmx_driver_api_tests.cpp
dmondrik: I did the best I could. In some cases especially where there are 2 calls to close with mixed success/failure, it's hard now.

source/tests/system/nidaqmx_driver_api_tests.cpp
dmondrik: Added a new header. That should help tremendously.

bkeryan (approved):
Dan, thank you for taking the time to revisit this idea and follow through on it. I realize this involved more churn than you wanted, but I hope that reducing ambiguity and improving interoperability (with code that displays error messages) will be worthwhile overall.

source/tests/system/nixnetsocket_driver_api_tests.cpp
```diff
@@ -1,6 +1,5 @@
 #include <gmock/gmock.h>
 #include <google/protobuf/util/time_util.h>
-#include <gtest/gtest.h>
```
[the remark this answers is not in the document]
dmondrik: I think having `test_helpers` include it is sufficiently obvious. I'd prefer to not go back and touch all these files yet again.

source/tests/system/rfmx_macros.h
```diff
@@ -48,11 +44,11 @@
 #define EXPECT_ERROR(expected_error_, message_substring_, session_, response_call_) \
```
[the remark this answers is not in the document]
dmondrik: I'd prefer to not make that change because for one, there would be several variants such as validating exact string versus substring. But the main reason I don't want this at this time is that I'd rather make a method that *returns* the caught exception (such as `TestAssert.Throws` in ASW) so we can more clearly separate the 'act' and 'assert' portions of the test.

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
astarche: Aw bummer. I agree with your choice though.

I think we made a mistake in writing the fake tests against the server API instead of the client API. Mostly because these tests are clumsier, especially now that we have the client wrappers. This looks like a good argument for that. For reference, there actually are a handful of fake tests written at a client level in [ni_fake_service_tests_endtoend](https://github.com/ni/grpc-device/blob/19157132294db5fcc44442712ef062a53cfd111b/source/tests/integration/ni_fake_service_tests_endtoend.cpp).

astarche (approved):

reckenro (approved):

source/tests/unit/ni_fake_non_ivi_service_tests.cpp
dmondrik: I'm going to submit these as-is. Should this warrant a github issue being opened for it? Do those issues cover both bugs and tech debt?

PR #716, opened 2022-08-31
Update RFmx Examples Error Handling

danestull:

Update RFmx Examples to new error handling

This updates the examples so they work appropriately with the new convention and cleans up the examples as well.
[Task 2124243](https://ni.visualstudio.com/DevCentral/_workitems/edit/2124243): Update examples (RFmx) to handle new errors

Examples still run as expected, when in an error state such as providing an non existent device the examples report errors correctly. See an example WLAN run:

Screen Shot 2022-08-31 at 2 50 34 PM


examples/nirfmxbluetooth/txp-basic.py
```diff
@@ -70,236 +71,232 @@
...
+    check_for_warning(txp_fetch_lecte_reference_period_powers_response, instr)
     reference_period_average_power_mean = (
         txp_fetch_lecte_reference_period_powers_response.reference_period_average_power_mean
     )
     reference_period_peak_absolute_power_deviation_maximum = (
         txp_fetch_lecte_reference_period_powers_response.reference_period_peak_absolute_power_deviation_maximum
     )
 
-    txp_fetch_lecte_transmit_slot_powers_array_response = raise_if_error(
+    txp_fetch_lecte_transmit_slot_powers_array_response = (
```
dmondrik, line 275: Nit: I know this is existing code, but that is _way too long_ of a variable name for me to sit by idly. If you remove the word `array` from it, it's still obvious what it's referring to. Can we at least get it short enough so the method name isn't split to the next line?

examples/nirfmxwlan/ofdm-modacc.py
```diff
@@ -76,233 +77,232 @@
...
+            instrument=instr,
+            selector_string="",
+            timeout=10.0,
         )
     )
+    check_for_warning(ofdm_mod_acc_fetch_frequency_error_mean_response, instr)
     frequency_error_mean = ofdm_mod_acc_fetch_frequency_error_mean_response.frequency_error_mean
-    ofdm_mod_acc_fetch_symbol_clock_error_mean_response = raise_if_error(
+
+    ofdm_mod_acc_fetch_symbol_clock_error_mean_response = (
```
dmondrik, line 284: Another inexplicably long name. This is the OFDM-MODACC example. Therefore, I suggest removing any `ofdm_mod_acc_` prefix from variable names in this example.

dmondrik (approved):
No need for me to re-review. Whatever you decide on my comments is fine.

examples/nirfmxbluetooth/txp-basic.py
```diff
@@ -70,236 +71,232 @@
...
-    return response
-
-
-def raise_if_error(response):
-    """Raise an exception if an error was returned."""
-    if response.status != 0:
-        error_response = client.GetError(
-            nirfmxbluetooth_types.GetErrorRequest(
-                instrument=instr,
+        warning_message = client.GetErrorString(
```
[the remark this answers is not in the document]
danestull: So the RFmx protos have `GetError` and `GetErrorString` where `GetError` takes in the instr and returns the last error info, meanwhile the `GetErrorString` request takes in the instr and the error code and returns the description. So for RFmx `GetErrorString` is the equivalent of `ErrorMessage` used in the other drivers like DC Power.

examples/nirfmxbluetooth/txp-basic.py
danestull: removed array

examples/nirfmxwlan/ofdm-modacc.py
danestull: removed these prefixes

examples/nirfmxbluetooth/acp-basic.py
```diff
@@ -322,18 +321,15 @@ def check_for_warning(response, instrument):
             print(f"Upper Margin (dB)                    : {upper_margin[i]}\n")
 except grpc.RpcError as rpc_error:
     error_message = rpc_error.details()
+    for key, value in rpc_error.trailing_metadata() or []:
```
dmondrik, line 324: The build process runs type checking on a few examples (@reckenro why not all?) which necessitated this change that you should also use:
```
for metadatum in rpc_error.trailing_metadata() or []:
    if metadatum.key == "ni-error":
        error_message += f"\nError status: {metadatum.value.decode('utf-8')}"
```

danestull: Got it, updated

reckenro: Yeah, I honestly forget the historical reason. If you look at the [validate_python.yml](https://github.com/ni/grpc-device/blob/main/.github/workflows/validate_python.yml) you can see it does the styleguide lint check on all examples. But it only runs validate_examples.py (which does the Mypy type checks among other things) on rfmx and xnet example folders.

The reason might've been issues with betterproto examples causing issues with mypy and/or us just focused on rfmx and xnet at the time and weren't expecting to change the others anytime soon so we didn't bother with them. I can look into it and see about making a Task/Bug/Tech Debt to sort it out (probably after these example updates go in).

astarche: The reason that it doesn't run on all examples is because the examples from before `validate_examples` was added weren't `mypy` compliant (most commonly reusing `response` variables) or had pypi deps beyond what the `validate_examples` script adds (including-but-not-limited-to `betterproto`). We just didn't go back and fix them, since it's most useful when examples are under development and changing to examples is kind of fussy and requires testing  against a server with simulated HW.

I think it'd be great to get all of the `validate_examples` running, especially since we're having to touch all of the examples for this release anyway. That's also helpful for running the examples and even testing some error cases (invalid resource). Currently that's only a convenience for manual testing, though potentially that could run on the HW CI (@DavidCurtiss ).

PR #717, opened 2022-09-01
Update switch/tclk/xnet/session examples to display error code

dmondrik:

We have changed how the error code gets back to the user. The examples should show the best practice of handling it.

In particular, as of [this PR which just went in](https://github.com/ni/grpc-device/pull/715), the status code will be part of the trailing metadata.

[Task 2124241](https://ni.visualstudio.com/DevCentral/_workitems/edit/2124241): Update examples (niswitch, nitclk, nixnet, session) to handle new errors

Built locally.


danestull (approved):

examples/nidcpower/measure-record.py
```diff
@@ -204,18 +203,15 @@ def _on_close(event):
 
 except grpc.RpcError as rpc_error:
     error_message = rpc_error.details()
+    for key, value in rpc_error.trailing_metadata() or []:
+        if key == "ni-error":
+            error_message += f"\nError status: {value}"
```
danestull, line 208: Does this value include the error description or message or are we just printing out the code now?

examples/nidcpower/measure-record.py
```diff
@@ -204,18 +203,15 @@ def _on_close(event):
 
 except grpc.RpcError as rpc_error:
     error_message = rpc_error.details()
+    for metadatum in rpc_error.trailing_metadata() or []:
+        if metadatum.key == "ni-error":
+            error_message += f"\nError status: {metadatum.value.decode('utf-8')}"
```
bkeryan, line 208: Did you retest the examples after adding `decode`? 

Based on the `grpcio` Cython source, I would expect `value` to be a `str` unless `key` ends in "-bin".
https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_cython/_cygrpc/metadata.pyx.pxi#L61

reckenro (approved):

examples/nidcpower/measure-record.py
dmondrik: OK the code I came up with in the end behaves correctly, passes mypy, and is formatted in a sane readable fashion:
```
    for entry in rpc_error.trailing_metadata() or []:
        if entry.key == "ni-error":
            value = entry.value if isinstance(entry.value, str) else entry.value.decode("utf-8")
            error_message += f"\nError status: {value}"
```
FYI @DavidCurtiss @reckenro @danestull @tdunkle

examples/nitclk/multi-device-configured-acquisition-tclk.py
```diff
@@ -80,23 +80,27 @@
...
+        )
+        sys.stderr.write(f"{warning_message.error_string}\nWarning status: {response.status}\n")
 
 
 try:
-    # Initialize the session
-    i = 1
-    for resource in RESOURCES:
+    # Initialize the sessions
+    for i, resource in enumerate(RESOURCES, start=1):
```
astarche, line 103: This is as good a place as any to comment: these examples look a lot better. Thanks!

astarche (approved):

examples/nidcpower/measure-record.py
bkeryan: I think `isinstance` is the key: it tells mypy that this code can handle `str`.

Apparently `assert isinstance(entry.value, str)` or `cast(str, entry.value)` will also work:
https://mypy.readthedocs.io/en/stable/type_narrowing.html

PR #718, opened 2022-09-01
Update nidaqmx example error handling.

tdunkle:

Updates the NI-DAQmx examples to use the new exception-handling approach first implemented in #709 (Updated NI-DCPower example to rely on exceptions).

Fixes [Task 2124239: Update examples (DAQmx) to handle new errors](https://ni.visualstudio.com/DevCentral/_workitems/edit/2124239/).

Builds locally. Running one of the examples on a VM produces this output:

```
PS C:\dev\grpc-device\examples\nidaqmx> python analog-input.py localhost 31763 dev1
Device identifier is invalid.
Device Specified: empty string

Task Name: my task

Status Code: -200220
Error status: -200220
```


examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -55,49 +55,29 @@ async def _main():
...
-            async def raise_if_error_async(awaitable_call):
-                """Await response, then raise an exception if an error was returned."""
-                response = await awaitable_call
-                await raise_if_error(response)
-                return response
-
-            create_response: nidaqmx_types.CreateTaskResponse = await raise_if_error_async(
-                client.CreateTask(nidaqmx_types.CreateTaskRequest())
-            )
+            client.CreateTask(nidaqmx_types.CreateTaskRequest())
```
dmondrik, line 58: You have to assign the return value to `create_response` so you can use it on the next line.

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -172,14 +146,16 @@ async def wait_for_done():
...
                 except asyncio.CancelledError:
                     pass
 
             await asyncio.gather(read_data(), wait_for_done())
 
         except grpc.RpcError as rpc_error:
             error_message = rpc_error.details()
+            for key, value in rpc_error.trailing_metadata() or []:
+                if key == "ni-error":
+                    error_message += f"\nError status: {value}"
```
dmondrik, line 158: Note that this likely needs to change. I've made some changes since my original post and I don't think it's final yet. Stay tuned.

examples/nidaqmx/counter-output.py
```diff
@@ -50,45 +50,34 @@
...
+            units=nidaqmx_types.DIGITAL_WIDTH_UNITS3_SECONDS,
+            idle_state=nidaqmx_types.LEVEL1_LOW,
+            initial_delay=1.0,
+            low_time=0.5,
+            high_time=1.0,
         )
     )
 
-    raise_if_error(client.StartTask(nidaqmx_types.StartTaskRequest(task=task)))
+    client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
```
dmondrik, line 69: @bkeryan what DAQ methods ought we check for warnings?

dmondrik (approved):
I'll try to make sure you know when I've got a solution that seems to work for all cases. Remember I won't be here tomorrow, so don't reset my vote.

examples/nidaqmx/analog-input-every-n-samples.py
tdunkle: Fixed.

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -119,13 +99,11 @@ async def raise_if_error_async(awaitable_call):
...
 
-            await raise_if_error_async(client.StartTask(nidaqmx_types.StartTaskRequest(task=task)))
+            client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
 
-            response = await raise_if_error_async(
-                client.GetTaskAttributeUInt32(
-                    nidaqmx_types.GetTaskAttributeUInt32Request(
-                        task=task, attribute=nidaqmx_types.TASK_ATTRIBUTE_NUM_CHANS
-                    )
+            client.GetTaskAttributeUInt32(
```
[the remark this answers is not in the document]
tdunkle: Fixed.

examples/nidaqmx/counter-output.py
bkeryan: I think the most important ones are StartTask and the various Read/Write methods.

Hardware timed AO and DO often uses IsTaskDone or WaitUntilTaskDone to check for errors and warnings, but I don't think these examples demonstrate these use cases.

StopTask can return DAQmxWarningStoppedBeforeDone (200010). It would be reasonable to move StopTask out of the finally block and check for warnings. ClearTask will do all necessary cleanup, including stopping the task, but I don't think it reports errors or warnings that occur during cleanup.

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -55,49 +55,29 @@ async def _main():
...
-                    )
-                    raise Exception(f"Error: {response.error_string}")
-
-            async def raise_if_error_async(awaitable_call):
-                """Await response, then raise an exception if an error was returned."""
-                response = await awaitable_call
-                await raise_if_error(response)
-                return response
-
-            create_response: nidaqmx_types.CreateTaskResponse = await raise_if_error_async(
```
bkeryan, line 72: Before: this assigned a local variable with a type annotation.
```
create_response: nidaqmx_types.CreateTaskResponse = ...
```

After: this incorrectly overwrites the type?
```
nidaqmx_types.CreateTaskResponse = ...
```

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -55,49 +55,29 @@ async def _main():
...
-                client.CreateAIVoltageChan(
-                    nidaqmx_types.CreateAIVoltageChanRequest(
-                        task=task,
-                        physical_channel=PHYSICAL_CHANNEL,
-                        min_val=-10.0,
-                        max_val=10.0,
-                        terminal_config=nidaqmx_types.InputTermCfgWithDefault.INPUT_TERM_CFG_WITH_DEFAULT_CFG_DEFAULT,
-                        units=nidaqmx_types.VoltageUnits2.VOLTAGE_UNITS2_VOLTS,
-                    )
+            client.CreateAIVoltageChan(
```
bkeryan, line 61: I haven't used Python asyncio but it looks like this creates a bunch of coroutines without actually executing them. I suspect that each one of this calls needs an `await`.

bkeryan (changes requested):

examples/nidaqmx/analog-input-every-n-samples.py
tdunkle: Whoops, didn't mean to remove the `create_response: `. I've fixed this.

examples/nidaqmx/analog-input-every-n-samples.py
tdunkle: Added `await`s everywhere that we were previously calling `raise_if_error_async()`.

examples/nidaqmx/analog-input-every-n-samples.py
tdunkle: I've updated the review so that every file's exception-handling uses the code block that you ended using for #717 (Update switch/tclk/xnet/session examples to display error code).

astarche:
Looks good once you get the details squared up with Dan and the tests/validate passing.

Can you run the success cases too? I believe most of these will run with the simulated PXIe-6341 that we use for system tests ([here's the nce file](https://github.com/ni/grpc-device/blob/main/source/tests/assets/grpc-device-daq-tests.nce) you could import; or just add one in MAX). Also consider the `validate_examples` script for running a bunch of tests, as I posted on the gRPC Teams channel. The DAQ tests probably don't pass mypy and may need additional deps, but hopefully you could get them running pretty easily (either ignoring, disabling, or fixing).

astarche (approved):

reckenro (approved):

examples/nidaqmx/counter-output.py
```diff
@@ -66,7 +74,8 @@
         )
     )
 
-    client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
+    start_task_response = client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
+    check_for_warning(start_task_response)
 
     client.WaitUntilTaskDone(
         nidaqmx_types.WaitUntilTaskDoneRequest(task=task, time_to_wait=10.0)
```
bkeryan, line 81: Check for warning after WaitUntilTaskDone.

examples/nidaqmx/analog-input.py
```diff
@@ -76,7 +84,8 @@
         )
     )
 
-    client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
+    start_task_response = client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
```
bkeryan, line 87: As I mentioned before, `StopTask` may return a specific warning. It's way at the bottom of each file in the `finally` block, but it doesn't really belong there because `ClearTask` is sufficient in error cases.

bkeryan (changes requested):

examples/nidaqmx/analog-input.py
tdunkle: I've removed the `StopTask` calls and am now checking the response of `ClearTask` for warnings.

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -55,49 +55,37 @@ async def _main():
...
 
-            async def raise_if_error(response):
-                """Raise an exception if an error was returned."""
-                if response.status:
-                    response = await client.GetErrorString(
-                        nidaqmx_types.GetErrorStringRequest(error_code=response.status)
+            def check_for_warning(response):
+                """Print to console if the status indicates a warning."""
+                if response.status > 0:
+                    warning_message = client.GetErrorStringResponse(
```
dmondrik, line 61: I think the method call should be just `GetErrorString` (response is part of the type it returns).

dmondrik: Also, is that the right method to be paired with `ErrorMessageRequest` below?

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -55,49 +55,37 @@ async def _main():
...
-            async def raise_if_error_async(awaitable_call):
-                """Await response, then raise an exception if an error was returned."""
-                response = await awaitable_call
-                await raise_if_error(response)
-                return response
-
-            create_response: nidaqmx_types.CreateTaskResponse = await raise_if_error_async(
-                client.CreateTask(nidaqmx_types.CreateTaskRequest())
-            )
+            create_response: nidaqmx_types.CreateTaskResponse = await client.CreateTask(nidaqmx_types.CreateTaskRequest())
```
dmondrik, line 66: Is it necessary to declare the `CreateTaskResponse` type here? It's not done anywhere else that I can see.

examples/nidaqmx/analog-input.py
```diff
@@ -51,46 +51,41 @@
...
-            nidaqmx_types.GetErrorStringRequest(error_code=response.status)
-        )
-        raise Exception(f"Error: {response.error_string}")
-
-
 try:
+    def check_for_warning(response):
+        """Print to console if the status indicates a warning."""
+        if response.status > 0:
+            warning_message = client.GetErrorStringResponse(
```
dmondrik, line 58: You've likely done this in all the files. I won't mention it elsewhere, just verify it globally in your PR.

examples/nidaqmx/analog-output.py
```diff
@@ -50,49 +50,49 @@
...
-    """Raise an exception if an error was returned."""
-    if response.status != 0:
-        response = client.GetErrorString(
-            nidaqmx_types.GetErrorStringRequest(error_code=response.status)
-        )
-        raise Exception(f"Error: {response.error_string}")
-
-
 try:
+    def check_for_warning(response):
```
dmondrik, line 54: In other examples we define the `check_for_warning` method above/before the try block. Might want to do the same for DAQ if possible.

examples/nidaqmx/dsa-shared-timebase-and-trig-analog-input-and-output.py
```diff
@@ -163,134 +165,98 @@ def generate_sinewave(elements, amplitude, frequency, phase):
...
-            async def raise_if_error_async(awaitable_call):
-                """Await response, then raise an exception if an error was returned."""
-                response = await awaitable_call
-                await raise_if_error(response)
-                return response
-
             # DAQmx Configure Tasks code
-            response: nidaqmx_types.CreateTaskResponse = await raise_if_error_async(
-                client.CreateTask(nidaqmx_types.CreateTaskRequest(session_name="Leader input task"))
+            response: nidaqmx_types.CreateTaskResponse = (
```
dmondrik, line 169: Here's another where you declare the type. In at least some cases it seems you went with a more well-scoped variable name, which would probably be easier to read IMO. But really, I don't want to cause more work if this passes the checks. It is what was there in the previous submission.

dmondrik (approved):

examples/nidaqmx/dsa-shared-timebase-and-trig-analog-input-and-output.py
tdunkle: Agreed that the variable name could be improved. Since all of the checks are passing and I think the variable names are out of scope for this PR, I'm going to leave them untouched for now.

examples/nidaqmx/analog-input-every-n-samples.py
tdunkle: Potentially not, but I'm also going to leave this untouched as it's out of scope for the current PR (which I'm hoping to merge today).

examples/nidaqmx/analog-input-every-n-samples.py
tdunkle: Switched to `GetErrorString`.

examples/nidaqmx/analog-input.py
tdunkle: Switched to `GetErrorString` throughout the PR.

tdunkle:
Dismissing @bkeryan who is out of office this week. Brad, please let me know if you have any concerns with this submission when you're back.

examples/nidaqmx/analog-input.py
bkeryan: `ClearTask` does not report warnings. Calling `StopTask` in success cases gives the task a chance to report warnings.

When in doubt, follow the LV shipping examples.
image

PR #719, opened 2022-09-01
Add nimxlcterminaladaptor_restricted service

achanyou:

Add a restricted service to expose remote access to the nimxlcTerminalAdaptor.dll functions used to get device terminal information on a host.

The python files for this service were manually created. The imports are from version 22.5 of nimxlcTerminalAdaptor.

Required for a committed feature to support grpc with RFmx SpecAn clients like Instrument Studio.

Added a system test to verify that session creation, refreshing terminal data and getting device container can be called and return without errors. Many of these functions only work in conjunction with other driver calls. More in-depth testing will be done later as the project progresses.


source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return ::grpc::Status::CANCELLED;
+    }
+    try {
+        auto grpc_session = request->session();
+        nimxlc_Session session = session_repository_->access_session(grpc_session.id(), grpc_session.name());
+
+        nierr_Status status;
+        status.code = 0;
+        status.capacity = 0;
+        status.json = "";
```
bkeryan, line 27: `status->json` (`status.json` here) is normally allocated on the heap using `status->reallocJson`. Initializing `status->json` to NULL is ok but initializing it to a string literal is risky. If capacity > 0, nimxlc uses code in `nierr_Status.c` that overwrites the buffer. This would corrupt the EXE's string constant table (or probably crash because string constants are usually read-only).

When nimxlc sets a fatal status, it uses code in `nierr_Status.c` that calls `status->reallocJson`. If you leave `reallocJson` uninitialized, setting a fatal status will crash.

Before the function returns, it should use `status->reallocJson` to free `status->json`. Otherwise, it will leak memory. Most clients of `nierr_Status` use the C++ `nierr::Status` class to do this.

I think it would be possible to use a fixed-size, stack-allocated buffer if you initialize `status->capacity` to the buffer size and `status->reallocJson` to point to a function that always returns false.

Cc: @irwand

bkeryan (changes requested):

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
```diff
@@ -0,0 +1,80 @@
...
+                library_->DeviceIterator_next(&deviceIterator);
+            }
+        }
+        library_->DeviceContainer_destroy(deviceContainer);
+
+        response->set_status((&status)->code);
+
+        ::grpc::Status grpcStatus;
+        if (nierr_Status_isFatal(&status)){
+            grpcStatus = ::grpc::Status(::grpc::INTERNAL, status.json);
```
bkeryan, line 62: It's mostly documented in the Google API documentation and not the gRPC documentation, but `grpc::Status::error_message` is really supposed to be a developer-facing, English error message. 
https://cloud.google.com/apis/design/errors
https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto

When you return a `grpc::Status` that has an error, most of the client APIs throw an exception, which makes it impossible for the client to access `response.status`.

Please strongly consider returning the `nierr_Status` as trailing metadata.
- `status->code`: use the key name "ni-error" like the the rest of grpc-device now does.
- `status->json`: 
  - You could use a key name similar to "ni-error-json" with a URL-encoded JSON string. (JSON is typically UTF-8, but string metadata must be ASCII.)
  - You could use a key name similar to "ni-error-json-bin" order to return the JSON as a binary string.

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
```diff
@@ -0,0 +1,80 @@
...
+        status.code = 0;
+        status.capacity = 0;
+        status.json = "";
+        nimxlc_DeviceContainer deviceContainer = library_->getDeviceContainer(session, &status);
+        if (nierr_Status_isNotFatal(&status))
+        {
+            nimxlc_DeviceIterator deviceIterator = library_->DeviceContainer_begin(deviceContainer);
+            while (!library_->DeviceIterator_isEnd(deviceContainer, deviceIterator) && nierr_Status_isNotFatal(&status))
+            {
+                auto device = response->add_container_out();
```
bkeryan, line 34: If adding data to the response object throws an exception, the server will leak memory. Please consider using RAII to call the cleanup functions.

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
bkeryan: BTW, the JSON does not contain the error code. You need both.

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
irwand: Why are we initializing status manually like that? why not use the provided `nierr_Status_initialize(&status)` ?

irwand: or.. yeah, just use the `nierr::Status` C++ wrapper as suggested.

generated/nimxlcterminaladaptor_restricted/nimxlcterminaladaptor_restricted.proto
```diff
@@ -0,0 +1,101 @@
...
+  int32 status = 1;
+}
+
+message RefreshTerminalCacheRequest {
+  nidevice_grpc.Session session = 1;
+}
+
+message RefreshTerminalCacheResponse {
+  int32 status = 1;
+  NIErrStatus c_status = 2;
```
astarche, line 68: Will `NIErrStatus` be needed in the response when this information is incorporated into the `status` and `trailing_metadata`?

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
astarche: Using the trailing metadata sounds good to me. Is there a way to get an "english error message" from the fatal status? If not, is it better to use the (url-encoded?) json or a generic "an error has occurred, refer to the trailing metadata" sort of message?

astarche (approved):
Looks good. I agree with Brad that the `status` handling should be updated to use the trailing metadata. I'm OK with that being a separate task.

reckenro (approved):

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
bkeryan: >  Is there a way to get an "english error message" from the fatal status? 

The `nierr_Status` doesn't contain an English error message. It can be translated into an English error message (or French, German, etc.) but this requires calling into internal libraries that you don't have access to here.

What does it contain?
- `status->code` is the error code, which is used to look up the static error message in the error XML files.
- `status->json` is a language-independent representation of the dynamic error message, which gets translated to 
English, Japanese, etc.:
  ```
   Property: AI.Min
   Requested Value: 1000.0
   Minimum Value: -10.0
   Maximum Value: 10.0
   ```

> If not, is it better to use the (url-encoded?) json or a generic "an error has occurred, refer to the trailing metadata" sort of message?

I think you should return a developer-facing error message like the latter. It could include more info such as the error code, but clients should read the trailing metadata rather than parsing the error message.

"Error -12345 has occurred in nimxlcterminaladaptor. Refer to the trailing metadata for details."

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
bkeryan: I think this is the bare minimum fix for now:
- Initialize `status.reallocJson` to point to a function with the correct signature that returns false.
- Initialize `status.json` to `nullptr`
- Update the rest of this code to tolerate a NULL `status.json`

It won't return any JSON, but at least it shouldn't crash.

Wedgetable:
What export version did the imports come from (21.8, 22.5..)? You might add that information to the PR description somewhere.

bkeryan:
> I may be a bit confused here, but why use the grpc status instead of just a status field in the response like most other driver methods? The fact that nierr_Status is being used to implement status chaining here seems like an implementation detail that should not leak into the proto file and a fatal status returned by the driver isn't actually a grpc error.

@cumitche This behavior has changed. The rest of the driver methods now report errors as errors.

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
achanyou: Added DeviceContainerPtr and TerminalContainerPtr objects that call the appropriate cleanup functions in the destructor.

source/custom/nimxlcterminaladaptor_restricted_service.custom.cpp
achanyou: I've created a follow-up task to pass the json string back as metadata. I'll also add a generic error message including the error code while doing this.

achanyou:
I got myself into trouble with git due to merging. After unsuccessfully trying to fix things, it looks like the simplest thing to do is to move to [another PR](https://github.com/ni/grpc-device/pull/726). Sorry for the inconvenience.

PR #726, opened 2022-09-07
(Repost) Add nimxlcterminaladaptor_restricted service

achanyou:
This is a repost of a [previous PR](https://github.com/ni/grpc-device/pull/719). I got myself into some trouble with git while trying to merge. Feedback from the earlier PR has been addressed, except for getting the nierr_Status json string back from the dll call and adding it to the grpc response metadata. That will be done in a follow-up PR.

Add a restricted service to expose remote access to the nimxlcTerminalAdaptor.dll functions used to get device terminal information on a host.

The python files for this service were manually created. The imports are from version 22.5 of nimxlcTerminalAdaptor.

Required for a committed feature to support grpc with RFmx SpecAn clients like Instrument Studio.

Added a system test to verify that session creation, refreshing terminal data and getting device container can be called and return without errors. Many of these functions only work in conjunction with other driver calls. More in-depth testing will be done later as the project progresses.


astarche (approved):

CMakeLists.txt
```diff
@@ -499,6 +502,7 @@ if(WIN32)
   list(
     APPEND 
       system_test_runner_sources
+        "source/tests/system/nimxlcterminaladaptor_restricted_driver_api_tests.cpp"
```
reckenro, line 505: The new test is [failing on our build](https://github.com/ni/grpc-device/runs/8253818570?check_suite_focus=true#step:4:870). We aren't blocking grpc-device PRs right now for failing System Tests since we have a few other intermittent but we should probably do one of two things for this new one.

We can either remove the new test file from the SystemTestsRunner along with any of the additions with `target_link_libraries` and `set_target_properties`.

Or we could try to update our windows grpcbot to pull in whatever feeds it needs to get these new tests to pass. The Windows feeds are defined [here](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Feng%2Fiac%2Fcfg_mgmt%2Fsalt%2Fsrv%2Fsalt%2Fgrpcbot%2Fwin%2Fresources%2Fdriver_feeds_win.yml&version=GBmain&_a=contents) in AzDo.

reckenro: And I'm fine with either update going in after this change, it's just something we'll need to address.

reckenro (approved):
Just had a note to figure out what to do with the system test. Other than that it looks good!

CMakeLists.txt
achanyou: I've created a task to address this next. Thanks for the detailed explanation!

PR #731, opened 2022-09-12
Format MI metadata to be same pretty format / ordering as hapigen output.

reckenro:

Formats the metadata files (attributes, config, enums, and functions) for the mi-drivers in the same style / ordering as the metadata output from the new hapigen grpc-device plugin.

I ran `python format-mi-metadata.py C:\dev\reckenro-git\grpc-device\source\codegen\metadata` to format all the metadata.

- This should make subsequent PRs / diffs for the mi metadata sourced from the new hapigen plugin much easier to parse. The data should be in the same order with the same formatting so the diff will show actual changes and not ordering/formatting changes.
- This should be a safe change since this is just formatting / re-ordering the metadata and there shouldn't be any functional changes.

Still builds and will check to make sure system-level tests still pass.


source/codegen/format-mi-metadata.py
```diff
@@ -0,0 +1,115 @@
...
+import sys
+from enum import Enum
+from pprint import pprint
+from typing import Set
+from unittest import FunctionTestCase
+
+from template_helpers import load_metadata
+
+
+class Formatter(object):
```
reckenro, line 14: Note: I took this Formatter class straight from hapigen: see [here](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/hapigen/hapigen/source/hapigen_utilities/hapigen_utilities/Formatter.py)

source/codegen/format-mi-metadata.py
```diff
@@ -0,0 +1,115 @@
+import argparse
```
reckenro, line 1: I can delete this file. It was just temporarily needed to auto-format all the mi-driver metadata files.

It doesn't hurt to keep it around. Still builds fine but might be noisy to keep it around. Thoughts, @astarche ?

source/codegen/metadata/niswitch/enums.py
```diff
@@ -1,6 +1,4 @@
-# -*- coding: utf-8 -*-
-# This file is generated from NI-SWITCH API metadata version 19.6.0d7
```
dmondrik, line 2: Is the metadata version now ignored? Seems a comment like this would still be helpful. But that's probably more a question for the generator.

dmondrik (approved):

source/codegen/metadata/niswitch/enums.py
dmondrik: I'm not sure what I was looking at, but this looks correct again. So nevermind?

source/codegen/format-mi-metadata.py
astarche: What about adding it as a workflow action? to run this and fail if it would change any MI drivers?

Anyway, I think I slightly prefer keeping it. We have a number of utility scripts and they're useful when they're useful.

astarche (approved):

source/codegen/format-mi-metadata.py
reckenro: I'll go ahead and submit it since it's useful when it's useful and doesn't harm anything like you said.

True, after we update the MI drivers we could put this in a workflow as a verification step. I'll make a note to circle back on that.

PR #732, opened 2022-09-13
Mutate mi driver metadata in format-mi-metadata script.

reckenro:

Follow up to #731. In a similar vein but updated the script to also mutate the mi metadata so the mutated metadata is stored in source.

For the new hapigen gprc-device plugin, we brought over the mutation code at the request of the reviewers for there to be minimal processing on the grpc-device side to consume the metadata from hapigen.
For that reason, we should go ahead and apply metadata mutation to the MI drivers to make the diffs easier when comparing to ones sourced from hapigen.

- Will make the diffs for grpc-device metadata from the hapigen plugin output easier to parse.
- No functional changes. Changed some ordering in the proto files due to the expanded Attribute enums being present before they're sorted. But other than that, no changes to generated files.

Checked and made sure only .proto files were re-generated (due to it sorting the pre-mutation attribute enums with the others by name).

System tests still pass.


source/codegen/metadata_validation.py
```diff
@@ -317,6 +336,8 @@ def _validate_enum(enum_name: str, used_enums: Set[str], metadata: dict):
                     raise Exception(
                         f"generate-mappings is False, but values have non-int types: {value_types}"
                     )
+            if metadata["config"]["driver_name"] in _ALREADY_MUTATED:
```
reckenro, line 339: Note: This is to avoid the duplicate enum rule running on Attribute enums that will now be present in the MI driver metadata.

source/codegen/metadata_validation.py
```diff
@@ -86,6 +98,11 @@ class RULES:
         Optional("additional_arguments_to_output_allocation"): [str],
         Optional("proto_only"): bool,
         Optional("input_passed_by_ptr"): bool,
+        Optional("cppName"): str,
+        Optional("determine_size_from"): [str],
+        Optional("is_size_param"): bool,
+        Optional("linked_params_are_optional"): bool,
+        Optional("mapped-enum"): str,
```
dmondrik, line 105: This entry `mapped-enum` is the only one in the list I see with the word separator being a dash instead of an underscore. Is that intentional? Can we fix it in the tool?

While we're at it we should probably also fix `cppName` which is the only one that doesn't have a word separator?

reckenro: Hmm, I don't think there was any intention behind it. Nor with `cppName`. I just don't think we standardized those with the rest.

Also, there's a handful of other dash-separated keys and values throughout the metadata files. We seemed to have added some for Enums along the way, like 'generate-mappings' and 'force-include'.

Then there were some pre-existing ones (from nimi-python and in turn hapigen) for the parameter's size mechanisms like 'ivi-dance-with-a-twist' and 'passed-in'.

reckenro: We can either

a) Hold off on them and standardize on underscore_seperated keys/values in a future clean-up.

or

b) I could address mapped-enum and cppName right now. Replace all instances with underscore separated versions. I would also then need to make corresponding updates to [metadata_mutation.py](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/hapigen/hapigen_grpc_device_plugin/source/hapigen_grpc_device_plugin/metadata_mutation.py&version=GBmain&line=173&lineEnd=174&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) used for the hapigen plugin.

dmondrik: I'd honestly prefer tech debt, at some theoretical point when we're not being quizzed every day as to progress.

reckenro: Sounds good. Made AB#2146456 in our Tech Debt backlog to look address this.

source/codegen/metadata_validation.py
astarche: Add this as a comment in the code.

source/codegen/metadata_validation.py
```diff
@@ -6,6 +6,18 @@
 from schema import And, Optional, Or, Schema, Use  # type: ignore
 
 
+_ALREADY_MUTATED = (
```
astarche, line 9: This should have both a technical and a historical coment. Something like "grpc device metadata mutations have been applied to MI drivers in source to minimize transformations that happen outside of hapigen"

Would it also be fair to say that it "makes sense" that all drivers would do this? Not that we should make the change, but  that this is probably a better model than what we're doing for the other drivers. At least for metadata that is generated/expanded/transformed by another process (like `hapigen` or `scrapigen`)

astarche (approved):

source/codegen/metadata_validation.py
reckenro: That comments sounds good. Added it.

And yes, I think it makes sense to do the mutations wherever the metadata is sourced (hapigen, scrapigen, or here for hand-written ones). It feels better and more transparent for the process creating the metadata to produce the final version used in grpc-device. Yeah, I'd say it would make sense to do the mutations for scrapigen-sourced drivers as well eventually.

PR #738, opened 2022-09-14
[Draft] Testing hapigen output for ni-dmm

reckenro:

Using it to test the output from hapigen for dmm. It's in progress in[ this PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/358221) in Azure

Intentional deletions:
- `Control4022` method
   - Likely not used, Marcos thought we shouldn't've added in first place, and Harsha agreed on removing it and breaking a client that had .001% chance of using it
- `Configuration4022` enum
   - Only used in Control4022 method which is now removed.
- `NIDMM_ATTRIBUTE_IO_SESSION`, `NIDMM_ATTRIBUTE_CLASS_DRIVER_CLASS_SPEC_MAJOR_VERSION `, and `NIDMM_ATTRIBUTE_CLASS_DRIVER_CLASS_SPEC_MINOR_VERSION` attributes
  - Likely not used, not in hapigen metadata, and Harsha agreed we can get rid of them

It shouldn't. That's why it's a draft request. Just using for diffs and running the build.

Manually checking metadata updates and generated file updates.
System tests still pass.


source/codegen/metadata/nidmm/attributes.py
```diff
@@ -1,645 +1,620 @@
...
     1050305: {
-        'access': 'read only',
-        'channel_based': False,
+        'codegen_method': 'public',
+        'grpc_type': 'string',
         'name': 'LOGICAL_NAME',
         'resettable': False,
         'type': 'ViString'
     },
-    1050322: {
```
reckenro, line 81: Figuring out how to add missing Attributes in hapigen.

source/codegen/metadata/nidmm/enums.py
```diff
@@ -1,83 +1,88 @@
...
+            },
+            {
+                'name': 'NIDMM_VAL_ACQUISITION_STATUS_NO_ACQUISITION_IN_PROGRESS',
+                'value': 4
             }
         ]
     },
     'ApertureTime': {
+        'codegen_method': 'public',
+        'generate-mappings': True,
```
reckenro, line 48: After [this PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/361109) goes in and I regen from hapigen, the values should go back to integers and `'generate-mappings': False,`

source/codegen/metadata/nidmm/enums.py
```diff
@@ -149,51 +158,64 @@
...
-                'name': 'NIDMM_VAL_OFFSET_COMP_OHMS_ON',
-                'value': 1
-            }
-        ]
-    },
     'Configuration4022': {
+        'codegen_method': 'public',
         'values': [
             {
-                'name': 'NIDMM_4022_CONFIG_GUARD',
```
reckenro, line 167: Working on straightening these names out with [this PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/361682)

generated/nidmm/nidmm.proto
```diff
@@ -119,7 +119,6 @@ enum NiDmmAttribute {
   NIDMM_ATTRIBUTE_SPECIFIC_DRIVER_PREFIX = 1050302;
   NIDMM_ATTRIBUTE_IO_RESOURCE_DESCRIPTOR = 1050304;
   NIDMM_ATTRIBUTE_LOGICAL_NAME = 1050305;
-  NIDMM_ATTRIBUTE_IO_SESSION = 1050322;
```
reckenro, line 122: `NIDMM_ATTRIBUTE_IO_SESSION`, `NIDMM_ATTRIBUTE_CLASS_DRIVER_CLASS_SPEC_MAJOR_VERSION `, and `NIDMM_ATTRIBUTE_CLASS_DRIVER_CLASS_SPEC_MINOR_VERSION` are now missing.

But these are from [IviDmmObsolete.h](https://github.com/ni/grpc-device/blob/main/imports/include/IviDmmObsolete.h). Should these be included in the enum or should we just remove them?

source/codegen/metadata/nidmm/attributes.py
reckenro: May not want to add these. See comment in proto file.

generated/nidmm/nidmm.proto
```diff
@@ -150,7 +147,7 @@ enum NiDmmAttribute {
   NIDMM_ATTRIBUTE_NUMBER_OF_AVERAGES = 1150032;
   NIDMM_ATTRIBUTE_LATENCY = 1150034;
   NIDMM_ATTRIBUTE_BUFFER_SIZE = 1150037;
-  NIDMM_ATTRIBUTE_FREQ_VOLTAGE_AUTO_RANGE_VALUE = 1150044;
```
reckenro, line 153: Changed Attribute name... need to fix

generated/nidmm/nidmm.proto
```diff
@@ -160,7 +157,7 @@ enum NiDmmAttribute {
   NIDMM_ATTRIBUTE_DC_BIAS = 1150053;
   NIDMM_ATTRIBUTE_SERIAL_NUMBER = 1150054;
   NIDMM_ATTRIBUTE_LC_NUMBER_MEAS_TO_AVERAGE = 1150055;
-  NIDMM_ATTRIBUTE_CONFIG_PRODUCT_NUMBER = 1150061;
```
reckenro, line 163: Changed Attribute name... need to fix.

generated/nidmm/nidmm.proto
```diff
@@ -197,11 +194,11 @@ enum NiDmmAttribute {
 }
 
 enum AcquisitionStatus {
-  ACQUISITION_STATUS_NIDMM_VAL_RUNNING_ANTICOLLISION = 0;
-  ACQUISITION_STATUS_NIDMM_VAL_FINISHED_WITH_BACKLOG = 1;
-  ACQUISITION_STATUS_NIDMM_VAL_FINISHED_WITH_NO_BACKLOG = 2;
-  ACQUISITION_STATUS_NIDMM_VAL_PAUSED = 3;
-  ACQUISITION_STATUS_NIDMM_VAL_NO_ACQUISITION_IN_PROGRESS = 4;
+  ACQUISITION_STATUS_NIDMM_VAL_ACQUISITION_STATUS_RUNNING = 0;
```
reckenro, line 197: Enum value name changed. Need to fix.

generated/nidmm/nidmm.proto
```diff
@@ -226,17 +223,12 @@ enum CalibrationType {
   CALIBRATION_TYPE_NIDMM_VAL_EXTERNAL_AREA = 1;
 }
 
-enum CompensatedOhms {
```
reckenro, line 229: Need to override OffsetCompensatedOhms to generate as CompensatedOhms. Same enum, different names.

generated/nidmm/nidmm.proto
```diff
@@ -226,17 +223,12 @@ enum CalibrationType {
...
-  COMPENSATED_OHMS_NIDMM_VAL_OFFSET_COMP_OHMS_ON = 1;
-}
-
 enum Configuration4022 {
-  CONFIGURATION4022_NIDMM_4022_CONFIG_GUARD = 0;
-  CONFIGURATION4022_NIDMM_4022_CONFIG_CURRENT_100_NANOAMP = 2;
-  CONFIGURATION4022_NIDMM_4022_CONFIG_CURRENT_10_MICROAMP = 3;
-  CONFIGURATION4022_NIDMM_4022_CONFIG_CURRENT_1_MILLIAMP = 7;
-  CONFIGURATION4022_NIDMM_4022_CONFIG_DISCONNECT_AGND = 8;
+  CONFIGURATION4022_NIDMM_VAL_4022_CONFIG_GUARD = 0;
```
reckenro, line 227: Enum value names changed.

generated/nidmm/nidmm.proto
```diff
@@ -414,13 +406,29 @@ enum NiDmmReal64AttributeValues {
...
+}
+
 enum PowerLineFrequencies {
   POWER_LINE_FREQUENCIES_UNSPECIFIED = 0;
   POWER_LINE_FREQUENCIES_NIDMM_VAL_50_HERTZ = 50;
   POWER_LINE_FREQUENCIES_NIDMM_VAL_60_HERTZ = 60;
 }
 
-enum RtdType {
+enum RTDType {
```
reckenro, line 431: Enum name changed.

generated/nidmm/nidmm.proto
```diff
@@ -557,7 +565,8 @@ message CheckAttributeViReal64Request {
   NiDmmAttribute attribute_id = 3;
   oneof attribute_value_enum {
     NiDmmReal64AttributeValues attribute_value = 4;
-    double attribute_value_raw = 5;
+    NiDmmReal64AttributeValuesMapped attribute_value_mapped = 5;
```
reckenro, line 568: Need to add field number overrides for this function to keep attribute_value_raw as 5

source/codegen/metadata/nidmm/functions.py
```diff
@@ -1747,14 +1807,15 @@
...
+        'codegen_method': 'public',
         'init_method': True,
         'parameters': [
             {
                 'cppName': 'resourceName',
                 'direction': 'in',
                 'grpc_type': 'string',
                 'name': 'resourceName',
-                'type': 'ViRsrc'
+                'type': 'ViString'
```
reckenro, line 1818: Note: This changed types but it shouldn't affect the underlying type in .proto or the underlying type passed into the DLL. This may be more accurate since it's just sourced from base hapigen metadata now.

reckenro:
There's conversation about what to do with Control4022 (4022Control) method. Marcos was surprised it's included here and would opt to remove it. Which would be a breaking change but I believe would result in an UNIMPLEMENTED gRPC status returned if an old client tried to call it.

Dan was opting to mark the method deprecated but I'm not sure if there's a way to do that in .proto files. We could look into marking all the request/response message fields as deprecated but I'm not sure if that's better than just introducing the breaking-change of removing the method. @astarche , thoughts on this?

generated/nidmm/nidmm.proto
reckenro: Marcos said in the ni-central PR: "I don't think IVIDMM_ATTR_IO_SESSION means anything in NI-DMM. I don't even know what it returns. It's for IVI drivers implemented on top of VISA iirc.

The other two I suppose are harmless."

I'm still trying to figure out what to do with these. These variants aren't in the other drivers so I think they were added erroneously here. I guess the question here is if we want / need to keep them for backwards compatibility or not.

astarche:
@reckenro It looks like there is [some support for deprecating messages in `proto`](https://github.com/protocolbuffers/protobuf/issues/1734), which seems like a decent solution here. We'd want that to generate code with some sort of IDE error/warning/message support in python/C# (C++ would be nice too). If it's `proto`-only I think that's not sufficient (I am making an assumption that this method "never should have been there" and is likely not used so that deleting it is an OK option).

I like the `deprecated` idea and think it's worth trying out to have it in the toolbox. If it's not working as hoped/advertised, or if if doesn't address Marcos's concerns, I think deleting is OK too.

generated/nidmm/nidmm.proto
harsha-bhushan: Let's leave them out. 
Same reasoning as before - if someone was using it, we break it in the next release & lead them to the correct way (of not using it)

source/codegen/metadata/nidmm/functions.py
dmondrik: The types are interchangeable so I'm not worried.

generated/nidmm/nidmm.proto
reckenro: Updated so that method and attributes are no longer generated.

generated/nidmm/nidmm.proto
```diff
@@ -53,7 +53,6 @@ service NiDmm {
   rpc ConfigureWaveformAcquisition(ConfigureWaveformAcquisitionRequest) returns (ConfigureWaveformAcquisitionResponse);
   rpc ConfigureWaveformCoupling(ConfigureWaveformCouplingRequest) returns (ConfigureWaveformCouplingResponse);
   rpc Control(ControlRequest) returns (ControlResponse);
-  rpc Control4022(Control4022Request) returns (Control4022Response);
```
dmondrik, line 56: The PR description should definitely call out which obsolete deletions are intentional.

reckenro, line 56: TODO: Figure out when we submit these changes how to remember to call out that we removed this function in the release notes.

dmondrik (approved):

generated/nidmm/nidmm.proto
reckenro: Added a part at the top of the PR description for "intentional deletions" to list all of these out and the rationale for deleting them.

generated/nidmm/nidmm.proto
```diff
@@ -557,6 +556,7 @@ message CheckAttributeViReal64Request {
   NiDmmAttribute attribute_id = 3;
   oneof attribute_value_enum {
     NiDmmReal64AttributeValues attribute_value = 4;
+    NiDmmReal64AttributeValuesMapped attribute_value_mapped = 6;
```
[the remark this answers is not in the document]
reckenro: Hmm, yeah I don't think it has an effect but we should fix it. I found the spot to fix in proto_helpers.py to sort the enum parameters by field number. I made a [separate task](https://dev.azure.com/ni/DevCentral/_workitems/edit/2166444) for the story, though, to knock it out in a separate PR here, parallel one in hapigen, and to wait for the in-flight draft PRs to go in to fix these occurrences in all the updated .proto files.

dmondrik:
These changes should be rendered obsolete by my PR #754
This PR should be abandoned.

PR #741, opened 2022-09-16
Replaced switch metadata with hapigen local export, and rebuilt

dmondrik:

This includes the effect of ni-central [PR 360432](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/360432) for NI-SWITCH which was submitted. 

This uses the actual metadata exported from the hapigen CI.

Rebuilt locally with no other observed changes.


generated/niswitch/niswitch.proto
```diff
@@ -7,7 +7,7 @@
 syntax = "proto3";
 
 option java_multiple_files = true;
-option java_package = "com.ni.grpc.niswitch";
+option java_package = "com.ni.grpc.switch";
```
reckenro, line 10: Cool, yeah we talked about this and I think it's fine to update the name here to be more in line with the other .proto files.

source/codegen/metadata/niswitch/config.py
```diff
@@ -1,30 +1,20 @@
...
+# -*- coding: utf-8 -*-
+# This file is generated from NI-SWITCH API metadata version 22.8.0d9999
 config = {
     'additional_headers': {
         'custom/ivi_errors.h': [
             'service.cpp'
         ]
     },
-    'api_version': '19.6.0d7',
+    'api_version': '22.8.0d9999',
```
reckenro, line 9: Oh yeah, this is what I was talking about earlier in person. This and the version printed in the comment above all these files. That's mainly why we don't want to submit this version I think. I think it's fine once the ni-central PR goes in and we update from the metadata in that export. Then we'll get an actual export version here for the api_version.

We also haven't talked about it but in my mind for all these MI drivers we can do updates to development versions of the API between grpc-device releases. But when we prepare a new release, we'll pull the metadata from a final export in ni-central.

source/codegen/metadata/niswitch/functions.py
```diff
@@ -301,7 +314,6 @@
                 'type': 'ViSession'
             },
             {
-                'attribute': 'SCAN_DELAY',
```
reckenro, line 304: Note: We talked about this offline and this 'attribute' key/value isn't used in grpc-device. Doesn't affect generated output. Safe to remove.

source/codegen/metadata/niswitch/functions.py
```diff
@@ -443,6 +461,8 @@
     },
     'ErrorMessage': {
         'cname': 'niSwitch_error_message',
+        'codegen_method': 'public',
+        'is_error_handling': True,
```
reckenro, line 465: Hmm, little surprised at the addition  of 'is_error_handling' here and a few other functions. But looking at the code this is only in some functions.py and the metadata_validation so it's not actually used in the code generation with reflects in no corresponding changes for these additions in the generated folder.

I'll just mark this resolved. We could look into it and removing it in the future but right now it's innocuous extra info we're not using.

source/codegen/metadata/niswitch/functions.py
```diff
@@ -962,7 +997,8 @@
         'returns': 'ViStatus'
     },
     'Init': {
-        'cname': 'niSwitch_init ',
```
reckenro, line 965: This is just a funny callout that this used to have a space in it which nothing complained about. You can see it reflected in some of the changes in the generated files where it got rid of the space like the old version of
```  return niSwitch_init (resourceName, idQuery, resetDevice, vi);```

**Comment by reckenro** [comment]
Looks great! Love to see only additions in the proto file and expected fixups and name changes in the other generated files.

Just made some callouts on other things in the metadata but marked them resolved.

File: generated/niswitch/niswitch.proto (hunk 1 above)
Comment (astarche) [inline generated/niswitch/niswitch.proto:10]: We'd have to tag these as `breaking_change`. But I expect it's OK to change these. Would be an "easy-to-fix source-breaking change" affecting a small group of clients (possibly none).

Comment (dmondrik) [reply generated/niswitch/niswitch.proto]: Verified it's fixed now (back to what it used to be).

File: generated/niswitch/niswitch.proto [hunk 6]
```diff
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is automatically generated. All manual edits will be lost.
+// This file is generated from NI-SWITCH API metadata version 22.8.0d130
```
reckenro, line 3: I think we'll need to revert updating the comment for now. You can just use the generated .proto file here instead of the one from hapigen.

I have this [Task](https://dev.azure.com/ni/DevCentral/_workitems/edit/2128588) to add validation against source proto files and what's generated in the generated folder. I think I'll also need to add this updated comment generation to the proto gen in grpc_device for all the .proto files. I can do that in a separate PR though.

dmondrik: Pushed a commit for what is built locally rather than what hapigen exported. This should pass the prebuild now.

reckenro (approved):

PR #742, opened 2022-09-16
[Draft] Hapigen output for NI SCOPE

danestull:

Testing hapigen generated data from [PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/358344?path=/src/hapigen/hapigen_grpc_device_plugin/source/hapigen_grpc_device_plugin/GrpcDeviceExpander.py) 

Just showing diffs

Local builds show generated proto file looks about right


source/codegen/metadata/niscope/config.py
```diff
@@ -1,101 +1,91 @@
...
     'additional_headers': {
         'custom/ivi_errors.h': [
             'service.cpp'
         ]
     },
-    'api_version': '20.5.0d7',
+    'api_version': '22.8.0d9999',
     'c_function_prefix': 'niScope_',
-    'c_header': 'niScope.h',
+    'c_header': 'niscope.h',
```
reckenro, line 11: Wrong capitalization for c_header.

source/codegen/metadata/niscope/functions.py
```diff
@@ -1458,7 +1384,41 @@
         ],
         'returns': 'ViStatus'
     },
+    'ErrorMessage': {
```
reckenro, line 1387: Wasn't here before. I think we'll just 'codegen_method': 'no' for this function in hapigen.

source/codegen/metadata/niscope/functions.py
```diff
@@ -1440,7 +1366,7 @@
                 'name': 'errorSource',
                 'size': {
                     'mechanism': 'fixed',
-                    'value': 642
+                    'value': 55
```
reckenro, line 1369: This is interesting that this changed. Needs further investigation.

source/codegen/metadata/niscope/enums.py
```diff
@@ -18,276 +19,163 @@
...
             {
-                'documentation': {
-                    'description': 'The oscilloscope is configured to emit a single CableSense pulse.'
-                },
                 'name': 'NISCOPE_VAL_CABLE_SENSE_MODE_ON_DEMAND',
                 'value': 1
             }
         ]
     },
-    'CalibrationTypes': {
```
reckenro, line 274: Missing

source/codegen/metadata/niscope/enums.py
```diff
@@ -695,217 +579,130 @@
             }
         ]
     },
-    'FIRFilterWindow': {
```
reckenro, line 698: Missing

source/codegen/metadata/niscope/enums.py
```diff
@@ -695,217 +579,130 @@
...
-                'documentation': {
-                    'description': 'Trigger on pulses of either positive or negative polarity.'
-                },
                 'name': 'NISCOPE_VAL_GLITCH_EITHER',
                 'value': 3
             }
         ]
     },
     'InputImpedance': {
-        'generate-mappings': True,
```
reckenro, line 863: Darn, I was assuming that we always had manually changed doubles to ints when possible for enums in grpc-device. Doesn't seem to be the case though. I can look into updating the hapigen-plugin to take in an option for enums to not try to coerce enum values to ints.

source/codegen/metadata/niscope/enums.py
```diff
@@ -1680,183 +1355,118 @@
...
             {
-                'documentation': {
-                    'description': '\nNI-SCOPE does not return an error when an overflow has occurred in the\nOSP block.\n'
-                },
                 'name': 'NISCOPE_VAL_ERROR_REPORTING_DISABLED',
                 'value': 2
             }
         ]
     },
-    'P2PStreamRelativeTo': {
```
reckenro, line 1743: Missing

source/codegen/metadata/niscope/attributes.py
```diff
@@ -1,2556 +1,1558 @@
...
         'type': 'ViReal64'
     },
     1150042: {
-        'access': 'read-write',
-        'channel_based': True,
-        'codegen_method': 'public',
-        'documentation': {
-            'description': '\nSpecifies the FIR window type. The possible choices are:\nNISCOPE_VAL_NONE\nNISCOPE_VAL_HANNING_WINDOW\nNISCOPE_VAL_HAMMING_WINDOW\nNISCOPE_VAL_TRIANGLE_WINDOW\nNISCOPE_VAL_FLAT_TOP_WINDOW\nNISCOPE_VAL_BLACKMAN_WINDOW\nThe symmetric windows are applied to the FIR filter coefficients to limit passband ripple in FIR filters.\nDefault: NISCOPE_VAL_NONE\n'
-        },
-        'enum': 'FIRFilterWindow',
```
reckenro, line 690: Missing 'enum' tag for this attribute. You can look [here](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/358221?_a=files&path=/src/nidmm/nidmmBaseDistribution/nidmmAPIMetadata/source/nidmmAPIMetadata/grpc_device/attributes/attributes.hapigen) for an example of how to add these tags.

source/codegen/metadata/niscope/attributes.py
```diff
@@ -1,2556 +1,1558 @@
...
     },
     1150335: {
-        'access': 'read-write',
-        'channel_based': False,
         'codegen_method': 'public',
-        'documentation': {
-            'description': '\nSpecifies the address to Push Message push Value to on the event specified by the NISCOPE_ATTR_P2P_NOTIFY_PUSH_MESSAGE_ON attribute.\n',
-            'note': 'This attribute can be used only with high-speed digitizers that support peer-to-peer streaming.'
-        },
-        'enum': 'AddressType',
```
reckenro, line 1858: Missing 'enum' tag for this attribute. You can look [here](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/358221?_a=files&path=/src/nidmm/nidmmBaseDistribution/nidmmAPIMetadata/source/nidmmAPIMetadata/grpc_device/attributes/attributes.hapigen) for an example of how to add these tags.

reckenro: Hmm, actually. I think the attribute below is correctly using this enum 'AddressType'. I don't think this one ever should have... Not sure how we want to deal with this for backwards compatibility. I'd opt to just remove the 'enum' tag like you have here but this will change the .proto.

source/codegen/metadata/niscope/attributes.py
```diff
@@ -1,2556 +1,1558 @@
...
         'name': 'SAMP_CLK_TIMEBASE_MULT',
-        'python_name': 'sample_clock_timebase_multiplier',
         'resettable': True,
         'type': 'ViInt32'
     },
     1150373: {
-        'access': 'read-write',
-        'channel_based': False,
         'codegen_method': 'public',
-        'enum': 'P2PStreamRelativeTo',
```
reckenro, line 2035: Missing 'enum' tag for this attribute. You can look [here](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/358221?_a=files&path=/src/nidmm/nidmmBaseDistribution/nidmmAPIMetadata/source/nidmmAPIMetadata/grpc_device/attributes/attributes.hapigen) for an example of how to add these tags.

source/codegen/metadata/niscope/attributes.py
```diff
@@ -1,2556 +1,1558 @@
...
-        'type_in_documentation': 'hightime.timedelta, datetime.timedelta, or float in seconds'
+        'type': 'ViReal64'
     },
     1250201: {
-        'access': 'read-write',
-        'channel_based': False,
-        'documentation': {
-            'description': '\nSpecifies the type of video signal, such as NTSC, PAL, or SECAM.\n'
-        },
-        'enum': 'VideoSignalFormat',
```
reckenro, line 2386: Missing 'enum' tag for this attribute. You can look [here](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/358221?_a=files&path=/src/nidmm/nidmmBaseDistribution/nidmmAPIMetadata/source/nidmmAPIMetadata/grpc_device/attributes/attributes.hapigen) for an example of how to add these tags.

source/codegen/metadata/niscope/functions.py
danestull: So i think this one is actually a correct change 642 is the max error description size, and 55 is the max function name size so therefore errorSource which is the parameter telling us the function source seems like 55 is more correct. It is strange that it didn't change in the service though

source/codegen/metadata/niscope/enums.py
danestull: These enums are only used by functions that in grpc_device are listed as codegen no such as `CalFetchDate` so since I put them as codegen no in hapigen as well they no longer generate therefore these enums now cause a build error since they aren't referenced so they have been removed.

source/codegen/metadata/niscope/enums.py
danestull: same as above

source/codegen/metadata/niscope/enums.py
danestull: same as above enums

source/codegen/metadata/niscope/enums.py
danestull: This has been added back since its referenced by an attribute now

source/codegen/metadata/niscope/enums.py
danestull: this has also been added back as its referenced by attribute now

source/codegen/metadata/niscope/attributes.py
danestull: left this as is for now

source/codegen/metadata/niscope/enums.py
danestull: I've already made some changes to the hapigen-plugin for Scope specific stuff so I can try updating this as well if you want to have a chat about what all needs to change with me

danestull: This is being handled in this [PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/366124)

danestull: This has been fixed

source/codegen/metadata/niscope/enums.py
```diff
@@ -680,186 +565,127 @@
                 'name': 'NISCOPE_VAL_READY_FOR_REF_EVENT',
                 'value': 10
             },
-            {
-                'name': 'NISCOPE_VAL_REF_CLOCK',
-                'value': 100
-            },
             {
                 'name': 'NISCOPE_VAL_5V_OUT',
-                'python_name': 'FIVE_V_OUT',
```
danestull, line 689: @reckenro are these python names in enums needed? There's a few places I see them and I don't know if they are needed and I need to make more hapigen changes or if I can leave them out

source/codegen/metadata/niscope/functions.py
```diff
@@ -1882,7 +1803,7 @@
                     'mechanism': 'custom-code',
                     'value': '(num_samples * actual_num_wfms)'
                 },
-                'type': 'NIComplexNumber_struct[]'
```
danestull, line 1885: @reckenro I think these changes are also ones I'm not sure if need to be handled it seems the typedef struct for niscope has both names so it seems like it should be alright without the `_struct` see niScope.h for reference.

reckenro: Yeah, I think these changes are okay because it's the same struct just with different aliases in the niScope.h file like you said. So we'll see extra diffs in the generated files with these type name changes but they're the same structs so it's fine.

I do have some changes to better handle and provide overrides for custom_type param 'type's: https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/367950. I needed that for FGEN but I don't think it's required for your changes.

generated/niscope/niscope.proto
```diff
@@ -979,10 +966,7 @@ message CheckAttributeViInt64Request {
   nidevice_grpc.Session vi = 1;
   string channel_list = 2;
   NiScopeAttribute attribute_id = 3;
-  oneof value_enum {
```
reckenro, line 982: Looking at this again I'm concerned with the oneof going away and value_raw's field number is changing from 5 to 4. We discussed the cause of this in the metadata files but summarizing this is because initially for grpc-device we had incorrectly marked the `P2P_NOTIFY_MESSAGE_PUSH_ADDR` attribute with the `'enum': 'AddressType'` tag which in turn causes `NiScopeInt64AttributeValues` to get generated and used here and in the `Set...` method. If we correctly remove that 'enum' tag now, we'll get this breaking change.

@harsha-bhushan , is this an acceptable breaking change? Or would we rather want to try to keep backwards compatibility here by still providing the `NiScopeInt64AttributeValues` enum input option (even if the values are kind of bogus) along with the `value_raw` option.

harsha-bhushan: Looking at the P2P_NOTIFY_MESSAGE_PUSH_ADDR attribute documentation, this specifies to which peer address a message should be pushed. It is unlikely someone would have used an enum for this & most likely they would have set a raw value. So, although it is a breaking change the impact may not be big, but rather beneficial? If someone was using the enum by mistake (which was incorrect), they will now be forced to use the raw value. Which seems the right thing to do.
(please correct me if i'm missing something)

reckenro:
Newer Draft PR here for hapigen-exported NI Scope metadata: https://github.com/ni/grpc-device/pull/760

Marking this one closed.

PR #743, opened 2022-09-19
Fix pipeline names.

csjall:
Fixed AzDo pipeline names.

Currently our main/release CI pipelines are not running on AzDo. 

From @davidcorrigan714:
The token looks valid. The pipeline that workflow is triggering looks to be “ni-central-grpc-device-main-desktop” which doesn’t exist. Looks like someone may have changed a dash to an underscore in the name.

TBD


zhindes (approved):
Can confirm these match the current names. It is a shame this [action](https://github.com/Azure/pipelines/blob/master/action.yml) doesn't have an option to use the ID of the pipeline, so that name changes don't break stuff. :(

astarche (approved):

reckenro (approved):

SparkingSpork (approved):

PR #748, opened 2022-09-23
Fix nimxlcTerminalAdaptor_restricted system test failure

achanyou:

Remove the hostname passed to the nimxlcTerminalAdaptor create session call from the system test, since the intention is for mxlcCore to look for devices on the localhost.

This fixes a failing system test.

Run Windows system test workflow.

### DIFF
--- a/source/tests/system/nimxlcterminaladaptor_restricted_driver_api_tests.cpp
+++ b/source/tests/system/nimxlcterminaladaptor_restricted_driver_api_tests.cpp
@@ -14,8 +14,6 @@ namespace tests {
 namespace system {
 namespace {
 
-constexpr auto DEFAULT_HOSTNAME = "localhost";
-
 class NiMxLcTerminalAdaptorRestrictedDriverApiTests : public Test {
  protected:
   NiMxLcTerminalAdaptorRestrictedDriverApiTests()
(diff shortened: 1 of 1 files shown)


cumitche (approved):

source/tests/system/nimxlcterminaladaptor_restricted_driver_api_tests.cpp
```diff
@@ -48,20 +48,15 @@ class NiMxLcTerminalAdaptorRestrictedDriverApiTests : public Test {
...
+  auto response = client::create_session(stub, "");
   EXPECT_EQ(0, response.status());
   EXPECT_EQ(0, response.c_status().code());
   auto session = response.handle();
   return session;
 }
 
-nidevice_grpc::Session init_session(const client::StubPtr& stub)
-{
-  return init_session(stub, DEFAULT_HOSTNAME);
```
astarche, line 62: It seems that the `DEFAULT_HOSTNAME` constant is never used and can be removed (or should be changed to `""` and used in the new code?).

astarche (approved):

source/tests/system/nimxlcterminaladaptor_restricted_driver_api_tests.cpp
achanyou: Thanks for catching that! I've removed the constant.

reckenro (approved):

PR #749, opened 2022-09-27
Include driver api_version in proto file comments.

reckenro:

Includes the driver version in the proto file comments at the top. 

We did this for the proto files generated in hapigen and want to do it in grpc-device as well. This is also required for a follow up task where we want to add a validation step so if any driver has a .proto file in it's metadata folder (or adjacent proto folder, TBD) that we validate it against the one generated here in grpc-device. When we add that it will ensure the .proto generation in grpc-device matches the generation done in hapigen for the same metadata files.

Manually inspected .proto file changes. Also, compared the niswitch.proto file to the one in the hapigen export and they line up now with that same comment at the top.


generated/nidcpower/nidcpower.proto
```diff
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is automatically generated. All manual edits will be lost.
+// This file is generated from NI-DCPower API metadata version 20.7.0d8
```
astarche, line 3: It looks like all the dev/weird versions are from MI components. Are they all going to go final as part of syncing up with hapigen? Did those get dev versions in the first place because we just took "whatever is in nimi-python"? And it has dev versions because of some auto-update process? Or some ad-hoc update process?

In practice, this probably only matters if there is an API stability bug or an API change/revert during dev. But .. it looks pretty weird.

Is there a task tracking that?

What is our intended update cadence in the future? Will we pull in new finals when they are available (or when requested)? Or will we attempt to update against dev versions?

astarche (approved):

generated/nidcpower/nidcpower.proto
reckenro: Yeah, this is just getting it from the 'api_version' in config.py. And right now for most, it's just what we got from nimi-python's config.py at whatever time we took them. I made a task to fix it on the hapigen side, though, [here](https://dev.azure.com/ni/DevCentral/_workitems/edit/2163782). So after that, the api_version's in config.py and the version put in this comment for .proto files shouldn't have the 'd' or 'f' build numbers at the end. I do think the build number versions are fine on the grpc-device metadata files (comment at top of functions.py) but they shouldn't be exposed on the proto files, I agree.

As for the intended update cadence in my mind, I think it's okay for main to be using dev versions but for any release I'd imagine we'd want all the grpc-device metadata files sourced from a final export. Now when we pull in finals into main in grpc-device is still not clear to me.

PR #750, opened 2022-09-29
Validate hapigen imported proto files against ones in generated folder.

reckenro:

Adds a script and corresponding GitHub Workflow to check to make sure proto files imported from elsewhere (right now just MI drivers from hapigen) match up with the corresponding .proto files in the generated folder.

- New validate_imported_protos.py that will take in path to "imported" protos directory and "generated" directory. It will compare all imported .proto files with their corresponding .proto file in the "generated" directory. If any mismatch, it will exit with code 1
- New validate_imported_protos.yml to run the validate_imported_protos.py script. Consumed in early parts of ci.yml workflow.
- Fixed naming of validate-linux-rt.py to validate_linux_rt.py while I was here
- Added all MI proto files to imported/codegen/protos folder. Copied from generated proto files. Adding them now so that when we make updates to a driver's MI metadata (from hapigen) we will now be forced to include the hapigen output .proto file in this directory or the build will fail.

[AB#2128588](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2128588)

This is to ensure that the proto generation in grpc-device matches whatever is creating the .proto files elsewhere (hapigen process). I think ideally we would single-source a "proto generator" component to take metadata and spit out .proto files used here and in hapigen but I think that was deemed too much effort right now so this should at least cover the generators being functionally equivalent.

Tested in my fork with the script failing and passing

Failing after altering three proto files in source/codegen/metadata:
https://github.com/reckenro/grpc-device/actions/runs/3152881724/jobs/5128663939

Success with matching proto files:
https://github.com/reckenro/grpc-device/actions/runs/3152909065/jobs/5128727223


source/codegen/validate_source_protos.py
```diff
@@ -0,0 +1,54 @@
...
+def _check_mismatched_protos(source: str, generated: str) -> bool:
+    any_mismatches = False
+    file_paths = glob.glob(f"{source}/**/*.proto", recursive=True)
+    for file_path in file_paths:
+        driver_dir_name = os.path.split(os.path.split(file_path)[0])[1]
+        file_name = os.path.basename(file_path)
+        driver_and_file_path = f"{driver_dir_name}/{file_name}"
+        generated_file_path = f"{generated}/{driver_and_file_path}"
+        compare_result = filecmp.cmp(file_path, generated_file_path)
+        if not compare_result:
```
dmondrik, line 22: When you have both `if` and `else` conditions, it's better to use positive logic rather than negative logic. In other words, flip the logic and the 2 blocks.

.github/workflows/validate_source_protos.yml
```diff
@@ -0,0 +1,25 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7.x'
+
+    - name: Install build dependencies
+      run: python -m pip install -r python_build_requirements.txt
+
+    - name: Validate source proto files against generated
+      run: |
+        python source/codegen/validate_source_protos.py -s source/codegen/protos -g generated
```
dmondrik, line 25: Since the script only grabs the proto files, you could technically put these .proto files in the same directories as the other metadata files, and make it easier on us when copying from hapigen exports.

dmondrik (approved):

.github/workflows/validate_source_protos.yml
reckenro: Yeah, I thought about that and thought it was weird having the .proto files alongside the metadata files. So my thought was to have them in this separate folder here and also update the .mak files in hapigen to output the .proto files to a "proto" folder in the exports as well.

I could also just update this submission to put the .proto files next to the metadata files in the source/codegen/metadata folder instead.

Thoughts?

astarche, line 25: Should these be `imports/hapigen/protos` instead of `source/codegen/protos`. Maybe `hapigen`-sourced metadata should also be in `imports`. And, either way, could there be a `README.md` in the directory explaining where they came from?

Similarly I'm not getting the name `*_source_protos`. `imported` or `hapigen` would work there. Maybe `dual_sourced`, if that's not too cute.

source/codegen/validate_source_protos.py
```diff
@@ -0,0 +1,54 @@
+"""Script for validating the source proto files against generated ones.
+
+Expects input source path to source/codegen/protos and generated folder path.
+"""
+
+import argparse
```
astarche, line 6: Plug to consider `click` ([Internal best practices doc](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/21764/Common-Modules-List)) when you can use external modules. (No need for a rewrite here; and always optional IMO).

astarche (changes requested):
await: I'd like to see the code and comments to reflect the intent of the protos/scripts and give some clues on how to work with them.

.github/workflows/validate_source_protos.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Validate Source Proto Files
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Repo
+      uses: actions/checkout@v2
+
+    - name: Setup python3
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7.x'
```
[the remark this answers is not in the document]
reckenro: File name changed to validate_imported_protos.yml but I changed it to `3.x` which should work.

source/codegen/validate_source_protos.py
reckenro: File name changed to validate_imported_protos.py but I inverted the `if` and `else` to have positive logic.

source/codegen/validate_source_protos.py
```diff
@@ -0,0 +1,54 @@
+"""Script for validating the source proto files against generated ones.
```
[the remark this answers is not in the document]
reckenro: Yeah, that's basically the workflow and things they will hit. I tried to capture this at the top of validate_imported_protos.py now (explaining a bit of the background, the problem, and fixes if the validation fails).

SparkingSpork (approved):

.github/workflows/validate_source_protos.yml
reckenro: Talked with Alex and he was okay with putting the .proto files alongside the metadata files. For now, that'll be source/codegen/metadata/<driver_name>. I went ahead and moved the proto files there and updated the script/workflow to look there.

Long term, we want any imported metadata/proto files to be somewhere other than source. Perhaps in the imports directory. That applies to both hapigen and scrapigen sourced metadata/proto files. But for now, keeping it in souce/codegen/metadata will be alright.

reckenro: Talked about offline some. Eventually we'll want the imported metadata and proto files (hapigen and scrapigen currently) somewhere outside of the source directory. But for now, we'll keep with the current paradigm and put the hapigen-sourced stuff here in source/codegen/metadata.

I did update all the file names and terminology in the files to use "imported" over "source" to avoid confusion some. Also, added small README.md in each of the hapigen metadata folders to give some insight about where they come from. Also, have a Imported_From_Hapigen.md living in source/codegen/metadata that each of them reference with some more info on what's going on and what to do if the build fails on the "validate imported protos" step.

source/codegen/validate_source_protos.py
reckenro: Oh neat. I went ahead and did it. I like `click`, don't think I've seen it before.

astarche (approved):
Great update. Thanks!

PR #752, opened 2022-09-30
Create validate_mi_mutation.yml

IPetersNI:
This Pull Request responds to/validates the need and use of formatting the metadata files(attributes.py ,config.py ,enums.py , and functions.py) with the same style through the creation of validate_mi_mutation.yml and format_mi_metadata.py

This change should be merged so that there is a standard for each metadata generated file which then makes it easier to manipulate them.

A local build was done to ensure building will succeed and the format_mi_metadata.py file was ran and compared to what is in the grpc-device repository locally.

[AB#2145221](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2145221)


source/codegen/format_mi_metadata.py
```diff
@@ -104,8 +104,8 @@ def _format_mi_metadata(metadata_dir: str):
                 f"# This file is generated from {api_name} API metadata version {api_version}"
             )
             pretty_metadata += "\r\n"
-            pretty_metadata += f"{metadata_name} = {pretty(actual_metadata)}"
-            with codecs.open(f"{path}{metadata_name}.py", "w", "utf-8-sig") as temp:
+            pretty_metadata += f"{metadata_name} = {pretty(actual_metadata)}\r\n"
+            with io.open(f"{path}{metadata_name}.py", "w", encoding="utf-8", newline="\n") as temp:
```
reckenro, line 108: Note: This fixes up the encoding to match what hapigen mutation / format uses. Also, the newlines in the hapigen output for the bulk of the exported files are just `\n` so we set that here as well.

.github/workflows/validate_mi_mutation.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Install build dependencies
+      run: python -m pip install -r python_build_requirements.txt
+
+    - name: Validate mi driver metadata mutations 
+      run: |
+        python source/codegen/format_mi_metadata.py source/codegen/metadata
+    
+    - name: Check for files dirtied by codegen
+      run: |
+        git diff --ignore-space-at-eol --exit-code
```
reckenro, line 29: Note: It was tricky getting the end of lines to match up completely between here and the hapigen output... especially when this runs on Windows vs. Ubuntu so we added `--ignore-space-at-eol` to just not care about the line endings.

source/codegen/format_mi_metadata.py
astarche: I'm surprised by this because both git and python have special features to give platform newlines. In git, the repo version will have unix newlines that are converted to platform newlines on checkout (by default). And python's [default file I/O is "universal newlines"](https://docs.python.org/3/library/functions.html#open-newline-parameter) where `\n` is always written out as a platform newline.

So I'd expect:
1. There should never be a `\r\n` in the python code. The line above should end with `\n`.
   * It looks like some of the existing code outside of this diff should be changed as well.
3. The `newline` param to `io.open` should not be used. The default `None` is correct ("universal newlines").

Are you sure that doesn't work? If the problem is that it doesn't matches `hapigen` (a) that's still surprising and (b) the `git diff` call is already ignoring line separators. So, it seems that this code is both wrong and does not match hapigen. Might as well be right and not match hapigen.

.github/workflows/validate_mi_mutation.yml
astarche: The `\r\n` code is definitely wrong on linux unless there's some double-negative there that I'm not understanding. If it's not breaking the linux build, I guess it's OK to leave it if you can't figure this out. But it's confusing enough that I bet someone will run into it again.

astarche (commented):
I'm glad to see this change to keep the data in sync!

await: I'm interested to hear the follow up on the `\r\n` newline business.

.github/workflows/validate_mi_mutation.yml
SparkingSpork: Not necessary in this PR, consider adding a `.gitattributes` to address the line ending behavior. See [here.](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings).

SparkingSpork (approved):
Looks good to me aside from the line endings discussion that's ongoing.

source/codegen/format_mi_metadata.py
reckenro: Yeah, I was getting a bit confused about all the newline stuff. Changed all the `\r\n` in this file to just `\n` and got rid of the `newline` arg to `io.open` so it just relies on the default being `None`. That at least makes it right in this script.

That should work and I updated the validation .yml file to just call `git diff --exit-code`. It doesn't need to ignore end lines like I initially thought.

As for the hapigen output, at least right now it will have `\r\n` for three lines (corresponding to two comment lines at the top and the ending blank line the .mako file has). It then uses `\n` for all the lines corresponding to the pretty formatted dictionary. You can see that .mako [here](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/hapigen/hapigen_grpc_device_plugin/source/templates/grpc_device.py.mako).

.github/workflows/validate_mi_mutation.yml
reckenro: I removed the `--ignore-space-at-eol` and it should still work. I think I was getting confused by the newline stuff and at some point put this in for it to work. But it seems like and should work without it.

python_build_requirements.txt
```diff
@@ -6,3 +6,4 @@ black>=21.10b0
 mypy>=0.910
 ni-python-styleguide~=0.1
 pycodestyle==2.7.0
+importlib-metadata==4.12.0
```
reckenro, line 9: Note: This is a somewhat unrelated change but the build started breaking today when running python styleguide check on the python files: https://github.com/ni/grpc-device/actions/runs/3175907824/jobs/5174555407

importlib-metadata released version 5.0.0 which removed some deprecated methods I think we were hitting in the styleguide check. @astarche , I'm not too familiar with this stuff. Does this seem like a reasonable way to resolve this?

astarche (approved):

PR #756, opened 2022-10-05
Fix RFmxInstr snapshot retrieval

cumitche:

Changed GetInitiaitedSnapshotStrings() and GetLatestConfigurationSnapshot() to call GetSnapshotInfoFromCache() instead of their own function again to complete the IVI dance.

The current implementation performs a standard IVI dance with a twist.  This is not what the driver expects.  The first call to GetInitiaitedSnapshotStrings() or GetLatestConfigurationSnapshot() will write the snapshot to a cache.  Calling it again may succeed if none of the sizes have changed while capturing a second snapshot, but the snapshot in the cache will remain.  Eventually, the snapshot cache will overflow its maximum size and it will return an internal error.  This change addresses the issue by reading the snapshot from the cache, which also removes it from the cache.

Tested GetInitiaitedSnapshotStrings() by successfully remotely monitoring an RFmx SpecAn session.


cumitche:
@astarche I looked at the XNET socket implementation, but I couldn't figure out how to simplify my custom method implementation based upon them.  Can you provide some more specifics?

astarche:
@cumitche . I like your solution of using custom service methods here. Looks good 👍 .

BUT I'll try to explain what I was thinking w/r/t nxsocket. I'm not sure if it would even work for this case. But if we ever have dozens of methods like this, we should look into it.

A good example is AddrInfo:
* [Some special metadata in functions.py](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nixnetsocket/functions.py#L183). Sets up `supports_standard_copy_convert` and `supports_standard_output_allocation`.
* [OutputConverter uses the library](https://github.com/ni/grpc-device/blob/f66d03eda39832da9c958f3bf8fd6854b146b8e6/source/custom/xnetsocket_converters.h#L649) when copying the output to the result.
* Note that there are also [some tricks for handling arrays and sizes](https://github.com/ni/grpc-device/blob/f66d03eda39832da9c958f3bf8fd6854b146b8e6/source/custom/xnetsocket_converters.h#L53).

So you'd need:
* A `<driver>_converters.h` file (and add to config.py)
* Add a type alias to `<driver>_converters.h`  (i.e., `using SnapshotArray = int32[]`) and use it in `functions.py` (so C++ has something to dispatch on for the standard converter overload method (i.e., `convert_to_grpc(SnapshotArray)`).
* Update `functions.py` to use the `supports_standard_copy_convert` metadata (... and figure out the right set of inputs.)
* Implement the converters that call into the library and get the snapshot from the cache.

Superficially, it looks like the hard part would be that there are SO MANY inputs to those methods. That's awkward to deal with, and every param is another opportunity for weird stuff. I'd also be concerned that all of the "boiler plate" that we're hoping to keep in the generated code would get pulled into the converter (so it'd just be a "custom" method in a different place and in a confusing order).

astarche (approved):

reckenro (approved):

source/custom/nirfmxinstr_restricted_service.custom.cpp
```diff
@@ -15,4 +22,191 @@ ::grpc::Status NiRFmxInstrRestrictedService::ConvertApiErrorStatusForNiRFmxInstr
...
+    response->mutable_snapshot_timestamp_array()->Resize(snapshot_timestamp_array_actual_size, 0);
+    uInt64* snapshot_timestamp_array = reinterpret_cast<uInt64*>(response->mutable_snapshot_timestamp_array()->mutable_data());
+    auto personality_id_array_size = personality_id_array_actual_size;
+    auto result_names_size = result_names_actual_size;
+    auto signal_names_size = signal_names_actual_size;
+    auto snapshot_identifiers_size = snapshot_identifiers_actual_size;
+    auto snapshot_timestamp_array_size = snapshot_timestamp_array_actual_size;
+    status = library_->GetSnapshotInfoFromCache(
+      instrument,
+      snapshot_info_cache_index,
```
cumitche, line 89: Calling this method with an index of 0 will return an internal error.  If the above call returns an index of 0, we should skip trying to retrieve the snapshot because there was nothing to cache.

cumitche: Added a check for snapshot_info_cache_index != 0 before calling GetSnapshotInfoFromCache().  I also tagged the snapshotInfoCacheIndex parameter in GetInitiaitedSnapshotStrings and GetLatestConfigurationSnapshot methods with 'include_in_proto': False.  This parameter is only used to communicate between the above C functions and their corresponding GetSnapshotInfoFromCache function call.

source/codegen/metadata/nirfmxinstr_restricted/functions.py
```diff
@@ -869,6 +871,118 @@
         ],
         'returns': 'int32'
     },
+    'GetSnapshotInfoFromCache': {
```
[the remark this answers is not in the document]
cumitche: I have been working on finding a way to add a test for GetInitiaitedSnapshotStrings().  The method can only be called in a session setup for monitoring and requires a separate standalone session actively using the device to return any snapshot information.  I will continue working on this and post a subsequent PR with a test if I can identify a test scenario of reasonable complexity.

PR #758, opened 2022-10-05
Import ni-fake metadata from hapigen export.

reckenro:

Like the other MI drivers, we should get the ni-fake grpc-device metadata from its hapigen export.

1. Gets ni-fake grpc-device metadata in sync with changes made in hapigen.
2. Want to get this version in since our hapigen work is nearing completion. There will be one upcoming ni-fake change to test out nested custom_types which is needed for DC Power. But that diff will be simpler if we get this baseline ni-fake change in first.

[AB#2173878](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2173878)

Unit and Integration tests against ni-fake pass locally.

Note: No new tests were added since there's no new grpc-device codegen capabilities to test. The pre-existing Unit and Integration tests should be sufficient.

### DIFF
--- a/generated/nifake/nifake.proto
+++ b/generated/nifake/nifake.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-FAKE API metadata version 1.2.0d9
+// This file is generated from NI-FAKE API metadata version 23.0.0
 //---------------------------------------------------------------------
 // Proto file for the NI-FAKE Metadata
 //---------------------------------------------------------------------
@@ -22,8 +22,10 @@ service NiFake {
   rpc AcceptViUInt32Array(AcceptViUInt32ArrayRequest) returns (AcceptViUInt32ArrayResponse);
   rpc BoolArrayInputFunction(BoolArrayInputFunctionRequest) returns (BoolArrayInputFunctionResponse);
   rpc BoolArrayOutputFunction(BoolArrayOutputFunctionRequest) returns (BoolArrayOutputFunctionResponse);
+  rpc Close(CloseRequest) returns (CloseResponse);
   rpc CloseExtCal(CloseExtCalRequest) returns (CloseExtCalResponse);
   rpc CommandWithReservedParam(CommandWithReservedParamRequest) returns (CommandWithReservedParamResponse);
+  rpc Control4022(Control4022Request) returns (Control4022Response);
   rpc CreateConfigurationList(CreateConfigurationListRequest) returns (CreateConfigurationListResponse);
   rpc DoubleAllTheNums(DoubleAllTheNumsRequest) returns (DoubleAllTheNumsResponse);
   rpc EnumArrayOutputFunction(EnumArrayOutputFunctionRequest) returns (EnumArrayOutputFunctionResponse);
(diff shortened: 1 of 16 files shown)


generated/nifake/nifake.proto
```diff
@@ -86,7 +94,6 @@ service NiFake {
   rpc ViUInt8ArrayInputFunction(ViUInt8ArrayInputFunctionRequest) returns (ViUInt8ArrayInputFunctionResponse);
   rpc ViUInt8ArrayOutputFunction(ViUInt8ArrayOutputFunctionRequest) returns (ViUInt8ArrayOutputFunctionResponse);
   rpc WriteWaveform(WriteWaveformRequest) returns (WriteWaveformResponse);
-  rpc Close(CloseRequest) returns (CloseResponse);
```
reckenro, line 89: Note: Re-sorted to be on line 25.

generated/nifake/nifake.proto
```diff
@@ -96,7 +103,7 @@ enum NiFakeAttribute {
   NIFAKE_ATTRIBUTE_READ_WRITE_STRING = 1000002;
   NIFAKE_ATTRIBUTE_READ_WRITE_COLOR = 1000003;
   NIFAKE_ATTRIBUTE_READ_WRITE_INTEGER = 1000004;
-  NIFAKE_ATTRIBUTE_FLOAT_ENUM = 1000005;
+  NIFAKE_ATTRIBUTE_FLOAT_ENUM_NAME_OVERRIDE = 1000005;
```
reckenro, line 106: There's going to be several "breaking" changes to ni-fake that was done to test out things in hapigen. Since we only use ni-fake for Unit and Integration tests and don't publish / add the service to the server executable these should be fine.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -161,7 +161,7 @@ TEST(NiFakeServiceTests, NiFakeService_InitWithOptionsAndResetServer_SessionIsCl
   const char* session_name = "sessionName";
   EXPECT_CALL(library, InitWithOptions)
       .WillOnce(DoAll(SetArgPointee<4>(kTestViSession), Return(kDriverSuccess)));
-  EXPECT_CALL(library, close(kTestViSession))
```
reckenro, line 164: Note: All the tests fixes were because the names changed in the metadata when updated from hapigen. Some were our doing intentional (name overrides for enums/attributes) but some were just the nature of updating the version of the ni-fake hapigen metadata we're based on (new fields, casing changes, etc.)

generated/nifake/nifake.proto
```diff
@@ -138,10 +175,10 @@ enum MobileOSNames {
 
 enum NiFakeInt32AttributeValues {
   NIFAKE_INT32_UNSPECIFIED = 0;
```
dmondrik, line 177: This one is unexpected (to me) since there's `GrpcColorOverride.GRPC_COLOR_OVERRIDE_UNSPECIFIED` above.

dmondrik (approved):

generated/nifake/nifake.proto
reckenro: So the "AttributeValues" enums are sort of special, they're populated for an attribute type from other enum values. But for UNSPECIFIED (0) enum value for the AttributeValue enums it'll just populate one with the `NIFAKE_INT32_UNSPECIFIED = 0;` instead of multiple UNSPECIFIED's from each enum. This might not be the best example since its just made up of one enum (GrpcColorOverride), but if you look at switch's proto for [NiSwitchInt32AttributeValues ](https://github.com/ni/grpc-device/blob/38d7abec10b2a79ec34c2595fa08f09b5c102afa/generated/niswitch/niswitch.proto#L154) you can see it just has one UNSPECIFIED instead of grabbing them from all the different enums that are used to populate it.

Anyway, this is unchanged behavior and you can see that this same sort of scheme also applies to `NiFakeReal64AttributeValuesMapped` where it just populates a `NIFAKE_REAL64_MAPPED_UNSPECIFIED = 0` and doesn't grab the name from `FLOAT_ENUM_UNSPECIFIED`.

astarche (approved):

PR #760, opened 2022-10-07
[Draft] Testing hapigen output for ni-scope

reckenro:

Using it to test the output from hapigen for scope. It's in progress in [this PR](TBD) in Azure.

It shouldn't. That's why it's a draft request. Just using for diffs and running the build.

Manually checking metadata updates and generated file updates.
System tests still pass.


generated/niscope/niscope.proto
```diff
@@ -979,10 +966,7 @@ message CheckAttributeViInt64Request {
   nidevice_grpc.Session vi = 1;
   string channel_list = 2;
   NiScopeAttribute attribute_id = 3;
-  oneof value_enum {
-    int64 value_raw = 4;
-    NiScopeInt64AttributeValues value = 5;
-  }
+  int64 value_raw = 4;
```
reckenro, line 969: Note: For this change, Harsha saw it in an earlier version of importing ni-scope [here](https://github.com/ni/grpc-device/pull/742#discussion_r980734958) and thought this was okay:

"Looking at the P2P_NOTIFY_MESSAGE_PUSH_ADDR attribute documentation, this specifies to which peer address a message should be pushed. It is unlikely someone would have used an enum for this & most likely they would have set a raw value. So, although it is a breaking change the impact may not be big, but rather beneficial? If someone was using the enum by mistake (which was incorrect), they will now be forced to use the raw value. Which seems the right thing to do.
(please correct me if i'm missing something)"

source/codegen/metadata/niscope/enums.py
```diff
@@ -18,276 +19,163 @@
...
             {
-                'documentation': {
-                    'description': 'The oscilloscope is configured to emit a single CableSense pulse.'
-                },
                 'name': 'NISCOPE_VAL_CABLE_SENSE_MODE_ON_DEMAND',
                 'value': 1
             }
         ]
     },
-    'CalibrationTypes': {
```
dmondrik, line 274: Why is `CalibrationTypes` missing?

source/codegen/metadata/niscope/enums.py
```diff
@@ -1427,20 +1257,6 @@
             }
         ]
     },
-    'NiScopeInt64AttributeValues': {
```
dmondrik, line 1430: Why is `NiScopeInt64AttributeValues` missing?

generated/niscope/niscope_library.cpp
```diff
@@ -580,7 +580,7 @@ ViStatus NiScopeLibrary::Disable(ViSession vi)
 #endif
 }
 
-ViStatus NiScopeLibrary::ErrorHandler(ViSession vi, ViStatus errorCode, ViChar errorSource[642], ViChar errorDescription[642])
+ViStatus NiScopeLibrary::ErrorHandler(ViSession vi, ViStatus errorCode, ViChar errorSource[55], ViChar errorDescription[642])
```
dmondrik, line 583: Neither of these buffer sizes make sense to me. Character arrays of a fixed size are usually 256 or something like that. Thoughts?

source/codegen/metadata/niscope/enums.py
reckenro: It was only referenced by `CalFetchDate` which was marked 'codegen_method': 'no'. So it wasn't referenced by anything else in the .proto file and shouldn't have been used for any arbitrary functions.

source/codegen/metadata/niscope/enums.py
reckenro: So there was only one Int64 attribute referencing an enum previously which was `P2P_NOTIFY_MESSAGE_PUSH_ADDR`. There was a conversation about it in Dane's original Draft PR and Harsha agreed that that attribute shouldn't have been marked with that enum as it didn't make sense as a possible input for that attribute.

So that 'enum' tag was removed from that attribute which results in that enum's values being removed from this `Int64AttributeValues`. Since it was the only enum here the whole Int64AttributeValues dissappears because it'd be empty and it wouldn't be referenced by any Request/Response messages anymore.

dmondrik:
Superseded by [my SCOPE PR](https://github.com/ni/grpc-device/pull/762).

reckenro:
Abandoning in Favor of Dan's PR using hapigen-built export for NI-SCOPE: [Dan's SCOPE PR](https://github.com/ni/grpc-device/pull/762).

PR #761, opened 2022-10-07
Support nested custom types in grpc-device

reckenro:

Supports nested custom_types in grpc-device.
[Corresponding AzDo PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/374154)

- Adds support for nested custom types in grpc-device.
- Updates ni-fake metadata and niFake.h / niFakeObsolete.h from recent 23.0.x export

Needed for a future update to DC Power which will pull in nested custom_types used in a few functions.

Unit and integration tests pass against ni-fake.
Added new unit test to test roundtrip function that has a nested custom_type input and output.


dmondrik (approved):

source/codegen/templates/proto_helpers.mako
```diff
@@ -118,7 +118,7 @@ message ${common_helpers.snake_to_pascal(function)}Response {
 message ${custom_type["grpc_name"]} {
 % for field in common_helpers.filter_parameters_for_grpc_fields(custom_type["fields"]):
 <%
-  field_type = field.get("enum", None) or common_helpers.get_grpc_type(field["type"], config)
+  field_type = field.get("enum", None) or field.get("grpc_type", None) or common_helpers.get_grpc_type(field["type"], config)
```
astarche, line 121: This is a pretty weird line. Is the issue here that `"grpc_type`" is set on all function params as a metadata mutation (or in source), but there's not an equivalent for custom type fields? So we have semi-reasonable code for accessing param types as `param["grpc_type"]` but can't do the same for fields here?

Would it be reasonable to change `metadata_mutation` to do the same for custom type fields?

If not, this is fine as a one off. It could make sense to add something to `common_helpers`, but there are already a lot of `get_grpc_whatever` methods in there. So I think that would make it worse.

astarche (approved):

source/codegen/metadata/niscope/config.py
```diff
@@ -21,41 +21,49 @@
             'fields': [
                 {
                     'grpc_name': 'absolute_initial_x',
+                    'grpc_type': 'double',
```
reckenro, line 24: Note: Ran the format_mi_metadata.py script on NI-Scope again to populate these. It was the only MI driver that had custom_types and hasn't yet been imported from a hapigen export.

PR #744, opened 2022-09-19
Updated from latest NI-DCPower exports

reckenro:
### What does this Pull Request accomplish?

- Updates NI-DCPower metadata/proto to come from hapigen export: `nidcpower_grpc_device 23.0.0d114`
- Updates NI-DCPower header from hapigen export: `nidcpower 23.0.0d114`
- Updates NI-DCPower `nidcpower.lib` from `nidcpower 23.0.0d114`

### Breaking changes
- Correction for wrong input on `ErrorQuery`: updated so `ErrorQueryRequest` doesn't have `error_message` field but instead the `ErrorQueryResponse` has it
- `ExportAttributeConfigurationBuffer` and `ImportAttributeConfigurationBuffer` `configuration` fields updated from `repeated fixed64` to `bytes`. This correction matches the header files and matches the similar function signatures in `nidmm.proto` and `niscope.proto`.

### Why should this Pull Request be merged?

Covers the NI-DCPower part of the following task:

[AB#2172471](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2172471) : Import 23.0.x hapigen exports into grpc-device (nidcpower, niscope)

### What testing has been done?

Built locally and NI-DCPower system tests still pass.

PR #762, opened 2022-10-11
Updated from latest NI-SCOPE exports

dmondrik:

Updates metadata using actual hapigen exports:
`niscope_grpc_device 23.0.0d42`

### Breaking changes

- Removed `enum CalibrationTypes` which was not used anywhere. So this shouldn't break any client.
- Removed `enum NiScopeInt64AttributeValues` which was only exposed in `CheckAttributeViInt64Request` and `SetAttributeViInt64Request` but was not needed. Instead of that enum, you should use raw values for those methods.

[Task 2172471](https://ni.visualstudio.com/DevCentral/_workitems/edit/2172471): Import 23.0.x hapigen exports into grpc-device (nidcpower, niscope)

Built locally just fine.


source/codegen/common_helpers.py
```diff
@@ -204,12 +204,14 @@ def get_input_and_output_custom_types(config, functions):
 
 def _get_nested_types(config, type_name):
     custom_types = set()
-    matching_custom_type = [
+    matching_custom_types = [
```
reckenro, line 207: Note: This change in `common_helpers.py` is more-so a fixup for the DC Power update. You can see the comment for these changes in that PR [here](https://github.com/ni/grpc-device/pull/744#discussion_r991607701).

That being said, I think it's good to have these changes in this PR as well to make sure the update isn't adversely affecting NI-SCOPE's generation with its new metadata.

reckenro (approved):

astarche (approved):

PR #764, opened 2022-10-11
Allow client to specify attach or initialize behavior

dmondrik:

This would be step one of several. We'll need to add the desired initialization behavior as an input/request parameter in each API. And then it will pass that value to `add_session`.

The purpose is to return an error if the user expects a specific session state and that guarantee cannot be satisfied.

It also changes the way in which exceptions internal to the server are handled. Now, the exception hierarchy allows more flexibility and control at the point at which the exception is generated, rather than in each method's catch block.

- Provides infrastructure by which session initialization can request specifically whether to create or attach.
- Simplifies exception handling in each method. The exception now contains the gRPC status object to be returned.

Built locally.


source/server/session_repository.cpp
```diff
@@ -27,12 +34,24 @@ int SessionRepository::add_session(
     const std::string& session_name,
     InitFunc init_func,
     CleanupSessionFunc cleanup_func,
-    uint32_t& session_id)
+    uint32_t& session_id,
+    SessionInitializationBehavior initializationBehavior,
+    bool* createdNewSession)
```
marcoskirsch, line 39: I like the term "create", but I hesitate to introduce it when it comes to sessions.

IVI (or at least NI's use of IVI) has the term "initialize" to mean a session was created. Never do the APIs say "create" nor "open".

That's why I think it's useful to stick with two verbes: "attach" and "initialize" rather than overload "initialize" to mean "sometimes create sometimes attach".

marcoskirsch: I am aware that this applies to our MI and RF APIs but DAQmx uses different terminology and I bet other drivers do as well.

dmondrik: Here are all the methods that call the session manager's `add_session` method:
```
nidaqmx: CreateTask, CreateWatchdogTimer, CreateWatchdogTimerTask/Ex, LoadTask
nidcpower: InitializeWithChannels, InitializeWithIndependentChannels
nidigitalpattern: Init, InitWithOptions
nidmm: Init, InitWithOptions
nifgen: Init, InitWithOptions, InitializeWithChannels
nirfmxbluetooth, nirfmxinstr, nirfmxlte, nirfmxnr, nirfmxspecan, nirfmxwlan: Initialize, InitializeFromNIRFSASession
nirfsa: Init, InitWithOptions
nirfsg: Init, InitWithOptions
niscope: Init, InitWithOptions
niswitch: Init, InitWithOptions, InitWithTopology
nisync: Init, InitExtCal
nitclk: N/A
nixnet: CreateSession, CreateSessionByRef, DbOpenDatabase, SystemOpen
nixnetsocket: Accept, IpStackCreate, IpStackOpen, Socket
```

source/protobuf/session.proto
```diff
@@ -27,6 +27,12 @@ service SessionUtilities {
   rpc ResetServer(ResetServerRequest) returns (ResetServerResponse);
 }
 
+enum SessionInitializationBehavior {
+  SESSION_INITIALIZATION_BEHAVIOR_ATTACH_OR_CREATE = 0;
```
bkeryan, line 31: Does grpc-device follow https://developers.google.com/protocol-buffers/docs/style?

> The zero value enum should have the suffix UNSPECIFIED.

source/server/session_repository.cpp
```diff
@@ -2,6 +2,13 @@
 
 #include <grpcpp/grpcpp.h>
 
+#if !defined(NI_SESSION_REQUESTED_NEW_SESSION_BUT_ONE_ALREADY_EXISTS_ERROR)
```
bkeryan, line 5: These are returned as driver error codes, right? I think it would make sense to throw an exception that is translated to gRPC status code `NOT_FOUND` or `ALREADY_EXISTS`, similar to https://github.com/ni/grpc-device/pull/687 . Returning a driver error code is more appropriate for driver errors.

image
image

astarche (approved):

astarche:
I think it's a good feature and the names work for me.

source/server/session_repository.cpp
dmondrik: I hope it was obvious that -1 and -2 were placeholders. That's why I put [RFC] in the title. I guess I could've made it a draft, but to me that says "I'm still working on this, don't comment yet" which is the opposite of what I wanted.

I would almost say yes to `ALREADY_EXISTS` but I can't agree to `NOT_FOUND` since we return that when the library's method pointer is NULL, which would usually indicate a newer gRPC-device server with an older driver runtime.

source/protobuf/session.proto
dmondrik: That would make sense. This was intended to be a conversation starter. I prefer being explicit about what unspecified means or allows, but sticking to convention is useful too. Looking at our enum values in various proto files, it seems if 0 is nonsensical then we include it with the suffix `UNSPECIFIED` but if it has a meaningful value then we use a name with meaning. @reckenro thoughts on this one? I'm leaning to sticking with the current name.

source/server/session_repository.cpp
dmondrik: What about `FAILED_PRECONDITION` instead of `NOT_FOUND`? I could almost argue that would be a meaningful status code for both cases. But would that be stretching it?

*The operation was rejected because the system is not in a state required for the operation's execution. For example, the directory to be deleted is non-empty, an rmdir operation is applied to a non-directory, etc. Service implementors can use the following guidelines to decide between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher level (e.g., when a client-specified test-and-set fails, indicating the client should restart a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an "rmdir" fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless the files are deleted from the directory.*

bkeryan: > I hope it was obvious that -1 and -2 were placeholders.

Yes. What is it a placeholder for? An infrastructure error like `niapalerr_resourceNotFound`? Are you planning to add an error code range for grpc-device specific errors? 

Anyway, this code used to return +1. See https://github.com/ni/grpc-device/issues/685 and https://github.com/ni/grpc-device/pull/687 . That was before you changed grpc-device to return errors as UNKNOWN, so the example programs would ask the driver to look up the error message for +1.

> I would almost say yes to ALREADY_EXISTS but I can't agree to NOT_FOUND since we return that when the library's method pointer is NULL, which would usually indicate a newer gRPC-device server with an older driver runtime.

The server returns NOT_FOUND when the operation is UNIMPLEMENTED by the driver? 😅

> What about FAILED_PRECONDITION instead of NOT_FOUND? I could almost argue that would be a meaningful status code for both cases. But would that be stretching it?

FAILED_PRECONDITION is plausible but ALREADY_EXISTS and NOT_FOUND seem clearer to me. The session/task is an entity that either already exists or was not found.

Note that gRPC status codes also include a developer-facing error message which can explain what was not found.

source/server/session_repository.cpp
dmondrik: I've changed it to consistently use "initialize" now since we have to pick one, and that term is used more often than not, at least in what we support in gRPC-device today.

source/protobuf/session.proto
dmondrik: I've gone with `UNSPECIFIED` since it matches precedent better. My comment about adding it to the Init structures is something we'll likely do in a follow-on PR since it will affect many proto files.

source/server/session_repository.cpp
dmondrik: Given our existing use of NOT_FOUND, I chose FAILED_PRECONDITION and ALREADY_EXISTS for this PR. I'm not opposed to those changing in the future, but it will be easier to make that change now that the gRPC status code is tied to what is thrown rather than what is caught.

source/server/shared_library.h
```diff
@@ -16,16 +18,18 @@ typedef HMODULE LibraryHandle;
 typedef void* LibraryHandle;
 #endif
 
-struct LibraryLoadException : std::runtime_error {
```
dmondrik, line 19: Note to reviewers: since `std::runtime_error` is a class, it feels really wrong to me for a struct to inherit from a class.

source/server/exceptions.h
```diff
@@ -1,14 +1,29 @@
...
+
 #include <stdexcept>
 
 namespace nidevice_grpc {
 
-struct ValueOutOfRangeException : std::runtime_error {
-  ValueOutOfRangeException(const char* message) : std::runtime_error(message) {}
-  ValueOutOfRangeException(const std::string& message) : std::runtime_error(message) {}
-  ValueOutOfRangeException(const ValueOutOfRangeException& other) : std::runtime_error(other) {}
+class NonDriverException : std::runtime_error {
```
bkeryan, line 10: Changing `struct` to `class` also changed from public inheritance to private inheritance. Can `catch` blocks match against private base classes? https://en.cppreference.com/w/cpp/language/try_catch seems to say they only match against public base classes.

dmondrik: And of course that was the reason for build failures on Unix (not sure why it didn't have a problem on Windows, but oh well):
```
/home/runner/work/grpc-device/grpc-device/source/tests/unit/session_resource_repository_tests.cpp:280:58: error: ‘virtual const char* std::runtime_error::what() const’ is inaccessible within this context
           EXPECT_STREQ(expected_message.c_str(), ex.what());
```

bkeryan (approved):

reckenro (approved):

reckenro:
@dmondrik , FYI the run_win_system_tests step for all CI runs is currently just stuck in the queue forever. It's a known issue and being worked on. I believe the system tests running on ubuntu should provide enough coverage though: the run_ubuntu_system_tests step ran and passed. If you agree, feel free to Squash and merge without waiting for the run_win_system_tests issue to be resolved and ran for these changes.

PR #765, opened 2022-10-11
Trigger AzDo CI after all other steps have completed

reckenro:

Puts the `trigger_azdo_ci` step at the very end of the CI workflow.

Before, it was triggerring the AzDo CI before the other steps necessarily finished. This proved problematic recently because the system_tests were stuck and the AzDo CI was triggered and pulled an older commit that had a finished CI as the basis for it's CI which was incorrect.

CI Workflow before:

image

CI Workflow now:

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,9 +41,11 @@ jobs:
     uses: ./.github/workflows/create_release.yml
   trigger_azdo_ci:
     needs:
-     - build_desktop
+     - run_win_system_tests
+     - run_ubuntu_system_tests
      - build_nilrt
      - create_client_artifacts
+     - create_release
     uses: ./.github/workflows/trigger_azdo_ci.yml
     secrets:
       AZDO_PIPELINE_TRIGGERS: ${{secrets.AZDO_PIPELINE_TRIGGERS}}


SparkingSpork (approved):
Approving because it gets the job done, but have we considered having the AzDO CI in a [separate workflow that's triggered](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) after the GitHub CI? That way you don't keep having to add additional needs fields whenever we add new independent steps.

Maybe not worth it depending on how often that actually changes.

astarche (approved):

reckenro:
> Approving because it gets the job done, but have we considered having the AzDO CI in a [separate workflow that's triggered](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) after the GitHub CI? That way you don't keep having to add additional needs fields whenever we add new independent steps.
> 
> Maybe not worth it depending on how often that actually changes.

That looks good, I don't think we've considered that. I think I want to get this sub-par fix in for now, though, so we start getting more logical AzDo CI's running to avoid confusion. But I'll look into separating it out like what you linked in my fork to test it out and probably post a separate PR to fix it up that way soon. Thanks!

astarche:
> Approving because it gets the job done, but have we considered having the AzDO CI in a [separate workflow that's triggered](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) after the GitHub CI? That way you don't keep having to add additional needs fields whenever we add new independent steps.

> Maybe not worth it depending on how often that actually changes.

I think that's worth considering. We actually had multiple workflows originally and I switched to this as part of a refactor I did. The single workflow is nice for managing granular tasks with natural dependencies. The features I specifically wanted were: it puts all of the artifacts in one place and gives you a single "required workflow" to keep track of for PRs. It also has a few UI advantages like those nice graphs that Ryan screenshotted above.

For a step like trigger which (a) depends on "everything", (b) does not produce artifacts, and (c) is not part of a PR requirement: Most of those benefits don't apply. And the `needs` everything problem is definitely a real issue (probably what broke this). So a workflow might be better in this case. I wonder a bit if this step *does* have a natural dependency (`create_release`) and we just missed it when updating. Also, is it possible to make the dependency more direct? If we're signalling the existence of the artifact, can we check the artifact (and include the path?) as part of the signal?

PR #767, opened 2022-10-19
Uptake changes in NI-FGEN metadata to support init and attach calls.

tdunkle:

This pull request updates the source and generated files for NI-FGEN to those from the d39 build in order to reflect [the changes made to support init and attach calls in NI gRPC Device Server](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/379186).

Fixes [AB#2186880](https://ni.visualstudio.com/DevCentral/_workitems/edit/2186880)

Local and PR build.

### DIFF
--- a/generated/nifgen/nifgen_service.h
+++ b/generated/nifgen/nifgen_service.h
@@ -184,4 +184,9 @@ private:
 
 } // namespace nifgen_grpc
 
+namespace nidevice_grpc {
+namespace converters {
+} // namespace converters
+} // namespace nidevice_grpc
+
 #endif  // NIFGEN_GRPC_SERVICE_H
--- a/source/codegen/metadata/nifgen/attributes.py
+++ b/source/codegen/metadata/nifgen/attributes.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FGEN API metadata version 23.0.0d13
+# This file is generated from NI-FGEN API metadata version 23.0.0d36
 attributes = {
     1050002: {
         'codegen_method': 'public',
--- a/source/codegen/metadata/nifgen/attributes.py
+++ b/source/codegen/metadata/nifgen/attributes.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FGEN API metadata version 23.0.0d36
+# This file is generated from NI-FGEN API metadata version 23.0.0d39
 attributes = {
     1050002: {
         'codegen_method': 'public',
--- a/generated/nifgen/nifgen.proto
+++ b/generated/nifgen/nifgen.proto
@@ -1703,7 +1703,7 @@ message InitRequest {
   string resource_name = 2;
   bool id_query = 3;
   bool reset_device = 4;
-  nidevice_grpc.SessionInitializationBehavior requested_behavior = 5;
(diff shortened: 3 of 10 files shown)


dmondrik:
await: The inputs are fine, but the outputs (generated code) doesn't build. Need to wait for Ryan's PR first, which is to fix the gRPC code generator. At that point it might be as simple as re-queueing the prebuild.

dmondrik (changes requested):
await: The inputs are fine, but the outputs (generated code) doesn't build. Need to wait for Ryan's PR first, which is to fix the gRPC code generator. At that point it might be as simple as re-queueing the prebuild.

reckenro:
> await: The inputs are fine, but the outputs (generated code) doesn't build. Need to wait for Ryan's PR first, which is to fix the gRPC code generator. At that point it might be as simple as re-queueing the prebuild.

I think you'll have to merge main and rebuild locally after my changes go in https://github.com/ni/grpc-device/pull/768 to get you're generated files correct. The build will complain that the generated files in your branch don't match what the PR build generates. Other than that, though, it should be good to go.

source/codegen/metadata/nifgen/attributes.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FGEN API metadata version 23.0.0d13
+# This file is generated from NI-FGEN API metadata version 23.0.0d36
```
dmondrik, line 2: Please do another update with the latest FGEN export from this morning.

reckenro: Yeah, should be d39. We changed some parameter names.

reckenro: Also, after you make the update to the imported metadata/proto make sure to build locally to get updated files in the generated folder. Otherwise, the PR build will fail.

dmondrik (approved):

source/codegen/metadata/nifgen/attributes.py
dmondrik: It would be nice to mention the version you ended up using somewhere in the PR description.

generated/nifgen/nifgen_service.h
```diff
@@ -184,4 +184,9 @@ class NiFgenService final : public NiFgen::Service {
 
 } // namespace nifgen_grpc
 
+namespace nidevice_grpc {
```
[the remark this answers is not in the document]
reckenro: Ah, yeah. So didn't hit this diff with ni-fake because it was already generating this section for some other custom_type conversions already. It's considering the new enum param (I bet because it's coming from nidevice_grpc package) as a custom type so in this determination [here](https://github.com/ni/grpc-device/blob/ff2901836e4c46c37f34b51f599fb2ac18b4d4a2/source/codegen/templates/service.h.mako#L128) it thinks it should produce this section because it will need to have custom conversion code.

I made a tech debt to look into and fix this: [AB#2192495](https://dev.azure.com/ni/DevCentral/_workitems/edit/2192495)

reckenro (approved):
Looks good. You'll just need to rebuild locally and push up to get the generated/nifgen/nifgen.proto in sync with what's in source/codegen/metadata/nifgen

astarche (approved):

PR #768, opened 2022-10-19
Update ni-fake imported metadata and support new initialization behavior for init methods.

reckenro:

- Updates ni-fake from the `nifake_grpc_device` export 23.0.0d45
- Adds codegen support for the new `requestedBehavior` input and `newSessionInitialized` output that will start showing up on init methods (starting with ni-fake).

Fixes [AB#2186876](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2186876)

These changes add support to handle new parameters on `init_method`s. Specifically handling a new input specifying the initialization behavior and an output boolean to signify if a session was initialized or not.

Existing ni-fake unit and integration tests pass.

Added unit tests for new session_repository behavior.
Added unit tests for success and error cases in ni-fake with the new initialize behavior.

### DIFF
--- a/generated/nifake/nifake.proto
+++ b/generated/nifake/nifake.proto
@@ -31,6 +31,7 @@ service NiFake {
   rpc DoubleAllTheNums(DoubleAllTheNumsRequest) returns (DoubleAllTheNumsResponse);
   rpc EnumArrayOutputFunction(EnumArrayOutputFunctionRequest) returns (EnumArrayOutputFunctionResponse);
   rpc EnumInputFunctionWithDefaults(EnumInputFunctionWithDefaultsRequest) returns (EnumInputFunctionWithDefaultsResponse);
+  rpc ErrorMessage(ErrorMessageRequest) returns (ErrorMessageResponse);
   rpc ExportAttributeConfigurationBuffer(ExportAttributeConfigurationBufferRequest) returns (ExportAttributeConfigurationBufferResponse);
   rpc FetchWaveform(FetchWaveformRequest) returns (FetchWaveformResponse);
   rpc GetABoolean(GetABooleanRequest) returns (GetABooleanResponse);
@@ -58,6 +59,7 @@ service NiFake {
   rpc GetCustomType(GetCustomTypeRequest) returns (GetCustomTypeResponse);
   rpc GetCustomTypeArray(GetCustomTypeArrayRequest) returns (GetCustomTypeArrayResponse);
   rpc GetEnumValue(GetEnumValueRequest) returns (GetEnumValueResponse);
+  rpc GetError(GetErrorRequest) returns (GetErrorResponse);
   rpc GetViInt32Array(GetViInt32ArrayRequest) returns (GetViInt32ArrayResponse);
   rpc GetViUInt32Array(GetViUInt32ArrayRequest) returns (GetViUInt32ArrayResponse);
   rpc GetViUInt8(GetViUInt8Request) returns (GetViUInt8Response);
@@ -71,6 +73,7 @@ service NiFake {
   rpc MethodWithGetLastErrorParam(MethodWithGetLastErrorParamRequest) returns (MethodWithGetLastErrorParamResponse);
   rpc MethodWithGrpcFieldNumber(MethodWithGrpcFieldNumberRequest) returns (MethodWithGrpcFieldNumberResponse);
   rpc MethodWithGrpcOnlyParam(MethodWithGrpcOnlyParamRequest) returns (MethodWithGrpcOnlyParamResponse);
+  rpc MethodWithProtoOnlyParameter(MethodWithProtoOnlyParameterRequest) returns (MethodWithProtoOnlyParameterResponse);
   rpc MultipleArrayTypes(MultipleArrayTypesRequest) returns (MultipleArrayTypesResponse);
   rpc MultipleArraysSameSize(MultipleArraysSameSizeRequest) returns (MultipleArraysSameSizeResponse);
   rpc MultipleArraysSameSizeWithOptional(MultipleArraysSameSizeWithOptionalRequest) returns (MultipleArraysSameSizeWithOptionalResponse);
@@ -85,6 +88,11 @@ service NiFake {
   rpc ReturnDurationInSeconds(ReturnDurationInSecondsRequest) returns (ReturnDurationInSecondsResponse);
   rpc ReturnListOfDurationsInSeconds(ReturnListOfDurationsInSecondsRequest) returns (ReturnListOfDurationsInSecondsResponse);
   rpc ReturnMultipleTypes(ReturnMultipleTypesRequest) returns (ReturnMultipleTypesResponse);
+  rpc SetAttributeViBoolean(SetAttributeViBooleanRequest) returns (SetAttributeViBooleanResponse);
+  rpc SetAttributeViInt32(SetAttributeViInt32Request) returns (SetAttributeViInt32Response);
+  rpc SetAttributeViInt64(SetAttributeViInt64Request) returns (SetAttributeViInt64Response);
+  rpc SetAttributeViReal64(SetAttributeViReal64Request) returns (SetAttributeViReal64Response);
+  rpc SetAttributeViString(SetAttributeViStringRequest) returns (SetAttributeViStringResponse);
   rpc SetCustomType(SetCustomTypeRequest) returns (SetCustomTypeResponse);
   rpc SetCustomTypeArray(SetCustomTypeArrayRequest) returns (SetCustomTypeArrayResponse);
   rpc StringValuedEnumInputFunctionWithDefaults(StringValuedEnumInputFunctionWithDefaultsRequest) returns (StringValuedEnumInputFunctionWithDefaultsResponse);
@@ -360,6 +368,16 @@ message EnumInputFunctionWithDefaultsResponse {
   int32 status = 1;
 }
 
+message ErrorMessageRequest {
+  nidevice_grpc.Session vi = 1;
+  sint32 error_code = 2;
+}
+
+message ErrorMessageResponse {
+  int32 status = 1;
+  string error_message = 2;
+}
+
 message ExportAttributeConfigurationBufferRequest {
   nidevice_grpc.Session vi = 1;
 }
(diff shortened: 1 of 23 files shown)


source/codegen/templates/service_helpers.mako
```diff
@@ -8,17 +8,28 @@
...
+  add_session_snippet = 'add_session(grpc_device_session_name, init_lambda, cleanup_lambda, session_id)'
+  if session_behavior_param and session_initialized_param:
+      session_behavior_param_name = common_helpers.get_cpp_local_name(session_behavior_param)
+      session_initialized_param_name = common_helpers.get_cpp_local_name(session_initialized_param)
+      add_session_snippet = f'add_session(grpc_device_session_name, init_lambda, cleanup_lambda, session_id, {session_behavior_param_name}, &{session_initialized_param_name})'
 %>\
 ${initialize_input_params(function_name, parameters)}
 ${initialize_output_params(output_parameters_to_initialize)}\
+% if session_initialized_param:
+      bool ${session_initialized_param_name};
```
dmondrik, line 31: I get the feeling this should be initialized to false. I can imagine that you'll get a warning with some compiler that it's possible to be used uninitialized otherwise. It can't know for sure that the call to `add_session` will always write to it.

source/server/session_resource_repository.h
```diff
@@ -133,7 +139,13 @@ int SessionResourceRepository<TResourceHandle>::add_dependent_session(
   // cleanup code will be executed when the initiating session is closed.
   // We're only responsible for cleaning up the map structures.
   auto status = add_session(
-      session_name, init_func, [](uint32_t id) {}, session_id, /* include_in_reverse_lookup = */ false);
+      session_name,
+      init_func,
+      [](uint32_t id) {},
+      session_id,
+      /* initialization_behavior = */ SESSION_INITIALIZATION_BEHAVIOR_UNSPECIFIED,
+      /* initialized_new_session = */ nullptr,
```
dmondrik, line 147: Much further down there's a local member `added_new_handle_` which seems to be tracking exactly the same thing. Should getting rid of that duplication be a separate PR?

source/codegen/common_helpers.py
```diff
@@ -228,6 +228,11 @@ def is_string_arg(parameter):
     return parameter["grpc_type"] in ["string", "bytes"]
 
 
+def is_nidevice_enum_param(grpc_type):
```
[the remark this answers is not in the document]
reckenro: Thanks, that sounds better for consistency. Updated.

source/codegen/templates/service_helpers.mako
reckenro: Yeah, you're probably right. Updated so it initializes the bool to false.

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -190,6 +253,7 @@ TEST(NiFakeServiceTests, NiFakeService_InitWithOptionsFails_NoSessionIsStored)
   EXPECT_EQ(message, status.error_message());
   nidevice_grpc::Session session = response.vi();
   EXPECT_EQ(0, session_repository.access_session(session.id(), ""));
+  EXPECT_FALSE(response.new_session_initialized());
```
dmondrik, line 256: I don't think this is a valid test when the gRPC status is other than OK.

source/server/session_resource_repository.h
reckenro: Oh, you're right. I forgot we already had that bookkeeping for another reason. I can make a follow up PR to try to just return that instead.

source/tests/unit/ni_fake_service_tests.cpp
dmondrik: I recommend adding an assertion that the field is true in the test method below `NiFakeService_InitWithOptionsAndResetServer_SessionIsClosed` which does initialize a new session. Or possibly choose a different one.

dmondrik (approved):
Thanks for adding the low-level session repository unit tests. I really should've had those in the previous PR. 

The new 'fake' unit tests make sense but need a little work to make sure you're testing the output response fields only when the gRPC status is OK.

source/server/session_resource_repository.h
dmondrik: Agreed, it would only pollute this PR. That cleanup would be tech debt, and this one is needed ASAP for its functionality.

source/tests/unit/ni_fake_service_tests.cpp
reckenro: Ah, good catch. Yeah, that was a bogus assertion just testing the proto message field's default value essentially.

I updated to remove this assert. I also added some more test variations to cover the success cases for the new options.

So now I assert this output is True in
`NiFakeService_InitWithOptionsSucceeds_CreatesAndStoresSession` and `NiFakeService_InitWithOptionsWithInitializeBehavior_CreatesAndStoresSession`
and assert it's False in `NiFakeServiceWithSession_InitWithOptionsWithAttachBehavior_ReturnsAlreadyInitializedSession`

generated/nifake/nifake.proto
```diff
@@ -687,11 +715,13 @@ message InitWithOptionsRequest {
   bool id_query = 3;
   bool reset_device = 4;
   string option_string = 5;
+  nidevice_grpc.SessionInitializationBehavior requested_behavior = 6;
```
reckenro, line 718: Note: This update to InitWithOptions for the Request and Response is the interesting bit and the new functionality tested in this PR. The other added functions were pulling in recent hapigen updates to change those methods from 'private' to 'public' for nimi-python purposes.

source/server/session_resource_repository.h
reckenro: Sounds good. Added this AB#2191089 tech debt to tackle it separately and assigned it to myself. I just put it in the tech debt bucket but can move it elsewhere if needed.

source/codegen/templates/service_helpers.mako
```diff
@@ -8,17 +8,28 @@
...
+  add_session_snippet = 'add_session(grpc_device_session_name, init_lambda, cleanup_lambda, session_id)'
+  if session_behavior_param and session_initialized_param:
+      session_behavior_param_name = common_helpers.get_cpp_local_name(session_behavior_param)
+      session_initialized_param_name = common_helpers.get_cpp_local_name(session_initialized_param)
+      add_session_snippet = f'add_session(grpc_device_session_name, init_lambda, cleanup_lambda, session_id, {session_behavior_param_name}, &{session_initialized_param_name})'
 %>\
 ${initialize_input_params(function_name, parameters)}
 ${initialize_output_params(output_parameters_to_initialize)}\
+% if session_initialized_param:
+      bool ${session_initialized_param_name} = false;
```
astarche, line 31: This is needed because `session_initialized_param_name` is not in `output_parameters_to_initialize`? Because `is_proto_only_parameter`? If there's a way to tweak the logic so that this goes in `output_parameters_to_initialize` (and perhaps tweak some names so the distinctions are clearer): I think that'd be better.

astarche (approved):
You can track follow up in separate tasks if needed to unblock other work.

source/server/session_resource_repository.h
```diff
@@ -29,6 +29,8 @@ class SessionResourceRepository {
       InitFunc init_func,
       CleanupSessionFunc cleanup_func,
       uint32_t& session_id,
+      SessionInitializationBehavior initialization_behavior = SESSION_INITIALIZATION_BEHAVIOR_UNSPECIFIED,
```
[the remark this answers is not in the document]
reckenro: Yes, but the only caller that passed in `include_in_reverse_lookup` was in `add_dependent_session` in this file. For that reason, I updated that call and changed the ordering to keep the `include_in_reverse_lookup` arg at the end so that all the generated calls don't have to specify it. They can just pass in the two new arguments and leave off that one to get the default `true` for it.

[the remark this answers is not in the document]
dmondrik: Also, this isn't code exposed to external clients either - it's just internal to the server implementation.

source/codegen/templates/service_helpers.mako
reckenro: It was in `output_parameters_to_initialize`, it was just that the `initialize_output_params` was filtering out proto_only params. I updated the `initialize_output_params` function to instead filter using a new method in common_helpers (get_params_needing_initialization) so that this new output param gets initialized the standard way through `initialize_output_params` now.

source/codegen/metadata/nifake/functions.py
```diff
@@ -1559,6 +1559,22 @@
                 'grpc_type': 'nidevice_grpc.Session',
                 'name': 'vi',
                 'type': 'ViSession'
+            },
+            {
+                'cppName': 'requestedBehavior',
```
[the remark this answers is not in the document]
reckenro: Yeah, that sounds reasonable. I think the names were updated in hapigen for the MI drivers to be `requestedBehavior`. Those updates haven't been brought over here yet, though. In any case, I made [AB#2191115](https://dev.azure.com/ni/DevCentral/_workitems/edit/2191115) to update them all to be `initializationBehavior`. I think I'll still update ni-fake in a separate PR after those hapigen updates go in, though. To get the bulk of this handling of the behavior in now.

source/codegen/templates/service_helpers.mako
```diff
@@ -8,17 +8,28 @@
 <%
   config = data['config']
   output_parameters = [p for p in parameters if common_helpers.is_output_parameter(p)]
-  session_output_param = next((parameter for parameter in output_parameters if parameter['grpc_type'] == 'nidevice_grpc.Session'), None)
+  input_parameters = [p for p in parameters if common_helpers.is_input_parameter(p)]
+  session_output_param = next((p for p in output_parameters if p['grpc_type'] == 'nidevice_grpc.Session'), None)
+  session_behavior_param = next((p for p in input_parameters if p['grpc_type'] == 'nidevice_grpc.SessionInitializationBehavior'), None)
+  session_initialized_param = next((p for p in output_parameters if common_helpers.is_proto_only_parameter(p) and p['type'] == 'bool'), None)
```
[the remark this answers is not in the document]
reckenro: True, for the time being it's the only `proto_only` bool output for init methods but there's no reason that will hold over time. Similar to the other one, I made a task [AB#219117](https://dev.azure.com/ni/DevCentral/_workitems/edit/2191117) to add some tag for these outputs instead to check like `'sessionInitializationIndicator: True`

PR #772, opened 2022-10-24
Restricted debug session property access

cumitche:

Adds the ability to set and query debug session properties from a custom restricted proto.
Adds server reset observer so services outside of the device enumerator can be notified of a server reset.

Adds the ability for NI applications that access these SysCfg properties for local devices to do so for remote devices.

Tested querying and setting each property.
Tested that methods return UNIMPLEMENTED if the feature toggle is not enabled.


source/protobuf/debugsessionproperties_restricted.proto
```diff
@@ -0,0 +1,59 @@
+syntax = "proto3";
```
astarche, line 1: Can you check how/if this is included in client drops?

I believe that we wanted the `restricted` protos to NOT be in the main client artifacts. But possibly wanted to create a separate `restricted` client zip artifact. But I'm not sure where that's at or what the mechanism is. And, I expect it would be different for a built in service.

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,244 @@
+#include "debug_session_properties_restricted_service.h"
```
astarche, line 1: I'd expect there to be some system tests for this if that's doable.

astarche:
Looking good at a high level. I'll dig more into the implementation tomorrow.

source/protobuf/debugsessionproperties_restricted.proto
cumitche: @jfriedri-ni is working on creating the restricted client artifact.  Currently, it would appear in the public client artifact.  How do you propose we decorate it to show that it should not be included?  Should I move it to a different directory, protobuf_restricted?  Is there some way we can associate it with the code readiness Restricted value?  Other ideas?

source/protobuf/debugsessionproperties_restricted.proto
```diff
@@ -0,0 +1,59 @@
...
+  rpc IsDebugSessionEnabled(IsDebugSessionEnabledRequest) returns (IsDebugSessionEnabledResponse);
+  rpc IsDebugSessionServerOutOfProcess(IsDebugSessionServerOutOfProcessRequest) returns (IsDebugSessionServerOutOfProcessResponse);
+  rpc IsDebugSessionSupported(IsDebugSessionSupportedRequest) returns (IsDebugSessionSupportedResponse);
+  rpc SetDebugSessionEnabled(SetDebugSessionEnabledRequest) returns (SetDebugSessionEnabledResponse);
+  rpc SetDebugSessionServerOutOfProcess(SetDebugSessionServerOutOfProcessRequest) returns (SetDebugSessionServerOutOfProcessResponse);
+}
+
+message IsDebugSessionEnabledRequest {
+  uint32 product_id = 1;
+  string serial_number = 2;
```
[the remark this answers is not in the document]
jonlugoatx: FYI @dmondrik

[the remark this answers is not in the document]
dmondrik: Agreed, adding the device name will make it much more robust. Simulated hardware can really throw what you'd normally expect right out the window.

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,244 @@
...
+  try {
+    auto library = get_syscfg_library_interface();
+    if (NISysCfg_Succeeded(status = open_or_get_localhost_syscfg_session(&session))) {
+      if (NISysCfg_Succeeded(status = library->CreateFilter(session, &filter))) {
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyIsDevice, NISysCfgBoolTrue);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyProductId, product_id);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertySerialNumber, serial_number);
+        if (NISysCfg_Succeeded(status = library->FindHardware(session, NISysCfgFilterModeMatchValuesAll, filter, NULL, &resources_handle))) {
+          if (NISysCfg_Succeeded(status) && (status = library->NextResource(session, resources_handle, &resource)) == NISysCfg_OK) {
+            library->GetResourceProperty(resource, property_id, &syscfg_value);
```
jonlugoatx, line 84: Consider factoring out a helper that takes in a lambda so you don't have to duplicate the filtered query and library handle cleanup.

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  try {
+    auto library = get_syscfg_library_interface();
+    if (NISysCfg_Succeeded(status = open_or_get_localhost_syscfg_session(&session))) {
+      if (NISysCfg_Succeeded(status = library->CreateFilter(session, &filter))) {
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyIsDevice, NISysCfgBoolTrue);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyProductId, product_id);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertySerialNumber, serial_number);
+        if (NISysCfg_Succeeded(status = library->FindHardware(session, NISysCfgFilterModeMatchValuesAll, filter, NULL, &resources_handle))) {
+          if (NISysCfg_Succeeded(status) && (status = library->NextResource(session, resources_handle, &resource)) == NISysCfg_OK) {
```
jonlugoatx, line 83: Would it make sense to check the filter for more than one match? If there are two matching devices, how should this behave?

source/server/debug_session_properties_restricted_service.h
```diff
@@ -0,0 +1,62 @@
...
+#include <debugsessionproperties_restricted.grpc.pb.h>
+
+#include "feature_toggles.h"
+#include "shared_library.h"
+#include "syscfg_library_interface.h"
+#include "syscfg_session_handler.h"
+
+namespace nidevice_restricted_grpc {
+
+static const char* kDebugSessionPropertyAccessFailedMessage = "The NI System Configuration API was unable to access the debug session property.";
```
jonlugoatx, line 16: This is only used in the `debug_session_properties_restricted_service.cpp`. Would it make more sense to define this there?

source/server/session_utilities_service.cpp
```diff
@@ -50,8 +52,37 @@ ::grpc::Status SessionUtilitiesService::ResetServer(::grpc::ServerContext* conte
   }
   bool is_server_reset = session_repository_->reset_server();
   response->set_is_server_reset(is_server_reset);
-  device_enumerator_->clear_syscfg_session();
+
+  {
+    std::unique_lock<std::shared_mutex> lock(observers_mutex_);
+    auto i = observers_.begin();
+    auto e = observers_.end();
+    for (; i != e; ++i)
```
jonlugoatx, line 60: Could you use `for (auto observer : observers_)` here?

jonlugoatx (approved):

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,244 @@
...
+  NISysCfgResourceHandle resource = NULL;
+  NISysCfgBool syscfg_value = NISysCfgBoolFalse;
+
+  try {
+    auto library = get_syscfg_library_interface();
+    if (NISysCfg_Succeeded(status = open_or_get_localhost_syscfg_session(&session))) {
+      if (NISysCfg_Succeeded(status = library->CreateFilter(session, &filter))) {
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyIsDevice, NISysCfgBoolTrue);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyProductId, product_id);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertySerialNumber, serial_number);
```
dmondrik, line 81: Since you say "match all" below, I suggest only including this in the filter if a non-empty serial number string was actually passed.

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,244 @@
...
+        library->CloseHandle(filter);
+      }
+    }
+  }
+  catch (nidevice_grpc::NonDriverException& ex) {
+    return ex.GetStatus();
+  }
+
+  if (NISysCfg_Failed(status)) {
+    return ::grpc::Status(::grpc::StatusCode::INTERNAL, kDebugSessionPropertyAccessFailedMessage);
```
dmondrik, line 99: This is not how we do this now. It should follow the convention of:
- Returning UNKNOWN status and the real already-looked-up error message in the Status object.
- Putting the actual status code in the trailer

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,244 @@
...
+      if (NISysCfg_Succeeded(status = library->CreateFilter(session, &filter))) {
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyIsDevice, NISysCfgBoolTrue);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertyProductId, product_id);
+        library->SetFilterProperty(filter, NISysCfgFilterPropertySerialNumber, serial_number);
+        if (NISysCfg_Succeeded(status = library->FindHardware(session, NISysCfgFilterModeMatchValuesAll, filter, NULL, &resources_handle))) {
+          if (NISysCfg_Succeeded(status) && (status = library->NextResource(session, resources_handle, &resource)) == NISysCfg_OK) {
+            library->SetResourceProperty(resource, property_id, syscfg_value);
+            NISysCfgBool changes_require_restart = NISysCfgBoolFalse;
+            char* detailed_changes = nullptr;
+            library->SaveResourceChanges(resource, &changes_require_restart, &detailed_changes);
```
dmondrik, line 171: You need to check the status here and return it if saving fails. Also, on failure, `detailed_changes` should be returned as the error message.

source/server/debug_session_properties_restricted_service.cpp
cumitche: Wouldn't including the empty serial number in the filter help ensure that I'm getting the simulated device I expect?  We would only pass an empty string here if that is what we received from calling EnumerateDevices.

dmondrik: I suppose it would be fine.

SparkingSpork (approved):

source/protobuf/debugsessionproperties_restricted.proto
astarche: That sounds good. I like moving it to a `restricted` sub-directory, or even filtering on `*_restricted.proto`. Associating it with a `CodeReadiness` directly sounds more complicated and isn't necessary for this change. I don't think we need to support the `NextRelease` or `RestrictedNextRelease` granularity.

reckenro (approved):

astarche (approved):

source/protobuf/debugsessionproperties_restricted.proto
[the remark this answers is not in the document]
cumitche: Created DeviceId message and added a string name field in addition to these fields.

source/server/debug_session_properties_restricted_service.cpp
cumitche: After talking with you, I asked Harsha about this.  He said that returning a static error message with the INTERNAL status for SysCfg operations was an intentional decision since the user is not attempting to interact with SysCfg but just perform the operation.  Providing detailed SysCfg errors in this context is not useful.  While this is a restricted proto file, I think these methods fulfill a similar function and should follow the same convention.

source/server/debug_session_properties_restricted_service.cpp
cumitche: Assigned status to the result of SaveResourceChanges().  I did not retain the detailed_changes for the error message due to the reasons I outlined in the above conversation.

source/server/debug_session_properties_restricted_service.cpp
cumitche: There is no way for a user to enter a state where multiple device resources will be returned for a specific set of user alias, serial number, and product id through normal use.  I do not think it is worth adding code here to handle this very unlikely scenario.

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,176 @@
...
+}
+
+::grpc::Status DebugSessionPropertiesRestrictedService::get_bool_property(
+  ::grpc::ServerContext* context,
+  NISysCfgResourceProperty property_id,
+  const nidevice_restricted_grpc::DeviceId& device_id,
+  bool* value)
+{
+  auto get_bool_lambda = [&] (nidevice_grpc::SysCfgLibraryInterface* library, NISysCfgResourceHandle resource, bool* save_changes) {
+    *save_changes = false;
```
dmondrik, line 127: Since it defaults to false, resetting it here seems superfluous.

source/server/debug_session_properties_restricted_service.cpp
```diff
@@ -0,0 +1,176 @@
...
+        library->CloseHandle(filter);
+      }
+    }
+  }
+  catch (nidevice_grpc::NonDriverException& ex) {
+    return ex.GetStatus();
+  }
+
+  if (NISysCfg_Failed(status)) {
+    return ::grpc::Status(::grpc::StatusCode::INTERNAL, kDebugSessionPropertyAccessFailedMessage);
```
dmondrik, line 114: The others use status code `UNKNOWN`. Use [ApiErrorAndDescriptionToStatus](https://github.com/ni/grpc-device/blob/main/source/server/converters.h#L325) instead of creating this yourself.

dmondrik (approved):

source/server/debug_session_properties_restricted_service.cpp
cumitche: Updated as suggested while leaving the static message in place.

source/server/debug_session_properties_restricted_service.cpp
cumitche: Removed unnecessary assignment.

source/server/debug_session_properties_restricted_service.cpp
cumitche: Refactored as suggested.

source/server/debug_session_properties_restricted_service.h
cumitche: This is used by the unit test, so I am leaving it in place.

source/server/session_utilities_service.cpp
cumitche: Updated as suggested.

source/server/debug_session_properties_restricted_service.cpp
cumitche: After further discussion, I've updated as suggested.

PR #773, opened 2022-10-26
Add enumerate_software functionality

Wedgetable:

Similar to `enumerate_devices`, this uses the NI System Configuration API to find NI software on the server and enables clients to get a list of these packages.

I'm still pretty new to this codebase, so I based my implementation heavily on that of the existing `enumerate_devices` functionality. Feel free to point out methods that could/should be shared, or other places that are too copy/pasta-ish.

Enables applications to check for certain packages and versions on the remote server - drivers especially - in order to know what functionality is available.

* Created unit tests for the new `SoftwareEnumerator` class and ran them
* Added integration and system tests similar to those for `DeviceEnumerator` and ran them
* Created `enumerate-software` example and used it to get a list of NI packages off a remote machine running `ni_grpc_device_server`


examples/session/enumerate-software.py
```diff
@@ -0,0 +1,78 @@
+r"""List out NI software installed on the server machine.
```
astarche, line 1: It'd be great to add this to the list of examples that are covered with validate_examples on the workflow here: https://github.com/ni/grpc-device/blob/d5d281adace5590683685c9cadea9d78046b3446/.github/workflows/validate_python.yml#L34

Note you can also use that script locally to check types and run the example (OR ... see why it's failing to install 😢 )

source/server/core_service_registrar.cpp
```diff
@@ -1,6 +1,7 @@
 #include "core_service_registrar.h"
```
astarche, line 1: Watch for conflicts with #772

source/server/syscfg_library.cpp
```diff
@@ -142,4 +145,42 @@ NISysCfgStatus SysCfgLibrary::GetResourceProperty(
...
+    char* id,
+    char* version,
+    char* title,
+    NISysCfgComponentType* itemType,
+    char** detailedDescription
+  )
+  {
+    if (!function_pointers_.NextComponentInfo) {
+    throw LibraryLoadException(kSysCfgApiNotInstalledMessage);
+  }
```
astarche, line 182: Formatting is off here. Do you have the [clang auto-formatter set up](https://github.com/ni/grpc-device/wiki/Development-Tips-and-Tricks#clang-format-on-save)?

source/server/software_enumerator.cpp
```diff
@@ -0,0 +1,92 @@
...
+        
+        library_->CloseHandle(installedComps);
+      }
+    }
+  }
+  catch (nidevice_grpc::NonDriverException& ex) {
+    return ex.GetStatus();
+  }
+
+  if (NISysCfg_Failed(status)) {
```
astarche, line 53: Dan had a comment on this in Curtis's PR: https://github.com/ni/grpc-device/pull/772#discussion_r1004745616

source/server/syscfg_library.h
```diff
@@ -87,6 +102,11 @@ class SysCfgLibrary : public SysCfgLibraryInterface {
       NISysCfgFilterHandle filter_handle,
       const char* expert_names,
       NISysCfgEnumResourceHandle* resource_enum_handle);
+  using GetInstalledSoftwareComponentsPtr = NISysCfgStatus (*)(
```
astarche, line 105: No Change Required.

I think these could all use the `using FunctionPtr = decltype(&Function)` pattern that the generated code uses.

astarche (approved):

reckenro (approved):

SparkingSpork (approved):

source/server/core_service_registrar.cpp
Wedgetable: I talked with Curtis yesterday, and there are a number of things he's adding in his PR that will help improve the code in my PR, so I'm going to let his go in first.

source/protobuf/session.proto
```diff
@@ -12,6 +12,10 @@ service SessionUtilities {
   rpc EnumerateDevices(EnumerateDevicesRequest)
       returns (EnumerateDevicesResponse);
 
+  // Provides a list of software installed on server under localhost
+  rpc EnumerateSoftware(EnumerateSoftwareRequest)
```
dmondrik, line 16: I would suggest calling it `EnumerateInstalledSoftware`. As we move forward with extending the API, there are cases where we want to consider available software that is not yet installed, and we have 'available' APIs for that.

source/protobuf/session.proto
```diff
@@ -54,6 +58,18 @@ message EnumerateDevicesResponse {
   repeated DeviceProperties devices = 1;
 }
 
+message SoftwareProperties {
+  string package_id = 1;
+  string version = 2;
```
dmondrik, line 63: To be more future-proof, I'd suggest calling this `package_version`. The reason is that we're likely to add "display version" in the near future, you know, the 22Q4 kind of stuff. This field needs to clearly be the dotted version that you can pass to normal version comparison routines.

source/server/software_enumerator.cpp
```diff
@@ -0,0 +1,58 @@
...
+  NISysCfgEnumSoftwareComponentHandle installedComps = NULL;
+  unsigned int numInstalledComps = 0;
+  char package_id[NISYSCFG_SIMPLE_STRING_LENGTH] = "";
+  char version[NISYSCFG_SIMPLE_STRING_LENGTH] = "";
+  char product_name[NISYSCFG_SIMPLE_STRING_LENGTH] = "";
+
+  try {
+    auto library = get_syscfg_library_interface();
+    if (NISysCfg_Succeeded(status = open_or_get_localhost_syscfg_session(&session))) {
+      if (NISysCfg_Succeeded(status = library->GetInstalledSoftwareComponents(session, NISysCfgIncludeItemsAllVisibleAndHidden, NISysCfgBoolFalse, &installedComps))) {
```
dmondrik, line 32: Consider: It might be a good idea to have some kind of input/request argument as to whether you want `NISysCfgIncludeItemsAllVisibleAndHidden` or just `NISysCfgIncludeItemsAllVisible`. The latter feels like it would be useful in more places.

source/server/software_enumerator.cpp
```diff
@@ -46,7 +48,8 @@ ::grpc::Status SoftwareEnumerator::enumerate_software(google::protobuf::Repeated
   }
 
   if (NISysCfg_Failed(status)) {
-    return ::grpc::Status(::grpc::StatusCode::INTERNAL, kSoftwareEnumerationFailedMessage);
+    std::string description(kSoftwareEnumerationFailedMessage);
+    return nidevice_grpc::ApiErrorAndDescriptionToStatus(context, status, description);
```
dmondrik, line 52: Thanks for using this pattern!

dmondrik (approved):

source/protobuf/session.proto
Wedgetable: Updated, thanks for the suggestion!

source/protobuf/session.proto
```diff
@@ -54,6 +62,24 @@ message EnumerateDevicesResponse {
...
+  string product_name = 3;
+}
+
+message EnumerateInstalledSoftwareRequest {}
+
+message EnumerateInstalledSoftwareResponse {
+  repeated SoftwareProperties software = 1;
+}
+
+message EnumerateInstalledSoftwareIncludeHiddenPackagesRequest {}
```
dmondrik, line 77: It feels a little weird to make this 2 separate methods instead of passing an argument. 
@harsha-bhushan thoughts on how this should be exposed?

source/server/software_enumerator.cpp
Wedgetable: Sounds good. I added a call for `EnumerateInstalledSoftwareIncludeHiddenPackages` and just used a boolean for now. Feel free to unresolve if something about my implementation seems off.

dmondrik: This part is OK, but the way it's exposed in the .proto is what concerns me. I added a comment there to see what others think.

source/protobuf/session.proto
Wedgetable: This is mostly just me being a newb to proto files. I agree that an additional argument would be better - let me revisit this.

Wedgetable: I changed the Request to take an argument (and tested everything out again). Let me know if this looks better.

harsha-bhushan: The single method with argument looks good.

harsha-bhushan:
We could avoid having a separate example for this function. I recommend combining it with device enumeration example & rename it appropriately. That way, going ahead, when we add more sys config functions, we can just put them in the same example. 
When you do that, you need not accept the 'hidden package argument' for the example itself, you can just pass it in the function as True.

harsha-bhushan (approved):

dmondrik:
> When you do that, you need not accept the 'hidden package argument' for the example itself, you can just pass it in the function as True

Actually for the example, I'd either not specify it, or make it False. Most users would care only about the visible packages.

examples/session/nisyscfg.py
```diff
@@ -0,0 +1,158 @@
...
+To run this example, install NI System Configuration on the server machine:
+  https://www.ni.com/en-in/support/downloads/drivers/download.system-configuration.html
+
+For instructions on how to use protoc to generate gRPC client interfaces, see our "Creating a gRPC
+Client" wiki page:
+  https://github.com/ni/grpc-device/wiki/Creating-a-gRPC-Client
+
+Running from the command line:
+
+usage: nisyscfg.py [-h]
```
dmondrik, line 14: I'm concerned about calling the example `nisyscfg`. It's not exposing the API. I'd rather keep distinct super-simple examples for enumerating devices and for enumerating software.

Wedgetable: > I'd rather keep distinct super-simple examples for enumerating devices and for enumerating software.

This conflicts with what @harsha-bhushan suggested above. I can either 
* Roll back to the two examples as I had previously, or 
* Rename the merged example here to not reference NISysCfg. I'm open to suggestions regarding the name. Maybe it could be called `server-resources` or something. Or 
* _\<some other suggestion here\>_

dmondrik: I'm OK with a single example that doesn't claim to *be* nisyscfg. What about `enumerate-hardware-software.py`?

Wedgetable: That's fair. Though from what Harsha said (and others I've heard), it sounds like we may add more functions that call nisyscfg, so for the name, I was just going for something generic. I'm also using subcommands to determine what function is called/information is printed: in my current changes, users can do something like `python nisyscfg.py enumerate-devices` and it just prints out the list of devices. So `python enumerate-hardware-software.py enumerate-devices` feels a little redundant to me. More ideas for the name:  `session-utilities.py`, `server-info.py`

I'm not keen on making the example print out **both** a list of devices and software packages, in case that was being asked, because the list of software especially can get _very_ long (even leaving out hidden packages). But if subcommands aren't desirable either, I'm okay with going back to separate example modules.

dmondrik: I don't have a strong opinion. Given the 2 names you suggested, I prefer `session-utilities.py`.

PR #774, opened 2022-10-26
Use custom implementation of FetchCaptureWaveformU32

DavidCurtiss:

Fixes a failure (and sometimes crash) in the function. [AB#2193950](https://ni.visualstudio.com/DevCentral/_workitems/edit/2193950)

Note that the test assets (and the conceptual test steps) were copied from one of the affected nimi-python tests.

Crashes are bad, m'kay?

Added test. Without the fix, it fails with data_size()==2 instead of 4.
Verified that the nimi-python test for it no longer crashes ni_grpc_device_server.


source/custom/nidigitalpattern_service.custom.cpp
```diff
@@ -17,4 +26,48 @@ ::grpc::Status NiDigitalService::ConvertApiErrorStatusForViSession(::grpc::Serve
   return nidevice_grpc::ApiErrorAndDescriptionToStatus(context, status, description);
 }
 
+::grpc::Status NiDigitalService::FetchCaptureWaveformU32(::grpc::ServerContext* context, const FetchCaptureWaveformU32Request* request, FetchCaptureWaveformU32Response* response)
+{
+  // This function is modified from the generated code to multiple the two 'ivi-dance' parameters
```
astarche, line 31: nit: multiple -> multiply

source/tests/system/nidigital_driver_api_tests.cpp
```diff
@@ -79,7 +113,23 @@ class NiDigitalDriverApiTest : public ::testing::Test {
...
     ::grpc::Status status = GetStub()->Close(&context, request, &response);
 
-    EXPECT_TRUE(status.ok());
+    EXPECT_SUCCESS(status);
+    expect_api_success(response.status());
+  }
+
+  void close_multi_instrument_driver_session()
+  {
+    if (!multi_instrument_driver_session_) {
```
astarche, line 122: Can this case actually happen?

astarche (approved):

source/tests/system/nidigital_driver_api_tests.cpp
DavidCurtiss: I'm sure it's just as likely as the equivalent check in `close_driver_session()`, from which I copied this function. Can it actually happen, though? Probably, yeah, if the initialization functions fail.

source/custom/nidigitalpattern_service.custom.cpp
DavidCurtiss: Fixed

reckenro (approved):

CMakeLists.txt
```diff
@@ -586,6 +586,11 @@ if(WIN32)
...
     set(SystemTestsRunner CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /DELAYLOAD:nicaiu.dll /DELAYLOAD:nidcpower_64.dll /DELAYLOAD:niDigital_64.dll /DELAYLOAD:nidmm_64.dll /DELAYLOAD:niFgen_64.dll /DELAYLOAD:nimxlcTerminalAdaptor.dll /DELAYLOAD:niRFmxBT.dll /DELAYLOAD:niRFmxInstr.dll /DELAYLOAD:niRFmxLTE.dll /DELAYLOAD:niRFmxNR.dll /DELAYLOAD:niRFmxSpecAn.dll /DELAYLOAD:niRFmxWLAN.dll /DELAYLOAD:niRFSA_64.dll /DELAYLOAD:niRFSG_64.dll /DELAYLOAD:niScope_64.dll /DELAYLOAD:niswitch_64.dll /DELAYLOAD:nisync.dll /DELAYLOAD:niTClk_64.dll /DELAYLOAD:nixnet.dll /DELAYLOAD:nixntipstack.dll")
 endif()
+if(_GRPC_DEVICE_NILRT_LEGACY_TOOLCHAIN)
+    target_link_libraries(SystemTestsRunner stdc++fs)
+    set(SystemTestsRunner CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DFS_EXPERIMENTAL")
```
DavidCurtiss, line 591: @astarche @reckenro The NILRT build doesn't seem to be passing `-DFS_EXPERIMENTAL` to the compiler -- the build of nidigital_driver_api_tests.cpp is erroring in the `#else` case of `#ifdef FS_EXPERIMENTAL`. Any idea what I'm doing wrong here?

SparkingSpork, line 591: This is not the proper way to set the `CMAKE_CXX_FLAGS` for a single target. The `set` command is of the format `set(<variable> <value>...)`. If you want to add flags for a specific compile target only, use `target_compile_definitions` instead. As per the documentation [here](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html#command:target_compile_definitions), that will add the definitions only to the target without touching the flags used for everything else. There are similar methods for adding options and includes.

SparkingSpork (changes requested):

CMakeLists.txt
DavidCurtiss: That worked, thanks!

DavidCurtiss: SparkingSpork got me straightened out. Fixed.

SparkingSpork (approved):

PR #777, opened 2022-11-01
Trigger AzDo CIs workflow after main GitHub CI is completely finished.

reckenro:

Triggers the AzDo CIs after the main GitHub CI completes.

[AB#2197310](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2197310)

1. Having it trigger as the last step of the main GitHub CI workflow is a little brittle if new steps are added.
2. This is better timing wise to make sure the GitHub CI workflow is completely finished before triggering the AzDo CIs, which depends on the finished CI workflow to get its artifacts.

It should also be merged because the PR # is lucky 🎰

Using PR build to test new workflow ordering/triggering.

### DIFF
--- a/.github/workflows/trigger_azdo_ci.yml
+++ b/.github/workflows/trigger_azdo_ci.yml
@@ -5,9 +5,6 @@ on:
     workflows: [CI]
     types:
       - completed
-    secrets:
-      AZDO_PIPELINE_TRIGGERS:
-        required: true
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/trigger_azdo_ci.yml
+++ b/.github/workflows/trigger_azdo_ci.yml
@@ -1,7 +1,10 @@
 name: Trigger AzDO CIs
 
 on:
-  workflow_call:
+  workflow_run:
+    workflows: [CI]
+    types:
+      - completed
     secrets:
       AZDO_PIPELINE_TRIGGERS:
         required: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,13 +39,3 @@ jobs:
       - build_nilrt
       - create_client_artifacts
     uses: ./.github/workflows/create_release.yml
-  trigger_azdo_ci:
-    needs:
-     - run_win_system_tests
-     - run_ubuntu_system_tests
-     - build_nilrt
-     - create_client_artifacts
-     - create_release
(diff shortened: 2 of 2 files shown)


.github/workflows/trigger_azdo_ci.yml
```diff
@@ -1,7 +1,10 @@
 name: Trigger AzDO CIs
 
 on:
-  workflow_call:
+  workflow_run:
+    workflows: [CI]
+    types:
+      - completed
     secrets:
```
reckenro, line 8: I'm not entirely sure if it needs this `secrets` section anymore or if that was just needed when it was triggered by a `workflow_call`.

.github/workflows/trigger_azdo_ci.yml
```diff
@@ -1,7 +1,10 @@
 name: Trigger AzDO CIs
 
 on:
-  workflow_call:
+  workflow_run:
```
reckenro, line 4: Note: I think we could also trigger these only on CI success (everything through the system_tests passing). See [these GitHub docs ](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow) for how that would be enabled.

For now, I'm keeping it as it was to always trigger the AzDo CIs upon GitHub CI completion. @csjall , does that sound reasonable?

astarche (approved):

.github/workflows/trigger_azdo_ci.yml
reckenro: I don't think that section was needed/useful. The secrets should be passed/available and the workflow already just references `secrets.AZDO_PIPELINE_TRIGGERS`.

reckenro: I don't think that section was needed/useful. The secrets should be passed/available and the workflow already just references `secrets.AZDO_PIPELINE_TRIGGERS`.

.github/workflows/trigger_azdo_ci.yml
csjall: Agreed. My main concern was with the timing. The adhoc export pipeline simply grab the latest artifacts. As long as we trigger the pipeline on completion of the CI I think it should be fine. We will look to system test result before we decide to release a specific release.

csjall (approved):

PR #778, opened 2022-11-01
Add binding address to server configuration file

csjall:
Currently the NI gRPC Device Server defaults to allow any remote connections "[::]". Add a new server configuration parameter (address) which allows users to restrict specific address or localhost "[::1]".

Currently incoming connection cannot be restricted without changing source code.
[AB#2190918](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2190918)

Add new test for localhost configuration.
Add new test for parsing address property.


reckenro:
Is there a task to mention this in our documentation somewhere? I'd imagine adjacent to [the spot in the README](https://github.com/ni/grpc-device#default-configuration-file-insecure) that mentions the default configuration or a new page linked from the README like the [Server Security Support](https://github.com/ni/grpc-device/wiki/Server-Security-Support) page is which mentions setting up the configuration with certificates.

reckenro (approved):
Looks good. As a side note, you can put AB# and then the AzDo work item number to cross-link GitHub PRs to tasks. So for this if you put AB#2190918 in the PR description it should get linked to the Task.

reckenro:
@SparkingSpork is out this week but I think we should still have @astarche review this as well before merging.

source/server/server_configuration_parser.cpp
```diff
@@ -237,11 +221,62 @@ std::string ServerConfigurationParser::read_keycert(const std::string& filename)
...
 
+std::string ServerConfigurationParser::parse_bind_address() const
+{
+  // Use default address if none is specified
+  std::string parsed_bind_address = kDefaultAddressPrefix;
+
+  auto it = config_file_.find(kAddressJsonKey);
+  if (it != config_file_.end()) {
+    try {
+      parsed_bind_address = it->get<std::string>();
```
jwp2000, line 232: It should probably allow for an address like "::" or "::1" which are valid IP addresses on their own. The caller appends ":" + <port>, which assumes the address returned here already includes the brackets if needed before appending the colon and port to form the full "endpoint address".  I'd be inclined to have the caller actually do this. Add brackets if the address contains a colon but no brackets.

source/server/server_configuration_parser.cpp
```diff
@@ -237,11 +221,62 @@ std::string ServerConfigurationParser::read_keycert(const std::string& filename)
...
+    }
+    catch (const nlohmann::json::type_error& ex) {
+      throw WrongPortTypeException(ex.what());
```
jwp2000, line 235: InvalidAddressException ?

source/config/localhost_config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "address": "[::1]",
+    "port": 0,
```
bkeryan, line 3: Consider keeping port 31763 so this is a localhost-only version of `insecure_config.json`.

source/server/server_configuration_parser.cpp
csjall: Throw WrongAddressTypeException if the bind address is not a string. Added unit tests for use case.

source/config/localhost_config.json
csjall: Added known port to localhost_config.json

source/server/server_configuration_parser.cpp
csjall: For the sake of simplicity we will only check for non-empty string and documented recommended usage for the new "address" property. I have a follow-up task to document syntax and provide some examples.

source/server/server_configuration_parser.h
```diff
@@ -19,7 +21,7 @@ static const char* kInvalidExePathMessage = "The server was unable to resolve th
 static const char* kInvalidMaxMessageSizeMessage = "The max message size must be an integer.";
 static const char* kInvalidFeatureToggleMessage = "Feature Toggles must be specified as boolean fields in the form \"feature_toggles\": { \"feature1\": true, \"feature2\": false }. \n\n";
 static const char* kInvalidCodeReadinessMessage = "code_readiness must be a string in [Release, RestrictedRelease, NextRelease, RestrictedNextRelease, Incomplete, Prototype].\n\n";
-static const char* kDefaultAddressPrefix = "[::]:";
+static const char* kDefaultAddressPrefix = "[::]";
```
bkeryan, line 24: This is not really a prefix anymore, so I think it makes sense to drop "Prefix" from the constant name.

csjall: Renamed to kDefaultAddress

PR #780, opened 2022-11-03
Export artifact that lists provided service interfaces

dmondrik:

Generates the file `server_capabilities.json` adjacent to the server executable. It is included in the exported artifacts.
- Currently the file contains only the provided service interfaces. This can be easily extended in the future.
- Chose to not put this in the same file as the server configuration, because it's intended for a totally different use.

[Task 2206702](https://ni.visualstudio.com/DevCentral/_workitems/edit/2206702): Export information with gRPC server about what proto files are supported

Built locally. It contains these lines:
```
{
    "provided_service_interfaces": [
        "nidaqmx_grpc.NiDAQmx",
        "nidcpower_grpc.NiDCPower",
        "nidigitalpattern_grpc.NiDigital",
        "nidmm_grpc.NiDmm",
        "nifgen_grpc.NiFgen",
        "nirfmxbluetooth_grpc.NiRFmxBluetooth",
        "nirfmxinstr_grpc.NiRFmxInstr",
        "nirfmxlte_grpc.NiRFmxLTE",
        "nirfmxnr_grpc.NiRFmxNR",
        "nirfmxspecan_grpc.NiRFmxSpecAn",
        "nirfmxwlan_grpc.NiRFmxWLAN",
        "nirfsa_grpc.NiRFSA",
        "nirfsg_grpc.NiRFSG",
        "niscope_grpc.NiScope",
        "niswitch_grpc.NiSwitch",
        "nisync_grpc.NiSync",
        "nitclk_grpc.NiTClk",
        "nixnet_grpc.NiXnet",
        "nixnetsocket_grpc.NiXnetSocket"
    ]
}
```

### DIFF
--- a/source/codegen/generate_server_impl_info.py
+++ b/source/codegen/generate_server_impl_info.py
@@ -1,0 +1,43 @@
+"""Service information generation."""
+
+import argparse
+import json
+
+from pathlib import Path
+from template_helpers import load_metadata
+
+
+def _generate_file(metadata_dir: Path, output_dir: Path) -> None:
+    service_instance_names = []
+    driver_modules = [
+        load_metadata(p)
+        for p in sorted(metadata_dir.iterdir())
+        if p.is_dir() and "fake" not in p.name
+    ]
+    for data in driver_modules:
+        config = data["config"]
+        service_name = f'{config["module_name"]}_grpc.{config["service_class_prefix"]}'
+        service_instance_names.append(service_name)
+    service_information_dict = {}
+    service_information_dict["provided_service_interfaces"] = service_instance_names
+    output_file_path = Path(output_dir).joinpath("server_impl_info.json")
+    with open(output_file_path, "w") as outfile:
+        json.dump(service_information_dict, outfile, indent=4)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate NI gRPC-device server implementation information."
+    )
+    parser.add_argument(
+        "metadata",
+        help="The path to the directory containing the metadata for the API being generated.",
+    )
+    parser.add_argument(
+        "--output",
+        "-o",
+        help="The path to the top-level directory to save the generated files. The API-specific sub-directories will be automatically created.",
+    )
+    args = parser.parse_args()
+    output_path = "." if args.output is None else args.output
+    _generate_file(Path(args.metadata), Path(output_path))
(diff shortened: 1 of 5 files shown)


dmondrik:
Although this file contains all interfaces, in theory on RT for example we could limit it to only the APIs that are supported. But I had to start somewhere. Would like feedback on it, though. @reckenro @astarche

DelpireNI (approved):
Looks good to me.

source/codegen/generate_server_impl_info.py
```diff
@@ -0,0 +1,43 @@
...
+from template_helpers import load_metadata
+
+
+def _generate_file(metadata_dir: Path, output_dir: Path) -> None:
+    service_instance_names = []
+    driver_modules = [
+        load_metadata(p)
+        for p in sorted(metadata_dir.iterdir())
+        if p.is_dir() and "fake" not in p.name
+    ]
```
reckenro, line 16: Should we be limiting this like `stage_client_files` does to not include the restricted services? That stage_client_files checks the driver_readiness ('code_readiness' in config.py) through `common_helpers.get_driver_readiness` [here](https://github.com/ni/grpc-device/blob/a4430019dd911c664e3080804d6a6685b980d6db/source/codegen/stage_client_files.py#L24) to exclude any that aren't marked explicitly or implicitly with `'code_readiness': 'Release'`.

bkeryan:
> Although this file contains all interfaces, in theory on RT for example we could limit it to only the APIs that are supported. But I had to start somewhere. Would like feedback on it, though. @reckenro @astarche

FYI, printing a list of interfaces to stdout would be more dynamic, but I don't think we need that. Generating a static file and installing it is adequate for our purposes and allows us to use this list at build time (though I understand your concerns about doing so).

bkeryan:
[bikeshed] Re: `server_impl_info.json`

Is this really "implementation info"? It describes the interface to the server (the package/service names used in its `.proto` files), not the implementation.

Consider something like `services.json` or `server_capabilities.json`.

dmondrik:
> Consider something like `services.json` or `server_capabilities.json`.

Since this is intended to potentially grow over time, 'services' is too limiting. But I do like the other name, `server_capabilities.json` so I'll go with that.

source/codegen/generate_server_impl_info.py
dmondrik: Excellent point, I'll get on it.

reckenro (approved):

PR #781, opened 2022-11-04
Update Trigger AzDo CI to only run after CI completes on main or releases branch.

reckenro:

- Trigger AzDo CI now only triggers after a CI is completed on main or a releases branch
- Trigger AzDo CI looks at the branch the CI completed on to know which AzDo Pipeline to trigger (previously it would always trigger the main one).

Noticed after my recent submission that the Trigger AzDo CI workflow was triggering after any completed CI, not just the ones on main or releases branches.

None. Hard to test without submission.

### DIFF
--- a/.github/workflows/trigger_azdo_ci.yml
+++ b/.github/workflows/trigger_azdo_ci.yml
@@ -5,6 +5,9 @@ on:
     workflows: [CI]
     types:
       - completed
+    branches:
+      - main
+      - 'releases/**'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_azdo_ci.yml
```diff
@@ -14,31 +17,31 @@ jobs:
     steps:
     - name: Trigger Main CI
       uses: Azure/pipelines@v1
-      if: ${{ github.repository == 'ni/grpc-device' && github.ref == 'refs/heads/main' }}
+      if: ${{ github.repository == 'ni/grpc-device' && github.event.workflow_run.head_branch == 'main' }}
```
reckenro, line 20: Updated these because the github.ref for workflow_run triggers will always be the repo's default branch (or main). Using the `github.eventworkflow_run.head_branch` instead will make sure we're triggering the right AzDo pipeline based on where the CI completed (main or a releases branch).

astarche (approved):

PR #784, opened 2022-11-08
Remove id from Session message

ccifra:

* Having both name and id is somewhat confusing
* It makes using the session in interactive tools like bloom RPC difficult, since oneof is a little complex to use.

I don't think ID should ever be used.  There are a lot of advantages to keeping the data on the wire "time invariant" or independent on server state.  For example, we are trying to do batching to improve throughput when doing "cloud-based measurements" as part of various LTIs and hopefully products.

Even if we remove id we can still maintain the current behavior by generating a name if the client does not specify. Clients who use name should not need to make any changes to their code. The wire format does not break for clients who use name, so we don't even need to require clients change. The wire format does not really break for clients that use id since they can't set it and our server will stop setting it. Clients who use id should be treating Session as an opaque structure since there is nothing they can do with name.

General improvement to grpc-device.

Existing test suite updated.

### DIFF
--- a/source/server/session_repository.cpp
+++ b/source/server/session_repository.cpp
@@ -20,7 +20,7 @@ SessionRepository::~SessionRepository()
   reset_server(/* cleanup_external_resources =*/false);
 }
 
-// Either returns (via session_id) an existing session with the specified name or initializes and
+// Either returns (via session_name) an existing session with the specified name or initializes and
 // adds a new session using the init_func provided. The init_func is only called when an existing
 // session with session_name is not found. It is expected to return a tuple, where the first value
 // is a status code, and the second value is the ID of a newly initialized session. The return
--- a/source/server/session_repository.cpp
+++ b/source/server/session_repository.cpp
@@ -26,17 +26,15 @@ SessionRepository::~SessionRepository()
 // is a status code, and the second value is the ID of a newly initialized session. The return
 // value is the status returned from the init_func, or 0 if an existing named session is found.
 int SessionRepository::add_session(
-    const std::string& session_name,
+    std::string& session_name,
     InitFunc init_func,
     CleanupSessionFunc cleanup_func,
-    uint32_t& session_id,
     SessionInitializationBehavior initialization_behavior,
     bool* initialized_new_session)
 {
   if (initialized_new_session) {
     *initialized_new_session = false;
   }
-  session_id = 0;
   std::unique_lock<std::shared_mutex> lock(repository_lock_);
   auto now = std::chrono::steady_clock::now();
   auto it = named_sessions_.find(session_name);
@@ -48,7 +46,6 @@ int SessionRepository::add_session(
   }
 
   if (it != named_sessions_.end()) {
-    session_id = it->second->id;
     it->second->last_access_time = now;
     return 0;
   }
@@ -57,15 +54,11 @@ int SessionRepository::add_session(
   if (status) {
     return status;
   }
-  session_id = next_id();
-  info->id = session_id;
+  session_name = next_id();
   info->cleanup_func = cleanup_func;
   info->last_access_time = now;
-  sessions_.emplace(session_id, info);
-  if (!session_name.empty()) {
-    info->name = session_name;
-    named_sessions_.emplace(session_name, info);
-  }
+  info->name = session_name;
+  named_sessions_.emplace(session_name, info);
   if (initialized_new_session) {
     *initialized_new_session = true;
   }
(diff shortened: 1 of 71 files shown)


ccifra:
This change is not quite ready for review.  I wanted to get the builds running and test suite running to see where things are at.

ccifra:
Ready for review.

source/server/session_repository.cpp
```diff
@@ -76,46 +71,37 @@ int SessionRepository::add_session(
...
-    }
-    sessions_.erase(it);
+  auto named_it = named_sessions_.find(name);
+  if (named_it != named_sessions_.end()) {
+    named_sessions_.erase(named_it);
   }
 }
 
-void SessionRepository::register_dependent_session(uint32_t id, uint32_t dependent_session_id, std::function<void()> cleanup)
+void SessionRepository::register_dependent_session(const std::string& name, const std::string& dependent_session_name, std::function<void()> cleanup)
```
astarche, line 97: I can't see the results from `run_win_system_tests` (I haven't spent much time that action, I'm not sure if that's normal). Could you tell if the `rfmx` tests ran and passed? Those are windows-only and those drivers are the main clients of dependent sessions.

astarche (approved):

source/server/session_repository.cpp
ccifra: I'll try and run them locally.  The RFMX tests unit and integration tests which use dependent sessions do pass. I did break something there, so it seems those tests are good.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -705,7 +705,7 @@ class NiDAQmxDriverApiTests : public Test {
   template <typename TRequest>
   void set_request_session_id(TRequest& request, const nidevice_grpc::Session& session)
   {
-    request.mutable_task()->set_id(session.id());
+    request.mutable_task()->set_name(session.name());
```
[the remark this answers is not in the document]
ccifra: I agree.  I was somewhat disappointed to see how much code decomposes the session instead of passing it around.  In my world the session was an opaque structure for most code.

source/server/session_repository.h
```diff
@@ -24,19 +24,18 @@ class SessionRepository {
...
-      uint32_t& session_id,
       SessionInitializationBehavior initialization_behavior = SESSION_INITIALIZATION_BEHAVIOR_UNSPECIFIED,
       bool* initialized_new_session = nullptr);
-  uint32_t access_session(uint32_t session_id, const std::string& session_name);
-  void remove_session(uint32_t id);
+  std::string access_session(const std::string& session_name);
+  void remove_session(const std::string& name);
 
-  void register_dependent_session(uint32_t id, uint32_t dependent_session_id, std::function<void()> cleanup);
+  void register_dependent_session(const std::string& id, const std::string& dependent_session_id, std::function<void()> cleanup);
```
reckenro, line 38: Nit: Not sure if it functionally matters, but can you match the name updates to what's done in the `session_repository.cpp` file: `id` -> `name` and `dependent_session_id` -> `dependent_session_name`?

reckenro (approved):

reckenro:
> It looks (to me) like all of the `examples` are OK. Either way, there should be a release task to run all the examples, especially with some major changes this release. This is something that `mypy` should catch, so another case for expanding which drivers pass `mypy` so we can `validate_examples` on them on the CI.
> 
> Nothing needed for this PR. @reckenro can you see if there's a good place to track that release task?

Yeah, I'm starting an email thread with our PO (Harsha) for where to track this among other tasks with the next release. I'm guessing it'll become a sustaining item in our backlog but whatever the case I'll capture this along with that Release related work.

source/server/session_repository.cpp
reckenro: Yeah, those tests seem good for that.

I am finding as I'm looking at other PRs that the `run_win_system_tests` either are failing with no description in the action or are now infinitely queuing so something's messed up there, probably with our system. Not sure yet but looking into it.

reckenro: I kicked off the pipeline to re-create the VM the `run_win_system_tests` connects to. After that completes, I'll re-queue that action on this PR to see if it actually runs this time.

source/server/session_repository.h
ccifra: Not sure how I missed these when I did my search. I will update.

ccifra: I removed the usages of id and session_id.

source/server/session_repository.cpp
ccifra: It looks like the system tests are still failing.
I ran what tests I could locally.  many of the RF tests fail for what looks like other reasons. I think maybe we need actual HW to run the tests?

ccifra: System tests finally ran and completed successfully.

PR #790, opened 2022-11-15
Skip converter namespace if no custom types

dmondrik:

In some cases, we were generating an empty namespace at the bottom of the service header. This PR now avoids doing that.

[Technical Debt 2192495](https://ni.visualstudio.com/DevCentral/_workitems/edit/2192495): [gRPC] Fix issue with empty namespace generating at bottom of _service.h

Built locally. Verified that the only generated change was this one for FGEN.

### DIFF
--- a/examples/session/session-reservation.py
+++ b/examples/session/session-reservation.py
@@ -80,7 +80,7 @@ try:
         )
     )
     vi = init_with_options_response.vi
-    print(f"Session initialized with name {session_name} and id {vi.id}.\n")
+    print(f"Session initialized with name {session_name}.\n")
 
     # Check if session is reserved by client 1.
     # Note: The reservation_id is defined by and has meaning only for the client + Session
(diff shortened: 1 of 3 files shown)


examples/session/session-reservation.py
```diff
@@ -80,7 +80,7 @@
         )
     )
     vi = init_with_options_response.vi
-    print(f"Session initialized with name {session_name} and id {vi.id}.\n")
+    print(f"Session initialized with name {session_name}.\n")
```
dmondrik, line 83: I know why this change is needed, given Chris' recent submission. I have no idea how his PR got through without including this. FYI @astarche @reckenro

reckenro: Agreed, this should've failed the `validate_python` step in Chris' PR or at least his merge to main. I created a Tech Debt [AB#2219712](https://dev.azure.com/ni/DevCentral/_workitems/edit/2219712) item to investigate and hopefully figure out / fix what's going on here.

reckenro (approved):

PR #792, opened 2022-11-15
Fix sizing of waveform_info array returned from NI-SCOPE's FetchArrayMeasurement.

reckenro:

Fixes an issue in the custom implementation for NI-SCOPE's FetchArrayMeasurement where it was using the wrong size for the returned waveform_info array.

Fixes [AB#2220088](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2220088)

Found during testing this method in nimi-python against grpc-device that there was something wrong on the grpc-device end. Comparing the nimi-python implementation and the custom implementation here revealed the discrepancy.

This update aligns with nimi-python and the [documentation for the method](https://github.com/ni/grpc-device/wiki/NI-SCOPE-Measurement-Functions#niscope_fetcharraymeasurement)

Updated and ran system test that was asserting the wrong expected array size before for this method.

### DIFF
--- a/source/tests/system/niscope_driver_api_tests.cpp
+++ b/source/tests/system/niscope_driver_api_tests.cpp
@@ -296,7 +296,7 @@ TEST_F(NiScopeDriverApiTest, NiScopeFetchArrayMeasurement_SendRequest_FetchCompl
   EXPECT_TRUE(status.ok());
   expect_api_success(response.status());
   EXPECT_EQ(expected_num_waveforms * expected_waveform_size, response.meas_wfm_size());
-  EXPECT_EQ(expected_waveform_size, response.wfm_info_size());
+  EXPECT_EQ(expected_num_waveforms, response.wfm_info_size());
 }
 
 TEST_F(NiScopeDriverApiTest, NiScopeFetchBinary16_SendRequest_FetchCompletesWithCorrectSizes)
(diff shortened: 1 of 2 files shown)


source/tests/system/niscope_driver_api_tests.cpp
```diff
@@ -296,7 +296,7 @@ TEST_F(NiScopeDriverApiTest, NiScopeFetchArrayMeasurement_SendRequest_FetchCompl
   EXPECT_TRUE(status.ok());
   expect_api_success(response.status());
   EXPECT_EQ(expected_num_waveforms * expected_waveform_size, response.meas_wfm_size());
-  EXPECT_EQ(expected_waveform_size, response.wfm_info_size());
+  EXPECT_EQ(expected_num_waveforms, response.wfm_info_size());
```
dmondrik, line 299: This looks reasonable to me but I would want some scope expert to sign off for this change. @marcoskirsch can you take a look at this?

dmondrik (approved):
I think it's fine as long as others agree.

astarche (approved):

marcoskirsch (approved):

source/tests/system/niscope_driver_api_tests.cpp
marcoskirsch: I _think_ this is correct.

reckenro: Sounds good. I'm pretty confident this is correct looking at the documentation and what nimi-python does. Without this change the corresponding nimi-python test using grpc-device failed on the assert about the wfm_info size and with this update it passes so I'll go ahead and submit this. I think we had it wrong on this side and no one had caught it yet.

PR #795, opened 2022-11-18
Fix example mypy type violations.

### LINKED ISSUE #793: Fix examples referencing Session.id
Some of the examples are still trying to reference the removed `id` field on the Session message. One example is nidcpower's [measure-record.py](https://github.com/ni/grpc-device/blob/main/examples/nidcpower/measure-record.py#L219)

These should be updated to no longer expose the AttributeError that will occur trying to access that removed field.

[AB#2222316](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2222316)

reckenro:

- Fixes mypy violations on all examples except DAQmx
- Primarily focused on avoiding referencing removed `id` field on `Session` message

Fixes #793 

Related AzDo Bug: [AB#2222316](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2222316)

Recently, a change went in to remove `id` from the `Session` message. We don't run validation on all the examples so there were a handful still trying to reference this `id` field.

Ran validate_examples.py against all examples with `--ignore-missing-imports` to avoid any errors related to numpy or matplotlib. All passed with updates except DAQmx which I'm not attempting to fix with this change.

### DIFF
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -29,7 +29,7 @@ jobs:
       with:
         poetry-version: 1.1.5
 
-    - name: Validate RFmx, XNET, and session examples
+    - name: Validate examples (except DAQmx)
       run: |
         python source/codegen/validate_examples.py -p "*dcpower*"
         python source/codegen/validate_examples.py -p "*digital*"
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -31,6 +31,13 @@ jobs:
 
     - name: Validate RFmx, XNET, and session examples
       run: |
-        python source/codegen/validate_examples.py -p "*rfmx*"
+        python source/codegen/validate_examples.py -p "*dcpower*"
+        python source/codegen/validate_examples.py -p "*digital*"
+        python source/codegen/validate_examples.py -p "*dmm*"
+        python source/codegen/validate_examples.py -p "*fgen*"
+        python source/codegen/validate_examples.py -p "*rf*"
+        python source/codegen/validate_examples.py -p "*scope*"
+        python source/codegen/validate_examples.py -p "*switch*"
+        python source/codegen/validate_examples.py -p "*tclk*"
         python source/codegen/validate_examples.py -p "*xnet*"
         python source/codegen/validate_examples.py -p "session"
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -31,13 +31,4 @@ jobs:
 
     - name: Validate examples (except DAQmx)
       run: |
-        python source/codegen/validate_examples.py -p "*dcpower*"
-        python source/codegen/validate_examples.py -p "*digital*"
-        python source/codegen/validate_examples.py -p "*dmm*"
-        python source/codegen/validate_examples.py -p "*fgen*"
-        python source/codegen/validate_examples.py -p "*rf*"
-        python source/codegen/validate_examples.py -p "*scope*"
-        python source/codegen/validate_examples.py -p "*switch*"
-        python source/codegen/validate_examples.py -p "*tclk*"
-        python source/codegen/validate_examples.py -p "*xnet*"
-        python source/codegen/validate_examples.py -p "session"
+        python source/codegen/validate_examples.py -e "nidaqmx"
--- a/examples/nidcpower/measure-record.py
+++ b/examples/nidcpower/measure-record.py
@@ -105,7 +105,7 @@ try:
         nidcpower_types.SetAttributeViInt32Request(
             vi=vi,
             attribute_id=nidcpower_types.NiDCPowerAttribute.NIDCPOWER_ATTRIBUTE_MEASURE_RECORD_LENGTH,
-            attribute_value=RECORD_LENGTH,
+            attribute_value_raw=RECORD_LENGTH,
         )
     )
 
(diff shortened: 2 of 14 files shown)


examples/nidcpower/measure-record.py
```diff
@@ -180,11 +180,11 @@ def _on_close(event):
 
             # Updating the precision of the fetched values.
             y_axis_new = []
-            for value in y_axis:
-                if value < VOLTAGE_LEVEL:
-                    y_axis_new.append(math.floor(value * 100) / 100)
+            for y_value in y_axis:
```
reckenro, line 183: I don't know if this was the best way to address this. But it was complaining on line 208 that we were trying to assign a string to a float variable because `value` was defined here.

examples/nidcpower/measure-record.py
```diff
@@ -105,7 +105,7 @@ def check_for_warning(response, vi):
         nidcpower_types.SetAttributeViInt32Request(
             vi=vi,
             attribute_id=nidcpower_types.NiDCPowerAttribute.NIDCPOWER_ATTRIBUTE_MEASURE_RECORD_LENGTH,
-            attribute_value=RECORD_LENGTH,
+            attribute_value_raw=RECORD_LENGTH,
```
reckenro, line 108: Mypy was complaining about trying to put RECORD_LENGTH which is just the literal value 10 in `attribute_value` which expects and enum.

examples/nidmm/continuous-acquire-graph-multiple-points.py
```diff
@@ -162,7 +162,7 @@ def _on_close(event):
                 fetch_multipoints_response = nidmm_client.FetchMultiPoint(
                     nidmm_types.FetchMultiPointRequest(
                         vi=vi,
-                        maximum_time=-1,
+                        maximum_time_raw=-1,
```
reckenro, line 165: Should use `_raw` when assigning literal values.

examples/nirfsa/getting-started-iq.py
```diff
@@ -64,7 +64,7 @@ def check_for_warning(response, vi):
     """Print to console if the status indicates a warning."""
     if response.status > 0:
         warning_message = client.ErrorMessage(
-            nirfsa_types.ErrorMessageRequest(vi=vi, error_code=response.status)
+            nirfsa_types.ErrorMessageRequest(vi=vi, status_code=response.status)
```
reckenro, line 67: For nirfsa, the `ErrorMessageRequest`'s field name is `status_code` and not `error_code`.

examples/nirfsa/getting-started-spectrum.py
```diff
@@ -68,7 +68,7 @@ def check_for_warning(response, vi):
     """Print to console if the status indicates a warning."""
     if response.status > 0:
         warning_message = client.ErrorMessage(
-            nirfsa_types.ErrorMessageRequest(vi=vi, error_code=response.status)
+            nirfsa_types.ErrorMessageRequest(vi=vi, status_code=response.status)
```
reckenro, line 71: For nirfsa, the `ErrorMessageRequest`'s field name is `status_code` and not `error_code`.

source/codegen/validate_examples.py
```diff
@@ -79,7 +79,7 @@ def _validate_examples(
             _system(f"poetry run black --check --line-length 100 {dir}")
 
             print(f" -> Running mypy")
-            _system(f"poetry run mypy {dir} --check-untyped-defs")
+            _system(f"poetry run mypy {dir} --check-untyped-defs --ignore-missing-imports")
```
reckenro, line 82: I did this temporarily to avoid all the type errors related to `numpy` or `matplotlib`. 

But I wanted to ask, should I leave this in and update the CI to run for all examples except DAQmx? Is there a downside to doing this?

examples/nidcpower/measure-record.py
astarche: Yeah, that's correct. Try this

`try.py`:
```python
def test():
    try:
        while True:
            x = "ok"
            raise IndexError
    except IndexError:
        print(x)
        x = 1
        print(x)


test()
```

Output:
```
(.venv) λ python try.py
ok
1
```

`python` variable scope is not the same as C++/C#/etc and `mypy` is generally trying to follow that or at least being conservative to prevent potential confusion if the typechecker doesn't match the runtime. There's a [`--allow-redefinition`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition) option, but I think it actually excludes this case. I'm not sure why that is, I think because the runtime behavior is particularly surprising. And because it is more often done on accident, as opposed to reusing in the same/equal scope, which is semi-idiomatic (if a little sketchy).

source/codegen/validate_examples.py
astarche: Yeah, I think this is a pretty good tradeoff. The worst case is a typo in some other `import` could cause us to silently skip type checking on a lot of files. Or a mistake in our setup code that causes the protoc types to be missing. So I think it's worth trying to remove this flag, but this change as-is is better than the current state.

astarche (approved):

.github/workflows/validate_python.yml
```diff
@@ -29,8 +29,15 @@ jobs:
...
+        python source/codegen/validate_examples.py -p "*dcpower*"
+        python source/codegen/validate_examples.py -p "*digital*"
+        python source/codegen/validate_examples.py -p "*dmm*"
+        python source/codegen/validate_examples.py -p "*fgen*"
+        python source/codegen/validate_examples.py -p "*rf*"
+        python source/codegen/validate_examples.py -p "*scope*"
+        python source/codegen/validate_examples.py -p "*switch*"
+        python source/codegen/validate_examples.py -p "*tclk*"
         python source/codegen/validate_examples.py -p "*xnet*"
         python source/codegen/validate_examples.py -p "session"
```
SparkingSpork, line 43: Is there a way to do this in a single line? Maybe having a flag to exclude certain things and include everything else?

SparkingSpork (approved):

.github/workflows/validate_python.yml
reckenro: Yeah, this was ugly. Updated to add a default for the `-p` arg to `"*"` so it includes everything by default. Added additional `-e` argument for exclusions so now this is one line `python source/codegen/validate_examples.py -e "nidaqmx"` to run everything except that one.

Issue #693, opened 2022-08-05
grpc-device should convert string encoding to/from UTF-8 when necessary


bkeryan:
Protocol Buffers requires strings to be encoded as UTF-8. However, many of the NI driver APIs supported by grpc-device use the Windows current ANSI code page, which is typically a single-byte encoding like ISO-8859-1 or a multi-byte encoding like Shift JIS.

NI driver APIs typically support internationalization for:
- User-defined names, such as device aliases, DAQmx global channel names, and DAQmx task names
- File paths
- Localized error messages
- Serialized configuration data for import/export

I haven't tested internationalization with grpc-device, but I would expect these results:
- Sending strings that are not structurally valid UTF-8 may result in errors or coerced string contents.
- Sending or receiving strings in languages that natively support Unicode (like C# and Python3) may result in [mojibake](https://en.wikipedia.org/wiki/Mojibake). 

Driver APIs:
- DAQmx: `nicaiu.dll` uses ANSI/MBCS. `nicai_utf8.dll` uses UTF-8, but requires a different import library than the one that is checked into `imports\lib\win64`. 
- MI, RF, XNET I think these all use ANSI/MBCS.
  - Exception: The MI `ImportAttributeConfigurationBuffer` and `ExportAttributeConfigurationBuffer` functions pass an array of bytes containing a UTF-8 JSON string. Related issue: https://github.com/ni/grpc-device/issues/694 

[AB#2108036](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2108036)

dmondrik:
Related, the test `QuotedLocalizedDescription_ApiErrorAndDescriptionToStatus_MessageIsEncoded` is not behaving in a sufficiently predictable way that meets everyone's expectations. While I'm baffled by the nuances of non-ANSI characters in C string constants, @bkeryan and Ben Craig are not convinced the code is doing anything to encode the characters.
- Brad thought the constant `std::string("\"chaîne localisée 😀\"")` in MSVC would issue a code page warning, but I can confirm it does not.
- I tried making the constant explicitly specify `u8` encoding, but Ben pointed out that code would cease to compile when we move to a newer C++ standard.
- I tried keeping a "normal" C string and using explicit Unicode characters `std::string("\"cha\u00EEne localis\u00E9e\U0001F600\"")` but that causes 2 problems:
    - Compiling: `warning C4566: character represented by universal-ch
[...]

PR #796, opened 2022-11-19
Convert between UTF-8 (gRPC) and local codepage (driver)

DavidCurtiss:

Converts to/from UTF-8 when calling drivers.

#693 "grpc-device should convert string encoding to/from UTF-8 when necessary"

Created unit and system tests; ran them on en-US Windows.
Hand-tested creating a nimi-python gRPC session connecting to a simulated device with various fancy names. On default en-US Windows, tested a few Windows-1252 extended characters (¶ë†). Changed the system locale to Japanese, then tested a device name containing a Japanese character (本).


source/server/converters.h
```diff
@@ -113,6 +114,75 @@ CType convert_from_grpc(const GrpcType& value)
...
+#if WIN32
+  int flags = MB_ERR_INVALID_CHARS | (GetACP() == 54936 ? 0 : MB_PRECOMPOSED);
+  int wcs_len = MultiByteToWideChar(CP_ACP, flags, value.c_str(), -1, NULL, 0);
+  if (wcs_len == 0) {
+    throw new std::runtime_error("Unknown character in string");
+  }
+  --wcs_len;  // don't include trailing null
+  std::wstring utf16value(wcs_len, '\0');
+  MultiByteToWideChar(CP_ACP, flags, value.c_str(), -1, &utf16value[0], wcs_len);
+  utf8::utf16to8(utf16value.begin(), utf16value.end(), std::back_inserter(*value_out));
```
bkeryan, line 174: Consider reserving `*value_out`'s capacity. An imperfect estimate would still avoid many unnecessary reallocations, though I guess the `is_ascii` check above also does that.

bkeryan (changes requested):
I think the Import/Export issue is a regression.

generated/nidcpower/nidcpower_service.cpp
```diff
@@ -3170,7 +3283,8 @@ namespace nidcpower_grpc {
       auto vi_grpc_session = request->vi();
       ViSession vi = session_repository_->access_session(vi_grpc_session.name());
       ViInt32 size = static_cast<ViInt32>(request->configuration().size());
-      ViInt8* configuration = (ViInt8*)request->configuration().c_str();
+      auto configuration_mbcs = convert_from_grpc<std::string>(request->configuration());
+      ViInt8* configuration = (ViInt8*)configuration_mbcs.c_str();
       auto status = library_->ImportAttributeConfigurationBuffer(vi, size, configuration);
```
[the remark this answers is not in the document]
DavidCurtiss: Fixed

source/server/converters.h
```diff
@@ -113,6 +114,75 @@ CType convert_from_grpc(const GrpcType& value)
...
+
+template <>
+inline std::string convert_from_grpc(const std::string& value)
+{
+  if (is_ascii(value.c_str())) {
+    return value;
+  }
+#if WIN32
+  std::wstring utf16value;
+  utf8::utf8to16(value.begin(), value.end(), std::back_inserter(utf16value));
```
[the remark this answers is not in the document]
DavidCurtiss: Done

source/server/converters.h
```diff
@@ -113,6 +114,75 @@ CType convert_from_grpc(const GrpcType& value)
...
+  int flags = GetACP() == 54936 ? 0 : 0x400;  // WC_NO_BEST_FIT_CHARS
```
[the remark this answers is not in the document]
DavidCurtiss: Done

source/server/converters.h
DavidCurtiss: Done

SparkingSpork (approved):
CMake changes look good to me.

generated/nidcpower/nidcpower_service.cpp
```diff
@@ -3354,7 +3466,8 @@ namespace nidcpower_grpc {
     try {
       auto vi_grpc_session = request->vi();
       ViSession vi = session_repository_->access_session(vi_grpc_session.name());
-      auto channel_name = request->channel_name().c_str();
+      auto channel_name_mbcs = convert_from_grpc<std::string>(request->channel_name());
```
bkeryan, line 3469: You should also update MeasureMultiple and MeasureMultipleLCR: https://github.com/ni/grpc-device/blob/main/source/custom/nidcpower_service.custom.cpp

In general, you should update the relevant custom entry points: https://github.com/ni/grpc-device/tree/main/source/custom

generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -7024,7 +7277,8 @@ namespace nidaqmx_grpc {
       return ::grpc::Status::CANCELLED;
     }
     try {
-      auto session_name = request->session_name().c_str();
+      auto session_name_mbcs = convert_from_grpc<std::string>(request->session_name());
+      auto session_name = session_name_mbcs.c_str();
 
       auto init_lambda = [&] () {
```
bkeryan, line 7283: Ok, this is correctly passing the MBCS session name to DAQmx. The session repository uses the UTF-8 session name, which seems fine because the session lookup code also uses the UTF-8 session name.

bkeryan (changes requested):
I skimmed the generated code for DAQmx, DCPower, DMM, Fgen, and Scope. I did not look at the rest.

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "third_party/json"]
 	path = third_party/json
 	url = https://github.com/nlohmann/json.git
+[submodule "third_party/utfcpp"]
```
reckenro, line 10: Can you add a section to [ThirdPartyNotices.txt](https://github.com/ni/grpc-device/blob/main/ThirdPartyNotices.txt) for this in a similar fashion to what was added for grpc and json?

reckenro (approved):
Looks good to me. Just need to update the ThirdPartyNotices.txt appropriately.

astarche (approved):

source/codegen/templates/service_helpers.mako
```diff
@@ -565,7 +575,11 @@ ${initialize_standard_input_param(function_name, parameter)}
   c_element_type_that_needs_coercion = service_helpers.get_c_element_type_for_array_that_needs_coercion(parameter)
 %>\
 % if c_type in ['ViConstString', 'const char[]']:
-      auto ${parameter_name} = ${request_snippet}.c_str();\
+      auto ${parameter_name}_mbcs = convert_from_grpc<std::string>(${request_snippet});
+      auto ${parameter_name} = ${parameter_name}_mbcs.c_str();\
+% elif grpc_type == 'string':
+      auto ${parameter_name}_mbcs = convert_from_grpc<std::string>(${request_snippet});
+      ${c_type_pointer} ${parameter_name} = (${c_type_pointer})${parameter_name}_mbcs.c_str();\
```
dmondrik, line 582: Above on line 463 in a similar case you're doing `const_cast` - should these be consistent? It would seem more semantically correct than a C-style cast.

source/server/converters.h
```diff
@@ -113,6 +114,79 @@ CType convert_from_grpc(const GrpcType& value)
...
+template <>
+inline std::string convert_from_grpc(const std::string& value)
+{
+  if (is_ascii(value.c_str())) {
+    return value;
+  }
+#if WIN32
+  std::wstring utf16value;
+  utf16value.reserve(value.length());  // often too much, but that's okay
+  utf8::utf8to16(value.begin(), value.end(), std::back_inserter(utf16value));
```
dmondrik, line 136: Consider: This could be done on Windows using `MultiByteToWideChar(CP_UTF8` since you have to use the Windows methods anyway. It still takes 2 conversions either way. Just a thought.

source/server/converters.h
```diff
@@ -113,6 +114,79 @@ CType convert_from_grpc(const GrpcType& value)
...
+  value_out->reserve(utf16value.length());  // not perfect, but better than nothing
+  utf8::utf16to8(utf16value.begin(), utf16value.end(), std::back_inserter(*value_out));
+#else
+  size_t wcs_len = mbstowcs(NULL, value.c_str(), 0);
+  if (wcs_len == (size_t)-1) {
+    throw new std::runtime_error("Unknown character in string");
+  }
+  std::wstring utf32value(wcs_len, '\0');
+  mbstowcs(&utf32value[0], value.c_str(), wcs_len);
+  value_out->reserve(utf32value.length());  // not perfect, but better than nothing
```
dmondrik, line 185: Does it function correctly if the actual size is greater than this? It might be a good idea to add to the comment that it's just a performance hint.

source/server/converters.h
```diff
@@ -113,6 +114,79 @@ CType convert_from_grpc(const GrpcType& value)
...
+  utf8::utf32to8(utf32value.begin(), utf32value.end(), std::back_inserter(*value_out));
+#endif
+}
+
```
dmondrik, line 189: At the bottom of this file, I think `ApiErrorAndDescriptionToStatus` needs the conversion for the `description` parameter.

dmondrik (changes requested):
await: What about strings in hand edited files under `source/custom`?

source/codegen/templates/service_helpers.mako
DavidCurtiss: 1. In addition to const_cast, it would also need a static_cast or reinterpret_cast, since the c_type_pointer is not a `char*`.
2. I made it this way to match the nearby code, 2 lines down.

source/server/converters.h
DavidCurtiss: I emulated what others have done in production code. I could research whether it was intentional or not, but since it works correctly like this, it's not worth the effort IMO.

generated/nidcpower/nidcpower_service.cpp
DavidCurtiss: Done

generated/nidigitalpattern/nidigitalpattern_service.cpp
```diff
@@ -1540,7 +1603,8 @@ namespace nidigitalpattern_grpc {
     try {
       auto vi_grpc_session = request->vi();
       ViSession vi = session_repository_->access_session(vi_grpc_session.name());
-      auto site = request->site().c_str();
+      auto site_mbcs = convert_from_grpc<std::string>(request->site());
```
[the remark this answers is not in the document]
DavidCurtiss: Done

.gitmodules
DavidCurtiss: Done

source/server/converters.h
DavidCurtiss: Yes. Updated the comment.

source/server/converters.h
DavidCurtiss: Done

dmondrik (approved):
LGTM

source/custom/nidcpower_service.custom.cpp
```diff
@@ -3,6 +3,8 @@
 #include <stdexcept>
 
 namespace nidcpower_grpc {
+using nidevice_grpc::converters::convert_from_grpc;
+using nidevice_grpc::converters::convert_to_grpc;
```
dmondrik, line 7: Consider: There are existing uses of `nidevice_grpc::converters::convert_to_grpc::` in this file you could clean up if you don't mind.

DavidCurtiss: Done

bkeryan (approved):

generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -55,7 +55,8 @@ namespace nidaqmx_grpc {
       return ::grpc::Status::CANCELLED;
     }
     try {
-      auto port_list = request->port_list().c_str();
+      auto port_list_mbcs = convert_from_grpc<std::string>(request->port_list());
+      auto port_list = port_list_mbcs.c_str();
       auto status = library_->AddCDAQSyncConnection(port_list);
```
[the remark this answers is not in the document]
dmondrik: Sounds to me like it'd be better for this release to do it the same for all of them, and consider how to add opt-out in the future as tech debt. But David if you can see a straightforward way to do it now, I'm fine with that too. Up to you.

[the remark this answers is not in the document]
DavidCurtiss: Created #802

PR #799, opened 2022-11-23
Add initialization behavior for non-MI driver init methods.

reckenro:

Adds the new initialization behavior input and newSessionInitialized output to init methods for non-MI Drivers.

This was done for MI drivers to give more control over initialization. We should offer non-MI drivers the same update. This shouldn't affect clients if they want to continue using the old behavior since leaving the input field blank (unspecified) will result in the same behavior as they had previously.

Mostly relying on successful build since this behavior has already been tested against ni-fake.


source/codegen/client_helpers.py
```diff
@@ -140,14 +147,21 @@ def _get_param_mechanism(param: dict) -> ParamMechanism:
     return ParamMechanism.COPY
 
 
+def _get_param_default(param: dict) -> str:
+    if param['grpc_type'] == 'nidevice_grpc.SessionInitializationBehavior':
+        return 'nidevice_grpc::SESSION_INITIALIZATION_BEHAVIOR_UNSPECIFIED'
```
reckenro, line 152: Note: I wanted the generated _client.h used throughout the tests to have a default for the SessioninitializationBehavior params so I didn't have to specify this default throughout the System tests.

source/codegen/metadata/nifake_non_ivi/functions.py
```diff
@@ -164,7 +164,7 @@
             },
             {
                 'direction': 'out',
-                'get_last_error': 'get_last_error',
+                'get_last_error': 'deprecated',
```
reckenro, line 167: Note: For the MI drivers, we updated all the 'get_last_error' values to 'deprecated'. I did the same for all the non-MI drivers as well so that's what these 'get_last_error' updates are about.

This doesn't functionally matter because we don't look at the value for this field in code generation, we're just looking for the 'get_last_error' key to decide to mark the field deprecated in the proto file.

generated/nirfmxbluetooth/nirfmxbluetooth.proto
```diff
@@ -1367,24 +1367,28 @@ message InitializeRequest {
...
   string option_string = 3;
+  nidevice_grpc.SessionInitializationBehavior initialization_behavior = 4;
 }
 
 message InitializeResponse {
   int32 status = 1;
   nidevice_grpc.Session instrument = 2;
   int32 is_new_session = 3;
   string error_message = 4 [deprecated = true];
+  bool new_session_initialized = 5;
```
reckenro, line 1378: TODO: Create a test for this, namely to make sure that there's scenarios where `is_new_session` and `new_session_initialized` differ and thus should both be represented in the response.

My thought is that they can be different because `is_new_session` is driver related while `new_session_initialized` is just about if the library's init method was called successfully and a new gRPC session was created (new mapping between named session and session handle): 

1. Initialize with `resource_name`/`option_string` and `sessionName = "SessionName1"`
   1a. `isNewSession` is True and `newSessionInitialized` is True
2. Init with same `resource_name`/`option_string` and different `sessionName = "SessionName2"`
   2a. `isNewSession` from driver is False while `newSessionInitialized` is True

astarche (approved):

generated/nirfmxbluetooth/nirfmxbluetooth.proto
reckenro: Added `Init_InitAgain_NewSessionInitializedValuesDiffer` to `nirfmxbluetooth_driver_api_tests.cpp` that confirms this behavior and the need for both output fields.

dmondrik (approved):
LGTM

reckenro:
Addressing the `build_nilrt` failure in #803

PR #803, opened 2022-11-29
Remove fetch-depth: 0 from build workflows.

reckenro:

Removes the `fetch-depth: 0` from the git checkout in the two `build_` files. This was previously needed to be able to reference the tags in the build but this is no longer used and thus no longer needed.

#799 keeps failing on the build-nilrt Workflow. It would miss the cache on the `Cache Host OS gRPC Support` step and therefore try the step to `Build Host OS gRPC Support` which was failing.

I don't want to tackle trying to fix the `Build Host OS gRPC Support` since I think it would be more difficult so instead looked to fix the cache miss.

Turns out, since we switched the fetch-depth to 0 (all history) we're getting to the point where the default abbreviation git chooses was bumped from 7 to 8 (you can see it briefly explained in [this section of the git 2.11.0 release notes](https://github.com/git/git/blob/7452749a781d84244ecd08c6f6ca7e5df67dfce8/Documentation/RelNotes/2.11.0.txt#L22)). That means for commands like `git rev-parse --short @:./third_party/grpc` that's used as the key for the cache used to be calculated as `58602e2` but is now `58602e20` causing the cache miss.

Since we don't need the `fetch-depth` to be 0 (all history) anymore, I'm removing it to get the caching working again and side step the build failure on the `Build Host OS gRPC Support` step.

Build succeeds.


reckenro:
Note: Once I had enough diffs in the full history like I did in the other PR it failed to fetch the cached gRPC Host OS Support here: https://github.com/ni/grpc-device/actions/runs/3578898259/jobs/6019550748

I'm addressing that issue with the change and not the failing of the Build Host OS gRPC Support step.

The reason I think it failed to get the cached build is because the history with the fetch-depth set to 0 hit some threshold so the abbreviation of the git commit using `git rev-parse --short @:./third_party/grpc` was increased. It used to be 7 characters (`58602e2`) but now is 8 (`58602e20`). This resulted in the cache miss and caused it to try to rebuild it.

We don't need to have the fetch-depth set to 0 (full history) anymore. It was used when we were using the git tags in the build but we arren't doing that anymore. My hope is that this will get us back to the shorter git commit and we can start using the cached gRPC build again.

dmondrik (approved):

astarche (approved):

SparkingSpork (approved):

PR #805, opened 2022-12-07
Use function pointers to call functions on all platforms

DavidCurtiss:

Uses loaded function pointers on Windows, too, where we had been using delayload'ed symbols.

Consistency, simplicity, and [AB#2209410](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2209410)

Using an equivalent code change, tested the [AB#2209410](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2209410) use case.


bkeryan:
Is this a temporary workaround or a permanent change?

If it's a temporary workaround, you could do something like `#if defined(_MSC_VER) && USE_DELAY_LOAD`.

If it's a permanent change, you may be able to delete the import libraries from disk and remove `CMakeLists.txt`'s references to `Delayimp`, the import libraries, and `/DELAYLOAD`.

DavidCurtiss:
> Is this a temporary workaround or a permanent change?

It's permanent, so I've now deleted the delayload stuff and the import libraries.

I also added some compilation tests, because, as @astarche previously pointed out to me over Teams, the former implementation was providing "compile time validation that we got the [function parameter] types ... close enough to compile against the header."

astarche (approved):
Looks great! Multiple nice improvements. Thanks for sorting out the details.

bkeryan (approved):

source/codegen/templates/compilation_test.cpp.mako
```diff
@@ -0,0 +1,36 @@
...
+% for method_name in service_helpers.filter_api_functions(functions, only_mockable_functions=False):
+<%
+  f = functions[method_name]
+  parameters = f['parameters']
+  return_type = f['returns']
+  parameter_list = service_helpers.create_params(parameters, expand_varargs=True)
+  argument_list = ', '.join(p['cppName'] for p in service_helpers.expand_varargs_parameters(parameters) if not p.get("proto_only", False))
+  c_name = service_helpers.get_cname(functions, method_name, c_function_prefix)
+%>\
+% if not service_helpers.is_private_method(f) and not service_helpers.is_restricted_driver_service(config):
```
reckenro, line 28: More of a curiosity but is there a reason to not include the private methods for these compilation tests?

DavidCurtiss: 1. This is the same filter that library.cpp was using.
2. At least some of the private methods are implemented by us, not the driver.

SparkingSpork (approved):

PR #810, opened 2022-12-14
Update MI metadata to match version in nimi-python.

reckenro:

For all the MI drivers in nimi-python, update the grpc-device metadata to be sourced from the same hapigen metadata as what the corresponding nimi-python metadata is based on.

In other words, nimi-python has NI-DCPower's metadata/proto sourced from the d318 version of the hapigen metadata (see [here](https://github.com/ni/nimi-python/blob/master/src/nidcpower/metadata/nidcpower.proto)) so we should align with that and pull in the same version of the metadata to grpc-device. Same goes for all the MI drivers.

Was looking at nimi-python metadata and noticed that it was working off a different version of the hapigen metadata (and resulting .proto file) than grpc-device was pulling in.

I think in general these should be in sync (since nimi-python metadata takes into account grpc-device metadata). I'm not sure if it practically matters at this point but thought we should get them on the same page.

Reviewed generated changes and they seem reasonable.
Unit and Integration tests pass locally. Relying on System Tests for everything else.

### DIFF
--- a/generated/nidcpower/nidcpower.proto
+++ b/generated/nidcpower/nidcpower.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-DCPower API metadata version 23.0.0d288
+// This file is generated from NI-DCPower API metadata version 23.0.0d318
 //---------------------------------------------------------------------
 // Proto file for the NI-DCPower Metadata
 //---------------------------------------------------------------------
@@ -43,7 +43,6 @@ service NiDCPower {
   rpc ConfigureDigitalEdgeSourceTriggerWithChannels(ConfigureDigitalEdgeSourceTriggerWithChannelsRequest) returns (ConfigureDigitalEdgeSourceTriggerWithChannelsResponse);
   rpc ConfigureDigitalEdgeStartTrigger(ConfigureDigitalEdgeStartTriggerRequest) returns (ConfigureDigitalEdgeStartTriggerResponse);
   rpc ConfigureDigitalEdgeStartTriggerWithChannels(ConfigureDigitalEdgeStartTriggerWithChannelsRequest) returns (ConfigureDigitalEdgeStartTriggerWithChannelsResponse);
-  rpc ConfigureLCRCompensation(ConfigureLCRCompensationRequest) returns (ConfigureLCRCompensationResponse);
   rpc ConfigureLCRCustomCableCompensation(ConfigureLCRCustomCableCompensationRequest) returns (ConfigureLCRCustomCableCompensationResponse);
   rpc ConfigureOutputEnabled(ConfigureOutputEnabledRequest) returns (ConfigureOutputEnabledResponse);
   rpc ConfigureOutputFunction(ConfigureOutputFunctionRequest) returns (ConfigureOutputFunctionResponse);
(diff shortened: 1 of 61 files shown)


generated/nidcpower/nidcpower.proto
```diff
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-DCPower API metadata version 23.0.0d288
+// This file is generated from NI-DCPower API metadata version 23.0.0d318
```
reckenro, line 3: Note: These functions weren't present in the previous release of grpc-device and were marked to not codegen in this [AzDo PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/392729?_a=files&path=/src/nidcpower/nidcpowerBaseDistribution/nidcpowerAPIMetadata/source/nidcpowerAPIMetadata/grpc_device/functions/functions.hapigen).

The resulting nimi-python metadata was updated based on those changes but the grpc-device repo was never updated so that's what these changes represent.

I believe these methods will be enabled before the 23.0 driver release so there should be another update to enable them here and in nimi-python. But until then, we should stay in sync with nimi-python and update to the same hapigen-generated metadata.

reckenro: Also, this matches the [proto file that's in nimi-python right now ](https://github.com/ni/nimi-python/blob/master/src/nidcpower/metadata/nidcpower.proto)(sourced on 23.0.0d318)

source/tests/unit/ni_fake_service_tests.cpp
```diff
@@ -1316,9 +1316,7 @@ TEST(NiFakeServiceTests, NiFakeService_ParametersAreMultipleTypes_CallsParameter
           an_int_enum,
           a_float,
           expected_float_enum_value,
-          expected_string_size,
-          _))
-      .With(Args<8, 7>(ElementsAreArray(a_string, expected_string_size)))
```
reckenro, line 1321: Note: These three tests were updated because the `ParametersAreMultipleTypes` signature changed in the base hapigen metadata to not include the `string size` input. So the expectation was updated accordingly.

generated/nifake/nifake.proto
```diff
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-FAKE API metadata version 23.0.0
+// This file is generated from NI-FAKE API metadata version 23.0.0d96
```
reckenro, line 3: Note: NI-FAKE hapigen metadata has changed some since we last updated here, but that shouldn't really matter. It still builds and the Unit/Integration tests still pass which is all we care about.

generated/nidcpower/nidcpower.proto
```diff
@@ -43,7 +43,6 @@ service NiDCPower {
   rpc ConfigureDigitalEdgeSourceTriggerWithChannels(ConfigureDigitalEdgeSourceTriggerWithChannelsRequest) returns (ConfigureDigitalEdgeSourceTriggerWithChannelsResponse);
   rpc ConfigureDigitalEdgeStartTrigger(ConfigureDigitalEdgeStartTriggerRequest) returns (ConfigureDigitalEdgeStartTriggerResponse);
   rpc ConfigureDigitalEdgeStartTriggerWithChannels(ConfigureDigitalEdgeStartTriggerWithChannelsRequest) returns (ConfigureDigitalEdgeStartTriggerWithChannelsResponse);
-  rpc ConfigureLCRCompensation(ConfigureLCRCompensationRequest) returns (ConfigureLCRCompensationResponse);
```
astarche, line 46: Looks like this would be a `breaking change` right?

astarche (approved):

generated/nidcpower/nidcpower.proto
dmondrik: I believe the DCPower methods will be exposed for gRPC only after 23.0 releases, in 23.3 for example.

generated/nidcpower/nidcpower.proto
reckenro: Hmm, yes and no. Compared to previous versions of this .proto in main -> Yes. Compared to previous versions of this .proto in the GitHub Releases -> No.

I'm wary of marking it a breaking-change because I think it'll be confusing in the next release notes. Thoughts?

dmondrik (approved):

generated/nidcpower/nidcpower.proto
dmondrik: We haven't released it, so we shouldn't mark it breaking.

PR #814, opened 2022-12-19
Import shared session.proto and nidevice.proto.

reckenro:

Import session.proto and nidevice.proto to `imports/protobuf` instead of having them in `source/proto`. Also add a README to the spot to document where the imported proto files are coming from.

We were hitting a circular dependency between grpc-device and hapigen. Hapigen used to be relying on grpc-device for the nidevice.proto and session.proto files while grpc-device was relying on hapigen for the specific MI driver metadata/proto.

Now that we have an export (see [here](https://p4.natinst.com/browser/perforce/build/exports/ni/hapi/hapigen_grpc_device_plugin/official/export/.latest/.latest/proto/)) for nidevice.proto and session.proto we should import from there instead of having them in source in this repo. We're also depending on this new export in hapigen to break the circular dependency.

So now hapigen can go final, then grpc-device, then dependents of grpc-device.

Unit and integration tests still pass locally. Trusting the build/PR tests for everything else.


astarche (approved):

SparkingSpork (approved):

dmondrik:
> So now hapigen can go final, then grpc-device, then dependents of grpc-device.

I might expect this sequence: hapigen goes final, then MI clients of hapigen using those proto files to generate MI proto files, then copy those final exported MI proto files here, then grpc-device goes final.

PR #819, opened 2023-01-10
Update grpc to v1.51.1

reckenro:

Updates grpc submodule to point to [v1.51.1](https://github.com/grpc/grpc/releases/tag/v1.51.1)

The `build_nilrt` action has been failing recently when trying to "Build Host OS gRPC Support" due to an issue in the current version of abseil-cpp the grpc submodule pulls in building against a recent gcc. I think we avoided this for a bit because it was using a cached version of this build but we recently started getting a cache miss for it and it would try to rebuild and fail (see [Illya's build](https://github.com/ni/grpc-device/actions/runs/3875868384/jobs/6611884632) failing for this reason)

Also had to replace references to `::google::protobuf::string` with `std::string` because that was removed in one of the protobuf updates.

Build passes and specifically the "Build Host OS gRPC Support" step now succeeds.


reckenro:
Hmm, I'm wondering now if I updated it correctly. I navigated to `third_party/grpc` and then did `git fetch` and then `git checkout v1.51.1` and then added/committed that. Is this the correct way to update a submodule?

astarche:
@reckenro it looks right to me from the diff (shows the commit for `v1.51.1`). I'm not 100% on the steps and it's been fussy for me in the past.

astarche (approved):

source/codegen/client_helpers.py
```diff
@@ -124,6 +124,8 @@ def _get_cpp_type_for_protobuf_type(protobuf_type: str) -> str:
 
     if protobuf_type in PROTOBUF_TYPE_TO_CPP_TYPE:
         cpp_type = PROTOBUF_TYPE_TO_CPP_TYPE[protobuf_type]
+        if cpp_type in ["string"]:
```
reckenro, line 127: This is to fix up the generated _client.h and _client.cpp files. Found out in some recent update to protobuf that `string` is no longer defined in `::google::protobuf`. You can see the issue raised [here](https://github.com/protocolbuffers/protobuf/issues/9943) but apparently it was intentional and the recommendation is just to use `std::string` instead which is what this update is about.

PR #820, opened 2023-01-10
Pull in 23.0 finals of MI driver grpc-device metadata

reckenro:

Pulls in the 23.0 final grpc-device metadata for the MI drivers.

We should transition to the final metadata before making a grpc-device release.

Build should suffice.

### DIFF
--- a/generated/nifgen/nifgen_compilation_test.cpp
+++ b/generated/nifgen/nifgen_compilation_test.cpp
@@ -562,7 +562,7 @@ ViStatus SelfTest(ViSession vi, ViInt16* selfTestResult, ViChar selfTestMessage[
   return niFgen_self_test(vi, selfTestResult, selfTestMessage);
 }
 
-ViStatus SendSoftwareEdgeTrigger(ViSession vi, ViInt32 trigger, ViString triggerId)
+ViStatus SendSoftwareEdgeTrigger(ViSession vi, ViInt32 trigger, ViConstString triggerId)
 {
   return niFgen_SendSoftwareEdgeTrigger(vi, trigger, triggerId);
 }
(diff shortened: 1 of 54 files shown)


generated/nifgen/nifgen_compilation_test.cpp
```diff
@@ -562,7 +562,7 @@ ViStatus SelfTest(ViSession vi, ViInt16* selfTestResult, ViChar selfTestMessage[
   return niFgen_self_test(vi, selfTestResult, selfTestMessage);
 }
 
-ViStatus SendSoftwareEdgeTrigger(ViSession vi, ViInt32 trigger, ViString triggerId)
+ViStatus SendSoftwareEdgeTrigger(ViSession vi, ViInt32 trigger, ViConstString triggerId)
```
reckenro, line 565: I'm not sure when this change got into the FGEN metadata but I believe it is correct. In the header file, it is a `ViConstString` for `triggerId` so this is more accurate and correct and avoids a casting to `ViString` in nifgen_service.cpp.

generated/niscope/niscope.proto
```diff
@@ -473,8 +473,8 @@ enum ExportableSignals {
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_ADVANCE_EVENT = 6;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_START_EVENT = 7;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_REF_EVENT = 10;
-  EXPORTABLE_SIGNALS_NISCOPE_VAL_5V_OUT = 13;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_REF_CLOCK = 100;
+  EXPORTABLE_SIGNALS_NISCOPE_VAL_5V_OUT = 13;
```
reckenro, line 477: Note to self: Find whatever in hapigen is causing these to jump around. Guessing its some index equivalency which causes the ordering to fluctuate each build.

dmondrik (approved):

astarche (approved):

PR #823, opened 2023-01-11
Bump version in CMakeLists.txt for next release in releases branch.

reckenro:

Bumps version in CMakeLists.txt in preparation for the 2.0.0 release.

Gets the version correct in the releases/2.0 branch.

None.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 1.5.2)
+  VERSION 2.0.0)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(CreateVirtualEnvironment)


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 1.5.2)
+  VERSION 2.0.0)
```
reckenro, line 5: I imagine doing it here in the releases/2.0 branch is correct. Then after we have a 2.0 release published we can update from 1.5.2 in the main branch.

I'd guess we'd also update it to 2.0.0 in main as well post-release but could also see us having some clarification that its a development version instead of a released one...

SparkingSpork: The way we do it for the RTOS repos (e.g., [see here](https://github.com/ni/meta-nilrt/blob/61b06f95243c342711731148054aff34f55706d5/conf/distro/nilrt.conf#L3)) is to set the version in main once we start development on that version. When we cut the release branch, it will already have that version and then we post a PR to bump the version in main afterwards. The main difference there is that the history of when the version bump occurred is preserved in both the release branch, the main branch, and any subsequent release branches. @amstewart - Would you have any advice on this? Relying on you as an "expert" here since you're the RTOS code owner. 

I'm not sure how that would play with the automation or how that compares to other open source projects at NI, but it seems reasonable to expect that if a "release" doesn't exist, the version is under development.

SparkingSpork (commented):
PR looks fine, just want to resolve the process questions before approving.

CMakeLists.txt
reckenro: Mmm, good point. That seems the better route to take. I'll assume we'll want to go that route and should've updated the version in main to this upcoming release a while ago as we were working towards it. I guess for now I'll update in main to 2.0.0, get that in, and then get it into the releases/2.0 branch from the commit in main instead.

amstewart: My recommendation would be to keep the mainline branch versioned like: `last.released.version+git` and the release branch versioned like `new.release.version`.

eg.
* `main` versioned `1.2.3+git`
* cut a release branch `release-1.3.0` with a commit that changes version to `1.3.0`.
* Put a commit into the `main` ref, reversioning it to `1.3.0+git`.

That way, builders using the mainline branch will always get a binary with a `+git` version suffix, unless they build from the release branch. And because the main branch's *number* components are the last release's version number, you don't have to decide what the next version's number will be until you are ready to cut that release's branch.

reckenro:
@harsha-bhushan , I think we had a discussion at some point on the next version of the grpc-device release but I wanted to point it out here. I believe the most substantial reason to bump the major version for this release was because of the overhaul in how error-reporting is done with Dan's PR: https://github.com/ni/grpc-device/pull/692.

Do you agree that the next release should be 2.0.0 instead of a minor version bump to 1.6.0?

CMakeLists.txt
reckenro: @amstewart I looked into it a little bit and don't think this VERSION we're using for the project in CMAKE allows for a suffix, just non-negative integers so we couldn't distinguish it with a `+git` suffix in main. See CMAKE project version docs [here](https://cmake.org/cmake/help/latest/command/project.html#options)

reckenro: I could add a DESCRIPTION to the project though with "git" or "Development Version" or something like that to denote it in the main branch if you think that would suffice.

SparkingSpork: This is strange to me. CMake doesn't have an option for version suffixes that I can find (and therefore doesn't fully support semantic versioning). 

This also raises another issue - we don't even tag our binaries with versions when they're built as far as I recall.

reckenro: So I guess we could have something like

ReleaseExecutable

for the executable created from a Releases branch and

MainExecutable

reckenro: Not great, but I think that's as close as we'll get with respect to this Executable versioning for Windows.

amstewart: > @amstewart I looked into it a little bit and don't think this VERSION we're using for the project in CMAKE allows for a suffix, just non-negative integers so we couldn't distinguish it with a `+git` suffix in main. See CMAKE project version docs [here](https://cmake.org/cmake/help/latest/command/project.html#options)

Oh. I saw [this implementation](https://github.com/grpc/grpc/blob/master/CMakeLists.txt#L31) in grpc and assumed that it did. But I suppose that is a custom variable they're using, rather than the cmake `VERSION`. Lame.

reckenro:
Abandoning this PR since for now we bumped both main and releases/2.0 branch Version to 2.0.0.

Created #828 linking back to this PR to investigate a way to distinguish it and provided a link to the discussion in this PR as possible seed ideas for differentiation.

PR #824, opened 2023-01-11
Bump to version 2.0.0 in main.

reckenro:

Bumps the executable version from 1.5.2 to 2.0.0 in the main branch.

We should mirror what RTOS does for releases. When we start development on the next release, update the version in main. Then when when we create the releases branch for it it will have the correct version. Then post-release, update the main branch to the upcoming release.

Relying on the build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 1.5.2)
+  VERSION 2.0.0)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(CreateVirtualEnvironment)


reckenro:
I think I'll likely abandon #823 in favor of submitting in the main branch first and then cherry-picking the change over to the releases/2.0 branch.

SparkingSpork (approved):

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 1.5.2)
+  VERSION 2.0.0)
```
reckenro, line 5: @harsha-bhushan , I think we had a discussion at some point on the next version of the grpc-device release but I wanted to point it out here. I believe the most substantial reason to bump the major version this release is because of the overhaul in how error-reporting is done with Dan's PR: https://github.com/ni/grpc-device/pull/692.

Do you agree that the next release should be 2.0.0 instead of a minor version bump to 1.6.0?

dmondrik (approved):

astarche (approved):

CMakeLists.txt
harsha-bhushan: I agree

harsha-bhushan (approved):

PR #825, opened 2023-01-12
Cherry-pick to releases/2.0 branch: "Bump to version 2.0.0 in main. (#824)"

reckenro:

Cherry-pick commit to main (#824) which updated CMAKE project version to 2.0.0 (used to show the executable version on Windows).

Releases branch should also update to 2.0.0

There was discussion in my first attempt to change the version in the releases branch directly (#823 ) about how to distinguish it in the main branch. Up to this point, we haven't had a distinction between the released 1.5.1 version and the version in main also showing 1.5.1 for this project / executable version. So I think for the release it should be fine to bump both to 2.0.0.

I think there should be an extra effort / investigation to figure out in main how to distinguish an executable coming from there, possibly with specifying `DESCRIPTION` on the CMAKE project and incorporating that in the executable details.

Confirmed the update updates the version shown with ni_grpc_device_server.exe

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 1.5.2)
+  VERSION 2.0.0)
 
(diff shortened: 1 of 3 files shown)


reckenro:
I think I messed up this PR: I just want it to pull in the 2.0.0 bump. I'll abandon this one and get it set up properly not to synchronize all pushes to main.

PR #829, opened 2023-01-18
[RFC] Expose GetStartTimestampInformation in niscope.proto

reckenro:

Exposes the private niScope_GetStartTimestampInformation in niscope.proto

[AB#2267151](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2267151)

While developing the grpc-device DVL plugin for LabVIEW calls, it was noted that other plugins call a private `niScope_GetStartTimestampInformation` to populate some of the return info for their WDT methods like `niScope_LVGenericFetchWDT_Info`. And currently, that method isn't exposed in grpc-device.

Still trying to determine if it makes sense to try to put this in the regular niscope.proto or if it warrants something more akin to the restricted services and the DVL plugin for grpc-device will have to pull that in as well for these WDT methods to work. Or some other option I'm not thinking of / prompting a discussion about if exposing a private method like this is appropriate at all.

Build passes and new system test for GetStartTimestampInformation succeeds.

### DIFF
--- a/generated/niscope/niscope.proto
+++ b/generated/niscope/niscope.proto
@@ -1707,10 +1707,10 @@ message GetStartTimestampInformationRequest {
 
 message GetStartTimestampInformationResponse {
   int32 status = 1;
-  uint32 sys_time_in128_bits_t1 = 2;
-  uint32 sys_time_in128_bits_t2 = 3;
-  uint32 sys_time_in128_bits_t3 = 4;
-  uint32 sys_time_in128_bits_t4 = 5;
+  uint32 sys_time_in_128_bits_t1 = 2;
+  uint32 sys_time_in_128_bits_t2 = 3;
+  uint32 sys_time_in_128_bits_t3 = 4;
+  uint32 sys_time_in_128_bits_t4 = 5;
   double device_time_in_absolute_time_units = 6;
 }
 
--- a/generated/niscope/niscope.proto
+++ b/generated/niscope/niscope.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-SCOPE API metadata version 23.0.0f168
+// This file is generated from NI-SCOPE API metadata version 23.3.0d9999
 //---------------------------------------------------------------------
 // Proto file for the NI-SCOPE Metadata
 //---------------------------------------------------------------------
@@ -84,6 +84,7 @@ service NiScope {
   rpc GetFrequencyResponse(GetFrequencyResponseRequest) returns (GetFrequencyResponseResponse);
   rpc GetNormalizationCoefficients(GetNormalizationCoefficientsRequest) returns (GetNormalizationCoefficientsResponse);
   rpc GetScalingCoefficients(GetScalingCoefficientsRequest) returns (GetScalingCoefficientsResponse);
+  rpc GetStartTimestampInformation(GetStartTimestampInformationRequest) returns (GetStartTimestampInformationResponse);
   rpc GetStreamEndpointHandle(GetStreamEndpointHandleRequest) returns (GetStreamEndpointHandleResponse);
   rpc ImportAttributeConfigurationBuffer(ImportAttributeConfigurationBufferRequest) returns (ImportAttributeConfigurationBufferResponse);
   rpc ImportAttributeConfigurationFile(ImportAttributeConfigurationFileRequest) returns (ImportAttributeConfigurationFileResponse);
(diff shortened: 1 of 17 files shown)


imports/include/niScopePrivate.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
```
reckenro, line 1: Note: I stripped everything else out except the one method I needed to get things building and also not expose too much.

source/codegen/metadata/niscope/niscope.proto
```diff
@@ -1700,6 +1701,19 @@ message GetScalingCoefficientsResponse {
   sint32 number_of_coefficient_sets = 3;
 }
 
+message GetStartTimestampInformationRequest {
+  nidevice_grpc.Session vi = 1;
+}
+
+message GetStartTimestampInformationResponse {
+  int32 status = 1;
+  uint32 sys_time_in128_bits_t1 = 2;
```
reckenro, line 1710: Note: I can add `grpc_name` in the metadata to get these parameter names formatted how we want.

source/codegen/metadata/niscope/config.py
```diff
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-SCOPE API metadata version 23.0.0f168
+# This file is generated from NI-SCOPE API metadata version 23.3.0d9999
 config = {
     'additional_headers': {
         'custom/ivi_errors.h': [
             'service.cpp'
-        ]
+        ],
+        'niScopePrivate.h': [
```
reckenro, line 8: Note: This is how I got it to pull in the private header. I didn't see a way to pull in multiple headers currently with the `c_header` key so just did it this way.

generated/niscope/niscope.proto
```diff
@@ -1700,6 +1701,19 @@ message GetScalingCoefficientsResponse {
   sint32 number_of_coefficient_sets = 3;
 }
 
+message GetStartTimestampInformationRequest {
+  nidevice_grpc.Session vi = 1;
+}
+
+message GetStartTimestampInformationResponse {
+  int32 status = 1;
+  uint32 sys_time_in128_bits_t1 = 2;
```
astarche, line 1710: Is this supposed to be `sys_time_in_128_bits_t1`?

source/codegen/metadata/niscope/functions.py
```diff
@@ -2672,6 +2672,54 @@
         ],
         'returns': 'ViStatus'
     },
+    'GetStartTimestampInformation': {
+        'codegen_method': 'public',
```
astarche, line 2676: I was expecting this to be `codegen_method` `private` and then you don't need to include the `private` header. But that also excludes it from the `proto` API? Do we have some other method to exclude it from the compilation test but include it in the `proto`?

astarche (approved):
This looks OK to me if you decide to go this direction (I notice that there's an open question in the description ...).

reckenro:
@harsha-bhushan , for this private `GetStartTimestampInformation` (it's in niScopePrivate.h) that's needed for some of the DVL work, should we be adding it as an entrypoint to niscope.proto or should I make a new restricted service like NiScopeRestricted to expose this method? It would look and function similar to the restricted services like the one added for RFmx SpecAn: https://github.com/ni/grpc-device/blob/main/generated/nirfmxspecan_restricted/nirfmxspecan_restricted.proto

generated/niscope/niscope.proto
reckenro: Yeah, I'll need to add `grpc_name` for these fields in the metadata.

source/codegen/metadata/niscope/functions.py
reckenro: I don't think so, not at the moment at least. I know we skip over them for the restricted services which is making me slightly lean towards that. I made a draft PR for what that would look like instead since it wasn't much more work than getting this one thrown together: #830 

I also tagged Harsha in a separate comment below to try and see which route to take.

harsha-bhushan:
Going by the function name & what it does, it seems ok to me to expose it. Restricted service just for this one function might be an overkill. It would be good to ensure that the driver experts don't have any reservations with it.

If you anticipate more such functions now or in the future, then having a restricted service might be a good approach.

Also, we typically document the functions exposed in the proto. But if my understanding is right, this function isn't something that a user would typically call (?) in their code. So, not having function documentation is acceptable.

harsha-bhushan (approved):

reckenro:
Sounds good. I have [this PR in AzDo](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/426457) to see if the NI-SCOPE folks agree to put it in niscope.proto directly or not. I'll update this to a real review or open up the alternative depending on the results of that PR discussion.

generated/niscope/niscope.proto
reckenro: Added `grpc_name` so these look normal now with `sys_time_in_128_bits_t<number>`

reckenro:
Abandoning this one in favor of the restricted service. There was more discussion in a related [AzDo PR Thread](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/426457?discussionId=4068478) and it was decided adding a restricted service made more sense.

PR #835, opened 2023-01-25
[DRAFT] [US2258982] Add nirfmxnr_restricted support 

WesleyTangNationalInstruments:

Adds support for nirfmxnr_restricted to grpc-device.

This adds entrypoints that are not included in the public header, specifically:

- ModAccFetchTransientPeriodLocationsTrace

[AB#2258982](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2258982)

Tests pass. Build passes.


source/codegen/metadata/nirfmxnr_restricted/config.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+config = {
+    "code_readiness": "RestrictedRelease",
```
reckenro, line 3: If/When you get back to this a heads up that I changed the restricted proto / service exposure a little bit with #842 

I changed the config.py expectations to separate the restricted and code_readiness concept so it now expects it in the form
```py
    "code_readiness": "Release",
    "is_restricted": True,
```

WesleyTangNationalInstruments: Thanks for letting me know, but I actually forgot to abandon this PR. Since we added the functionality we needed in the public server, there is no longer a need for a restricted service for NR.

PR #838, opened 2023-01-30
Update README to reflect current driver support / testing.

reckenro:

Updates the README driver support section to reflect the recent updates (MI drivers to 23.0 / 2023 Q1).

Also, noticed that FGEN has its tests marked not to run on Linux when they should be able to. So I removed the code there that skipped FGEN tests on Linux.

Keeping documentation up to date.

Doesn't skip FGEN system tests anymore on Linux.

Trusting the build for the FGEN test updates.

### DIFF
--- a/source/tests/system/nifgen_driver_api_tests.cpp
+++ b/source/tests/system/nifgen_driver_api_tests.cpp
@@ -586,7 +586,7 @@ TEST_F(NiFgenDriverApiTest, ExportTriggerMode_ResetAndImportConfiguration_Trigge
 TEST_F(NiFgenDriverApiTest, AllocateWaveform_WriteWaveformComplexF64_WaveformWrittenSuccessfully)
 {
 #ifndef WIN32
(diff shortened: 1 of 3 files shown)


source/tests/system/nifgen_driver_api_tests.cpp
```diff
@@ -588,6 +585,9 @@ TEST_F(NiFgenDriverApiTest, ExportTriggerMode_ResetAndImportConfiguration_Trigge
 
 TEST_F(NiFgenDriverApiTest, AllocateWaveform_WriteWaveformComplexF64_WaveformWrittenSuccessfully)
 {
+#ifndef WIN32
```
reckenro, line 588: Note: Found out this test won't work on Linux but the rest should. Failed on Linux when I didn't have this skip and looked at the [NI-FGEN documentation](https://www.ni.com/pdf/manuals/ni-fgen-linux-2023-q1.html) which mentioned OSP functionality wasn't supported on Linux so it made sense.

PR #840, opened 2023-02-01
Initial RFmx GSM service

phindman:

* Add GSM metadata to define simple service with Initialize, Close, GetError, and GetErrorString methods
* Add system level tests for GSM Init/Close

Like we did with the other RFmx personalities, we want to have an initial service in place for GSM.
Closes [AB#2275156](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2275156) 

Builds, and new System-level tests pass.
image


phindman:
I see that some of the system tests failed on Windows.  Based on the error message for the GSM tests, I suspect that the agent that actually executes the tests (grpcbot-win-10?) doesn't have RFmxGSM installed.  Is there another change (in ni-central?) that I should make to get that installed?

DavidCurtiss:
> Is there another change (in ni-central?) that I should make to get that installed?

@phindman 
The driver feeds that get installed come from [here](https://ni.visualstudio.com/DevCentral/_git/ni-central?path=/eng/iac/cfg_mgmt/salt/srv/salt/grpcbot/win/resources/driver_feeds_win.yml).

On a related note, there's also an effort to get RFmx NR licensed on the test machine, discussed [here](https://teams.microsoft.com/l/message/19:658a48f1bcd240dc9c56f5eb35761898@thread.tacv2/1675278885122?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=22cecbce-4450-4339-bbf5-85ade50ab7fb&parentMessageId=1675278885122&teamName=gRPC&channelName=grpc-device&createdTime=1675278885122&allowXTenantAccess=false).

source/codegen/metadata/nirfmxgsm/config.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+config = {
+    'api_version': '21.0.0',
```
reckenro, line 3: [Minor] Is this the right version of nirfmxgsm you're pulling in here? This gets pulled in as a comment at the top of the generated .proto file.

source/tests/system/nirfmxgsm_driver_api_tests.cpp
```diff
@@ -0,0 +1,82 @@
...
+    device_server_->ResetServer();
+  }
+
+  const std::unique_ptr<NiRFmxGSM::Stub>& stub() const
+  {
+    return stub_;
+  }
+
+  template <typename TResponse>
+  void EXPECT_SUCCESS(const nidevice_grpc::Session& session, const TResponse& response)
```
reckenro, line 46: This might be more helpful later as more tests are added, but there's some Macros in "rfmx_macros.h" for EXPECT_SUCCESS and related helpers you might want to re-use going forward instead of this.

source/tests/system/nirfmxgsm_session_tests.cpp
```diff
@@ -0,0 +1,185 @@
...
+  EXPECT_EQ(init_response_one.instrument().name(), init_response_two.instrument().name());
+
+  nidevice_grpc::Session session_one = init_response_one.instrument();
+  nidevice_grpc::Session session_two = init_response_two.instrument();
+  rfmxgsm::CloseResponse close_response_one, close_response_two;
+  status_one = call_close(session_one, false, &close_response_one);
+  EXPECT_TRUE(status_one.ok());
+  EXPECT_EQ(0, close_response_one.status());
+
+  try {
```
reckenro, line 156: More of an FYI as I think this is fine, but you can use EXPECT_THROW as an alternative to a try-catch with a `FAIL()` assertion. It gets kind of messy if you want to validate properties on the exception but here's [an example](https://github.com/ni/grpc-device/blob/ca6a68aa7a18f41419e33f70c75ac9c95e9ad732/source/tests/unit/session_repository_tests.cpp#L162) in some of our unit tests that do it.

reckenro (approved):

source/tests/system/nirfmxgsm_driver_api_tests.cpp
phindman: I based this file on an older version of nirfmxlte_driver_api_tests.cpp, so I'll update it.

source/tests/system/nirfmxgsm_session_tests.cpp
phindman: I'm not sure I could use EXPECT_THROW in any of these places without it getting a little messy.  The current session tests for other RFmx personalities also use FAIL() in this way, so I'm going to leave this as is.

source/codegen/metadata/nirfmxgsm/config.py
phindman: I changed the version to 23.0.0 in commit 888d28ec40e24341c1968e9deddf4901c236ab5c

source/tests/system/nirfmxgsm_driver_api_tests.cpp
phindman: Updated in commit 5df2e1418bf5d0c607d07e769eae02c2aa604918

astarche (approved):

PR #842, opened 2023-02-02
Expose restricted services on the server.

reckenro:

Exposes restricted services on the server. Rely on "code_readiness" for whether to enable / disable services accross the board.
- Introduced separation of "code_readiness" and "is_restricted" in the config.py files
- Updated all config.py files with "code_readiness": "RestrictedRelease" to have both `"code_readiness": "Release"` and `"is_restricted": True`
- In places determining which proto files / services to advertise key off code_readiness **and** the new "is_restricted" config properties.
- The checking against a header / creating that compilation_test file now keys off the new "is_restricted" config setting as well to keep the previous behavior for restricted services
- Strip out CodeReadiness enum values `kRestrictedRelease` and `kRestrictedNextRelease`

Internal clients need access to some of these services so this will simplify that use case but more importantly soon some driver LV APIs will expect these restricted services to be present/accessible so this seems like a good default instead of expecting the client of those LV APIs to configure the grpc-device server themselves.

We'll continue not to advertise the restricted proto files / services (in the GitHub releases or installer). Internal clients that use them will continue to get the restricted proto files through the export stream we have set up.

Build and unit tests pass.


source/tests/unit/server_configuration_parser_tests.cpp
```diff
@@ -613,7 +613,7 @@ TEST_P(ServerConfigurationParserCodeReadinessTests, CodeReadinessConfiguration_P
   EXPECT_EQ(std::get<1>(GetParam()), readiness);
 }
 
-TEST_P(ServerConfigurationParserCodeReadinessTests, CodeReadinessConfiguration_ParseFeatureToggles_FeatureOfCorrespodingReadinessIsEnabled)
+TEST_P(ServerConfigurationParserCodeReadinessTests, CodeReadinessConfiguration_ParseFeatureToggles_FeatureOfCorrespondingReadinessIsEnabled)
```
reckenro, line 616: Note: This was a typo fix but also points out that these tests for parsing the "code_readiness" exist (this test and the one directly above) including "RestrictedRelease" that's in the new `localhost_restricted_config.json` file.

reckenro:
Alternative: We could also interpret that "code_readiness" field being blank in the `server_conifg.json` file as meaning RestrictedRelease and avoid making a new .json file. Thoughts?

source/server/server_configuration_parser.cpp
```diff
@@ -164,12 +164,12 @@ FeatureToggles::CodeReadiness ServerConfigurationParser::parse_code_readiness()
       using ReadinessMap = std::unordered_map<std::string, FeatureToggles::CodeReadiness>;
       const auto READINESS_MAP = ReadinessMap{
           {"release", FeatureToggles::CodeReadiness::kRelease},
-          {"restrictedrelease", FeatureToggles::CodeReadiness::kRestrictedRelease},
-          {"restricted_release", FeatureToggles::CodeReadiness::kRestrictedRelease},
+          {"restrictedrelease", FeatureToggles::CodeReadiness::kRelease},
```
reckenro, line 167: I think separating the restricted release stuff out of the CodeReadiness concept is correct. But for the config files, should we still accept the "restricted" strings to support already created config files?

SparkingSpork (approved):

astarche (approved):

source/server/server_configuration_parser.cpp
reckenro: I'm going to say yes for now. We can revisit this later if it's deemed confusing and we don't care about supporting anyone who happened to really on "restricted.." for code_readiness in their config files.

SparkingSpork: Just had a thought on this - If you want to go above and beyond, you could print a warning message when a configuration file is using these indicating they're deprecated. I don't know how much effort that would be so feel free to just take this approach.

source/codegen/stage_client_files.py
```diff
@@ -42,7 +47,9 @@ def is_release_ready(self, module_path: Path) -> bool:
         if module_name not in self._module_to_readiness:
             raise KeyError(f"No module config.py metadata for module_name: {module_path}")
 
-        return self._module_to_readiness[module_name] == "Release"
+        return (
+            not self._module_to_restricted and self._module_to_readiness[module_name] == "Release"
```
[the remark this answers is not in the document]
reckenro: Oh yes, oops. Updated that.

And I should've ran this file before. Anyway, ran it after this update with and without the extra `--ignore-release-readiness` and it produces the same results as before.

PR #844, opened 2023-02-07
Include meas_wfm_size in niscope.proto for FetchArrayMeasurement.

reckenro:

Adds `meas_wfm_size` to the `FetchArrayMeasurementRequest`.
Mark the new `meas_wfm_size` as `optional` in the proto request.

Updates custom service handler to
1. Use the request's `meas_wfm_size` was set to a non-negative value, use it.
2. If it wasn't set or if it was set to a negative number, get the size using `ActualMeasWfmSize`.

This makes it a backwards compatible change and provides the option for a 0-fetch request.

[AB#2292398](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2292398)

While we were implementing the LV SCOPE APIs to go through grpc-device we found scenarios in their tests where they were passing in a different `meas_wfm_size` to `FetchArrayMeasurement` than what we'd get from calling `ActualMeasWfmSize`. We couldn't forward that `meas_wfm_size` input before since it wasn't exposed on the .proto file and thus we got incorrect behavior for these legitimate client scenarios.

Build passes.
Scope system tests pass including new one exercising setting or not setting `meas_wfm_size`.

### DIFF
--- a/generated/niscope/niscope.proto
+++ b/generated/niscope/niscope.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-SCOPE API metadata version 23.0.0f168
+// This file is generated from NI-SCOPE API metadata version 23.3.0d9999
 //---------------------------------------------------------------------
 // Proto file for the NI-SCOPE Metadata
 //---------------------------------------------------------------------
@@ -473,8 +473,8 @@ enum ExportableSignals {
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_ADVANCE_EVENT = 6;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_START_EVENT = 7;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_REF_EVENT = 10;
-  EXPORTABLE_SIGNALS_NISCOPE_VAL_REF_CLOCK = 100;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_5V_OUT = 13;
+  EXPORTABLE_SIGNALS_NISCOPE_VAL_REF_CLOCK = 100;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_SAMPLE_CLOCK = 101;
 }
 
@@ -1439,6 +1439,7 @@ message FetchArrayMeasurementRequest {
     ArrayMeasurement array_meas_function = 4;
     sint32 array_meas_function_raw = 5;
   }
+  sint32 meas_wfm_size = 6;
 }
 
 message FetchArrayMeasurementResponse {
--- a/generated/niscope/niscope.proto
+++ b/generated/niscope/niscope.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-SCOPE API metadata version 23.3.0d9999
+// This file is generated from NI-SCOPE API metadata version 23.3.0d127
 //---------------------------------------------------------------------
 // Proto file for the NI-SCOPE Metadata
 //---------------------------------------------------------------------
--- a/generated/niscope/niscope.proto
+++ b/generated/niscope/niscope.proto
@@ -1440,6 +1440,7 @@ message FetchArrayMeasurementRequest {
     sint32 array_meas_function_raw = 5;
   }
   sint32 meas_wfm_size = 6;
+  bool info_only = 7;
 }
 
 message FetchArrayMeasurementResponse {
--- a/generated/niscope/niscope.proto
+++ b/generated/niscope/niscope.proto
@@ -1439,8 +1439,7 @@ message FetchArrayMeasurementRequest {
     ArrayMeasurement array_meas_function = 4;
     sint32 array_meas_function_raw = 5;
   }
-  sint32 meas_wfm_size = 6;
-  bool info_only = 7;
+  optional sint32 meas_wfm_size = 6;
 }
 
 message FetchArrayMeasurementResponse {
--- a/source/codegen/metadata/niscope/config.py
+++ b/source/codegen/metadata/niscope/config.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-SCOPE API metadata version 23.0.0f168
+# This file is generated from NI-SCOPE API metadata version 23.3.0d9999
 config = {
     'additional_headers': {
         'custom/ivi_errors.h': [
             'service.cpp'
         ]
     },
-    'api_version': '23.0.0f168',
+    'api_version': '23.3.0d9999',
     'c_function_prefix': 'niScope_',
     'c_header': 'niScopeCal.h',
     'close_function': 'Close',
--- a/source/codegen/metadata/niscope/config.py
+++ b/source/codegen/metadata/niscope/config.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-SCOPE API metadata version 23.3.0d9999
+# This file is generated from NI-SCOPE API metadata version 23.3.0d127
 config = {
     'additional_headers': {
         'custom/ivi_errors.h': [
             'service.cpp'
         ]
     },
-    'api_version': '23.3.0d9999',
+    'api_version': '23.3.0d127',
     'c_function_prefix': 'niScope_',
     'c_header': 'niScopeCal.h',
     'close_function': 'Close',
--- a/source/custom/niscope_service.custom.cpp
+++ b/source/custom/niscope_service.custom.cpp
@@ -228,8 +228,10 @@ void CheckStatus(int status)
         break;
     }
 
-    ViInt32 measurement_waveform_size;
-    CheckStatus(library_->ActualMeasWfmSize(vi, array_meas_function, &measurement_waveform_size));
+    ViInt32 measurement_waveform_size = request->meas_wfm_size();
+    if (measurement_waveform_size < 0) {
+      CheckStatus(library_->ActualMeasWfmSize(vi, array_meas_function, &measurement_waveform_size));
+    }
 
     ViInt32 num_waveforms;
     CheckStatus(library_->ActualNumWfms(vi, channel_list, &num_waveforms));
--- a/source/custom/niscope_service.custom.cpp
+++ b/source/custom/niscope_service.custom.cpp
@@ -228,11 +228,14 @@ void CheckStatus(int status)
         break;
     }
 
-    ViInt32 measurement_waveform_size = request->meas_wfm_size();
-    if (request->info_only()) {
-      measurement_waveform_size = 0;
+    // meas_wfm_size is marked optional in the proto file.
+    // If it was set to a non-negative value, use it.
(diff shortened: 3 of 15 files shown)


generated/niscope/niscope.proto
```diff
@@ -473,8 +473,8 @@ enum ExportableSignals {
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_ADVANCE_EVENT = 6;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_START_EVENT = 7;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_READY_FOR_REF_EVENT = 10;
-  EXPORTABLE_SIGNALS_NISCOPE_VAL_REF_CLOCK = 100;
   EXPORTABLE_SIGNALS_NISCOPE_VAL_5V_OUT = 13;
+  EXPORTABLE_SIGNALS_NISCOPE_VAL_REF_CLOCK = 100;
```
reckenro, line 477: Note: This should be the last time this enum value bounces around. I fixed hapigen metadata that was causing this enum value's position to bounce around.

generated/niscope/niscope.proto
```diff
@@ -1439,6 +1439,7 @@ message FetchArrayMeasurementRequest {
     ArrayMeasurement array_meas_function = 4;
     sint32 array_meas_function_raw = 5;
   }
+  sint32 meas_wfm_size = 6;
```
reckenro, line 1442: Note: Didn't need any special metadata to get this param in the right spot which was nice so this shouldn't be a breaking change.

source/codegen/metadata/niscope/config.py
```diff
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-SCOPE API metadata version 23.0.0f168
+# This file is generated from NI-SCOPE API metadata version 23.3.0d9999
```
reckenro, line 2: await: This is coming from the output of the local changes in this [AzDo PR](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/434965). Once that goes in I'll update this to come from an actual export.

source/custom/niscope_service.custom.cpp
```diff
@@ -228,8 +228,10 @@ ::grpc::Status NiScopeService::FetchArrayMeasurement(::grpc::ServerContext* cont
         break;
     }
 
-    ViInt32 measurement_waveform_size;
-    CheckStatus(library_->ActualMeasWfmSize(vi, array_meas_function, &measurement_waveform_size));
+    ViInt32 measurement_waveform_size = request->meas_wfm_size();
+    if (measurement_waveform_size < 0) {
```
reckenro, line 232: I _think_ this is the desired behavior. @alexdubois-ni , do you happen to know if this is correct / in line with expectations? nimi-python, for example, will use `ActualMeasWfmSize` if the input is `None`. But here, the default for `meas_wfm_size` on the request will be 0.

Should I change this check to be `if (measurement_waveform_size < 0)`? Or is 0 an acceptable input for `meas_wfm_size`?

alexdubois-ni: If I had to guess, the nimi_python behavior is likely just for extra added usability, and not a part of the direct C API behavior (@marcoskirsch may know). In the end, grpc_server should just be calling down into the C API right? In this case, I don't think we need a special case, and we can let FetchArrayMeasurement perform its usual behavior (not immediately sure if that would just be to error or to attempt to return what is available), unless I'm missing something here.

reckenro: Hmm. I think there's an expectation that passing -1 behaves similar to passing -1 for `numSamples` in the other Fetch calls and will return all available samples. If I pull out the conditional and pass -1 through to the C API we'll get an error back. I feel like handling this case at this layer feels more appropriate than doing it at the grpc-device proxy layer... (avoids an extra grpc call client side to `ActualMeasWfmSize`) and in general making the client make that extra call.

bkeryan: If existing clients get 0 samples back because they didn't explicitly specify -1, then this seems like a breaking change.

reckenro: Yeah, it's definitely functionally breaking even if it's still wire-compatible and source-code compatible. Which is why I'm hoping `meas_wfm_size` being 0 isn't a legitimate case so the default/unspecified still gets the all available samples behavior but we'll see.

alexdubois-ni: Unfortunately, I don't see any immediate documentation for the C entrypoint that claims whether this is a valid case to pass in 0. I see some potential handling of this case when I dig into the C entrypoint though:

- https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/niscope/niscopeBaseDistribution/niScopeCommon/source/niScopeCommon/implMeasurements.cpp&version=GC4c97276e12d37fd3ab279e878c8635e1632aa9cd&line=1237&lineEnd=1238&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents. Passes this down into the following I believe once you follow it down to the underlying implementation:
- https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/niscope/niscopeBaseDistribution/niScopeCommon/source/niScopeCommon/baseMeasurements.cpp&version=GC4c97276e12d37fd3ab279e878c8635e1632aa9cd&line=268&lineEnd=269&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents, at this point, it seems to detect that if it's set to 0, then try to set the size to fetch to be all available samples.

I'd recommend digging around the C entrypoint by calling it with 0, or at least ensuring all cases in the LabVIEW API and Python/gRPC match your expectations are well tested with this input to match expectations. But, this is my immediate expectation.

alexdubois-ni: Perhaps @ni-balexand or @ni-jfitzger know more about the expectations of this entrypoint.

astarche (approved):

source/custom/niscope_service.custom.cpp
ni-jfitzger: The measurement library is one of my least-interacted-with features. 
I do know that a 0-sample fetch (regular non-measurement fetch) is possible. In such a case, we return just timestamps. We may or may not return absoluteInitialX, relativeInitialX for a measurement fetch. I don't know. I recommend trying this out with some real hardware.

reckenro: Yeah, it seems like a 0-sample fetch is a legitimate scenario. But we don't want to break backwards compatibility (especially now that nimi-python is consuming grpc-device).

I talked to Alex Starche offline some and came up with adding an additional `info_only` proto only parameter if the client wants to fall into that 0-sample fetch case.

So now this section of the custom server implementation looks like
```cpp
    ViInt32 measurement_waveform_size = request->meas_wfm_size();
    if (request->info_only()) {
      measurement_waveform_size = 0;
    }
    else if (measurement_waveform_size <= 0) {
      CheckStatus(library_->ActualMeasWfmSize(vi, array_meas_function, &measurement_waveform_size));
    }
```

This supports backwards-compatibility since it will maintain previous behavior of not specifying `meas_wfm_size` using the `ActualMeasWfmSize` since `info_only` defaults to False if not specified and `meas_wfm_size` defaults to 0 if not specified.

It also supports the 0-sample fetch case if the user sets `info_only` on the request to true.

marcoskirsch: The `info_only` idea seems like it would work, but it's also feels pretty hacky.

> (avoids an extra grpc call client side to ActualMeasWfmSize)

What if we have the extra round trip? How big of a deal is that?

marcoskirsch: Reminder – this is a public repo, and links to dev.azure.com or AzDO work items aren't reachable by people who can access these PRs. If possible, I think it's nice to put a bit more context here for everyone to read.

bkeryan: A more systematic way to avoid breaking compatibility is to wrap the new field in a message such as [`google.protobuf.Int32Value`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/wrappers.proto) to allow distinguishing between 0 and unset.

Adding a new method (e.g. `FetchArrayMeasurement2` or `FetchArrayMeasurementEx`) would also work.

ni-jfitzger: @reckenro 
[nimi-python will error if we try to set a measurement waveform size other than None](https://github.com/ni/nimi-python/blob/master/generated/niscope/niscope/session.py#L2228) for gRPC, so I don't think we need to worry about breaking backwards compatibility with nimi-python.

```Python
        if isinstance(self._interpreter, _library_interpreter.LibraryInterpreter):
            if meas_wfm_size is None:
                meas_wfm_size = self._actual_meas_wfm_size(array_meas_function)
        else:
            if meas_wfm_size is not None:
                raise ValueError('The argument "meas_wfm_size" must be None when using gRPC.')
```

bkeryan: @ni-jfitzger GrpcStubInterpreter.fetch_array_measurement ignores the `measurement_waveform_size` parameter and does not set it on the `FetchArrayMeasurementRequest` message: https://github.com/ni/nimi-python/blob/master/generated/niscope/niscope/_grpc_stub_interpreter.py#L252

```
    def fetch_array_measurement(self, channel_list, timeout, array_meas_function, measurement_waveform_size):  # noqa: N802
        response = self._invoke(
            self._client.FetchArrayMeasurement,
            grpc_types.FetchArrayMeasurementRequest(vi=self._vi, channel_list=channel_list, timeout=timeout, array_meas_function_raw=array_meas_function.value),
        )
        return response.meas_wfm, [waveform_info.WaveformInfo(x) for x in response.wfm_info]
```

ni-jfitzger: Why would we do a 0-sample fetch when meas_wfm_size is unset? We should only do that if it's 0. Are you saying it defaults to 0?

reckenro: Yes, in gRPC unset numeric types default to 0: https://developers.google.com/protocol-buffers/docs/proto3#default

I'll explore the options @bkeryan laid out above, though, as an alternative to this additional `info_only` param since currently there's no way to distinguish the `sint32` `meas_wfm_size` being unset or explicitly set to 0.

reckenro: Thanks @bkeryan for the recommendations. I didn't realize that optional tag was brought to proto3. I think that's the best option.

I updated to mark the new `meas_wfm_size` as optional in the proto file and handle it in the following matter in the service handler:
1. If it was set to a non-negative value, use it.
2. If it wasn't set **or** if it was set to a negative number, use the ActualMeasWfmSize.

astarche (approved):

source/codegen/metadata/niscope/config.py
reckenro: Updated to pull from 23.3.0d127

alexdubois-ni (approved):
Approach seems reasonable to me, I am not a code expert in grpc-device, so I will let the repo owners review that portion :)

PR #849, opened 2023-02-13
Add system level spot test for RFmx WCDMA

zhqinbo:

Add RFmx WCDMA system level spot test cases in system API test.

[AB#2280303](https://ni.visualstudio.com/DevCentral/_workitems/edit/2280303)

Run RFmx WCDMA system session test cases and system API test cases in Windows with RFmx 23.00 version driver. All test cases pass
image


phindman:
LGTM

source/tests/system/nirfmxwcdma_driver_api_tests.cpp
```diff
@@ -267,8 +267,8 @@ TEST_F(NiRFmxWCDMADriverApiTests, CdaFromExample_FetchData_DataLooksReasonable)
   EXPECT_GT(0.0, cda_fetch_symbol_evm_response.mean_symbol_power());
   EXPECT_NE(0.0, cda_fetch_symbol_evm_response.chip_rate_error());
   EXPECT_SUCCESS(session, cda_fetch_code_domain_power_response);
-  EXPECT_LT(0.0, cda_fetch_code_domain_power_response.total_power());
-  EXPECT_EQ(0.0, cda_fetch_code_domain_power_response.total_active_power());
+  EXPECT_LE(0.0, cda_fetch_code_domain_power_response.total_power());
```
zhqinbo, line 270: I just find the power value is not a fixed one. If you run the test cases for 10 more times, sometimes you will get a different value.  I changed the assertion to tolerate expected.

phindman (approved):

source/tests/system/nirfmxwcdma_driver_api_tests.cpp
reckenro: Yeah, that seems reasonable but it seems to have still failed on the PR build with:

"...\nirfmxwcdma_driver_api_tests.cpp(271): error: Expected: (0.0) <= (cda_fetch_code_domain_power_response.total_active_power()), actual: 0 vs -4.76837e-07"

So might need a little more tweaking to tolerate the range of values we're getting for the response. I know also in the past, we used EXPECT_NEAR to handle cases where a response always hovered around a specific value but if it's a wide range we may just have to change what we're asserting on. Here's some documentation on [EXPECT_NEAR](https://github.com/google/googletest/blob/main/docs/reference/assertions.md#expect_near-expect_near) if you go down that route.

reckenro (approved):
Looks good to me! I would just work on that CdaFromExample_FetchData_DataLooksReasonable test a little more to see it not fail on the PR build. Looks like the assertions are a little too strict still at the moment.

source/tests/system/nirfmxwcdma_driver_api_tests.cpp
zhqinbo: Thanks. Update the test with EXPECT_NEAR()

PR #853, opened 2023-02-15
Update niRFmxLTE to 23.0

WesleyTangNationalInstruments:

Updates niRFmxLTE support to 23.0 API. Metadata files were sourced from `grpc-device-scrapigen` (see[ this PR](https://github.com/ni/grpc-device-scrapigen/pull/109)).

The adds the following:

- Enums
     - `SemUplinkMaskType`
          -   `SEM_UPLINK_MASK_TYPE_NS28`, `SEM_UPLINK_MASK_TYPE_CANS09`, `SEM_UPLINK_MASK_TYPE_CANS10`
          - Note: `SEM_UPLINK_MASK_TYPE_NS27` was changed to `SEM_UPLINK_MASK_TYPE_NS27_OR_NS43`
     - `MODACC_SYMBOL_CLOCK_ERROR_ESTIMATION_ENABLED`
     - `MODACC_PRE_FFT_ERROR_ESTIMATION_INTERVAL`
     - `MODACC_TIMING_TRACKING_ENABLED`
     - `MODACC_PHASE_TRACKING_ENABLED`

- Attributes
     -  `MODACC_RESULTS_MAXIMUM_PEAK_IQ_ORIGIN_OFFSET`
     - `MODACC_PRE_FFT_ERROR_ESTIMATION_INTERVAL`
     - `MODACC_SYMBOL_CLOCK_ERROR_ESTIMATION_ENABLED`
     - `MODACC_TIMING_TRACKING_ENABLED`
     - `MODACC_PHASE_TRACKING_ENABLED`
     - `MODACC_RESULTS_MAXIMUM_PEAK_FREQUENCY_ERROR`

[AB#2299243](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2299243)

Build passes.


source/codegen/metadata/nirfmxlte/enums.py
```diff
@@ -1675,12 +1731,24 @@
                 'value': 7
             },
             {
-                'name': 'NS27',
+                'name': 'NS27_OR_NS43',
```
reckenro, line 1734: I'm worried about this change because it will break backwards compatibility.

I created a small [Draft PR in my fork here](https://github.com/reckenro/grpc-device/pull/26) that shows the grpc-device metadata changes you can make to support duplicate values in an enum in the proto file. I'm less sure about how to address this in scrapigen but I think using the option of allowing duplicate values for this enum will avoid the breaking change at least from the grpc-device perspective. @astarche, thoughts on this?

reckenro (changes requested):
Everything looks good besides the callout for the breaking change. Want to see what Alex thinks and the resolution to that before approving.

source/codegen/metadata/nirfmxlte/enums.py
WesleyTangNationalInstruments: Is introducing a breaking change really a problem grpc-device needs to worry about when it comes to updating driver APIs? If the intention of this code is to essentially be the gRPC API to NI drivers, I would think that if the driver is making the decision to add a new breaking API change means that grpc-device should reflect those changes as accurately as possible rather than making independent changes specific to grpc-device to allow backwards compatibility where the other APIs do not.

For example, it may be the case that an enum name change like this reflects a functionality change behind the scenes and that the old name no longer accurately reflects what the hardware is doing and allowing both the old value and the new value to specify the same new behavior may be misleading or incorrect. I am not familiar enough with the RFmxLTE instrument to know if this particular change from `NS27` to `NS27_OR_NS43` represents such a case.

I think in this case is pretty trivial and easy to fix as suggested, but I want to just push back a little on whether deviating from the driver C API is the correct thing to do, especially because updating RFmxInstr to 23.0 will have many renaming-based breaking changes like this for both attributes and enums.

@JasonC512, do you have any thoughts on this?

reckenro: That's a good point. I think I had some tunnel vision since historically we've tried to avoid breaking changes but the scenarios we've faced so far have moreso been corrections on the grpc-side and not bubbling up from the driver.

I think with that being the case, it seems odd to have to enforce no breaking changes to grpc-device while other derivatives of the Driver API will be breaking ostensibly for good reasons like the scenario you laid out and otherwise it seems like a decision the driver team has made and should be upheld regardless.

I think adding the breaking change here is probably more correct, like you suggest, and we should just clarify it on the PR description and add the 'breaking-change' label to the PR to make sure it's called out. Thanks for pushing back on this.

I'll let @astarche and @JasonC512 chime in though with their thoughts as well to see if this direction makes sense to them too.

astarche: Yeah, taking the name change from RFmx sounds OK to me. Not sure why they decided to break that, but as far as I'm concerned we can preserve their logic.

The effect of the break is the same for grpc as it is for C (old compiled code continues to work likely as expected; old source code breaks and needs easy fixes).

While enum aliases are possible in proto, they're discouraged and cause problems for replay/json use cases. So the mitigation is actually worse in proto. We could add a way to add a new enum entry that maps to a different C numeric value, but we've never pursued that and this doesn't seem like the time for it.

Breakages for grpc-device will get worse as we have more products built on it, so that could be a concern. Since this is not binary breaking, it seems like that would be OK. Unless someone is shipping json replay features.

Overall, I don't know of a reason for grpc to be more conservative than C.

astarche (approved):

reckenro (approved):

PR #859, opened 2023-02-21
Fix RFmx WCDMA API Test failure issue

zhqinbo:
The slot phase API test for RFmx WCDMA may fail due to an unstable assertion check. 

[AB#2280303](https://ni.visualstudio.com/DevCentral/_workitems/edit/2280303)

System Test for RFmx WCDMA

### DIFF
--- a/source/tests/system/nirfmxwcdma_driver_api_tests.cpp
+++ b/source/tests/system/nirfmxwcdma_driver_api_tests.cpp
@@ -639,7 +639,8 @@ TEST_F(NiRFmxWCDMADriverApiTests, SlotPhaseFromExample_FetchData_DataLooksReason
   EXPECT_LT(0.0, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.dx());
   EXPECT_EQ(38400, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit_size());
   EXPECT_EQ(38400, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit().size());
-  // EXPECT_EQ(0.0, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit(0));
+  auto linear_fit_zero = slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit(0);
+  EXPECT_TRUE(isnan(linear_fit_zero) || 0.0 == linear_fit_zero);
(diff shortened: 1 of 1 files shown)


source/tests/system/nirfmxwcdma_driver_api_tests.cpp
```diff
@@ -639,7 +639,7 @@ TEST_F(NiRFmxWCDMADriverApiTests, SlotPhaseFromExample_FetchData_DataLooksReason
   EXPECT_LT(0.0, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.dx());
   EXPECT_EQ(38400, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit_size());
   EXPECT_EQ(38400, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit().size());
-  EXPECT_EQ(0.0, slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit(0));
```
zhqinbo, line 642: @reckenro @astarche, I just find the fetched value of chip_phase_error_linear_fit(0) may be a "nan" value. That cause this assertion fail. The probability of occurrence for such case is very low, may be 5%. But it may block the CI Test. Most of time the value is 0. I don't how to handle such case. Here I comments this line. If you have good suggestion. Please let me know. Thanks!

reckenro: Yeah, I don't have a strong preference in cases like these whether to keep an assertion or not. I don't know how much value there is keeping an assertion so I'll leave it to you.

If you still want to assert on one or the other values occurring you could with `EXPECT_TRUE` like this:
```cpp
auto linear_fit_zero = slot_phase_fetch_chip_phase_error_linear_fit_trace_response.chip_phase_error_linear_fit(0);
EXPECT_TRUE(std::isnan(linear_fit_zero) || 0.0 == linear_fit_zero);
```

And if the assertion's not worth it I'd leave a brief clarification in the comment about why we're skipping over it:
```cpp
// Not asserting on chip_phase_error_linear_fit(0) because it's intermittently nan.
```

astarche (approved):

reckenro (approved):

source/tests/system/nirfmxwcdma_driver_api_tests.cpp
zhqinbo: Thanks for your suggestion. I would like to assert one of the two values occurring. When doing the hardware test, I  check the nan value occurring.

PR #877, opened 2023-03-03
Add system-level spot tests for RFmx TD-SCDMA

phindman:

Add RFmx TD-SCDMA system-level spot test cases that are based on the C examples.

Closes [AB#2316692](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2316692)

Built locally and ran the TDSCDMA tests:
image


source/tests/system/nirfmxtdscdma_driver_api_tests.cpp
```diff
@@ -64,6 +72,565 @@ TEST_F(NiRFmxTDSCDMADriverApiTests, Init_Close_Succeeds)
...
+  EXPECT_SUCCESS(session, client::cfg_frequency_reference(stub(), session, "", FREQUENCY_REFERENCE_SOURCE_ONBOARD_CLOCK, 10E+6));
+  EXPECT_SUCCESS(session, client::cfg_rf(stub(), session, "", 1.91E+9, 0.00, 0.00));
+  EXPECT_SUCCESS(session, client::cfg_iq_power_edge_trigger(stub(), session, "", "0", IQ_POWER_EDGE_TRIGGER_SLOPE_RISING, -20.00, 0.00, TRIGGER_MINIMUM_QUIET_TIME_MODE_AUTO, 16E-6, IQ_POWER_EDGE_TRIGGER_LEVEL_TYPE_RELATIVE, true));
+  EXPECT_SUCCESS(session, client::select_measurements(stub(), session, "", MEASUREMENT_TYPES_PVT , true));
+  EXPECT_SUCCESS(session, client::cfg_midamble_shift(stub(), session, "", MIDAMBLE_AUTO_DETECTION_MODE_MIDAMBLE_SHIFT, 16, 8));
+  EXPECT_SUCCESS(session, client::pvt_cfg_measurement_method(stub(), session, "", PVT_MEASUREMENT_METHOD_NORMAL));
+  EXPECT_SUCCESS(session, client::pvt_cfg_averaging(stub(), session, "", PVT_AVERAGING_ENABLED_FALSE, 10, PVT_AVERAGING_TYPE_RMS));
+  EXPECT_SUCCESS(session, client::initiate(stub(), session, "", ""));
+
+  const auto pvt_fetch_measurement_status_response = client::pvt_fetch_measurement_status(stub(), session, "", 10.00);
```
phindman, line 473: @ThangamV-NI - First, some background: When adding RFmx personalities to ni/grpc-device we have been adding spot tests that are based on the C examples for the personality.  These tests use simulated devices (a PXI-5663E), and run as part of PR/CI builds of the ni/grpc-device repository.

But I can't get this test, based on the PvT example, to work.  Everything up to this point seems to work (no errors returned) but these pvt_fetch functions always return an error: "-357830: Burst Timing Detection Failed. 0 midamble(s) found.".  The goal of these tests is simply to verify that the grpc-device connection to RFmx TD-SCDMA is working, rather than verify that RFmx TD-SCDMA itself is working, so I'm considering commenting out everything from here to the end of the test and re-enabling the test, but it would be nice if we could call these functions without error.  Do you know what might be going wrong, or have advice on what to try?

ThangamV-NI: Hi, the error from Tdscdma Pvt Fetch is expected on simulated hardware. You need actual hardware with proper waveform generation to make the Pvt Fetch to work. So, you can comment out the piece of tests.

reckenro (approved):

source/tests/system/nirfmxtdscdma_driver_api_tests.cpp
phindman: Thanks @ThangamV-NI, that is reassuring.

I decided to leave this test as is rather than comment out the code that we expect to fail and enable the test.  This way it still compiles that code (which helps verify that the proto file is correct) but since the test is disabled the test output doesn't give the false impression that the test fully passed.

PR #880, opened 2023-03-09
Update DAQmx enums and attributes scrapigen metadata & generate proto files

vigkre:

-  Created this PR to test the `grpc-device` with the updated scrapigen metadata (enums.py and attributes.py)

### Notes:

- Added `ENUMS_SHOULD_BE_USED` suppression for enums which is not used in` grpc-device` but are required in `nidaqmx-python` files.
- Added `Optional` parameters in metadata_validation.py file to align with newly added fields in `enums.py` and `attributes.py` metadata files

### Testing done
- Regenerated the files in `generated/nidaqmx` directory.
- No change in Proto files 😊


generated/nidaqmx/nidaqmx.proto
```diff
@@ -6109,7 +6109,6 @@ enum TriggerInt32AttributeValues {
   TRIGGER_INT32_TRIGGER_TYPE4_TIME = 15996;
   TRIGGER_INT32_TRIGGER_TYPE4_NONE = 10230;
   TRIGGER_INT32_TRIGGER_TYPE5_DIG_EDGE = 10150;
-  TRIGGER_INT32_TRIGGER_TYPE5_SOFTWARE = 10292;
```
bkeryan, line 6112: This breaks source compatibility.

source/codegen/metadata_validation.py
```diff
@@ -173,11 +189,13 @@ class RULES:
 
 ENUM_SCHEMA = Schema(
     {
+        Optional("python_name"): str,
         "values": [
             {
                 "name": str,
                 "value": Or(str, int, float),
                 Optional("python_name"): str,
+                Optional("python_value_name"): str,
```
bkeryan, line 198: Remove this. We are going to use `python_name` which is already optional.

bkeryan (changes requested):
This looks pretty good, but there are a couple of unexplained compatibility breakages that you should fix.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3328,7 +3333,7 @@ enum TimestampEvent {
   TIMESTAMP_EVENT_START_TRIGGER = 12491;
   TIMESTAMP_EVENT_REFERENCE_TRIGGER = 12490;
   TIMESTAMP_EVENT_ARM_START_TRIGGER = 14641;
-  TIMESTAMP_EVENT_FIRST_SAMPLE_TIMESTAMP = 16130;
+  TIMESTAMP_EVENT_FIRST_SAMP = 16130;
```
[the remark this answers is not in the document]
vigkre: Resolved

source/codegen/metadata/nidaqmx/enums_addon.py
```diff
@@ -5,6 +5,15 @@
 }
 
 enums_validation_suppressions = {
```
astarche, line 7: Consider: Is there a helpful comment that could be added here? I know, in general, the DAQ metadata is used for more than just grpc so contains fields/enums/etc that grpc does not use. Is there anything special about these new fields that helps understand why they are in the metadata but not used by functions or attributes?

astarche (approved):

source/codegen/metadata/nidaqmx/enums_addon.py
bkeryan: Yes. Here are some suggestions for line comments.

CouplingTypes: Overlaps with Coupling1 and Coupling2

DAQmxErrors and DAQmxWarnings: Used by nidaqmx-python

FillMode: Same as GroupBy

Impedance1: Should be a mapped enum, but grpc-device did not support those when this was added and adding it now would break compatibility

Save: Save as SaveOptions

SignalModifiers: Same as InvertPolarity

TaskMode: Same as TaskControlAction

TermCfg: Overlaps with InputTermCfg, InputTermCfg2, etc.

TriggerUsageTypes: Overlaps with TriggerType4-10

WDTTaskAction: Same as WatchdogControlAction

source/codegen/metadata/nidaqmx/enums.py
```diff
@@ -71,6 +71,7 @@
         ]
     },
     'AIMeasurementType': {
```
bkeryan, line 73: Please add "DAQmx" or "nidaqmx" to the PR title.

bkeryan (approved):

reckenro (approved):

PR #881, opened 2023-03-09
[PR0] Update NIDAQmx.h and other nidaqmx metadata files to v23.0

vigkre:
Updates daqmx metadata files to version 23.0

The source files that are updated to the version include,
- NIDAQmx.h 
- enums.py
- attributes.py
- functions.py

The files are updated to the version in this https://github.com/ni/grpc-device-scrapigen/pull/136

- Added a suppression logic `ENUMS_SHOULD_BE_USED` to exclude enums which are not used in` grpc-device` but are required in `nidaqmx-python` files.

- Source files were out of date for daqmx, as we have updated the source files for daqmx in scrapigen repo https://github.com/ni/grpc-device-scrapigen/pull/136. We are keeping grpc-device updated to latest version.

### Testing done 
- Regenerated the files in `generated/nidaqmx` directory and add to the PR


bkeryan (changes requested):

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -6957,7 +6992,6 @@
         'returns': 'int32'
     },
     'GetExtendedErrorInfo': {
-        'codegen_method': 'private',
```
bkeryan, line 6960: Removing codegen_method=private from GetExtendedErrorInfo adds this method to nidaqmx.proto. We shouldn't do that because this function must be called on the same thread that produced the error, and gRPC clients can't control which server thread handles each RPC request.

@IPetersNI @reckenro Was this file edited in https://github.com/ni/grpc-device/pull/755 without updating scrapigen?

We probably need a special case in scrapigen to fix this.

source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -6014,7 +6049,7 @@
                 'name': 'stateArray',
                 'size': {
                     'mechanism': 'passed-in-by-ptr',
-                    'value': 'arraySize'
+                    'value': 'arraySizePtr'
```
bkeryan, line 6052: This changes the field name in nidaqmx.proto from `array_size` to `array_size_ptr`, which breaks source compatibility.

None of the other field names in `.proto` files end in `_ptr`:
```
PS C:\dev\grpc-device> rg "_ptr" -t proto --type-add proto:*.proto
generated\nirfmxnr\nirfmxnr.proto
2894:  repeated nidevice_grpc.NIComplexNumberF32 pdsch_ptrs_constellation = 2;
2994:  repeated nidevice_grpc.NIComplexNumberF32 pusch_ptrs_constellation = 2;
```

We may need a special case in scrapigen to fix this.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -2948,6 +2953,1884 @@ enum CurrentUnits2 {
...
+  DA_QMX_ERRORS_REMOTE_SENSE = -209888;
+  DA_QMX_ERRORS_OVER_TEMPERATURE_PROTECTION_ACTIVATED = -209887;
+  DA_QMX_ERRORS_MULTI_TASK_CFG_SAMP_RATE_NOT_SUPPORTED_WITH_PROP_SET = -209886;
+  DA_QMX_ERRORS_MULTI_TASK_CFG_SAMP_RATE_CONFLICTING_PROP = -209885;
+  DA_QMX_ERRORS_NO_COMMON_SAMP_RATE_FOUND_NO_REPEAT_SAMPS = -209884;
+  DA_QMX_ERRORS_NO_COMMON_SAMP_RATE_FOUND = -209883;
+  DA_QMX_ERRORS_MULTI_TASK_CFG_DOES_NOT_SUPPORT_MULTI_DEV_TASK = -209882;
+  DA_QMX_ERRORS_MULTI_TASK_SAMP_RATE_CFG_NOT_SUPPORTED = -209881;
+  DA_QMX_ERRORS_DEBUG_SESSION_NOT_ALLOWED_TIMING_SOURCE_REGISTERED = -209880;
+  DA_QMX_ERRORS_DEBUG_SESSION_NOT_ALLOWED_WHEN_LOGGING = -209879;
```
bkeryan, line 2986: We don't need the error/warning enums in `nidaqmx.proto`.

Also, the enum values are converted to snake_case incorrectly: `DA_QMX_...`

We may need a special case in the grpc-device codegen to exclude these.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3066,8 +4949,8 @@ enum GpsSignalType1 {
 }
 
 enum GroupBy {
-  GROUP_BY_GROUP_BY_CHANNEL = 0;
-  GROUP_BY_GROUP_BY_SCAN_NUMBER = 1;
+  GROUP_BY_CHANNEL = 0;
+  GROUP_BY_SCAN_NUMBER = 1;
 }
```
bkeryan, line 4973: This breaks source compatibility. (However, if it didn't, I would prefer the shorter names.)

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3135,7 +5018,7 @@ enum LogicFamily {
   LOGIC_FAMILY_UNSPECIFIED = 0;
   LOGIC_FAMILY_2POINT_5_V = 14620;
   LOGIC_FAMILY_3POINT_3_V = 14621;
-  LOGIC_FAMILY_5_V = 14619;
+  LOGIC_FAMILY_5V = 14619;
 }
```
bkeryan, line 5041: This breaks source compatibility and it's inconsistent with the other values in this enum.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3672,7 +5570,7 @@ enum ChannelInt32AttributeValues {
   CHANNEL_INT32_LEVEL1_LOW = 10214;
   CHANNEL_INT32_LOGIC_FAMILY_2POINT_5_V = 14620;
   CHANNEL_INT32_LOGIC_FAMILY_3POINT_3_V = 14621;
-  CHANNEL_INT32_LOGIC_FAMILY_5_V = 14619;
+  CHANNEL_INT32_LOGIC_FAMILY_5V = 14619;
```
bkeryan, line 5592: Same as `LOGIC_FAMILY_5V` above.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3852,9 +5751,9 @@ enum DeviceInt32AttributeValues {
   DEVICE_INT32_ANT_STATUS_ABSENT = 15994;
   DEVICE_INT32_ANT_STATUS_OVERCURRENT = 15995;
   DEVICE_INT32_BUS_TYPE_PCI = 12582;
-  DEVICE_INT32_BUS_TYPE_PC_IE = 13612;
+  DEVICE_INT32_BUS_TYPE_PCIE = 13612;
   DEVICE_INT32_BUS_TYPE_PXI = 12583;
-  DEVICE_INT32_BUS_TYPE_PX_IE = 14706;
+  DEVICE_INT32_BUS_TYPE_PXIE = 14706;
```
bkeryan, line 5781: This breaks source compatibility but arguably it fixes a bug.

@astarche @reckenro

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3987,12 +5888,13 @@ enum PhysicalChannelInt32AttributeValues {
...
+  PHYSICALCHANNEL_INT32_AI_MEASUREMENT_TYPE_POWER = 16201;
   PHYSICALCHANNEL_INT32_AO_OUTPUT_CHANNEL_TYPE_VOLTAGE = 10322;
   PHYSICALCHANNEL_INT32_AO_OUTPUT_CHANNEL_TYPE_CURRENT = 10134;
   PHYSICALCHANNEL_INT32_AO_OUTPUT_CHANNEL_TYPE_FUNC_GEN = 14750;
-  PHYSICALCHANNEL_INT32_AO_POWER_UP_OUTPUT_BEHAVIOR_10322 = 10322;
-  PHYSICALCHANNEL_INT32_AO_POWER_UP_OUTPUT_BEHAVIOR_10134 = 10134;
-  PHYSICALCHANNEL_INT32_AO_POWER_UP_OUTPUT_BEHAVIOR_12527 = 12527;
+  PHYSICALCHANNEL_INT32_AO_POWER_UP_OUTPUT_BEHAVIOR_VOLTAGE = 10322;
+  PHYSICALCHANNEL_INT32_AO_POWER_UP_OUTPUT_BEHAVIOR_CURRENT = 10134;
+  PHYSICALCHANNEL_INT32_AO_POWER_UP_OUTPUT_BEHAVIOR_HIGH_IMPEDANCE = 12527;
```
bkeryan, line 5922: This breaks source compatibility but it fixes a bug.

bkeryan (changes requested):

source/codegen/metadata/nidaqmx/enums.py
```diff
@@ -1245,81 +1276,13187 @@
...
+                'value': 10342
+            },
+            {
+                'name': 'FROM_CUSTOM_SCALE',
+                'value': 10065
+            }
+        ]
+    },
+    'DAQmxErrors': {
+        'force-include': True,
```
bkeryan, line 1364: Where did `force-include` come from? I don't see it in the scrapigen main branch or in https://github.com/ni/grpc-device-scrapigen/pull/136

bkeryan: If I copy the metadata from the scrapigen trunk, `force-include` is removed from 4 enums:

```
diff --git a/source/codegen/metadata/nidaqmx/enums.py b/source/codegen/metadata/nidaqmx/enums.py
index 0961b121..e7516ea5 100644
--- a/source/codegen/metadata/nidaqmx/enums.py
+++ b/source/codegen/metadata/nidaqmx/enums.py
@@ -1361,7 +1361,6 @@ enums = {
         ]
     },
     'DAQmxErrors': {
-        'force-include': True,
         'values': [
             {
                 'documentation': {
@@ -13539,7 +13538,6 @@ enums = {
         ]
     },
     'DAQmxWarnings': {
-        'force-include': True,
         'values': [
             {
                 'documentation': {
@@ -15926,7 +15924,6 @@ enums = {
         ]
     },
     'PowerIdleOutputBehavior': {
-        'force-include': True,
         'values': [
             {
                 'documentation': {
@@ -15945,7 +15942,6 @@ enums = {
         ]
     },
     'PowerOutputState': {
-        'force-include': True,
         'values': [
             {
                 'documentation': {
@@ -18480,4 +18476,4 @@ enums = {
             }
         ]
     }
-}
\ No newline at end of file
+}
```

imports/include/NIDAQmx.h
```diff
@@ -2817,6 +2859,8 @@ int32 __CFUNC     DAQmxCreateCOPulseChanFreq     (TaskHandle taskHandle, const c
 int32 __CFUNC     DAQmxCreateCOPulseChanTime     (TaskHandle taskHandle, const char counter[], const char nameToAssignToChannel[], int32 units, int32 idleState, float64 initialDelay, float64 lowTime, float64 highTime);
 int32 __CFUNC     DAQmxCreateCOPulseChanTicks    (TaskHandle taskHandle, const char counter[], const char nameToAssignToChannel[], const char sourceTerminal[], int32 idleState, int32 initialDelay, int32 lowTicks, int32 highTicks);
 
+int32 __CFUNC     DAQmxCreateAIPowerChan       (TaskHandle taskHandle, const char physicalChannel[], const char nameToAssignToChannel[], float64 voltageSetpoint, float64 currentSetpoint, bool32 outputEnable);
```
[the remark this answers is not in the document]
vigkre: Updated the nidaqmx metadata to the version in https://github.com/ni/grpc-device-scrapigen/pull/136 and added the generated files to the PR

source/codegen/metadata/nidaqmx/enums.py
vigkre: I have added `force-include` based on the comment provided by @reckenro here. https://github.com/ni/grpc-device-scrapigen/pull/112#discussion_r1116023009.

I have updated the code to latest and here I can see only `DAQmxErrors` and `DAQmxWarnings` need to be handled. So, I think we need a special case in the grpc-device codegen to exclude these.

vigkre: I have added `force-include` based on the comment provided by @reckenro here. https://github.com/ni/grpc-device-scrapigen/pull/112#discussion_r1116023009.

I have updated the code to latest and here I can see only `DAQmxErrors` and `DAQmxWarnings` need to be handled. So, I think we need a special case in the grpc-device codegen to exclude these.

source/codegen/metadata/nidaqmx/functions.py
bkeryan: Can we override this in https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nidaqmx/functions_addon.py ?

source/codegen/metadata/nidaqmx/enums.py
reckenro: Oh yes, what Brad said is correct. I should've clarified a bit in that original comment.

If we wanted these enums to be in the .proto file we'd have to either reference them in functions.py/attributes.py or mark the enums 'force-include'.

We don't want to do that, though, so in grpc-device we'll have to either ease up the metadata_validation to not care about enums that are "unused" or add a validation suppression for these specific enums so it skips over this check just for them. I think the latter makes more sense because in general it's nice for most of our drivers to notice enums in enums.py that aren't used and thus not included in the .proto file.

For the suppression, it'll be similar to the one we have for [not allowing duplicate enum values](https://github.com/ni/grpc-device/blob/ad35fedd1975a3efbd84ffda3d3f2934377f0e9a/source/codegen/metadata_validation.py#L359) and that's marked up in the driver's enums_addons.py like [here](https://github.com/ni/grpc-device/blob/0161a4ccedbdba6aa0f53b8f87308a3805a5f2ff/source/codegen/metadata/nixnet/enums_addon.py#L7). You should add a new rule for ENUMS_SHOULD_BE_USED or something and mark up the enums_addon.py to note this suppression for the enums and then we'll skip over throwing the exception [here](https://github.com/ni/grpc-device/blob/ad35fedd1975a3efbd84ffda3d3f2934377f0e9a/source/codegen/metadata_validation.py#L363) if the rule is set for the enum.

source/codegen/metadata/nidaqmx/functions.py
reckenro: I think it was. David suggested a follow up with [this comment](https://github.com/ni/grpc-device/pull/755#issuecomment-1267934620) to update scrapigen but I'm guessing that didn't happen (I didn't follow it after that comment):
> > ... thoughts on if we should update scrapigen for this (I'm guessing yes) and possibly more info on the how.

> Yes, we should. If it would work to just remove the function, then the best path forward is to copy the function from DAQmx.h into DAQmx_exlcude.h. But if it's important to keep it present but mark it private, we'd write special code to do that in src/scrapigen/daqapi/filterfunctions.py.

As for if we want to just patch this grpc-device side (to not leak this into nidaqmx-python) I'm not sure if adding it to functions_addon.py would work. At least currently, those addon files are only used for metdata_validation suppressions. @astarche , does that sound right? I don't remember us actually plugging those `_override_metadata` dictionaries in anywhere.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -68,6 +68,7 @@ service NiDAQmx {
   rpc CreateAIPosEddyCurrProxProbeChan(CreateAIPosEddyCurrProxProbeChanRequest) returns (CreateAIPosEddyCurrProxProbeChanResponse);
   rpc CreateAIPosLVDTChan(CreateAIPosLVDTChanRequest) returns (CreateAIPosLVDTChanResponse);
   rpc CreateAIPosRVDTChan(CreateAIPosRVDTChanRequest) returns (CreateAIPosRVDTChanResponse);
+  rpc CreateAIPowerChan(CreateAIPowerChanRequest) returns (CreateAIPowerChanResponse);
```
reckenro, line 71: I can't comment at the top of the file but these .proto files have a comment for the version which right now is:
```proto
// This file is generated from NI-DAQMX API metadata version 20.0.0
```

Can you update the api_version in [config.py to 23.0.0](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nidaqmx/config.py#L3) to in turn bump it here as well?

I know this isn't apparent so I added an [issue to scrapigen](https://github.com/ni/grpc-device-scrapigen/issues/144) to move these config.py files over to scrapigen as it makes more sense to bump the version over there and copy over like the other main metadata files. But this is for all of the metadata sourced from scrapigen, not just daqmx, so for now I'd say just bump here in grpc-device and hopefully down the line we'll fix it up to move config.py over there.

vigkre: Updated the api_version in config.py file

generated/nidaqmx/nidaqmx.proto
vigkre: Added a validation suppression rule to exclude these. 

https://github.com/ni/grpc-device/pull/881/commits/b23fc4a4e0cdcaaf477b37b8eebe1af046b5df87

source/codegen/metadata/nidaqmx/enums.py
vigkre: Thanks @bkeryan and @reckenro for the suggestions. Sorry that I misunderstood the conversation and added `force-include` property before.

Now, I have removed the same and added a validation suppression rule to exclude these.

https://github.com/ni/grpc-device/pull/881/commits/b23fc4a4e0cdcaaf477b37b8eebe1af046b5df87

generated/nidaqmx/nidaqmx.proto
vigkre: Sure Brad. So, are we good with this change? or we should fix this as well?

generated/nidaqmx/nidaqmx.proto
vigkre: Sure Brad. So, are we good with this change? or we should fix this as well?

generated/nidaqmx/nidaqmx.proto
vigkre: Sure Brad. So, are we good with this change? or we should fix this as well?

source/codegen/metadata/nidaqmx/functions.py
vigkre: Understood @bkeryan . We can add some special case in scrapigen to fix this and the other similar changes that happened.

As these changes happened because of the version update for the source files. Shall we handle these changes and create a separate PR in scrapigen ?

Any thoughts?

source/codegen/metadata_validation.py
```diff
@@ -205,7 +207,10 @@ def validate_metadata(metadata: dict):
         attribute_enums = _get_attribute_enums(metadata)
         used_enums = function_enums.union(attribute_enums)
         for enum_name in metadata["enums"]:
-            _validate_enum(enum_name, used_enums, metadata)
+            if not _rule_is_suppressed(
+                metadata, RULES.ENUMS_SHOULD_BE_USED, ["enums", enum_name]
+            ):
```
reckenro, line 212:
```suggestion
            if not _rule_is_suppressed(metadata, RULES.ENUMS_SHOULD_BE_USED, ["enums", enum_name]):
```
We have Black do a formatting check on our .py files. I believe it's failing the build on this line. Ran black locally on your changes and it seems to prefer one line for this if statement.

vigkre: Thanks for the suggestion Ryan. I have configured black and formatted.

source/codegen/metadata_validation.py
```diff
@@ -31,6 +31,8 @@ class RULES:
     # In general, enums shouldn't have duplicate values. This is helpful for catching typos although
     # there are a few enums that have legitimate duplicates.
     ENUMS_SHOULD_NOT_HAVE_DUPLICATE_VALUES = "ENUMS_SHOULD_NOT_HAVE_DUPLICATE_VALUES"
+    # This is helpful in avoiding enums inclusion in the .proto files.
```
astarche, line 34: This comment isn't helpful. Maybe something like:

> The `ENUMS_SHOULD_BE_USED` validation rule is intended to catch enums that are not mapped to functions or attributes because of incomplete or mishandled metadata. Note that any unreferenced enums will not be included in the proto file. These enums should be reviewed, and if they are correctly included-but-not-referenced and are not required to be in the proto file, the rule can be suppressed.

astarche (approved):

source/codegen/metadata/nidaqmx/functions.py
bkeryan: Please fix this. Please don't check unwanted compatibility breakage into main.

generated/nidaqmx/nidaqmx.proto
bkeryan: No. This change is not okay. It breaks the tests and most of the examples, so it will also break most clients of nidaqmx.proto. Did you run any grpc-device DAQmx tests or examples?

```
PS C:\dev\grpc-device> rg GROUP_BY_GROUP_BY
source\tests\system\nidaqmx_driver_api_tests.cpp
261:    request.set_fill_mode(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
276:    request.set_data_layout(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
287:    request.set_data_layout(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
303:    request.set_fill_mode(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
312:    request.set_data_layout(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
328:    request.set_fill_mode(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
337:    request.set_data_layout(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
351:    request.set_fill_mode(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
362:    request.set_data_layout(GroupBy::GROUP_BY_GROUP_BY_CHANNEL);
1148:      GroupBy::GROUP_BY_GROUP_BY_CHANNEL,
1178:      GroupBy::GROUP_BY_GROUP_BY_CHANNEL,

examples\nidaqmx\analog-output.py
83:            data_layout=nidaqmx_types.GROUP_BY_GROUP_BY_CHANNEL,

examples\nidaqmx\digital-output.py
81:            data_layout=nidaqmx_types.GROUP_BY_GROUP_BY_CHANNEL,

examples\nidaqmx\dsa-shared-timebase-and-trig-analog-input-and-output.py
386:                    data_layout=nidaqmx_types.GROUP_BY_GROUP_BY_CHANNEL,
398:                    data_layout=nidaqmx_types.GROUP_BY_GROUP_BY_CHANNEL,
469:                                fill_mode=nidaqmx_types.GroupBy.GROUP_BY_GROUP_BY_CHANNEL,
482:                                fill_mode=nidaqmx_types.GroupBy.GROUP_BY_GROUP_BY_CHANNEL,

examples\nidaqmx\digital-input.py
80:            fill_mode=nidaqmx_types.GROUP_BY_GROUP_BY_CHANNEL,

examples\nidaqmx\analog-input.py
105:            fill_mode=nidaqmx_types.GROUP_BY_GROUP_BY_CHANNEL,

examples\nidaqmx\analog-input-every-n-samples.py
135:                                    fill_mode=nidaqmx_types.GroupBy.GROUP_BY_GROUP_BY_CHANNEL,
```

generated/nidaqmx/nidaqmx.proto
bkeryan: This change is not okay.

generated/nidaqmx/nidaqmx.proto
bkeryan: This change is not okay.

generated/nidaqmx/nidaqmx.proto
bkeryan: I think this change is okay. @astarche @reckenro do you agree?

generated/nidaqmx/nidaqmx.proto
bkeryan: I think this change is okay. @astarche @reckenro do you agree?

bkeryan (changes requested):

generated/nidaqmx/nidaqmx.proto
reckenro: I think this change is okay as well. @maxxboehme , does this sound good to you?

If we do make this change can you add the `source-breaking` label to this PR?

generated/nidaqmx/nidaqmx.proto
reckenro: I think this change is okay as well. @maxxboehme , does this sound good to you?

@vigkre , if we do make this change can you make sure the `source-breaking` label to this PR?

source/codegen/metadata/nidaqmx/functions.py
astarche: I think there's something like `grpc_name` to give a different name in grpc only, if that's the desired result.

generated/nidaqmx/nidaqmx.proto
reckenro: Checked with Maxx offline about this one as well and he agreed it was fine. Added that `source-breaking` label and marking this thread resolved.

source/codegen/metadata/nidaqmx/functions.py
astarche: That's correct. Personally I would add a new tag and have a `metadata_mutation` in `grpc-device` that inserts `codegen_method: private` when that flag is set. Naming is always a challenge for these things. `relies_on_thread_local_state: True` or `is_get_last_error_method: True`. Maybe something like `remote: False` or "remotable: False`, but those are pretty generic and I'm not sure if remotable is a word ...

source/codegen/metadata/nidaqmx/functions.py
reckenro: And in scrapigen, RFmx does it generally for param names like "instrumentHandle" that we just want to be "instrument" in grpc. You can see the scrapigen code for that [here](https://github.com/ni/grpc-device-scrapigen/blob/51d2dccebe162ed1476457003187870c6389f25d/src/scrapigen/rfmxnr/transformfunctions.py#L77). You should be able to do something similarly for this parameter name.

source/codegen/metadata/nidaqmx/functions.py
vigkre: Added a fix in https://github.com/ni/grpc-device-scrapigen/pull/127/commits/2ff22d5b025bf6287e43aff545216a1ba34eca4a to mark the methods private.

Reference: https://github.com/ni/grpc-device-scrapigen/blob/main/src/scrapigen/transform/transformfunctions.py#L458

source/codegen/metadata/nidaqmx/functions.py
vigkre: Added a fix in https://github.com/ni/grpc-device-scrapigen/pull/127/commits/2ff22d5b025bf6287e43aff545216a1ba34eca4a to remove Ptr from the value and name

generated/nidaqmx/nidaqmx.proto
vigkre: Added a fix in https://github.com/ni/grpc-device-scrapigen/pull/127/commits/2ff22d5b025bf6287e43aff545216a1ba34eca4a to remove function that was doing this explicitly

generated/nidaqmx/nidaqmx.proto
vigkre: Added a fix in https://github.com/ni/grpc-device-scrapigen/pull/127/commits/2ff22d5b025bf6287e43aff545216a1ba34eca4a

generated/nidaqmx/nidaqmx.proto
vigkre: Added a fix in https://github.com/ni/grpc-device-scrapigen/pull/127/commits/2ff22d5b025bf6287e43aff545216a1ba34eca4a

reckenro (approved):

PR #891, opened 2023-03-27
Update daqmx metadata and generate proto files

samchris007:
- Update latest scrapigen daqmx metadata files (attributes.py, enums.py, functions.py) from this PR: https://github.com/ni/grpc-device-scrapigen/pull/150 into grpc-device.

Implements [Task 2296058](https://dev.azure.com/ni/DevCentral/_workitems/edit/2296058): Verify if the generated metadata is not affecting the existing generators

Generated the files and can't find any changes in the generated files.


astarche (approved):

reckenro (approved):

reckenro (changes requested):
We're in the process of making the 2.1 release and I don't want anything going into main until we've bumped the exports from main to be 2.2. I'll reset myself and approve after the export version is bumped (should be in a day or so).

source/codegen/metadata_validation.py
```diff
@@ -113,6 +113,14 @@ class RULES:
         Optional("linked_params_are_optional"): bool,
         Optional("mapped-enum"): str,
         Optional("is_optional"): bool,
+        Optional("optional"): bool,
```
bkeryan, line 116: grpc-device uses `is_optional` and nidaqmx-python uses `optional`. Someone is going to mix these up.

source/codegen/metadata_validation.py
```diff
@@ -113,6 +113,14 @@ class RULES:
...
+        Optional("ctypes_data_type"): Or(str, None),
+        Optional("has_explicit_buffer_size"): bool,
+        Optional("is_list"): bool,
```
bkeryan, line 119: FYI, there is `use_array` above, but I'm not sure if it means the same thing. None of the metadata in grpc-device currently uses it.

source/codegen/metadata_validation.py
```diff
@@ -113,6 +113,14 @@ class RULES:
...
+        Optional("python_data_type"): str,
+        Optional("python_type_annotation"): str,
+        Optional("python_description"): str,
+        Optional("default"): Or(str, bool, None),
```
bkeryan, line 123: This has been renamed to `python_default` because the enum values are Python-specific. 

However, I just noticed that this schema supports `default_value`: `Optional("default_value"): Or(str, bool, int, float, None),`

Based on that, I think we should rename this to `python_default_value` and add `int` and `float` to the list of types. That way, it will be a Python-specific override for `default_value`.

bkeryan (changes requested):

source/codegen/metadata_validation.py
samchris007: Sure. Renaming `optional` to `is_optional` in Scrapigen is affecting grpc-device metadata validation logic. 
@bkeryan Instead, Can we rename it to `is_optional_in_python`?

source/codegen/metadata_validation.py
samchris007: Got it. I have created a PR in Scrapigen which renames `python_default` to `python_default_value`. 
https://github.com/ni/grpc-device-scrapigen/pull/155

source/codegen/metadata_validation.py
samchris007: Generally, `is_list` explains whether the parameter is a list or not.
And having key name to `is_list` seems to be more apt than `use_array`.

reckenro (approved):

source/codegen/metadata_validation.py
bkeryan: I think the distinction is pretty much language-specific: in the C API, it's an array, and in the Python API, it's a list. But I don't expect you to change this.

bkeryan (changes requested):

source/codegen/metadata_validation.py
samchris007: Understood. I have renamed to `is_optional_in_python` in this PR https://github.com/ni/grpc-device-scrapigen/pull/155 and updated the metadata here as well.

bkeryan (approved):

PR #893, opened 2023-03-27
Update CMakeLists.txt

christag-ni:

Bumps the version in CMake file

To accurately reflect the version grpc-device is in

No testing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 2.1.0)
+  VERSION 2.2.0)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(CreateVirtualEnvironment)


reckenro (changes requested):
Change looks fine. 

I thought about this more after we discussed it and I don't think we should submit anything into main until we've bumped the exports from main to be 2.2 (in AzDo). This is mainly so if someone gets the latest 2.1 exports they'll get the finals from the release branch.

christag-ni:
> Change looks fine.
> 
> I thought about this more after we discussed it and I don't think we should submit anything into main until we've bumped the exports from main to be 2.2 (in AzDo). This is mainly so if someone gets the latest 2.1 exports they'll get the finals from the release branch.

Should I then create a PR for the ni-central changes, and once those changes are submitted, we submit this change in grpc-device?

reckenro:
> Should I then create a PR for the ni-central changes, and once those changes are submitted, we submit this change in grpc-device?

Yes, sorry for switching these around. But yes, I think we're good to make that AzDo change now that Wes's cherry-pick went in and we're firm that we're not going to make further changes to the releases/2.1 branch.

reckenro (approved):

PR #895, opened 2023-03-30
Add CalibrationOperationsRestrictedService

cumitche:

- Pull DeviceId message out of debugsessionproperties_restricted.proto into its own proto file for reuse.
- Pull SysCfg resource accessor logic out of DebugSessionPropertiesRestrictedService into its own class and unit tests for reuse.
- Add calibration operations restricted proto, service, and unit tests.

This change adds a restricted service to query calibration information from NI System Configuration allowing remote access to external and self-calibration date/times and temperatures.

Created unit tests for calibration operations restricted service.
Ran a locally built executable on a machine with a PXIe-5841 and compared the calibration information available in MAX to the values received from the service in a .NET 6 gRPC client test application.
CalInfo
Note: the times returned by the restricted service are in UTC, hence the offset from the time zone adjusted values show in MAX.


source/protobuf_restricted/calibrationoperations_restricted.proto
```diff
@@ -0,0 +1,40 @@
...
+service CalibrationOperationsRestricted {
+  rpc GetCalibrationInformation(GetCalibrationInformationRequest) returns (GetCalibrationInformationResponse);
+}
+
+message GetCalibrationInformationRequest {
+  DeviceId device_id = 1;
+}
+
+message GetCalibrationInformationResponse {
+  bool supports_internal_calibration_available = 1;
```
reckenro, line 22: For these, you can consider using the `optional` field rule for the singular fields instead of an extra `_available` bool for each. I think it would give similar behavior but am not sure for this restricted service situation if it'd be worth it or not.

source/tests/unit/calibration_operations_tests.cpp
```diff
@@ -0,0 +1,638 @@
...
+  nidevice_restricted_grpc::CalibrationOperationsRestrictedService service(&mock_library);
+  ::grpc::ServerContext context;
+  nidevice_restricted_grpc::GetCalibrationInformationRequest request;
+  nidevice_restricted_grpc::GetCalibrationInformationResponse response;
+  configure_mock_library_calibration_properties_do_not_exist(&mock_library);
+
+  ::grpc::Status status = service.GetCalibrationInformation(&context, &request, &response);
+
+  EXPECT_TRUE(status.ok());
+  EXPECT_FALSE(response.supports_internal_calibration_available());
```
reckenro, line 257: Is there any value in asserting on the defaults set for the actual field in `CalibrationOperationsRestrictedService::GetCalibrationInformation` even when the `_available` field is false? I don't imagine them changing though and I'm assuming they're all correct in that file (False for bools, 0 for numerics, and empty for repeated fields) so may not be any value adding them in the tests.

But thought I'd call it out since the singular fields are explicitly being set in the service if it's available or not.

reckenro (approved):

source/protobuf_restricted/calibrationoperations_restricted.proto
cumitche: Using `optional` instead of `_available` fields is a great idea.  I completely forgot that this capability was built into protocol buffers.

source/tests/unit/calibration_operations_tests.cpp
cumitche: I'll revisit these when I change the fields to `optional`.

source/protobuf_restricted/calibrationoperations_restricted.proto
reckenro: Shoutout to Brad who pointed it out as an option in a recent Scope change. Might be worth noting that it was [added to proto3 with protobuf version v3.15.0 (Early 2021)](https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0) but we added it in niscope.proto and in general all the protoc's we use should be past that version.

cumitche: Added `optional` to all scalar basic type fields.  The repeated and message fields don't need it because it is already possible to determine their field presence.

I built InstrumentStudio with these changes and the version of the gRPC NuGets we are using has this support.

source/tests/unit/calibration_operations_tests.cpp
cumitche: Added checks for default values.

astarche (approved):

reckenro:
> FYI @cumitche , I have a PR up to fix the failing CI: #901

@cumitche , submitted #901 but didn't realize it uses the .github/... actions from the PR branch and not from main so you'll need to merge main to get the CI fixed.

reckenro:
@cumitche , feel free to Squash and merge when you're ready. Charlie's been out. He's back tomorrow but I don't think he needs eyes on this.

Also, the `run_win_system_tests` is pretty busted right now (can't connect to VM to run them on). Have an open Teams channel / Bug to sort it out but I'm not too concerned with your change skipping that check since you're only adding unit tests that run during `build_desktop` steps.

cumitche:
I want to pin InstrumentStudio's dependency on this prior to merging so that I can commit the dependency update and retrofit together, but I haven't been able to get the PR build to succeed for that.  I'll need to wait until tomorrow to merge this.

PR #900, opened 2023-04-03
Update sync_github_issues_to_azdo.yml to new owning squad

reckenro:

Changes the area path for syncing Issues to Bugs in AzDo to reflect the new grpc-device owning squad.

Driver team Customer Facing squad should have Bugs assigned to them.

None. I'll make a fake Issue to make sure it syncs to the right spot after this is merged to main.

### DIFF
--- a/.github/workflows/sync_github_issues_to_azdo.yml
+++ b/.github/workflows/sync_github_issues_to_azdo.yml
@@ -15,7 +15,7 @@ jobs:
           github_token: "${{ secrets.GH_REPO_TOKEN }}"
           ado_organization: "ni"
           ado_project: "DevCentral"
-          ado_area_path: "DevCentral\\Product RnD\\PlatformSW\\Test System SW\\CBS\\HW Config"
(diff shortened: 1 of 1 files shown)


astarche (approved):

.github/workflows/sync_github_issues_to_azdo.yml
bkeryan, line 1: FYI, look at https://github.com/ni/measurementlink-python/blob/main/.github/workflows/sync_github_issues_to_azdo.yml if you want to see how to sync comments and automatically set the work item type to "bug", "technical debt", or "user story" based on GitHub labels.

reckenro: Ah, thanks! I think with this change, though, I'll just update our current one so it points to the proper place.

FYI @maxxboehme  and @danielhuani , in case you want to make further improvements / modifications to the syncing you can look at how MeasurementLink does it.

reckenro:
Ah, it didn't work. @davidcorrigan714 , is there something we have to do with the secrets if we're updating the AzDo path here?

PR #901, opened 2023-04-03
Update github hosted runners from ubuntu-18.04 to ubuntu-latest.

reckenro:

Updates two GitHub actions using `ubuntu-18.04` to use `ubuntu-20.04` instead.

GitHub runners have deprecated ubuntu-18.04 so all actions using them will start failing today: https://github.com/actions/runner-images/issues/6002

Trusting the build.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -27,7 +27,7 @@ jobs:
           }
         - {
             name: "Ubuntu Build", artifact: "Linux.tar.xz",
-            os: ubuntu-18.04,
+            os: ubuntu-latest,
             build_type: "Release", cc: "gcc-9", cxx: "g++-9",
             cmake_platform: "",
             glibc_version: "2_27",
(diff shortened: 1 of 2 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           }
         - {
             name: "Ubuntu Build", artifact: "Linux.tar.xz",
-            os: ubuntu-18.04,
+            os: ubuntu-latest,
```
reckenro, line 30: For both of these, is `ubuntu-latest` appropriate? I assumed so but we can also update it to `ubuntu-20.04` or `ubuntu-22.04`.

reckenro: Actually, I thought about it and probably the smallest version bump here is more desirable so I updated these to `ubuntu-20.04`.

astarche (approved):

PR #909, opened 2023-04-07
Fix mypy violations in examples.

reckenro:

Fixes recent mypy violations popping up for examples. I think this is because mypy bumped to version 1.2.0 and is catching a scenario it previously missed.

The specific failure was complaining that the left operand of a concatenate could be None.

See this [failed validate_python step](https://github.com/ni/grpc-device/actions/runs/4639441570/jobs/8210400991) in a recent PR ci.yml run.

Fixes build by making validate_python.yml pass.

Ran the `python source/codegen/validate_examples.py -e "nidaqmx"` script locally and it now passes with these changes.
Also the `validate_python` step run in this PR's ci.yml run should validate it as well.

### DIFF
--- a/examples/nidaqmx/analog-input-every-n-samples.py
+++ b/examples/nidaqmx/analog-input-every-n-samples.py
@@ -169,7 +169,7 @@ async def _main():
             check_for_warning(stop_task_response)
 
         except grpc.RpcError as rpc_error:
-            error_message = rpc_error.details()
+            error_message = str(rpc_error.details() or "")
             for key, value in rpc_error.trailing_metadata() or []:  # type: ignore
                 if key == "ni-error":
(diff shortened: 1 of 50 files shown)


examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -169,7 +169,7 @@ async def wait_for_done():
             check_for_warning(stop_task_response)
 
         except grpc.RpcError as rpc_error:
-            error_message = rpc_error.details()
+            error_message = str(rpc_error.details() or "")
```
reckenro, line 172: Same change in every file. This makes sure error_message is a string because before when it tried to do `error_message += f"\nError status: {details}"` mypy v1.2.0 was complaining that `error: Unsupported left operand type for + ("None")  [operator]`

PR #917, opened 2023-05-02
Update nidaqmx latest scrapigen metadata 

samchris007:

This PR updates latest scrapigen metadata of nidaqmx and generate files.

Implements [Task 2380733](https://dev.azure.com/ni/DevCentral/_workitems/edit/2380733): Update latest scrapigen metadata into grpc-device

Manually generated the files and verified.


generated/nidaqmx/nidaqmx_library.h
```diff
@@ -147,7 +147,7 @@ class NiDAQmxLibrary : public nidaqmx_grpc::NiDAQmxLibraryInterface {
...
+  int32 GetAnalogPowerUpStates(const char deviceName[], const char channelName[], float64* state, int32 channelType);
```
bkeryan, line 150: This is a regression. These are `execl`-style varargs functions, so they require passing each array element as a separate parameter, plus a nullptr to terminate the list of varargs. Grpc-device handles these functions by generating the library method with a large, fixed number of parameters and then using functions like `get_lines_if` to fill all of the parameters with either valid values or nullptr.

bkeryan (changes requested):

generated/nidaqmx/nidaqmx_library.h
samchris007: Sure. I have reverted those changes by removing `proto_only` field for those parameters in scrapigen in this PR.
https://github.com/ni/grpc-device-scrapigen/pull/169

astarche (approved):

reckenro (approved):

generated/nidaqmx/nidaqmx_library.h
bkeryan: Much better, thanks.

PR #918, opened 2023-05-10
Add additional metadata to NIDAQmx errors

Sakthi-SM:
**What does this Pull Request accomplish?**
This PR updates the `read` and  `write` methods in NIDAQmx to send additional metad
ata when an error occurs.
* In `read` methods, additionally `samps_per_chan_read` will be added to the error info.
* In `write` methods, additionally `samps_per_chan_wrtiten` will be added to the error info.

**Why should this Pull Request be merged?**
Implements [Task 2323110](https://dev.azure.com/ni/DevCentral/_workitems/edit/2323110): Update the nidaqmx gRPC server to return RPC exceptions with data on the number of samples read/written in a read/write operation.

**What testing has been done?**
Manually generated the files and verified the changes made.


generated/nidaqmx/nidaqmx_service.h
```diff
@@ -435,6 +435,8 @@ class NiDAQmxService final : public NiDAQmx::WithCallbackMethod_RegisterSignalEv
   NiDAQmxLibraryInterface* library_;
   ResourceRepositorySharedPtr session_repository_;
   ::grpc::Status ConvertApiErrorStatusForTaskHandle(::grpc::ServerContext* context, int32_t status, TaskHandle task);
+  ::grpc::Status ConvertApiErrorStatusWithReadParametersForTaskHandle(::grpc::ServerContext* context, int32_t status, int32_t samps_per_chan_read, TaskHandle task);
+  ::grpc::Status ConvertApiErrorStatusWithWriteParametersForTaskHandle(::grpc::ServerContext* context, int32_t status, int32_t samps_per_chan_written, TaskHandle task);
```
bkeryan, line 439: "WithReadParameters" implies that there are multiple read parameters, but there is only one.

[style, subjective] After seeing this in context, I think it makes more sense to tack the read parameter (singular) on the end, because it is extra:
```
::grpc::Status ConvertApiErrorStatusForTaskHandleWithSampsPerChanRead(::grpc::ServerContext* context, int32_t status, TaskHandle task, int32_t samps_per_chan_read);
```

source/codegen/metadata/nidaqmx/functions_addon.py
```diff
@@ -1,4 +1,139 @@
 functions_override_metadata = {
+    'ReadAnalogF64': {
+        'python_codegen_method': 'CustomCode_Read',
+    },
+    'ReadAnalogScalarF64': {
+        'python_codegen_method': 'CustomCode_Read_Scalar',
+    },
```
bkeryan, line 7: This feature is not Python-specific. nidaqmx-python is the first DAQmx gRPC client to use these additional gRPC metadata keys, but other DAQmx gRPC clients will also benefit.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -10,4 +10,16 @@ ::grpc::Status NiDAQmxService::ConvertApiErrorStatusForTaskHandle(::grpc::Server
   return nidevice_grpc::ApiErrorAndDescriptionToStatus(context, status, description);
 }
 
+::grpc::Status NiDAQmxService::ConvertApiErrorStatusWithReadParametersForTaskHandle(::grpc::ServerContext* context, int32_t status, int32_t samps_per_chan_read , TaskHandle task)
```
bkeryan, line 13: [style] Don't leave a space before a comma

source/codegen/common_helpers.py
bkeryan, line 1: This code doesn't make it clear that these are DAQmx-specific hacks. I think the best solution is to put the DAQmx-specific details in the metadata, but if you can't do that, please rename/comment in the helpers and template so that it's clear that this is for DAQmx.

bkeryan (changes requested):

generated/nidaqmx/nidaqmx_service.cpp
```diff
@@ -12932,7 +12932,7 @@ namespace nidaqmx_grpc {
       int32 samps_per_chan_read {};
       auto status = library_->ReadAnalogF64(task, num_samps_per_chan, timeout, fill_mode, read_array, array_size_in_samps, &samps_per_chan_read, reserved);
       if (!status_ok(status)) {
-        return ConvertApiErrorStatusForTaskHandle(context, status, task);
+        return ConvertApiErrorStatusWithReadParametersForTaskHandle(context, status, samps_per_chan_read, task);
       }
```
bkeryan, line 12936: On further thought: how much value do the helper functions add, compared to this?
```
      if (!status_ok(status)) {
        context->AddTrailingMetadata("ni-samps-per-chan-read", std::to_string(samps_per_chan_read));
        return ConvertApiErrorStatusForTaskHandle(context, status, task);
      }
```

Duplicating this line adds negligible bloat ((1 virtual method call + 1 function call) x 20-30 functions) and it's all in the `if (!status_ok(status))` path.

generated/nidaqmx/nidaqmx_service.h
Sakthi-SM: Removed these as per the last suggestion

source/codegen/common_helpers.py
Sakthi-SM: Updated the metadata based on the latest suggestion and removed the unnecessary changes

generated/nidaqmx/nidaqmx_service.cpp
Sakthi-SM: As per your suggestion, a simple `boolean` will not be enough as we would still need the logic to figure out the read/write non-scalar functions. To do so, we need the additional metadata in `functions.addon`. Even then the parameter to be used for populating the error varies in each function. To simplify this, we have added a property called `return_on_error_key` in the functions metadata under the corresponding parameter. This will ensure simplicity and the generic approach that we are looking for.

source/codegen/metadata/nidaqmx/functions_addon.py
Sakthi-SM: Removed these additional metadata.

source/codegen/common_helpers.py
```diff
@@ -1186,3 +1186,10 @@ def get_params_needing_initialization(parameters: List[dict]) -> List[dict]:
     * Outputs that are calculated/populated after the API call.
     """
     return [p for p in parameters if not (is_return_value(p) or is_get_last_error_output_param(p))]
+
+
+def get_parameter_for_error_generation(params):
```
reckenro, line 1191: You'll need the `"""` docstring for this public method like the others. That's what's tripping up the validate_python part of the build, it fails when running `ni-python-styleguide` against this.

source/codegen/common_helpers.py
```diff
@@ -1186,3 +1186,10 @@ def get_params_needing_initialization(parameters: List[dict]) -> List[dict]:
...
+    for param in params:
+        if param["direction"] == "out" and "return_on_error_key" in param:
```
reckenro, line 1193: Minor: You can re-use the `is_output_parameter` method defined towards the top of this file.
```suggestion
        if is_output_parameter(param) and "return_on_error_key" in param:
```

source/codegen/templates/service_helpers.mako
```diff
@@ -773,6 +774,9 @@ ${set_response_values(normal_outputs, init_method)}\
   else:
     method_call = f'return ConvertApiErrorStatusFor{cpp_handle_type}(context, status, {session});'
 %>\
+      %if error_parameter is not None:
+      context->AddTrailingMetadata("${error_parameter["return_on_error_key"]}", std::to_string(${common_helpers.pascal_to_snake(error_parameter["name"])}));
+      %endif
```
bkeryan, line 779: I don't think the helper function helps very much. It duplicates the output check and constrains this to supporting one return_on_error_key parameter.

```suggestion
      %for parameter in output_parameters:
      %if "return_on_error_key" in parameter:
      context->AddTrailingMetadata("${parameter["return_on_error_key"]}", std::to_string(${common_helpers.pascal_to_snake(parameter["name"])}));
      %endif
      %endfor
```

source/codegen/templates/service_helpers.mako
```diff
@@ -756,6 +756,7 @@ ${set_response_values(normal_outputs, init_method)}\
   config = data['config']
   input_parameters = [p for p in parameters if common_helpers.is_input_parameter(p)]
   resource_handle_types = service_helpers.get_resource_handle_types(config)
+  error_parameter = common_helpers.get_parameter_for_error_generation(parameters)
```
bkeryan, line 759:
```suggestion
  input_parameters = [p for p in parameters if common_helpers.is_input_parameter(p)]
  output_parameters = [p for p in parameters if common_helpers.is_output_parameter(p)]
  resource_handle_types = service_helpers.get_resource_handle_types(config)
```

generated/nidaqmx/nidaqmx_service.cpp
bkeryan: I agree, specifying the gRPC metadata key name in the API metadata is more generic.

bkeryan (changes requested):

source/codegen/templates/service_helpers.mako
bkeryan: FYI, you could filter using `"return_on_error_key" in p` here too, or filter using `%for` / `%if`. It doesn't matter much either way.

source/codegen/common_helpers.py
Sakthi-SM: Removed this helper method and re-used `is_output_parameter` in mako template.

source/codegen/common_helpers.py
Sakthi-SM: Removed this helper method

source/codegen/templates/service_helpers.mako
Sakthi-SM: Removed the helper method and updated the mako as per the suggestion

astarche (approved):

bkeryan (approved):

PR #923, opened 2023-05-17
examples: Add multi-threaded DAQmx event example

bkeryan:

Add a multi-threaded DAQmx event example using `concurrent.futures`.

Add `-aio` suffix to asyncio examples.

Demonstrate how to use DAQmx events without asyncio. This is closer to what we're doing in nidaqmx-python.

Tested event examples.

Searched multi-threaded example for `await`, `asyncio`, `aio`, etc.

### DIFF
--- a/examples/nidaqmx/analog-input-every-n-samples.py
+++ b/examples/nidaqmx/analog-input-every-n-samples.py
@@ -28,8 +28,8 @@ If they are not passed in as command line arguments, then by default the server 
 "localhost:31763", with "Dev1/ai0" as the physical channel name.
 """
 
-import concurrent.futures
 import sys
+from concurrent.futures import ThreadPoolExecutor
 
 import grpc
 import nidaqmx_pb2 as nidaqmx_types
@@ -51,10 +51,9 @@ def _main():
     client = None
     task = None
 
-    with (
-        grpc.insecure_channel(f"{SERVER_ADDRESS}:{SERVER_PORT}") as channel,
-        concurrent.futures.ThreadPoolExecutor() as executor,
-    ):
+    with grpc.insecure_channel(
+        f"{SERVER_ADDRESS}:{SERVER_PORT}"
+    ) as channel, ThreadPoolExecutor() as executor:
         try:
             client = grpc_nidaqmx.NiDAQmxStub(channel)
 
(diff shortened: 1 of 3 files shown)


examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -47,11 +47,14 @@
...
-async def _main():
+def _main():
     client = None
     task = None
 
-    async with grpc.aio.insecure_channel(f"{SERVER_ADDRESS}:{SERVER_PORT}") as channel:
+    with (
+        grpc.insecure_channel(f"{SERVER_ADDRESS}:{SERVER_PORT}") as channel,
+        concurrent.futures.ThreadPoolExecutor() as executor,
+    ):
```
bkeryan, line 57: Hmm, ok
```
Run ni-python-styleguide lint source/codegen/*.py examples
examples/nidaqmx/analog-input-every-n-samples.py:55:6[7](https://github.com/ni/grpc-device/actions/runs/5008134058/jobs/8975642212?pr=923#step:5:8): E999 SyntaxError: invalid syntax
examples/nidaqmx/analog-input-every-n-samples-aio.py:25:101: W505 doc line too long (101 > 100 characters)
Error: Process completed with exit code 1.
```

bkeryan: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement

> Changed in version 3.10: Support for using grouping parentheses to break the statement in multiple lines.

astarche: That pipeline is still using `3.7`. Should be `3.9` right?

I guess the new syntax would break `mypy` (run by `validate_examples`, for which we exclude daq) but not `styleguide`? Or maybe not even `mypy`.

I think having this example require `3.10` is fine, even if `3.9` is the de facto version. Anyone that's interested in this is likely on newer python or can find their way around. And this syntax is useful, especially for this kind of code.

astarche (approved):

reckenro (approved):

examples/nidaqmx/analog-input-every-n-samples.py
bkeryan: Python 3.7 reaches end-of-life next month, but 3.8 and 3.9 will be around for a while: https://devguide.python.org/versions/#versions

The strange thing is I've been testing with 3.9, so I'm not sure if the version listed in the docs is correct.

Anyway, I don't think this is worth breaking compatibility with 3.8 and 3.9, so I'll update the example.

bkeryan:
This test failure is not Python-related.
```
[ RUN      ] NiFgenSessionTest.InitializeSessionWithDeviceAndNoSessionName_CreatesDriverSession
unknown file: Failure
C++ exception with description "MAX:  (Hex 0x80040303) Internal error: The requested object was not found in the configuration database. Please note the steps you performed that led to this error and contact technical support at http://ni.com/support.

Component Name: 
File Name: 
Line Number: 0

Status Code: -2147220733" thrown in the test body.
[  FAILED  ] NiFgenSessionTest.InitializeSessionWithDeviceAndNoSessionName_CreatesDriverSession (481 ms)
```

PR #927, opened 2023-05-18
codegen: Report DAQmx event registration failure by closing stream

bkeryan:
### What does this Pull Request accomplish?

The NI gRPC Device Server uses server streaming to return a response for each
event generated by DAQmx. However, when a DAQmx event registration function
(like DAQmxRegisterDoneEvent) returns an error, the server reports the error by
sending a streaming response where `response.status` is the error code.

Implications:

- It is difficult for clients to synchronously detect whether event registration
  failed.
  - In the success case, no responses are generated until DAQmx notifies the
     server of an event, so waiting indefinitely for a response only works in the
     error case.
  - Waiting indefinitely for initial metadata does not guarantee that the error
     will be reported at the same time because the server sends initial metadata
     before reporting the error.
 - Rich error descriptions are not reported for event registration errors, only
   an error code.
 - For done events, a fatal `response.status` is ambiguous: it can mean that
   event registration failed or that the acquisition/generation encountered an
   error.
 - The server does not close the stream when a event registration error occurs,
   but it will never send any responses other than the one containing the error
   code.

Reporting event registration errors by closing the stream solves these problems:

- After the client waits for initial metadata, they can check whether the event
  stream is done/closed without any additional waiting.
- When an event registration error occurs, the event stream is closed with a
  gRPC status containing the rich error message and `ni-error` trailing metadata
  indicating the error code.
- The done event stream is only closed when an event registration error occurs,
  when either side cancels, or when gRPC closes it for another reason. It is not
  closed when the DAQmx acquisition/generation encounters an error.
- The server closes the stream, so clients know to stop reading events.

This is a breaking change. However, it only affects error cases, and clients
should already handle gRPC errors when reading the event stream.

### Why should this Pull Request be merged?

Fixes #924 

Prerequisite for fixing [AB#2395958](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2395958): GrpcStubInterpreter does not report event registration errors to caller

### What testing has been done?

Manually tested with updated Python examples, which will be in a separate PR.
Ran updated C++ tests with `SystemTestsRunner.exe --gtest_filter=*DAQmx*`

PR #928, opened 2023-05-18
examples: Check for DAQmx event registration errors

bkeryan:

Change the DAQmx event examples to explicitly check for stream errors after waiting for initial metadata.
If the stream was closed with an error, it is "done" and reading a response will throw an exception.

Demonstrates how to benefit from #927 

Tested examples as written.

Tested error handling by changing `EVERY_N_SAMPLES_EVENT_TYPE_ACQUIRED_INTO_BUFFER` to `EVERY_N_SAMPLES_EVENT_TYPE_TRANSFERRED_FROM_BUFFER` and adding a `return` after the error check to prevent the worker thread/async task from being started.

### DIFF
--- a/examples/nidaqmx/analog-input-every-n-samples-aio.py
+++ b/examples/nidaqmx/analog-input-every-n-samples-aio.py
@@ -62,6 +62,11 @@ async def _main():
                     f"{warning_message.error_string}\nWarning status: {response.status}\n"
                 )
 
+        async def check_for_stream_error(stream: grpc.aio.UnaryStreamCall) -> None:
+            """Raise an exception if the stream was closed with an error."""
+            if stream.done() and await stream.code() != grpc.StatusCode.OK:
+                _ = await stream.read()
+
         try:
             create_task_response = await client.CreateTask(nidaqmx_types.CreateTaskRequest())
             task = create_task_response.task
@@ -98,20 +103,16 @@ async def _main():
             )
 
             # Wait for initial_metadata and check for stream errors to ensure that the callback is
-            # registered before starting the task.
+            # registered successfully before starting the task.
             await every_n_samples_stream.initial_metadata()
-            if every_n_samples_stream.done():
-                every_n_samples_response = await every_n_samples_stream.read()
-                check_for_warning(every_n_samples_response)
+            await check_for_stream_error(every_n_samples_stream)
 
             done_event_stream = client.RegisterDoneEvent(
                 nidaqmx_types.RegisterDoneEventRequest(task=task)
             )
 
             await done_event_stream.initial_metadata()
-            if done_event_stream.done():
-                done_response = await done_event_stream.read()
-                check_for_warning(done_response)
+            await check_for_stream_error(done_event_stream)
 
             start_task_response = await client.StartTask(nidaqmx_types.StartTaskRequest(task=task))
             check_for_warning(start_task_response)
--- a/examples/nidaqmx/analog-input-every-n-samples-aio.py
+++ b/examples/nidaqmx/analog-input-every-n-samples-aio.py
@@ -62,7 +62,7 @@ async def _main():
                     f"{warning_message.error_string}\nWarning status: {response.status}\n"
                 )
 
-        async def check_for_stream_error(stream: grpc.aio.UnaryStreamCall) -> None:
+        async def check_for_stream_error(stream):
             """Raise an exception if the stream was closed with an error."""
             if stream.done() and await stream.code() != grpc.StatusCode.OK:
                 _ = await stream.read()
--- a/examples/nidaqmx/analog-input-every-n-samples-aio.py
+++ b/examples/nidaqmx/analog-input-every-n-samples-aio.py
@@ -97,15 +97,21 @@ async def _main():
                 )
             )
 
-            # Wait for initial_metadata to ensure that the callback is registered before starting
-            # the task.
+            # Wait for initial_metadata and check for stream errors to ensure that the callback is
+            # registered before starting the task.
             await every_n_samples_stream.initial_metadata()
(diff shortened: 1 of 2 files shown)


examples/nidaqmx/analog-input-every-n-samples-aio.py
```diff
@@ -101,15 +101,21 @@ def check_for_warning(response):
                 )
             )
 
-            # Wait for initial_metadata to ensure that the callback is registered before starting
-            # the task.
+            # Wait for initial_metadata and check for stream errors to ensure that the callback is
+            # registered before starting the task.
             await every_n_samples_stream.initial_metadata()
+            if every_n_samples_stream.done():
```
bkeryan, line 107: If wait for initial metadata doesn't update the done flag for whatever reason, then the error is not lost forever. The stream reader task/thread will raise an exception and handle it. Both asyncio and futures ensure that an unhandled exception in the stream reader task/thread will be reported to the user.

examples/nidaqmx/analog-input-every-n-samples-aio.py
```diff
@@ -101,15 +101,21 @@ def check_for_warning(response):
...
+                every_n_samples_response = await every_n_samples_stream.read()
+                check_for_warning(every_n_samples_response)
```
bkeryan, line 109: I added `check_for_warning` for consistency with other code that gets responses, but on second thought it probably doesn't make sense here.

examples/nidaqmx/analog-input-every-n-samples-aio.py
```diff
@@ -101,15 +101,21 @@ def check_for_warning(response):
                 )
             )
 
-            # Wait for initial_metadata to ensure that the callback is registered before starting
-            # the task.
+            # Wait for initial_metadata and check for stream errors to ensure that the callback is
+            # registered before starting the task.
             await every_n_samples_stream.initial_metadata()
+            if every_n_samples_stream.done():
+                every_n_samples_response = await every_n_samples_stream.read()
```
bkeryan, line 108: I should test what happens if the event stream is closed with an OK status. 

For aio, I think `read()` returns a [`grpc.aio.EOF`](https://github.com/grpc/grpc/blob/cf6e5ed8434a260ccc8d526cc86a016b34afee49/src/python/grpcio/grpc/_cython/_cygrpc/aio/common.pyx.pxi#L57), which seems like some kind of placeholder.

For non-aio, I'm using `next()`, which probably raises `StopIteration`. Specifying a default value should prevent that.

examples/nidaqmx/analog-input-every-n-samples.py
bkeryan, line 1:
```
Run ni-python-styleguide lint source/codegen/*.py examples
examples/nidaqmx/analog-input-every-n-samples.py:55:67: E999 SyntaxError: invalid syntax
examples/nidaqmx/analog-input-every-n-samples-aio.py:25:[1](https://github.com/ni/grpc-device/actions/runs/5009958794/jobs/8979400169#step:5:1)01: W505 doc line too long (101 > 100 characters)
Error: Process completed with exit code 1.
```

astarche:
daqmx is still the only driver that's excluded from [the `validate_examples` pipeline step](https://github.com/ni/grpc-device/blob/f276a8e1edae1143a118b30f7c208629eb67b4d8/.github/workflows/validate_python.yml#L34). I think maybe that was because of the async examples?

Note that script can also *run* the examples if you run it locally with a "server url" param.

No action required, just pointing out a helpful tool that's not helping here.

astarche (approved):

reckenro (approved):

bkeryan:
> daqmx is still the only driver that's excluded from [the `validate_examples` pipeline step](https://github.com/ni/grpc-device/blob/f276a8e1edae1143a118b30f7c208629eb67b4d8/.github/workflows/validate_python.yml#L34). I think maybe that was because of the async examples?
> 
> Note that script can also _run_ the examples if you run it locally with a "server url" param.
> 
> No action required, just pointing out a helpful tool that's not helping here.

I got `validate_examples` working this morning:

#930 
#931

examples/nidaqmx/analog-input-every-n-samples-aio.py
bkeryan: aio calls don't have an `exception()` method, so I used `await stream.code()` to check for errors. I still have to use `await stream.read()` to raise an exception, so I extracted a `check_for_stream_error()` helper function.

examples/nidaqmx/analog-input-every-n-samples.py
```diff
@@ -64,6 +64,11 @@ def check_for_warning(response):
                     f"{warning_message.error_string}\nWarning status: {response.status}\n"
                 )
 
+        def check_for_stream_error(stream: grpc.Future) -> None:
```
bkeryan, line 67: The gRPC Python documentation says the stream object is a `Future`, and I can call `Future` methods at run time, but grpc-stubs doesn't agree:
```
examples/nidaqmx/analog-input-every-n-samples.py:70: error: Exception must be derived from BaseException  [misc]
examples/nidaqmx/analog-input-every-n-samples.py:110: error: Argument 1 to "check_for_stream_error" has incompatible type "CallIterator[RegisterEveryNSamplesEventResponse]"; expected "Future[Any]"  [arg-type]
examples/nidaqmx/analog-input-every-n-samples.py:117: error: Argument 1 to "check_for_stream_error" has incompatible type "CallIterator[RegisterDoneEventResponse]"; expected "Future[Any]"  [arg-type]
Found 3 errors in 1 file (checked 7 source files)
```

I filed a grpc-stubs bug: https://github.com/shabbyrobe/grpc-stubs/issues/41

I guess I'll use `Any` for now.

PR #934, opened 2023-05-23
Fix DAQmx proto file with misplaced attributes

Sakthi-SM:
In NI-DAQmx, the attributes with bitfield enums are placed under the wrong type during the proto file generation. This PR fixes this and places these attributes under the right type.

This fixes [Bug 2395661](https://dev.azure.com/ni/DevCentral/_workitems/edit/2395661): Attributes with bitfield enum are listed under the wrong type in DAQmx proto file

The changes have been tested manually


zhindes (approved):
I assume this is simply pulling in some fixes to scrapigen? Do you mind linking that in your PR description?

generated/nidaqmx/nidaqmx.proto
```diff
@@ -1699,34 +1699,19 @@ enum DeviceBoolAttribute {
...
-  DEVICE_ATTRIBUTE_CO_SAMP_MODES = 12255;
-  DEVICE_ATTRIBUTE_NAV_SUPPORTED_MEAS_TYPES = 12323;
-  DEVICE_ATTRIBUTE_NAV_TRIG_USAGE = 12324;
-  DEVICE_ATTRIBUTE_AI_DIG_FLTR_TYPES = 12551;
+  DEVICE_ATTRIBUTE_PRODUCT_CATEGORY = 10665;
+  DEVICE_ATTRIBUTE_CI_CURRENT_UTC_OFFSET = 12320;
+  DEVICE_ATTRIBUTE_NAV_MODE = 12325;
+  DEVICE_ATTRIBUTE_NAV_ALT_REF = 12329;
+  DEVICE_ATTRIBUTE_NAV_ANT_STATUS = 12334;
+  DEVICE_ATTRIBUTE_NAV_CURRENT_UTC_OFFSET = 12336;
```
zhindes, line 1714: FYI - I don't think these were ever officially released or used in any products, but its not a big deal.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -1776,6 +1761,21 @@ enum DeviceUInt32ArrayAttribute {
...
+  DEVICE_ATTRIBUTE_AO_SAMP_MODES = 12253;
+  DEVICE_ATTRIBUTE_CI_SAMP_MODES = 12254;
+  DEVICE_ATTRIBUTE_CO_SAMP_MODES = 12255;
+  DEVICE_ATTRIBUTE_NAV_SUPPORTED_MEAS_TYPES = 12323;
+  DEVICE_ATTRIBUTE_NAV_TRIG_USAGE = 12324;
+  DEVICE_ATTRIBUTE_AI_DIG_FLTR_TYPES = 12551;
+}
+
 enum ExportSignalDoubleAttribute {
   EXPORTSIGNAL_DOUBLE_ATTRIBUTE_UNSPECIFIED = 0;
```
bkeryan, line 1780: FYI, it's kind of weird that this shuffled the order of the enum blocks, but I think I see why. `get_attributes_split_by_sub_group` processes the attributes in `attributes.py` order and sorts them into dictionaries. The order of data types in `categorized_attributes` depends on the position of the first attribute with that data type within `attributes.py`.

source/codegen/common_helpers.py
```diff
@@ -915,12 +915,25 @@ def get_attribute_groups(data):
...
     return [AttributeGroup(service_class_prefix, attributes, config)]
 
+def _transform_attributes(attribute_groups):
+    # This is handles the specific attributes that uses bitfield enum.
+    # Currently this affects only NI-DAQmx generation.
+    for attribute_group in attribute_groups:
+        for id,data in attribute_group.attributes.items():
+            if "bitfield_enum" in data:
+                data["type"] = BITFIELD_ENUM_RETURN_TYPES.get(data["type"], data["type"])
+    return attribute_groups
```
bkeryan, line 932: Side effects or return value: choose one.

When I see code like `return _transform_attributes(attribute_groups)`, I think "that must return a transformed copy of the attribute groups". However, `_transform_attributes` modifies the attribute groups that you pass in.

Do you need to mutate the metadata? You can achieve the same result with a non-destructive change in `get_attributes_split_by_sub_group`:

image

source/codegen/common_helpers.py
```diff
@@ -915,12 +915,25 @@ def get_attribute_groups(data):
...
+
+BITFIELD_ENUM_RETURN_TYPES={
+    "int32[]": "int32",
+}
```
bkeryan, line 936: I don't expect this dictionary to grow. In this case, I think it may be more maintainable to assert int32[] and hardcode int32.

generated/nidaqmx/nidaqmx.proto
bkeryan, line 1: `DeviceInt32AttributeValues` contains the non-bitfield `TriggerUsage` enum, which you can only use in an array. 

```
enum DeviceInt32AttributeValues {
  option allow_alias = true;
  DEVICE_INT32_UNSPECIFIED = 0;
  DEVICE_INT32_AI_MEASUREMENT_TYPE_VOLTAGE = 10322;
...
  DEVICE_INT32_PRODUCT_CATEGORY_TEST_SCALE_CHASSIS = 16180;
  DEVICE_INT32_PRODUCT_CATEGORY_TEST_SCALE_MODULE = 16181;
  DEVICE_INT32_PRODUCT_CATEGORY_UNKNOWN = 12588;
  DEVICE_INT32_TRIGGER_USAGE_ADVANCE = 12488;
  DEVICE_INT32_TRIGGER_USAGE_PAUSE = 12489;
  DEVICE_INT32_TRIGGER_USAGE_REFERENCE = 12490;
  DEVICE_INT32_TRIGGER_USAGE_START = 12491;
  DEVICE_INT32_TRIGGER_USAGE_HANDSHAKE = 10389;
  DEVICE_INT32_TRIGGER_USAGE_ARM_START = 14641;
}
```

We don't necessarily have to delete it, but I think `DeviceInt32AttributeValues` should also contain the bitfield version, `TriggerUsageTypes`:
https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nidaqmx/enums.py#L18280

BTW, it has an "ENUMS_SHOULD_BE_USED" validation suppression: 😅
https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nidaqmx/enums_addon.py#L22

Likewise for `InputTermCfg` and `TermCfg`.

bkeryan (changes requested):

source/codegen/common_helpers.py
Sakthi-SM: Removed this based on the previous suggestion

generated/nidaqmx/nidaqmx.proto
Sakthi-SM: Brad, can you point out the value of these `TermCfg` enum attributes to be defined in the proto file?
image

bkeryan (changes requested):

source/codegen/metadata/nidaqmx/__init__.py
```diff
@@ -9,6 +10,7 @@
     "functions" : functions,
     "functions_validation_suppressions": functions_validation_suppressions,
     "attributes" : attributes,
+    "additional_attributes": additional_attributes_used_in_proto,
```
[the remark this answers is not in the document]
bkeryan: However, I think the metadata already contains these enums, so you probably don't need to update `__init__.py`. You need to update the codegen to interpret the `bitfield_enum` key correctly.

generated/nidaqmx/nidaqmx.proto
Sakthi-SM: Added these `bitfield_enums` to the proto based on that metadata in `enums.py`

source/codegen/metadata/nidaqmx/__init__.py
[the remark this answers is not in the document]
Sakthi-SM: Removed these additional metadata and their corresponding changes

source/codegen/templates/proto.mako
```diff
@@ -49,6 +49,9 @@ service ${service_class_prefix} {
 ${mako_helper.define_attribute_enum(group.name, define_attribute_enum, attributes, config)}\
 %   endfor
 % endfor
+% for name, bitfield_enum in common_helpers.get_bitfield_enum_groups(data).items():
+${mako_helper.define_bitfield_enum(name, bitfield_enum)}\
+%endfor
```
bkeryan, line 54: I don't think we need to add entirely new template code for these enum values. `GetDeviceAttributeInt32` and `GetPhysicalChanAttributeInt32` already have function enums.

I think there are two ways to update the existing codegen:

**A. Generate only the bitfield enum values**
```diff
diff --git a/source/codegen/common_helpers.py b/source/codegen/common_helpers.py
index 38185e38..c283d444 100644
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -501,6 +501,10 @@ def get_attribute_enums_by_type(attributes):
         if "enum" in attribute:
             attribute_type = get_underlying_type(attribute)
             enum_name = attribute["enum"]
+            if "bitfield_enum" in attribute:
+                enum_name = attribute["bitfield_enum"]
+                if enum_name.startswith("_"):  # nidaqmx-python uses a leading underscore
+                    enum_name = enum_name[1:]
             attribute_enums_by_type[attribute_type].add(enum_name)
     return attribute_enums_by_type
```

This will delete 
```
  DEVICE_INT32_COUPLING1_AC = 10045;
  DEVICE_INT32_COUPLING1_DC = 10050;
  DEVICE_INT32_COUPLING1_GND = 10066;
```
and add
```
  DEVICE_INT32_COUPLING_TYPES_AC = 1;
  DEVICE_INT32_COUPLING_TYPES_DC = 2;
  DEVICE_INT32_COUPLING_TYPES_GROUND = 4;
  DEVICE_INT32_COUPLING_TYPES_HF_REJECT = 8;
  DEVICE_INT32_COUPLING_TYPES_LF_REJECT = 16;
  DEVICE_INT32_COUPLING_TYPES_NOISE_REJECT = 32;
```
Removing `DEVICE_INT32_COUPLING1_AC` breaks compatibility, but I don't think it is a useful enum value.

Also, when some attributes use the enum and other attributes use the bitfield_enum, it generates both:
```
  DEVICE_INT32_TRIGGER_USAGE_ADVANCE = 12488;
  DEVICE_INT32_TRIGGER_USAGE_PAUSE = 12489;
  DEVICE_INT32_TRIGGER_USAGE_REFERENCE = 12490;
  DEVICE_INT32_TRIGGER_USAGE_START = 12491;
  DEVICE_INT32_TRIGGER_USAGE_HANDSHAKE = 10389;
  DEVICE_INT32_TRIGGER_USAGE_ARM_START = 14641;
  DEVICE_INT32_TRIGGER_USAGE_TYPES_ADVANCE = 1;
  DEVICE_INT32_TRIGGER_USAGE_TYPES_PAUSE = 2;
  DEVICE_INT32_TRIGGER_USAGE_TYPES_REFERENCE = 4;
  DEVICE_INT32_TRIGGER_USAGE_TYPES_START = 8;
  DEVICE_INT32_TRIGGER_USAGE_TYPES_HANDSHAKE = 16;
  DEVICE_INT32_TRIGGER_USAGE_TYPES_ARM_START = 32;
```
However, we only do that for one of the navigation attributes:
```
        12324: {
            'access': 'read',
            'c_function_name': 'DevNavTrigUsage',
            'calling_convention': 'StdCall',
            'ctypes_data_type': 'numpy.int32',
            'enum': 'TriggerUsage',
            'handle_parameters': {
                'deviceHandle': {
                    'accessor': 'self._name',
                    'ctypes_data_type': 'ctypes.c_char_p',
                    'cvi_name': 'deviceName'
                }
            },
            'has_explicit_read_buffer_size': True,
            'has_explicit_write_buffer_size': True,
            'is_list': True,
            'is_python_object': False,
            'name': 'NAV_TRIG_USAGE',
            'python_data_type': 'TriggerUsage',
            'python_description': 'Indicates the trigger types supported by the device for a given measurement type.',
            'resettable': False,
            'type': 'int32[]'
        },
```

**B. Generate both the enum and bitfield enum values**
```diff
diff --git a/source/codegen/common_helpers.py b/source/codegen/common_helpers.py
index 38185e38..a61dae03 100644
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -502,6 +502,12 @@ def get_attribute_enums_by_type(attributes):
             attribute_type = get_underlying_type(attribute)
             enum_name = attribute["enum"]
             attribute_enums_by_type[attribute_type].add(enum_name)
+        if "bitfield_enum" in attribute:
+            attribute_type = get_underlying_type(attribute)
+            enum_name = attribute["bitfield_enum"]
+            if enum_name.startswith("_"):  # nidaqmx-python uses a leading underscore
+                enum_name = enum_name[1:]
+            attribute_enums_by_type[attribute_type].add(enum_name)
     return attribute_enums_by_type
```
This will generate both:
```
  DEVICE_INT32_COUPLING1_AC = 10045;
  DEVICE_INT32_COUPLING1_DC = 10050;
  DEVICE_INT32_COUPLING1_GND = 10066;
  DEVICE_INT32_COUPLING_TYPES_AC = 1;
  DEVICE_INT32_COUPLING_TYPES_DC = 2;
  DEVICE_INT32_COUPLING_TYPES_GROUND = 4;
  DEVICE_INT32_COUPLING_TYPES_HF_REJECT = 8;
  DEVICE_INT32_COUPLING_TYPES_LF_REJECT = 16;
  DEVICE_INT32_COUPLING_TYPES_NOISE_REJECT = 32;
```

@astarche @reckenro @zhindes Anyone opposed to option A?

generated/nidaqmx/nidaqmx.proto
```diff
@@ -2835,6 +2835,31 @@ enum WriteStringAttribute {
...
 }
 
+enum TriggerUsageTypesAttribute {
+  TriggerUsageTypes_ADVANCE = 1;
+  TriggerUsageTypes_PAUSE = 2;
+  TriggerUsageTypes_REFERENCE = 4;
+  TriggerUsageTypes_START = 8;
+  TriggerUsageTypes_HANDSHAKE = 16;
+  TriggerUsageTypes_ARM_START = 32;
+}
```
bkeryan, line 2845: Problems with this:
- These new enums are not used by any request/response messages.
- This doesn't extend the enums that are used by functions like `GetPhysicalChanAttributeInt32`, such as `PhysicalChannelInt32AttributeValues` (note the "Values" suffix).
- The suffix "Attribute" is incorrect because this is not an attribute.
- The enum value names should be UPPER_SNAKE_CASE.

Updating `get_attribute_enums_by_type` will address all of these problems.

generated/nidaqmx/nidaqmx.proto
```diff
@@ -3913,9 +3913,12 @@ enum DeviceInt32AttributeValues {
   DEVICE_INT32_CO_OUTPUT_TYPE_PULSE_TIME = 10269;
   DEVICE_INT32_CO_OUTPUT_TYPE_PULSE_FREQ = 10119;
   DEVICE_INT32_CO_OUTPUT_TYPE_PULSE_TICKS = 10268;
-  DEVICE_INT32_COUPLING1_AC = 10045;
-  DEVICE_INT32_COUPLING1_DC = 10050;
-  DEVICE_INT32_COUPLING1_GND = 10066;
+  DEVICE_INT32_COUPLING_TYPES_AC = 1;
```
Sakthi-SM, line 3916: @bkeryan the field Ids are not unique in the newly added fields (`DEVICE_INT32_COUPLING_TYPES`  and `DEVICE_INT32_TRIGGER_USAGE_TYPES`). Any suggestions on how to resolve this?

source/codegen/templates/proto.mako
Sakthi-SM: I have implemented the changes based on option A.

generated/nidaqmx/nidaqmx.proto
bkeryan: The codegen automatically specifies `allow_alias` when the enum contains duplicate values. See https://github.com/ni/grpc-device/blob/7927feb2d590821a8066b82ae5b34677f312944b/source/codegen/proto_helpers.py#L26

Is this causing a problem for your work?

bkeryan (approved):

generated/nidaqmx/nidaqmx.proto
astarche: Attribute values enums will almost always have overlaps and `allow_alias`. It should be rare otherwise. `allow_alias` is not recommended in general.

source/codegen/templates/proto.mako
reckenro: I think option A makes sense and I agree it doesn't seem like the old enum values like ` DEVICE_INT32_COUPLING1_AC` were that useful.

reckenro (approved):

generated/nidaqmx/nidaqmx.proto
Sakthi-SM: No Brad, this doesn't cause any issue. But as a general standard the field Ids should be unique in proto file, hence I wanted to confirm this.

PR #935, opened 2023-05-24
Handle timeout behavior of read functions in NI-DAQmx

Sakthi-SM:

In Ni-DAQmx functions, whenever a timeout exception has occurred instead of raising a gRPC exception, the error and the partially read data should be sent through the gRPC response.
,  

This implements [Bug 2399915](https://dev.azure.com/ni/DevCentral/_workitems/edit/2399915): Partial reads report samps-per-chan-read via DaqReadError but don't update read_array

The changes have been tested manually


source/codegen/templates/service_helpers.mako
```diff
@@ -777,7 +778,11 @@ ${set_response_values(normal_outputs, init_method)}\
       %for error_parameter in error_parameters:
       context->AddTrailingMetadata("${error_parameter["return_on_error_key"]}", std::to_string(${common_helpers.pascal_to_snake(error_parameter["name"])}));
       %endfor
+      %if len(error_parameters)>0 and is_read_function:
+      if (!status_ok(status) && status != DAQmxErrorSamplesNotYetAvailable) {
```
bkeryan, line 782: Instead of a DAQmx-specific hack, I think we should add a key to the function metadata that opts in to this feature and specifies the timeout error constant. Other drivers such as [nisync](https://github.com/ni/grpc-device/blob/main/source/custom/nisync_service.custom.cpp) could benefit.

How about `'timeout_error': 'DAQmxErrorSamplesNotYetAvailable'`?

bkeryan (changes requested):

generated/nidaqmx/nidaqmx_service.cpp
bkeryan, line 1: Please add a test case for this behavior to `nidaqmx_driver_api_tests.cpp`.

source/codegen/templates/service_helpers.mako
bkeryan: Looks good

generated/nidaqmx/nidaqmx_service.cpp
bkeryan: One way to do this is to use a counter output task to generate the sample clock, like [test_timeout_stream](https://github.com/ni/nidaqmx-python/blob/master/tests/legacy/test_read_exceptions.py#L133). This is 100% reliable because you control when the sample clock stops.

A simpler way to do this is to read data from a finite acquisition with a timeout that is too short:
- Acquire 60000 samples at 1000 S/s. It would take 60 seconds to acquire this many samples.
- Read 60000 samples with a timeout of 100 ms. This will wait about 100 ms then return a timeout error.
- The number of samples read should be about 100, but you can only do fuzzy validation because any delays between the calls to start_task() and read_analog_f64() will increase the number of samples read. It's probably safe to validate that it returns between 80 and 1000 samples though; delays only affect the upper bound, and 1 second seems like a lot of slack.
- `AIVoltageChannel_CfgSampClkTimingAndAcquireData_Succeeds` is a good starting point. Change it to use `ACQUISITION_TYPE_FINITE_SAMPS`, set the finite sample count and read sample count differently, and change the validation.

astarche (approved):

reckenro (approved):

bkeryan (changes requested):

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -277,6 +292,7 @@ class NiDAQmxDriverApiTests : public Test {
   ::grpc::Status read_analog_f64(
       int32 samps_per_chan,
       uint32 array_size_in_samps,
+      double timeout,
```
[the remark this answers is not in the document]
bkeryan: > That way, we don't need to specify this parameter **when throwing away the response.**
(emphasis added)

When you are **not** throwing away the response, you must specify the optional parameters that come before the optional response parameter.

```
D:\a\grpc-device\grpc-device\source\tests\system\nidaqmx_driver_api_tests.cpp(1143,22): error C2664: 'grpc::Status ni::tests::system::`anonymous-namespace'::NiDAQmxDriverApiTests::read_analog_f64(ni::tests::system::`anonymous-namespace'::int32,google::protobuf::uint32,double,nidaqmx_grpc::ReadAnalogF64Response &)': cannot convert argument 3 from 'nidaqmx_grpc::ReadAnalogF64Response' to 'double' [D:\a\grpc-device\grpc-device\build\SystemTestsRunner.vcxproj]
D:\a\grpc-device\grpc-device\source\tests\system\nidaqmx_driver_api_tests.cpp(1143,60): message : No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called [D:\a\grpc-device\grpc-device\build\SystemTestsRunner.vcxproj]
D:\a\grpc-device\grpc-device\source\tests\system\nidaqmx_driver_api_tests.cpp(278,18): message : see declaration of 'ni::tests::system::`anonymous-namespace'::NiDAQmxDriverApiTests::read_analog_f64' [D:\a\grpc-device\grpc-device\build\SystemTestsRunner.vcxproj]
D:\a\grpc-device\grpc-device\source\tests\system\nidaqmx_driver_api_tests.cpp(1143,22): message : while trying to match the argument list '(const int, const int, nidaqmx_grpc::ReadAnalogF64Response)' [D:\a\grpc-device\grpc-device\build\SystemTestsRunner.vcxproj]
```

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1519,28 +1501,26 @@ TEST_F(NiDAQmxDriverApiTests, AIVoltageChannel_CfgSampClkTimingAndAcquireData_Su
 TEST_F(NiDAQmxDriverApiTests, AIVoltageChannel_CfgSampClkTimingAndAcquireDataWithMinimumTimeout_SampleNotAvailableStatusIsReturned)
```
bkeryan, line 1501: I think sample clock timing is a precondition, not the function you are testing. Also, "finite" is more important than "sample clock".

This paraphrases the error name.

You can omit "is".

```suggestion
TEST_F(NiDAQmxDriverApiTests, AIFiniteAcquisition_ReadWithTimeoutTooSmall_SamplesNotYetAvailableStatusReturned)
```

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1519,28 +1501,26 @@ TEST_F(NiDAQmxDriverApiTests, AIVoltageChannel_CfgSampClkTimingAndAcquireData_Su
...
 {
   create_ai_voltage_chan(0.0, 1.0);
-  const auto TIMEOUT = 100.0;
+  const auto TIMEOUT = 100e-3;
   const auto SAMPLE_RATE = 1000.0;
   const auto SAMPLES_PER_CHAN = 60000UL;
   const auto NUM_SAMPS = 60000;
-  const auto EXPECTED_ERROR_STATUS = -200284;
 
   CfgSampClkTimingResponse response;
```
bkeryan, line 1509: `response` is unused, so delete it.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1519,28 +1501,26 @@ TEST_F(NiDAQmxDriverApiTests, AIVoltageChannel_CfgSampClkTimingAndAcquireData_Su
...
-  auto config_status = cfg_samp_clk_timing(
+  cfg_samp_clk_timing(
       create_cfg_samp_clk_timing_request(SAMPLE_RATE, Edge1::EDGE1_RISING, AcquisitionType::ACQUISITION_TYPE_FINITE_SAMPS , SAMPLES_PER_CHAN));
   start_task();
```
bkeryan: Along with the suggested rename, move these to the "arrange" section.

(i.e. move them above the 1st blank line)

bkeryan (changes requested):

bkeryan (approved):
Once the test passes, this looks good to go.

bkeryan:
@Sakthi-SM These system test failures do not appear to be related to this PR.

Windows:
```
[ RUN      ] NiRFmxWCDMADriverApiTests.ModAccMarkerModeFromExample_FetchData_DataLooksReasonable
D:\a\grpc-device\grpc-device\source\tests\system\nirfmxwcdma_driver_api_tests.cpp(330): error: Expected: (0.0) < (mod_acc_fetch_reference_waveform_response.reference_waveform(0).real()), actual: 0 vs -0.00347697
[  FAILED  ] NiRFmxWCDMADriverApiTests.ModAccMarkerModeFromExample_FetchData_DataLooksReasonable (13146 ms)
```

Linux:
```
[ RUN      ] NiFgenDriverApiTest.PerformReset_CompletesSuccessfuly
/home/runner/work/grpc-device/grpc-device/source/tests/system/nifgen_driver_api_tests.cpp:71: Failure
Value of: status.ok()
  Actual: false
Expected: true
/home/runner/work/grpc-device/grpc-device/source/tests/system/nifgen_driver_api_tests.cpp:101: Failure
Value of: status.ok()
  Actual: false
Expected: true
[  FAILED  ] NiFgenDriverApiTest.PerformReset_CompletesSuccessfuly (112 ms)
```

PR #940, opened 2023-05-25
codegen: Support DAQmx event unregistration

### LINKED ISSUE #925: NI gRPC Device Server doesn't support unregistering DAQmx events
The NI gRPC Device Server uses server streaming to return a response for each event generated by DAQmx. However, it doesn't yet support unregistering DAQmx events.

Closing the event stream does not unregister the event with DAQmx. DAQmx continues to generate events, and the NI gRPC Device Server ignores the events. 

DAQmx only allows registering or unregistering events when the task is not running, so closing the event stream would not be an appropriate way to unregister the event with DAQmx:
- If you unregister the event at the wrong time, DAQmx returns an event unregistration error and continues to generate events.
- The server would not be able to report unregistration errors with the stream closed.

Addressing this is required to fix nidaqmx-python bug [AB#2395984](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2395984): "GrpcStubInterpreter does not unregister events with DAQmx". We will have the same problem when supporting gRPC in other versions of the DAQmx API.

[AB#2396359](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2396359)

bkeryan:

Add new RPCs to support DAQmx event unregistration:
- `UnregisterDoneEvent`
- `UnregisterEveryNSamplesEvent`
- `UnregisterSignalEvent`

Unregistering a DAQmx event configures the DAQmx task to stop producing the event. This allows test applications to enable/disable events without clearing and re-creating the task, which is required for sharing tasks between multiple MeasurementLink measurement services. It also allows test applications to unregister and reregister the same event with different settings, such as varying the every N samples interval or changing the signal event type. 

Events cannot be unregistered while the task is in the running state. The task must be in the verified, reserved, or committed states.

Canceling the event stream does not automatically unregister the event. If it did, canceling the event stream while the task is running would produce an error, but it would have no way to report the error because the stream is canceled. After the event stream is canceled, the NI gRPC Device Server ignores any additional events produced by DAQmx.

Unregistering an event does not automatically close or cancel the corresponding event stream at this time. The test application is currently responsible for canceling the event reader after unregistering the event. However, test architects are encouraged to handle stream closure/cancellation as well as stream errors.

Fixes #925 

Required to fix AB#2395984: GrpcStubInterpreter does not unregister events with DAQmx

Ran DAQmx system tests on Windows and Linux.
Ran nidaqmx-python tests on Windows with https://github.com/ni/nidaqmx-python/pull/377

### DIFF
--- a/generated/nidaqmx/nidaqmx.proto
+++ b/generated/nidaqmx/nidaqmx.proto
@@ -382,6 +382,9 @@ service NiDAQmx {
   rpc StopTask(StopTaskRequest) returns (StopTaskResponse);
   rpc TaskControl(TaskControlRequest) returns (TaskControlResponse);
   rpc TristateOutputTerm(TristateOutputTermRequest) returns (TristateOutputTermResponse);
+  rpc UnregisterDoneEvent(UnregisterDoneEventRequest) returns (UnregisterDoneEventResponse);
+  rpc UnregisterEveryNSamplesEvent(UnregisterEveryNSamplesEventRequest) returns (UnregisterEveryNSamplesEventResponse);
+  rpc UnregisterSignalEvent(UnregisterSignalEventRequest) returns (UnregisterSignalEventResponse);
   rpc UnreserveNetworkDevice(UnreserveNetworkDeviceRequest) returns (UnreserveNetworkDeviceResponse);
   rpc WaitForNextSampleClock(WaitForNextSampleClockRequest) returns (WaitForNextSampleClockResponse);
   rpc WaitForValidTimestamp(WaitForValidTimestampRequest) returns (WaitForValidTimestampResponse);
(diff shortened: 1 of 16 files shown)


generated/nidaqmx/nidaqmx.proto
```diff
@@ -382,6 +382,9 @@ service NiDAQmx {
   rpc StopTask(StopTaskRequest) returns (StopTaskResponse);
   rpc TaskControl(TaskControlRequest) returns (TaskControlResponse);
   rpc TristateOutputTerm(TristateOutputTermRequest) returns (TristateOutputTermResponse);
+  rpc UnregisterDoneEvent(UnregisterDoneEventRequest) returns (UnregisterDoneEventResponse);
```
astarche, line 385: I think we don't need to update the examples because `Unregister` is only required/encouraged if you want to reuse the task. Is that right?

astarche (approved):

generated/nidaqmx/nidaqmx.proto
bkeryan: Hmm. `Unregister` is entirely optional, but it is not required if you want to reuse the task with the current event configuration/streams. You can reuse the event streams for multiple acquisitions. However, the examples currently cancel the event streams as a form of flow control. Maybe we should rethink that aspect of the examples.

bkeryan: Anyway, I'm marking this resolved because no example changes are needed for this PR.

bkeryan:
The failing tests are not DAQmx related.
```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] NiRFmxInstrDriverApiTests.InitializeFromNIRFSAArray_SelfCalibrate_Succeeds
[  FAILED  ] NiRFmxLTEDriverApiTests.NBIoTModAccFromExample_FetchData_DataLooksReasonable

 2 FAILED TESTS
  YOU HAVE 3 DISABLED TESTS
```

Issue #936, opened 2023-05-24
DAQmx system tests crash when cleaning up leaked watchdog task


bkeryan:
[`DOWatchdogTask_StartTaskAndWatchdogTask_Succeeds`](https://github.com/ni/grpc-device/blob/main/source/tests/system/nidaqmx_driver_api_tests.cpp#L1631) leaks a watchdog task. On Linux, this crashes during server shutdown..

There are two bugs here:
- The test should not leak a watchdog task.
- Leaking a task should not crash the server during shutdown. 

Output and stack trace:
```
[----------] Global test environment tear-down
[==========] 63 tests from 2 test suites ran. (2718 ms total)
[  PASSED  ] 63 tests.
[Thread 0x7ffff692c640 (LWP 3545) exited]
[Thread 0x7ffff693e640 (LWP 3544) exited]

Thread 1 "SystemTestsRunn" received signal SIGSEGV, Segmentation fault.
0x0000555555b4f51e in nidaqmx_grpc::NiDAQmxService::CreateWatchdogTimerTaskEx(grpc::ServerContext*, nidaqmx_grpc::CreateWatchdogTimerTaskExRequest const*, nidaqmx_grpc::CreateWatchdogTimerTaskExResponse*)::{lambda(void*)#2}::operator()(void*) const ()
(gdb) bt
#0  0x0000555555b4f51e in nidaqmx_grpc::NiDAQmxService::CreateWatchdogTimerTaskEx(grpc::ServerContext*, nidaqmx_grpc::CreateWatchdogTimerTaskExRequest const*, nidaqmx_grpc::CreateWatchdogTimerTaskExResponse*)::{lambda(void*)#2}::operator()(void*) const ()
#1  0x0000555555ba0778 in void std::__invoke_impl<void, nidaqmx_grpc::NiDAQmxService::CreateWatchdogTimerTaskEx(grpc::ServerContext*, nidaqmx_grpc::CreateWatchdogTimerTaskExRequest const*, nidaqmx_grpc::CreateWatchdogTimerTaskExResponse*)::{lambda(void*)#2}&, void*>(std::__invoke_other, nidaqmx_grpc::NiDAQmxService::CreateWatchdogTimerTaskEx(grpc::ServerContext*, nidaqmx_grpc::CreateWatchdogTimerTaskExRequest const*, nidaqmx_grpc::CreateWatchdogTimerTaskExResponse*)::{lambda(void*)#2}&, void*&&) ()
[...]

PR #944, opened 2023-05-30
codegen: Delete LibraryAndService

bkeryan:

Now that the service object uses `shared_ptr` to share ownership of the library object, we no longer need `LibraryAndService` to ensure that the library object exists for the lifetime of the service object.

This is optional part 4 of a fix for #936

Manually tested DAQmx task leak scenario with the full fix.


bkeryan:
I used Graphviz to make a diagram showing object ownership:

server_lifetime

Source:
```
// Usage: dot -Tpng server_lifetime.dot -o server_lifetime.png

digraph {
  // register_all_services.cpp
  "core server" -> "session repository";
  "core server" -> "service";
  "session resource repository" [label="session resource\nrepository"];
  "service" -> "session resource repository";
  "session resource repository" -> "session repository";

  // *_service_registrar.cpp
  "service" -> "library";

  // *_service.cpp
  "service" -> "cleanup lambda" [style="dotted" label="creates"];
  "cleanup lambda" -> "library";
  "session repository" -> "cleanup lambda";

  // session_resource_repository.h
  //
  // (This cleanup lambda wraps the service's cleanup lambda, but
  // combine them for the diagram.)
  "cleanup lambda" -> "resource map";
}
```

Would it be reasonable to store the .dot file & the generated .png in the wiki, or would it make more sense to check them into the repo?

astarche:
@bkeryan Personally I like putting this sort of thing in `source` (like under a `docs` or `specs` folder). But the most similar thing in `grpc-device` is the [Internal Development wiki[(https://github.com/ni/grpc-device/wiki/internal-development). So, either way is fine by me, but going with the wiki is more consistent.

astarche (approved):
Thanks for taking this on and for putting in the effort to stage the changes intuitively!

reckenro (approved):

bkeryan:
> @bkeryan , I agree with Alex that I think starting from a clean slate I'd prefer it in source like in specs or docs folder but as things are I think it makes sense somewhere in the [Internal Development section of the wiki](https://github.com/ni/grpc-device/wiki/internal-development).

I added it to the wiki: https://github.com/ni/grpc-device/wiki/Server-Design

It turns out that GitHub supports [Mermaid](https://mermaid.js.org/), which are a Markdown-inspired diagram markup language, so the diagram source is embedded in the wiki page.

PR #953, opened 2023-06-23
Update timeout to 50 minutes

### LINKED ISSUE #952: Windows system test runner occasionally hits timeout
Seen in this PR: https://github.com/ni/grpc-device/pull/950

I think we're just starting to hit (since we've been adding driver support / corresponding tests) the arbitrary 27 minute time limit we started with for the Windows system tests: https://github.com/ni/grpc-device/blob/main/.github/workflows/run_win_system_tests.yml#L28

[AB#2429618](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2429618)

shastriUF:

Fixes #952

With the increase in number of  devices supported by grpc-device, we need to update our test timeouts to avoid running into the limit.

PR builds

### DIFF
--- a/.github/workflows/run_win_system_tests.yml
+++ b/.github/workflows/run_win_system_tests.yml
@@ -25,4 +25,4 @@ jobs:
     - name: Run System Tests
       run: .\SystemTestsRunner.exe
       working-directory: ./build/Release
-      timeout-minutes: 27
+      timeout-minutes: 50


.github/workflows/run_win_system_tests.yml
```diff
@@ -25,4 +25,4 @@ jobs:
     - name: Run System Tests
       run: .\SystemTestsRunner.exe
       working-directory: ./build/Release
-      timeout-minutes: 27
+      timeout-minutes: 50
```
SparkingSpork, line 28: Honestly, I'm wondering if we should just drop the timeout. I think the default is[ 360 minutes per job](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) (e.g, this entire file and its set of tests) and we're just setting a limit for a specific step. Since these are public GitHub runners on a public GitHub repository, I don't think we're actually being charged anything by letting it go longer (see [here](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration)).

@reckenro or @astarche - Have we ever had a test run issue that the timeout properly saved us time on?

SparkingSpork: Hmm. Actually this runs on a self-hosted runner. Do we have some kind of restriction around the time we can use on those?

shastriUF: In #952 @reckenro mentioned 27mins was arbitrary, so I arbitrarily bumped it up to 50

shastriUF: The tests in the CI finished in 29mins, so we were just over.. Anyway, I say we just go with 50 for now so it can handle +10mins easily, with an additional 10min buffer at that point

reckenro: Yeah, if I recall the 27 mins was with the idea that we could get two runs under an hour I think because we were charged (or at least we think we were charged) per hour for those self-hosted runners. I don't know if that was ever verified and we need enough time for the tests to run so I think this bump to 50 minutes is reasonable.

reckenro (approved):

.github/workflows/run_win_system_tests.yml
reckenro: And to Charlie's point, I don't recall a test run issue where the timeout kicked in and saved us time. But personally, I still like having some timeout around in case it pops up.

SparkingSpork (approved):

PR #963, opened 2023-07-18
Use Python 3.8 in validate_python workflow

ni-jfitzger:

The validate_python workflow is erroring with `FAILED STEP (256): [poetry add --dev --python ">=3.6.2" black]` because [the latest release of black does not have a Python 3.7 wheel](https://pypi.org/simple/black/).
By bumping the version used to Python 3.8, this fixes the workflow.

* PR Checks should always be able to pass with the code that's in main.
* Python has ended support of Python 3.7.

PR Checks

### DIFF
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Setup python3
       uses: actions/setup-python@v2
       with:
-        python-version: '3.7.x'
+        python-version: '3.8.x'
 
     - name: Install build dependencies
       run: python -m pip install -r python_build_requirements.txt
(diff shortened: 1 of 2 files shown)


SparkingSpork (approved):

.github/workflows/validate_python.yml
bkeryan, line 1: Is it still necessary for the script to specify `--python ">=3.6.2"`?
https://github.com/ni/grpc-device/blob/main/source/codegen/validate_examples.py#L65

ni-jfitzger: I would think not.

reckenro (approved):

bkeryan (approved):

PR #965, opened 2023-07-19
Add override specifier to inherited library class functions

ni-jfitzger:

Add the [override specifier](https://en.cppreference.com/w/cpp/language/override) to library class functions that exist in the interface class it inherits from.

Although we own the parent class, it's generally good practice to use `override` when you override a method so that you get a compiler error if there's no function to override (usually because the signature of the function in the parent class changes).

Built it locally


generated/nidaqmx/nidaqmx_library.h
```diff
@@ -18,399 +18,399 @@ class NiDAQmxLibrary : public nidaqmx_grpc::NiDAQmxLibraryInterface {
...
...const char lines66[], int32 expState66, const char lines67[], int32 expState67, const char lines68[], int32 expState68, const char lines69[], int32 expState69, const char lines70[], int32 expState70, const char lines71[], int32 expState71, const char lines72[], int32 expState72, const char lines73[], int32 expState73, const char lines74[], int32 expState74, const char lines75[], int32 expState75, const char lines76[], int32 expState76, const char lines77[], int32 expState77, const char lines78[], int32 expState78, const char lines79[], int32 expState79, const char lines80[], int32 expState80, const char lines81[], int32 expState81, const char lines82[], int32 expState82, const char lines83[], int32 expState83, const char lines84[], int32 expState84, const char lines85[], int32 expState85, const char lines86[], int32 expState86, const char lines87[], int32 expState87, const char lines88[], int32 expState88, const char lines89[], int32 expState89, const char lines90[], int32 expState90, const char lines91[], int32 expState91, const char lines92[], int32 expState92, const char lines93[], int32 expState93, const char lines94[], int32 expState94, const char lines95[], int32 expState95);
```
bkeryan, line 137: It's kind of weird that the interface only contains the mockable functions, but that is probably the most normal thing about this line of code. 😅

ni-jfitzger: Yeah, that function signature is a real doozy.

reckenro (approved):

astarche (approved):

reckenro:
Carrying forward with this in #995 to get past merge conflicts and see if the tests are looking better now.

reckenro:
Abandoning this one in favor of the new one (#995)  based on newer version of main.

PR #973, opened 2023-07-20
Add SetRuntimeEnvironment call to library constructors

ni-jfitzger:

Add an <DriverPrefix>Library::SetRuntimeEnvironment call upon library load, before init is called.

(Based on #962, tweaked the template code to be a little cleaner)

Implemented via the following changes:
* Update library.cpp.mako to call SetRuntimeEnvironment if the function exists
* Generate a version.h with some of the values found in version.rc, defined as string constants (couldn't figure out any other way to get access to the data easily)
* Update CMakeLists.txt and generate

As developers, we want to know how much nigrpc_device_server is used and which versions.

* Built ni_grpc_device_server.exe locally and called it with the nimi-python system tests in CEIP test mode.
   * Verified that the Runtime Environment data was logged correctly.
* Manually launched the server and called it, opening and closing sessions, as well as the python process that I called it from, multiple times. Confirmed that data was only logged once. This makes sense because the service that loads the library is supposed to remain loaded for the lifetime of the server instance.

TODOs:
* Before submitting this, merge changes from main after #964 is submitted and add a unit test. I might just close this and create a separate PR.


CMakeLists.txt
```diff
@@ -758,4 +770,4 @@ endif()
 # add_test( NAME UnitTests COMMAND UnitTestsRunner )
 add_test( NAME UnitTests COMMAND UnitTestsRunner )
 add_test( NAME IntegrationTests COMMAND IntegrationTestsRunner )
-add_test( NAME SystemTests COMMAND SystemTestsRunner )
+add_test( NAME SystemTests COMMAND SystemTestsRunner )
```
SparkingSpork, line 773: Missing new line at end of the file.

SparkingSpork (approved):

bkeryan (approved):

reckenro:
Sorry I missed this on the metadata PR, but do we want this for NI-TClk too? If so, I think having a follow up PR to update its metadata and regen makes sense.

reckenro (approved):

ni-jfitzger:
> @ni-jfitzger , sorry I missed this on the metadata PR, but do we want this for NI-TClk too? If so, I think having a follow up PR to update its metadata and regen makes sense.

NI-TClk doesn't have devices or sessions. It calls into other drivers. So, it has no telemetry.

ni-jfitzger:
Some of the System Tests sure are flaky.

PR #977, opened 2023-07-27
Add SetRuntimeEnvironment unit tests

ni-jfitzger:

* Add unit tests for SetRuntimeEnvironment behavior.
* Add a SharedLibraryInterface class and MockSharedLibrary
  * Needed to support the unit tests
  * Was not able to figure out a way to include the swap method, but it doesn't seem to be used anywhere

Impelmentation Notes:
* Tests were implemented with NI-SWITCH, instead of NI-FAKE due to functions missing from nifake.h that were present in the metadata.
  * I looked into adding these to nifake.h, but after iterating on the PR several times (having to add more and more: attributes, enums, etc, I decided it wasn't worth the effort).
* Updated Library Classes
  * They now take in a shared_ptr<SharedLibraryInterface> instead of creating their own SharedLibrary object, to support the unit tests
 * Another constructor was added to automatically create the shared_pointer<SharedLibrary> object to preserve existing behavior
  * Added an attribute and method to track, determine whether SetRuntimeEnvironment has been called (for testing purposes)

There are no plans to add system-level testing for SetRuntimeEnvironment in grpc-device, so this is the only testing we'll have to make sure it doesn't get broken.

Ran unit tests locally


source/codegen/templates/library.cpp.mako
```diff
@@ -15,18 +15,26 @@ linux_so_version_suffix = f'.{linux_abi_version}' if linux_abi_version else ''
 windows_library_info = config['library_info']['Windows']['64bit']
 windows_library_name = windows_library_info['name']
 
+class_name = f"{service_class_prefix}Library"
+
+intialization_list = "p_shared_library_(pSharedLibrary)"
```
bkeryan, line 20: spelling

source/codegen/templates/library.cpp.mako
```diff
@@ -35,10 +43,13 @@ static const char* kLibraryName = "lib${linux_library_name}.so${linux_so_version
...
 namespace ${config["namespace_component"]}_grpc {
 
-${service_class_prefix}Library::${service_class_prefix}Library() : shared_library_(kLibraryName)
+${class_name}::${class_name}() : ${class_name}(std::make_shared<nidevice_grpc::SharedLibrary>()) {}
+
+${class_name}::${class_name}(std::shared_ptr<nidevice_grpc::SharedLibraryInterface> pSharedLibrary) : ${intialization_list}
 {
-  shared_library_.load();
-  bool loaded = shared_library_.is_loaded();
+  p_shared_library_->set_library_name(kLibraryName);
```
bkeryan, line 50: Why 'p'?

source/codegen/templates/library.h.mako
```diff
@@ -60,8 +70,11 @@ class ${service_class_prefix}Library : public ${namespace_prefix}::${service_cla
 % endfor
   } FunctionLoadStatus;
 
-  nidevice_grpc::SharedLibrary shared_library_;
+  std::shared_ptr<nidevice_grpc::SharedLibraryInterface> p_shared_library_;
   FunctionPointers function_pointers_;
+% if set_runtime_environment_supported:
+  bool runtime_environment_set; // needed to test that we properly call SetRuntimeEnvironment
```
bkeryan, line 76: Missing trailing underscore

source/codegen/templates/library.h.mako
```diff
@@ -37,6 +44,9 @@ class ${service_class_prefix}Library : public ${namespace_prefix}::${service_cla
 %>\
   ${return_type} ${method_name}(${service_helpers.create_params(parameters, expand_varargs=True)});
 % endfor
+% if set_runtime_environment_supported:
+  bool get_runtime_environment_set(); // needed to test that we properly call SetRuntimeEnvironment
```
bkeryan, line 48: "is"/"was" sounds more natural for Booleans than "get".

source/tests/unit/library_set_runtime_environment_tests.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+  auto library = std::make_shared<niswitch_grpc::NiSwitchLibrary>(p_shared_library);
+  EXPECT_TRUE(library->get_runtime_environment_set());
+}
+
+TEST(LibrarySetRuntimeEnvironmentTests, Library_SetRuntimeEnvironmentNotFound_DoesNotCallSetRuntimeEnvironment)
+{
+  auto p_shared_library = std::make_shared<MockSharedLibrary>();
+  EXPECT_CALL(*p_shared_library, set_library_name(_))
+    .Times(1);
```
bkeryan, line 49: `.Times(1)` is the default for most of these calls.
https://google.github.io/googletest/gmock_for_dummies.html#cardinalities-how-many-times-will-it-be-called

source/tests/unit/library_set_runtime_environment_tests.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+  const char* func_name = "niSwitch_SetRuntimeEnvironment";
+
+  ON_CALL(*p_shared_library, get_function_pointer(_))
+    .WillByDefault(Return(nullptr));
+  // We must return an address for SetRuntimeEnvironment or we won't call it.
+  ON_CALL(*p_shared_library, get_function_pointer(func_name))
+    .WillByDefault(Return(&FunctionToPointAt));
+  EXPECT_CALL(*p_shared_library, get_function_pointer(_))
+    .Times(AtLeast(1));
```
bkeryan, line 37: You can simplify this by removing the `ON_CALL` and using `.WillRepeatedly(Return(...))` or `.WillOnce(Return(...))` on the `EXPECT_CALL`.
https://google.github.io/googletest/gmock_for_dummies.html#cardinalities-how-many-times-will-it-be-called

`ON_CALL` is more commonly used to specify default behavior in a setup/constructor or to specify behavior for `NiceMock`.

source/codegen/templates/library.cpp.mako
ni-jfitzger: Because it's a pointer.
I'm open to changing the name.

source/tests/unit/library_set_runtime_environment_tests.cpp
ni-jfitzger: I'll give it a shot. I actually tried that first and couldn't seem to get it working.

source/codegen/templates/library.cpp.mako
bkeryan: Does any other pointer variable in this codebase have a "p" prefix? When in Rome, do as the Romans do.

reckenro (approved):
Looks good to me. And Alex is out through next week so no need to wait on his review.

source/codegen/templates/library.cpp.mako
ni-jfitzger: Updated the name for the attribute and the constructor parameter.

ni-jfitzger:
> Linux CI failures:
> 
> * `/home/runner/work/grpc-device/grpc-device/./generated/niswitch/niswitch_library.h:20:3: error: extra qualification ‘niswitch_grpc::NiSwitchLibrary::’ on member ‘NiSwitchLibrary’ [-fpermissive]`
> * `/home/runner/work/grpc-device/grpc-device/third_party/gtest/googlemock/include/gmock/gmock-actions.h:682:40: error: invalid conversion from ‘void (*)()’ to ‘const void*’ [-fpermissive]`

Removed the unnecessary qualifer and added const_cast<const void*>(reinterpret_cast<void*>())

generated/nidaqmx/nidaqmx_library.h
```diff
@@ -17,7 +17,7 @@ namespace nidaqmx_grpc {
 class NiDAQmxLibrary : public nidaqmx_grpc::NiDAQmxLibraryInterface {
  public:
   NiDAQmxLibrary();
-  NiDAQmxLibrary::NiDAQmxLibrary(std::shared_ptr<nidevice_grpc::SharedLibraryInterface>);
+  NiDAQmxLibrary(std::shared_ptr<nidevice_grpc::SharedLibraryInterface>);
```
bkeryan, line 20: Two minor nitpicks I didn't notice before:
- The parameter is unnamed
- This single-parameter constructor is missing `explicit`, so it enables implicit conversion from `shared_ptr<SharedLibraryInterface>` to `NiDAQmxLibrary&`, etc.

bkeryan (approved):

generated/nidaqmx/nidaqmx_library.h
ni-jfitzger: >The parameter is unnamed

Is it not common to just specify parameter types in the function declaration?
I can add the name.

>This single-parameter constructor is missing explicit, so it enables implicit conversion from shared_ptr<SharedLibraryInterface> to NiDAQmxLibrary&, etc.

Nice catch.

bkeryan: > Is it not common to just specify parameter types in the function declaration?

I usually see unnamed parameters used to work around unused variable warnings.

ni-jfitzger:
@reckenro @bkeryan Are the systems tests waiting for approval from an owner before running?

bkeryan:
> @reckenro @bkeryan Are the systems tests waiting for approval from an owner before running?

No, this is an intermittent issue with the self-hosted GitHub runner. I'll send you more info offline.

PR #981, opened 2023-08-25
add RFmxInstr_InitializeWithChannel API  to Grpc device

alenkani:

Created an entry point  RFmxInstr_InitializeWithChannel API  to Grpc device
https://dev.azure.com/ni/DevCentral/_workitems/edit/2497354/

This entry point will be called from SFP, Hence it is required to define in the grpc-device

Built it locally


source/codegen/metadata/nirfmxinstr_restricted/functions.py
```diff
@@ -1212,5 +1212,61 @@
             }
 	],
 	'returns': 'int32'
-    }
+    },
+
+    'InitializeWithChannel': {
```
reckenro, line 1217: I would've expected this method to go in the main nirfmxinstr.proto file since it's a public method on the driver API. So I'd expect to bump the resources pulled into scrapigen for Instr to pull in this new method. You can see [this PR in scrapigen](https://github.com/ni/grpc-device-scrapigen/pull/182) which was bumping RFmx SpecAn metadata files to 23.5 as an example.

reckenro (changes requested):
I think this method belongs in the non-restricted proto file for RFmx Instr.

source/codegen/metadata/nirfmxinstr_restricted/functions.py
alenkani: this is not public API , Its hidden API

alenkani: Addressing build failures stemming from the unavailability of certain functions within the 'nirfmxinstr_service.h' header file in the restricted context, i have opted to migrate this API to the main proto file

astarche: Does this need to be added to `scrapigen` as well? I remember instr is tricky because the rfmx metadata is (was?) incomplete. Maybe @ThangamV-NI can help make sure those stay in sync.

astarche (approved):

PR #985, opened 2023-08-28
Improved codegen for pointers.

dmondrik:

- There was existing logic to convert `ViAddr` to `fixed64` as a gRPC type, which is good. However, it didn't convert input or output parameters of that type. Now it works as expected by using `reinterpret_cast`.
- There was existing logic to hard-cast types such as ViInt16 using the old C-style cast. This updates such code to use `static_cast` instead.

Precursor to planned API extensions.

Added tests for each of these as part of the fake extensions and verified it compiles.
- See `nifake_extension.proto` and `nifake_extension_service.cpp` for examples.

### DIFF
--- a/source/codegen/metadata/nifake_extension/enums.py
+++ b/source/codegen/metadata/nifake_extension/enums.py
@@ -1,20 +1,3 @@
 # -*- coding: utf-8 -*-
-enums = {
-    'LargeValue': {
-        'codegen_method': 'public',
-        'values': [
-            {
-                'name': 'ZERO',
-                'value': 0
-            },
-            {
-                'name': 'POSITIVE',
-                'value': 1073684512
-            },
-            {
-                'name': 'NEGATIVE',
-                'value': 3221168161
-            }
-        ]
-    }
-}
+# This file is generated from NI-FAKE API metadata version 1.2.0d9
+enums = {}
--- a/source/codegen/metadata/nifake_extension/enums.py
+++ b/source/codegen/metadata/nifake_extension/enums.py
@@ -1,3 +1,20 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FAKE API metadata version 1.2.0d9
-enums = {}
+enums = {
+    'LargeValue': {
+        'codegen_method': 'public',
+        'values': [
+            {
+                'name': 'ZERO',
+                'value': 0
+            },
+            {
+                'name': 'POSITIVE',
+                'value': 1073684512
+            },
+            {
+                'name': 'NEGATIVE',
+                'value': 3221168161
+            }
+        ]
+    }
+}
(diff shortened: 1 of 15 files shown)


danielhuani (approved):

source/codegen/templates/proto_helpers.mako
```diff
@@ -26,8 +26,9 @@ enum ${common_helpers.get_attribute_enum_name(group_name, sub_group, config)} {
 %   for attribute in attributes:
 <%
    attribute_name = attributes[attribute]["name"]
+   signed_attribute_value = attribute if attribute <= 0x7FFFFFFF else attribute - 2**32
```
reckenro, line 29: I think all the changes in this file go away if we go with representing numbers either with negative sign in metadata or with `'generate-mappings': True`

source/codegen/metadata/nifake_extension/enums.py
```diff
@@ -1,3 +1,20 @@
...
+                'name': 'ZERO',
+                'value': 0
+            },
+            {
+                'name': 'POSITIVE',
+                'value': 1073684512
+            },
+            {
+                'name': 'NEGATIVE',
+                'value': 3221168161
```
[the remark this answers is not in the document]
dmondrik: Well we could change scrapigen to do that. I don't like the idea of the separate mappings, because it's really not needed since these are 32-bit values.

source/codegen/metadata/nifake_extension/enums.py
```diff
@@ -1,3 +1,20 @@
 # -*- coding: utf-8 -*-
-# This file is generated from NI-FAKE API metadata version 1.2.0d9
-enums = {}
+enums = {
```
astarche, line 2: Putting this in `nifake` would be better, assuming there's not a downside to that I'm missing. `nifake_extension` is intended to prove we can support a service that operates on sessions from a different service. So it doesn't need any new functionality unless it's related somehow to cross-driver sessions. `nifake` is a better home for covering general codegen features.

source/codegen/metadata/nifake_extension/enums.py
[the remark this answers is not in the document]
astarche: I agree `generate-mappings` doesn't sound as good. If it's easy to fix in `scrapigen` I'd prefer that so that the metadata files are a valid-source-of-truth. If that doesn't work or isn't worth it, I'm alright with this or whatever y'all agree on. You could also consider heavy-handed-verbosity like an `overflow_to_i32` flag or something like that.

astarche (approved):

source/codegen/metadata/nifake_extension/enums.py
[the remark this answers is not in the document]
dmondrik: Fixing in scrapigen.

source/codegen/templates/proto_helpers.mako
dmondrik: Yep, reverted this file.

source/codegen/metadata/nifake_extension/enums.py
dmondrik: Since I'm reverting this enum file anyway, I'll consider it for future changes. Ryan felt the other change (to functions) was reasonable to do here.

reckenro (approved):

PR #988, opened 2023-08-29
Update niRFmxSpecAn to 23.8

ThangamV-NI:

Updates niRFmxSpecAn support to 23.8 API.

This adds the following:

**Methods**
RFmxSpecAn_MarkerCfgYLocation

**Attributes**
RFMXSPECAN_ATTR_IM_RESULTS_WORST_CASE_INTERMOD_ABSOLUTE_POWER
RFMXSPECAN_ATTR_IM_RESULTS_LOWER_INTERMOD_RELATIVE_POWER
RFMXSPECAN_ATTR_IM_RESULTS_UPPER_INTERMOD_RELATIVE_POWER
RFMXSPECAN_ATTR_IM_RESULTS_WORST_CASE_INTERMOD_RELATIVE_POWER

**Enums Values**
RFMXSPECAN_VAL_MARKER_MARKER_TYPE_FIXED

https://dev.azure.com/ni/DevCentral/_workitems/edit/2476252

Locally built grpc-device.


reckenro (changes requested):
Can you separate out the SpecAn changes from the BT ones for this PR?

ThangamV-NI:
> Can you separate out the SpecAn changes from the BT ones for this PR?

Resolved it now.

generated/nirfmxspecan/nirfmxspecan.proto
```diff
@@ -5614,8 +5620,8 @@ message IMFetchIntermodMeasurementRequest {
 message IMFetchIntermodMeasurementResponse {
   int32 status = 1;
   int32 intermod_order = 2;
-  double lower_intermod_power = 3;
-  double upper_intermod_power = 4;
+  double lower_intermod_absolute_power = 3;
```
astarche, line 5623: This is a source breaking change. Existing clients will still be compatible. But when they pull in the new proto file and re-protoc, they will most likely need to make source changes. If they are using a typed language (or using a type checker) this will show up as a compile error. In a dynamically typed language it will be a runtime error.

Is that OK to the RFmx team? Is it OK to your clients?

I will usually allow these because it's more of an inconvenience than a functional problem and we haven't gotten pushback on it. But it's definitely not preferred.

FYI @maxxboehme @reckenro

astarche (approved):

reckenro (approved):

generated/nirfmxspecan/nirfmxspecan.proto
reckenro: Right, we've mostly left it to the driver teams to make that call and like Alex said there hasn't been pushback so far. If we do decide to continue with changing the field names we'll just need to add the "source-breaking" label to this PR. That will make it easier to note in the next release.

ThangamV-NI: We have decided not to rename the parameters.

PR #991, opened 2023-08-30
Add VISA metadata to generate proto and services files

danielhuani:

1.  Copy the visa metadata generated by Scrapigen and use it to generate the proto file and service files
2.  Implemented custom functions: FindRsrc, ParseRsrc, Open, Close
3.  Added an example for finding visa resources

[User Story 2477322](https://ni.visualstudio.com/DevCentral/_workitems/edit/2477322): [grpc-device server] Generate visa.proto from python metadata (Complete)
[User Story 2479692](https://ni.visualstudio.com/DevCentral/_workitems/edit/2479692): [grpc-device server] Implement custom functions for visa in grpc-device (In progress)

Manually run the find-resources python example and tested open close in python client.
image


source/custom/visa_service.custom.cpp
```diff
@@ -0,0 +1,440 @@
...
+      if (request->object_handle().has_vi())
+      {
+        auto vi_grpc_session = request->object_handle().vi();
+        ViSession vi = session_repository_->access_session(vi_grpc_session.name());
+        session_repository_->remove_session(vi_grpc_session.name());
+        status = library_->Close(vi);
+        if (!status_ok(status)) {
+        return ConvertApiErrorStatusForViObject(context, status, vi);
+      }
+      }
```
dmondrik, line 77: Fix alignment. Will a style checker complain? Same below in this method.

source/custom/visa_service.custom.cpp
```diff
@@ -0,0 +1,440 @@
...
+  {
+    if (context->IsCancelled()) {
+      return ::grpc::Status::CANCELLED;
+    }
+    try {
+      ViStatus status;
+      if (request->object_handle().has_vi())
+      {
+        auto vi_grpc_session = request->object_handle().vi();
+        ViSession vi = session_repository_->access_session(vi_grpc_session.name());
```
dmondrik, line 71: Declare the local variable as `ViObject vi` above the if clause. The lines to actually call `Close()` and check the status can move below the if/else. That way the only part inside the conditional is how it gets assigned and what happens (one block removes the session, the other does not).

source/custom/visa_service.custom.cpp
```diff
@@ -0,0 +1,440 @@
...
+        return ConvertApiErrorStatusForViObject(context, status, object_handle);
+        }
+      }
+      response->set_status(status);
+      return ::grpc::Status::OK;
+    }
+    catch (nidevice_grpc::NonDriverException& ex) {
+      return ex.GetStatus();
+    }
+    catch (const DriverErrorException& ex) {
```
dmondrik, line 92: Nothing in this method throws this type (unlike the others that use `GetResourceManagerSession`). It's unclear to me whether it's better to be precise or make them all the same for the sake of safety and readability.

dmondrik (approved):
Fine with the specified changes, no need for me to re-review.

examples/visa/find-resources.py
```diff
@@ -0,0 +1,67 @@
...
+channel = grpc.insecure_channel(f"{SERVER_ADDRESS}:{SERVER_PORT}")
+client = grpc_visa.VisaStub(channel)
+
+try:
+    # Query for resources.
+    find_resources_resp = client.FindRsrc(visa_types.FindRsrcRequest(expression="?*"))
+
+    # Get information about each resource.
+    for resource in find_resources_resp.instrument_descriptor:
+        parse_resp = client.ParseRsrcEx(visa_types.ParseRsrcExRequest(resource_name=resource))
```
dmondrik, line 50: I forgot to fix this, it should be:
`parse_resp = client.ParseRsrc(visa_types.ParseRsrcRequest(resource_name=resource))`

SparkingSpork (changes requested):
We typically add the accompanying system tests for new drivers when adding a new driver. Can you add tests similar to the other drivers? I think for this PR a simple test that you can open and close a session is sufficient. See the tests under `source/tests/system` for some examples.

dmondrik:
> We typically add the accompanying system tests for new drivers when adding a new driver. Can you add tests similar to the other drivers? I think for this PR a simple test that you can open and close a session is sufficient. See the tests under `source/tests/system` for some examples.

@SparkingSpork I don't think that's going to work so well. Other drivers work with things like simulated resources. VISA doesn't have that.

bkeryan:
> @SparkingSpork I don't think that's going to work so well. Other drivers work with things like simulated resources. VISA doesn't have that.

Some ideas:
- Use ASRL1::INSTR. If it's not present, call GTEST_SKIP. This would have limited functionality, since the real COM1 may not be connected to an instrument.
- Use the VISA resource manager to find a valid resource. If none are present, call GTEST_SKIP.
- Install a virtual COM port driver on grpcbot. This would provide a pair of virtual COM ports that are connected together in software.

SparkingSpork:
> > @SparkingSpork I don't think that's going to work so well. Other drivers work with things like simulated resources. VISA doesn't have that.
> 
> Some ideas:
> 
> * Use ASRL1::INSTR. If it's not present, call GTEST_SKIP. This would have limited functionality, since the real COM1 may not be connected to an instrument.
> * Use the VISA resource manager to find a valid resource. If none are present, call GTEST_SKIP.
> * Install a virtual COM port driver on grpcbot. This would provide a pair of virtual COM ports that are connected together in software.

Another option would be to use a VISA TCP/IP SOCKET address and spin up a localhost socket to talk to.

bkeryan:
> Another option would be to use a VISA TCP/IP SOCKET address and spin up a localhost socket to talk to.

That is a much better option.

examples/visa/find-resources.py
danielhuani: Commit 2

source/tests/system/visa_session_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+};
+
+TEST_F(VisaSessionTest, OpenSessionWithInstrumentDescriptor_CreatesDriverSession)
+{
+  visa::OpenResponse response;
+  ::grpc::Status status = call_open(kInstrumentDescriptor, visa::LOCK_STATE_NO_LOCK, kVisaTestSession, 0, &response);
+
+  EXPECT_TRUE(status.ok());
+  EXPECT_EQ(0, response.status());
+  EXPECT_NE("", response.vi().name());
```
dmondrik, line 60: If we request a specific session name, this should test equality against that. (In the next test where you specify an empty string, this check makes sense there.)

source/tests/system/visa_session_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+namespace system {
+
+namespace visa = visa_grpc;
+
+const int kInvalidRsrc = -1073807343;
+const int kInvalidVisaSessionWarning = 1073676418;
+const char* kVisaErrorInstrumentDescriptorNotFoundMessage = "Insufficient location information or the device or resource is not present in the system.";
+const char* kInstrumentDescriptor = "TCPIP::www.ni.com::80::SOCKET";
+const char* kVisaTestSession = "SessionName";
+const char* kVisaTestInvalidInstrumentDescriptor = "";
```
dmondrik, line 16: I would make this "FOO::BAR"

source/tests/system/visa_session_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  EXPECT_TRUE(status.ok());
+  EXPECT_EQ(0, close_response.status());
+}
+
+TEST_F(VisaSessionTest, OpenWithErrorFromDriver_ReturnsDriverErrorWithUserErrorMessage)
+{
+  EXPECT_THROW_DRIVER_ERROR_WITH_SUBSTR({
+    visa::OpenResponse init_response;
+    call_open("kInstrumentDescriptor", visa::LOCK_STATE_NO_LOCK, "", 0, &init_response);
```
dmondrik, line 93: This should use `kVisaTestInvalidInstrumentDescriptor` without quotes.

dmondrik (approved):

source/custom/visa_service.custom.cpp
danielhuani: Commit 3

source/custom/visa_service.custom.cpp
danielhuani: Dan has an elegant implementation for Close in his new commit in the draft PR, and will post a PR with the set/get attributes implementation after this.

source/custom/visa_service.custom.cpp
danielhuani: Commit 5

source/tests/system/visa_session_tests.cpp
danielhuani: Commit 5

source/tests/system/visa_session_tests.cpp
danielhuani: Commit 5

source/tests/system/visa_session_tests.cpp
danielhuani: Commit 5

SparkingSpork (approved):

astarche (approved):

reckenro (approved):

PR #992, opened 2023-09-01
Implement support for get/set attribute

dmondrik:

- Updates to the latest metadata from scrapigen.
- Implement support for get/set attribute.
- Make it clearer for methods that support both events and sessions that they have distinct proto entries.

User Story 2479692: [grpc-device server] Implement custom functions for visa in grpc-device

- All VISA tests pass.
- Manual testing as well.


source/custom/visa_service.custom.cpp
```diff
@@ -39,6 +43,275 @@ static ViSession GetResourceManagerSession(visa_grpc::VisaService::LibraryShared
   return s_defaultRM;
 }
 
+static AttributeValueData::DataCase GetAttributeType(ViAttr attributeID)
```
astarche, line 46: I take it that we're not likely to add new attributes to visa? This would be hard to keep up-to-date in custom code. We could probably figure out how to generate it if necessary.

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -0,0 +1,192 @@
...
+    visa::CloseResponse response;
+    ::grpc::Status status = GetStub()->Close(&context, request, &response);
+
+    EXPECT_TRUE(status.ok());
+    EXPECT_EQ(VI_SUCCESS, response.status());
+  }
+
+  void set_bool_attribute(visa::VisaAttribute attribute, bool value)
+  {
+    ::grpc::ClientContext context;
```
astarche, line 79: I don't have a problem with writing the driver API tests like this, but it's usually much easier to write them with the generated clients. All of the services added since daq (I think) use those for the driver API.

source/custom/visa_service.custom.cpp
astarche: Or perhaps a test that loops over all attributes in the proto

source/custom/visa_service.custom.cpp
```diff
@@ -308,62 +571,57 @@ ::grpc::Status VisaService::ReadAsync(::grpc::ServerContext* context, const Read
...
+    else if (request->attribute_value().has_value_u16()) {
+      attribute_value = static_cast<ViAttrState>(request->attribute_value().value_u16());
+    }
+    else if (request->attribute_value().has_value_u8()) {
+      attribute_value = static_cast<ViAttrState>(request->attribute_value().value_u8());
+    }
+    else {
+      throw nidevice_grpc::NonDriverException(::grpc::INVALID_ARGUMENT, "The value for attribute_value was not specified or out of range");
+    }
+    auto status = library_->SetAttribute(vi, attributeID, attribute_value);
```
astarche, line 613: Are you OK with the behavior if the user-provided value does not match the type of the attribute ID? It looks like the worst case would be truncation or maybe roll over?

DAQ has a varargs generic accessor API where incorrect types could be a crash or other undefined behavior. For that we had a check in the service to make sure they matched and there's a config token to allow unknown attributes (i.e., potentially added in future versions). Although I think that's only valid for "raw" attribute IDs.

astarche (approved):

generated/visa/visa.proto
```diff
@@ -496,13 +495,21 @@ message ClearResponse {
...
+  nidevice_grpc.Session vi = 1;
 }
 
 message CloseResponse {
   int32 status = 1;
 }
 
+message CloseEventRequest {
+  uint32 event_handle = 1;
+}
```
bkeryan, line 507: This event handle is generated on the server by the call to WaitForEvent, right?

Consider allowing the client to assign a stable name for the event. See @ccifra's reasoning in  https://github.com/ni/grpc-device/pull/784:

> I don't think ID should ever be used. There are a lot of advantages to keeping the data on the wire "time invariant" or independent on server state. For example, we are trying to do batching to improve throughput when doing "cloud-based measurements" as part of various LTIs and hopefully products.

bkeryan: It looks like job identifiers for async calls are also server-generated tokens.

source/custom/visa_service.custom.cpp
dmondrik: The server is a 64-bit process, and `ViAttrState` is a 64-bit value. All the others will be smaller and will increase in size so they won't be truncated. If the user passes an attribute size that's smaller than needed - let's say a timeout of just over 4 seconds, 4096 specifically, and they pass it as i16 or u16 - that's no different than putting the constant 4096 on the stack. So I'm not worried here. And if they pass a *too-large* value, such as `0xFFFFFFFF00001000ULL`, well VISA is going to end up taking the 64-bit value and truncating it there. So I'm not really worried about that either.

generated/visa/visa.proto
dmondrik: Async jobs and events are so temporal, I would argue strongly against naming them. There's really no way that ever makes sense to me. 

I agree completely that long-lived objects such as sessions should always be named.

source/custom/visa_service.custom.cpp
dmondrik: Auto generation from `attributes.py` sounds more promising. If we have that, I don't think a test makes that much sense.

generated/visa/visa.proto
bkeryan: I should have said "stable identifier". It doesn't have to be a name. Allowing the client to choose a numeric identifier would enable batching of multiple calls, record/replay use cases, etc.

Would it be possible to implement this later? Currently, the server has a single flat namespace for event handles, so clients would not be able to choose their own event handles without causing collisions. If event handles were scoped by session, this would be less of a problem (assuming the session is accessed by only one client at a time, and each client completes any events or jobs before another client takes ownership of the session).

dmondrik: After the weekend, I still don't think it makes sense for the 99% case. Consider LV code. The user never gets access to the job ID or the event ID. A single LV read/write operation encapsulates all of [begin, wait, get completion data from the event, close the event]. Even service requests are hidden behind a VI that does [wait, close the event].

I would hope this would be something that one could add later if there were a useful business case, but I still can't grasp that in my mind at this time.

bkeryan: Ok, thank you for considering it.

SparkingSpork (approved):

reckenro (approved):

source/custom/visa_service.custom.cpp
dmondrik: Once you get the .mako hooked in, auto generation from our dictionary turns out to be quite easy. Thanks for the suggestion.

source/tests/system/visa_driver_api_tests.cpp
dmondrik: Yep, the client code does indeed look nicer. I modified most VISA driver tests to take advantage of it.

PR #994, opened 2023-09-08
Close VISA resource manager on server reset

dmondrik:

It closes the VISA resource manager when the server is reset.
- It does this by registering a fake session name (that really should be internal-only) that cleans up the static variable as well as closing the session.
- But it's lightweight enough that we don't need to worry about closing it in normal I/O session usage.

The server being reset should clean up as much as possible.

Added `visa_resource_manager_tests`
- Instantiating `OpenRequest`, `OpenResponse` etc. had the side effect of requiring many more .cpp files to be added to the integration test executable. Otherwise I got unresolved link errors.
image

### DIFF
--- a/source/custom/visa_service.custom.cpp
+++ b/source/custom/visa_service.custom.cpp
@@ -32,7 +32,7 @@ class DriverErrorException : public std::runtime_error {
   }
 };
 
-static ViSession GetResourceManagerSession(visa_grpc::VisaService::LibrarySharedPtr library)
+static ViSession GetResourceManagerSession(visa_grpc::VisaService::ResourceRepositorySharedPtr repository, visa_grpc::VisaService::LibrarySharedPtr library)
 {
   std::unique_lock<std::shared_mutex> lock(s_resource_manager_lock);
   if (!s_defaultRM) {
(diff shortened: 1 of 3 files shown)


SparkingSpork (approved):

source/custom/visa_service.custom.cpp
```diff
@@ -32,14 +32,19 @@ class DriverErrorException : public std::runtime_error {
...
   std::unique_lock<std::shared_mutex> lock(s_resource_manager_lock);
   if (!s_defaultRM) {
     ViStatus status = library->OpenDefaultRM(&s_defaultRM);
     if (!status_ok(status)) {
       throw DriverErrorException(status);
     }
+    // This should be cleaned up only when the user explicitly resets the server to close all sessions.
+    std::string manager_name = "INTERNAL::VISA::RESOURCE_MANAGER";
+    auto init_lambda = [&] () { return std::make_tuple(0, s_defaultRM); };
+    auto cleanup_lambda = [library](ViSession id) { library->Close(id); s_defaultRM = VI_NULL; };
```
bkeryan, line 46: https://www.ni.com/docs/en-US/bundle/ni-visa-api-ref/page/ni-visa-api-ref/viclose.html says 

> Calling viClose() on a VISA Resource Manager session will also close all I/O sessions associated with that resource manager session.

Does it matter what order you close the resource manager and I/O sessions? If you happen to close the resource manager first, does closing the I/O sessions gracefully return `VI_ERROR_INV_OBJECT` and everything's ok?

If the order does matter, I am aware of two solutions:
- Use `register_dependent_session` / `add_dependent_session`
- Add a shared_ptr dependency from the I/O session's cleanup lambda to the resource manager session (but I guess dependent sessions are supposed to solve the same problem)

reckenro (approved):

source/custom/visa_service.custom.cpp
dmondrik: We had looked at the dependent stuff first, but that has the opposite relationship that we want, so it doesn't work. The order of closing doesn't cause problems - VISA must recognize that the I/O sessions are no longer valid and just return a warning.

danielhuani: Maybe add an integration test case to cover the case where closing the resource manger while IO sessions are still available, and then calling the IO sessions (which are closed together with the resource manager) will return the expected warning.

danielhuani (approved):

source/custom/visa_service.custom.cpp
dmondrik: That would be a system test because it relies on the real installed driver. And that's just testing VISA functionality not gRPC functionality. So I don't think it really adds any value.

astarche (approved):

PR #995, opened 2023-09-15
Add override specifier to inherited library class functions

reckenro:

Add the [override specifier](https://en.cppreference.com/w/cpp/language/override) to library class functions that exist in the interface class it inherits from.

Although we own the parent class, it's generally good practice to use `override` when you override a method so that you get a compiler error if there's no function to override (usually because the signature of the function in the parent class changes).

Built it locally


reckenro:
Note: This is the same as #965 , but we're trying to see if we can get changes in today and that was in a fork so it wasn't as simple to merge main and regen from there so just duplicating that small `library.h.mako` change in a new branch and regening.

astarche (approved):

PR #912, opened 2023-04-12
Multi-personality RFmx examples.

yardov:
### What does this Pull Request accomplish?

How to mix multiple RFmx personalities using gRPC. It was not clear to me before if we needed to initialize each personality and keep track of the instrument variable.

### Why should this Pull Request be merged?

Show customers how to do this programming.

### What testing has been done?

Simple execution of the code. This mixes other two existing examples in the same one.

PR #930, opened 2023-05-19
examples: Fix nidaqmx mypy errors (except aio-related errors)

bkeryan:
### What does this Pull Request accomplish?

Fix non-aio-related mypy errors for nidaqmx examples.

The currently released version of mypy-protobuf (3.4.0) doesn't support aio yet, but it looks like support will be available in a future release.

### Why should this Pull Request be merged?

Demonstrate type-checking best practices (I hope).

Prerequisite for updating the CI workflow to validate nidaqmx examples in https://github.com/ni/grpc-device/pull/931

### What testing has been done?

Ran `validate_examples.py -p nidaqmx` before and after.

PR #996, opened 2023-09-15
Potential MyPy solutions for new example.

reckenro:
### Purpose

This is just a draft PR to show a potential solution to the MyPy violations being hit in #912 . This is targeted to that branch.

Merged main but the actual changes to fix the MyPy violations in that other PR are the last three commits here: https://github.com/ni/grpc-device/pull/996/files/4b741b1916f3c2b04ecb14889f2184e9c60a3074..159f657090ff99ee9d8354986bf27a544dc34e05

### My steps for running MyPy locally

Pre-req
1. Navigate to main grpc-device repo folder
2. Run `python -m pip install -r python_build_requirements.txt`
3. Run `python -m pip install poetry`

Validate new nirf example
1. Build
2. Navigate to build directory
3. Run `python ..\source\codegen\validate_examples.py -p "*nirf"`


examples/nidaqmx/analog-input.py
```diff
@@ -110,15 +112,15 @@ def check_for_warning(response):
...
+        f"Acquired {len(read_response.read_array)} samples",
+        f"({read_response.samps_per_chan_read} samples per channel)",
     )
 except grpc.RpcError as rpc_error:
     error_message = str(rpc_error.details() or "")
-    for key, value in rpc_error.trailing_metadata() or []:  # type: ignore
-        if key == "ni-error":
-            details = value if isinstance(value, str) else value.decode("utf-8")
-            error_message += f"\nError status: {details}"
+    for entry in rpc_error.trailing_metadata() or []:
```
bkeryan, line 120: Did you test that this works at run time? I thought we tried this and had problems with inconsistency between the sync and aio APIs.

bkeryan: grpcio's sync `MetadataType` is `Sequence[Tuple[str, Union[str, bytes]]]` which does not allow using named `key` and `value` fields: https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_typing.py#L39

I think the underlying cython code uses a plain old tuple at run time: https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_cython/_cygrpc/metadata.pyx.pxi#L61

grpc-stubs uses a class with `key` and `value` fields, which doesn't support destructuring. This seems like a bug in grpc-stubs. A `Tuple` or `NamedTuple` would support destructuring. https://github.com/shabbyrobe/grpc-stubs/blob/master/grpc-stubs/__init__.pyi#L416

bkeryan: > This is in main,

Hmm, ok, looks like I made that change in https://github.com/ni/grpc-device/pull/930. I'm pretty sure I tested it at run time, not just mypy.

reckenro: Yeah, sorry. Should've clarified up front in the PR description what this PR was about. Added that now.

examples/nirf/rfsa-rfsg-NR-analysis-only-modacc.py
```diff
@@ -299,15 +304,19 @@ def check_for_warning(response, instrument):
         )
     )
     # Once commit is called, the recommended acquisition settings can be queried
-    response = rfmxclient.GetAttributeF64(
+    # For these updates, the proto request (NR's GetAttributeF64)
+    # must match the RequestMessage type (come from nirfmxnr_types).
+    # But for these, since you're getting an INSTR attribute,
+    # I think you should use an instrclient's GetAttribute... RPC calls.
+    response = instrclient.GetAttributeF64(
         nirfmxinstr_types.GetAttributeF64Request(
```
yardov, line 312: Hi @reckenro , if I use instrclient, do I need to use instr types? Sometimes the variables I need are in the specific standard (NR, WCDMA, etc) but I am confused that they are used in a instrclient way.
In this case, I think you are right, they come from instr but in general how do I know? following the proto file?

reckenro: For the most part, yes. So in the nirfmxinstr.proto the [`GetAttirbuteF64` rpc method](https://github.com/ni/grpc-device/blob/main/generated/nirfmxinstr/nirfmxinstr.proto#L46) expects a `GetAttributeF64Request` from the same proto file [here](https://github.com/ni/grpc-device/blob/main/generated/nirfmxinstr/nirfmxinstr.proto#L720). So the rpc method I think always aligns it's Request and Response messages with ones from the same proto file.

And on that Request message for the attribute_id it expects an `NiRFmxInstrAttribute`: https://github.com/ni/grpc-device/blob/main/generated/nirfmxinstr/nirfmxinstr.proto#L723

I think the one thing for these RFmx client's is that they can share the Session used between them if I remember correctly, but most other messages used will come from the same .proto file.

reckenro: And for the most part I reference the .proto files. I think there are ways to get set up with a better python programming experience with typehints and such but I haven't set that up personally.

reckenro:
Abandoning since MyPy fixes have been incorporated in #912

PR #1004, opened 2023-09-29
Implement Read for VISA and add loopback read/write unit test

danielhuani:

- Implemented custom Read() function
- Created `tcp_echo_server` that could write back whatever input to a TCP socket
- Added unit tests to validate Read() and Wrtie() function for VISA.

Support Read and Write for VISA

Unit tests pass on both Windows and Ubuntu.


source/custom/visa_service.custom.cpp
```diff
@@ -320,28 +320,25 @@ ::grpc::Status VisaService::Read(::grpc::ServerContext* context, const ReadReque
...
   }
-#if 1
-  return ::grpc::Status(grpc::StatusCode::DO_NOT_USE, "Custom code not implemented yet");
-#else
   try {
     auto vi_grpc_session = request->vi();
     ViSession vi = session_repository_->access_session(vi_grpc_session.name());
     ViUInt32 count = request->count();
-    std::string buffer(return_count, '\0');
+    std::string buffer(256 - 1, '\0');
```
dmondrik, line 327: The size of the buffer needs to be `count` bytes. This is not a string and doesn't need to be 0 initialized. Read operations can contain purely binary data. I'd suggest this:
`std::vector<ViByte> buffer(count);`

source/tests/utilities/tcp_echo_server.h
```diff
@@ -0,0 +1,58 @@
...
+#ifdef _WIN32
+    closesocket(socket_fd_);
+#else
+    close(socket_fd_);
+#endif
+    }
+
+    void start()
+    {
+        do_read();
```
dmondrik, line 33: While it makes sense for read to call write, it doesn't make sense for write to call read again. That will blow up the stack after some time. Rather, have it return. Here, just do an infinite loop:
```
for (;;) {
   do_read();
}
```

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -153,6 +153,149 @@ TEST_F(VisaDriverApiTest, SetNonTcpipAttribute_ReturnsAttributeError)
...
+  #ifdef _WIN32
+    WSADATA wsa_data;
+    WSAStartup(MAKEWORD(2, 2), &wsa_data);
+  #endif
+    server_fd_ = socket(AF_INET, SOCK_STREAM, 0);
+    EXPECT_NE(-1, server_fd_);
+    struct sockaddr_in server_address;
+    server_address.sin_family = AF_INET;
+    server_address.sin_addr.s_addr = inet_addr("127.0.0.1");
+    server_address.sin_port = htons(std::stoi(portNumber));
```
dmondrik, line 190: You should specify port 0 when starting the server. Then use `getsockname` for example to find the port that the OS assigned you. It will be different on every run, but that's OK.

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -153,6 +153,149 @@ TEST_F(VisaDriverApiTest, SetNonTcpipAttribute_ReturnsAttributeError)
...
+TEST_F(VisaDriverLoopbackTest, WriteTwice_ReadOnce_Matches)
+{
+  std::string writeData = "Visa gRPC read/write test";
+  write(writeData + "1 ");
+  write(writeData + "2");
+  std::string expectedReadData = writeData + "1 " + writeData + "2";
+  read(expectedReadData.size(), expectedReadData, VI_SUCCESS_MAX_CNT);
+}
+
+TEST_F(VisaDriverLoopbackTest, ReadMoreThanWritten_ReturnTimeout)
```
dmondrik, line 262: Excellent test! I'm glad this works. I'd like to make the test method name a little clearer that it returns a response with what did get read along with the status. Maybe this?
`ReadMoreThanWritten_ReturnTimeoutErrorWithDataInResponsePacket`

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -153,6 +153,149 @@ TEST_F(VisaDriverApiTest, SetNonTcpipAttribute_ReturnsAttributeError)
...
+}
+
+TEST_F(VisaDriverLoopbackTest, WriteOnce_ReadTwice_Matches)
+{
+  std::string writeData = "Visa gRPC read/write test";
+  write(writeData);
+  read(10, writeData.substr(0, 10), VI_SUCCESS_MAX_CNT);
+  read(writeData.size() - 10, writeData.substr(10, writeData.size()), VI_SUCCESS_MAX_CNT);
+}
+
```
dmondrik, line 290: One test that might be useful is to set the attribute TERMCHAR_EN to true. Then write "Hello\nWorld\n" and do read(1024) twice. The first will return "Hello\n" and the second read will return "World\n" and both will be VI_SUCCESS_TERMCHAR.

source/custom/visa_service.custom.cpp
dmondrik: I think this would also be one of the very few places where you should add a catch block for `std::bad_alloc&`. In that case you should return the gRPC unknown code with NI error code `VI_ERROR_ALLOC`.

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -153,6 +153,149 @@ TEST_F(VisaDriverApiTest, SetNonTcpipAttribute_ReturnsAttributeError)
...
+TEST_F(VisaDriverLoopbackTest, ReadWithoutWrite_ThrowsError)
+{
+  EXPECT_THROW_DRIVER_ERROR(
+    read(10, "", -1),
+    VI_ERROR_TMO);
+}
```
dmondrik, line 296: Another error case would be to specify a size that's insanely large and should cause an error, maybe `ULONG_MAX`? That should trigger the `VI_ERROR_ALLOC` scenario I mention elsewhere.

dmondrik (changes requested):

source/custom/visa_service.custom.cpp
danielhuani: std::vector<ViBtye> buffer(count) is still allocated on the stack, I am not sure why it needs to check for `std::bad_alloc`. Does the `Read` in VISA throw? Since it returns an error, and it is a C API, it probably doesn't throw, right?

dmondrik: That's not quite correct. The *vector* object is on the stack, yes, but that's only 24 bytes (3 pointers to heap locations).

danielhuani: It turned out in the binding file `visa.pb.h`, it defines the read buffer as `::PROTOBUF_NAMESPACE_ID::internal::ArenaStringPtr buffer_;` which actually is a String type. I think that's why it generates the method with the std::string for the buffer, and we have to stick with it.

dmondrik: Saying the proto type is `bytes` is technically correct. We must be extra cautious to not use "strlen" or related methods anywhere on this. In particular, this points out an important test case for a buffer such as `A\x00B\x00C` and make sure all 5 bytes are properly included.

dmondrik: I'm generating the client side and realized I have to be careful because of this issue as well. It's a C/C++ binding limitation, not a protobuf limitation.

source/custom/visa_service.custom.cpp
```diff
@@ -320,28 +320,25 @@ ::grpc::Status VisaService::Read(::grpc::ServerContext* context, const ReadReque
...
     ViUInt32 return_count{};
     auto status = library_->Read(vi, (ViByte*)buffer.data(), count, &return_count);
-    if (!status_ok(status)) {
+    if (!status_ok(status) && return_count == 0) {
       return ConvertApiErrorStatusForViSession(context, status, vi);
     }
     response->set_status(status);
+    buffer.resize(return_count);
     response->set_buffer(buffer);
```
dmondrik, line 335: Don't bother resizing the buffer. Instead, use the explicit:
`response->set-buffer(buffer.data(), return_count);`

source/tests/system/visa_driver_api_tests.cpp
danielhuani: Does 'VI_ERROR_ALLOC` happen when there is no memory to allocate?

danielhuani: Then it would be hard to test, correct? If the system can't allocate memory, the system is barely able to complete the test.

source/tests/utilities/tcp_echo_server.h
```diff
@@ -0,0 +1,164 @@
...
+#ifdef _WIN32
+    closesocket(socket_fd_);
+#else
+    close(socket_fd_);
+#endif
+    }
+
+    void start()
+    {
+        while(!stop_session_) {
```
dmondrik, line 39: Nit: 1 space after control statements: `while (`

source/tests/utilities/tcp_echo_server.h
```diff
@@ -0,0 +1,164 @@
...
+    , stop_session_(false)
+    {
+    }
+
+    ~TcpEchoSession() {
+        stop_session_ = true;
+#ifdef _WIN32
+    closesocket(socket_fd_);
+#else
+    close(socket_fd_);
```
dmondrik, line 33: This should probably call `stop()` so that logic is single-sourced. Also, the calls to close the socket should be indented.

source/tests/utilities/tcp_echo_server.h
```diff
@@ -0,0 +1,164 @@
...
+        if (server_fd_ == -1) {
+            return -1;
+        }
+        struct sockaddr_in server_address;
+        std::memset(&server_address, 0, sizeof(server_address));
+        server_address.sin_family = AF_INET;
+        server_address.sin_addr.s_addr = inet_addr("127.0.0.1");
+        server_address.sin_port = htons(0);
+
+        if (bind(server_fd_, (struct sockaddr*)&server_address, sizeof(server_address)) == 0) {
```
dmondrik, line 122: Change this to be:
```
if (bind .... != 0) {
   return -1;
}
```
That way errors are handled immediately and the rest of the control flow can be at the left margin.

dmondrik (approved):
There's only a few small things remaining. Thanks for iterating on this!

source/tests/utilities/tcp_echo_server.h
```diff
@@ -0,0 +1,161 @@
...
+    void start()
+    {
+        while (!stop_session_) {
+            do_read();
+        }
+    }
+
+    void stop()
+    {
+	    stop_session_ = true;
```
reckenro, line 46: Nit: `stop_session_ = true;` is indented too much here.

We don't enforce it as much as python formatting but you can run clang-format on this file if you want. We have the .clang-format file specified here and instructions on setting it up to format on save in VS Code if you want to do that in this [Wiki section](https://github.com/ni/grpc-device/wiki/Development-Tips-and-Tricks#clang-format-on-save).

reckenro (approved):

source/tests/utilities/tcp_echo_server.h
danielhuani: Yeah, I did this change on Ubuntu which seems have weird indention configuration in its vim. Fixed.

astarche (approved):

PR #1005, opened 2023-10-03
We must close the Find handle to avoid a leak.

dmondrik:

Close the handle returned from `viFindRsrc` when we're done with it.

This fixes a resource leak.

`FindRsrc_OpensResourceManagerOnceAndClosesFindHandle`

### DIFF
--- a/source/custom/visa_service.custom.cpp
+++ b/source/custom/visa_service.custom.cpp
@@ -153,10 +153,12 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     for (ViUInt32 index = 1; index < return_count; ++index) {
       status = library_->FindNext(find_handle, descriptor);
       if (!status_ok(status)) {
+        library_->Close(find_handle);
         return ConvertApiErrorStatusForViSession(context, status, rsrc_manager_handle);
       }
       response->add_instrument_descriptor(descriptor);
     }
+    library_->Close(find_handle);
     response->set_status(status);
     return ::grpc::Status::OK;
   }
--- a/source/tests/integration/visa_resource_manager_tests.cpp
+++ b/source/tests/integration/visa_resource_manager_tests.cpp
@@ -27,6 +27,19 @@ OpenResponse call_open(VisaService& service, const char* session_name)
     return response;
 }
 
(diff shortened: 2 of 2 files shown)


danielhuani (approved):
It looks good to me.
Why does FindRsrc need to open the resource handle? Does ParseRsrc open the resource handle?

astarche (approved):

dmondrik:
> It looks good to me. Why does FindRsrc need to open the resource handle? Does ParseRsrc open the resource handle?

Why does it need to open the resource manager? Because the first parameter to viFindRsrc is the resource manager handle. This is one of 3 operations on a resource manager. The other 2 are Parse and Open, and yes they both do the same.

danielhuani:
> > It looks good to me. Why does FindRsrc need to open the resource handle? Does ParseRsrc open the resource handle?
> 
> Why does it need to open the resource manager? Because the first parameter to viFindRsrc is the resource manager handle. This is one of 3 operations on a resource manager. The other 2 are Parse and Open, and yes they both do the same.

I see, so it is the resource manager handle. Does Parse method need to close the resource manager as well, similar as this change to the FindRsrc?

dmondrik:
> I see, so it is the resource manager handle. Does Parse method need to close the resource manager as well, similar as this change to the FindRsrc?

This PR does not close the resource manager handle. This is yet another handle, specific to the response from viFindRsrc. Consider 2 distinct calls with different expressions to viFindRsrc that return multiple resources each. The "find handle" is the only way to distinguish the results, used only for repeated calls to viFindNext, and must be closed separately.

reckenro (approved):

danielhuani:
> > I see, so it is the resource manager handle. Does Parse method need to close the resource manager as well, similar as this change to the FindRsrc?
> 
> This PR does not close the resource manager handle. This is yet another handle, specific to the response from viFindRsrc. Consider 2 distinct calls with different expressions to viFindRsrc that return multiple resources each. The "find handle" is the only way to distinguish the results, used only for repeated calls to viFindNext, and must be closed separately.

Extended question not related to this PR, but why does the `find handle` need to be opened, and is not passed as a string instead? The resource name could be good to distinguish the resources, right?

PR #1009, opened 2023-10-05
Implement UsbControlIn

danielhuani:

Implement UsbControlIn the same way as Read

Add support for UsbControlIn method for visa

None.

### DIFF
--- a/source/custom/visa_service.custom.cpp
+++ b/source/custom/visa_service.custom.cpp
@@ -344,7 +344,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     std::string* buffer = response->mutable_buffer();
     buffer->resize(count);
     ViUInt32 return_count{};
-    auto status = library_->Read(vi, reinterpret_cast<ViByte*>(buffer->data()), count, &return_count);
+    auto status = library_->Read(vi, (ViByte*)buffer->data(), count, &return_count);
     if (!status_ok(status) && return_count == 0) {
       return ConvertApiErrorStatusForViSession(context, status, vi);
     }
@@ -461,7 +461,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     std::string* buffer = response->mutable_buffer();
     buffer->resize(w_length);
     ViUInt16 return_count{};
-    auto status = library_->UsbControlIn(vi, bm_request_type, b_request, w_value, w_index, w_length, reinterpret_cast<ViByte*>(buffer->data()), &return_count);
+    auto status = library_->UsbControlIn(vi, bm_request_type, b_request, w_value, w_index, w_length, (ViByte*)buffer->data(), &return_count);
     if (!status_ok(status) && return_count == 0) {
       return ConvertApiErrorStatusForViSession(context, status, vi);
     }
--- a/source/custom/visa_service.custom.cpp
+++ b/source/custom/visa_service.custom.cpp
@@ -162,7 +162,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     response->set_status(status);
     return ::grpc::Status::OK;
   }
-  catch (nidevice_grpc::NonDriverException& ex) {
+  catch (const nidevice_grpc::NonDriverException& ex) {
     return ex.GetStatus();
   }
   catch (const DriverErrorException& ex) {
@@ -188,7 +188,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     response->set_status(status);
     return ::grpc::Status::OK;
   }
-  catch (nidevice_grpc::NonDriverException& ex) {
+  catch (const nidevice_grpc::NonDriverException& ex) {
     return ex.GetStatus();
   }
 }
@@ -210,7 +210,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     response->set_status(status);
     return ::grpc::Status::OK;
   }
-  catch (nidevice_grpc::NonDriverException& ex) {
+  catch (const nidevice_grpc::NonDriverException& ex) {
     return ex.GetStatus();
   }
 }
@@ -276,7 +276,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     response->set_new_session_initialized(new_session_initialized);
     return ::grpc::Status::OK;
   }
-  catch (nidevice_grpc::NonDriverException& ex) {
+  catch (const nidevice_grpc::NonDriverException& ex) {
     return ex.GetStatus();
   }
   catch (const DriverErrorException& ex) {
@@ -321,7 +321,7 @@ static ViStatus GetAttributeValue(ViObject vi, ViAttr attributeID, VisaService::
     nidevice_grpc::converters::trim_trailing_nulls(*(response->mutable_alias_if_exists()));
     return ::grpc::Status::OK;
   }
-  catch (nidevice_grpc::NonDriverException& ex) {
+  catch (const nidevice_grpc::NonDriverException& ex) {
     return ex.GetStatus();
   }
   catch (const DriverErrorException& ex) {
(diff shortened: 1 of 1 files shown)


astarche (approved):

source/custom/visa_service.custom.cpp
```diff
@@ -458,21 +457,23 @@ ::grpc::Status VisaService::UsbControlIn(::grpc::ServerContext* context, const U
...
+    if (!status_ok(status) && return_count == 0) {
       return ConvertApiErrorStatusForViSession(context, status, vi);
     }
     response->set_status(status);
-    response->set_buffer(buffer);
+    response->set_buffer(buffer.data(), return_count);
     response->set_return_count(return_count);
     return ::grpc::Status::OK;
   }
+  catch (std::bad_alloc) {
```
dmondrik, line 471: We should catch by const reference:
`catch (const std::bad_alloc&) {`
Please fix the other catch blocks in this file as well.

source/custom/visa_service.custom.cpp
```diff
@@ -447,9 +447,8 @@ ::grpc::Status VisaService::UsbControlIn(::grpc::ServerContext* context, const U
...
     return ::grpc::Status::CANCELLED;
   }
-#if 1
-  return ::grpc::Status(grpc::StatusCode::DO_NOT_USE, "Custom code not implemented yet");
-#else
+  ViSession vi = VI_NULL;
+
   try {
     auto vi_grpc_session = request->vi();
     ViSession vi = session_repository_->access_session(vi_grpc_session.name());
```
dmondrik, line 454: This re-declares the intended variable, shadowing it. Just assign instead.

source/custom/visa_service.custom.cpp
```diff
@@ -458,21 +457,23 @@ ::grpc::Status VisaService::UsbControlIn(::grpc::ServerContext* context, const U
     ViUInt16 w_value = request->w_value();
     ViUInt16 w_index = request->w_index();
     ViUInt16 w_length = request->w_length();
-    std::string buffer(return_count, '\0');
+    std::vector<ViByte> buffer(w_length);
```
dmondrik, line 460: This means the buffer is allocated twice, once here and again in the response. I think we can optimize this just a bit:
```
std::string* buffer = response->mutable_buffer();
buffer->resize(w_length);
...
auto status = library_->UsbControlIn(... buffer->data(), &return_count);
...
buffer->resize(return_count);
```

dmondrik: And you should be able to do the same for Read.

dmondrik (changes requested):

source/custom/visa_service.custom.cpp
danielhuani: Commit 2.

Nice to save the extra allocation and copy. Do we consider to apply it for all the methods? Even beyond Visa?

source/custom/visa_service.custom.cpp
danielhuani: Commit 2.

dmondrik (approved):

source/custom/visa_service.custom.cpp
```diff
@@ -341,21 +341,22 @@ ::grpc::Status VisaService::Read(::grpc::ServerContext* context, const ReadReque
     auto vi_grpc_session = request->vi();
     vi = session_repository_->access_session(vi_grpc_session.name());
     ViUInt32 count = request->count();
-    std::vector<ViByte> buffer(count);
+    std::string* buffer = response->mutable_buffer();
+    buffer->resize(count);
     ViUInt32 return_count{};
-    auto status = library_->Read(vi, buffer.data(), count, &return_count);
+    auto status = library_->Read(vi, (ViByte*)buffer->data(), count, &return_count);
```
maxxboehme, line 347: When in C++ we should use C++ style casts.

maxxboehme (changes requested):

source/custom/visa_service.custom.cpp
danielhuani: This is a workaround, reinterpret_cast failed to compile on nilrt.

bkeryan: Have you tried `const_cast`? The non-const overload of [basic_string::data](https://en.cppreference.com/w/cpp/string/basic_string/data) was added in C++17.

danielhuani: Something like `reinterpret_cast<ViByte*>(const_cast<char*>(buffer->data())`? Is there a easy way to run build on nilrt to try things instead of adding a commit to the PR?

dmondrik: I am not a fan of just making C++ happy when it's not readable. The C style cast here is readable.

This is fundamentally a cast from "char*" to "unsigned char*". Yes the problem is that the non-const overload isn't available there in RT. I think casting it twice only makes it confusing. My 2¢ worth.

dmondrik: Two other possibilities that should work equally well and would be less confusing to me than above. My preference would be for the first:
`auto status = library_->Read(vi, reinterpret_cast<ViByte*>(&buffer[0]), count, &return_count);`
-or-
`auto status = library_->Read(vi, reinterpret_cast<ViByte*>(&buffer.front()), count, &return_count);`

danielhuani: Commit 3

reckenro (approved):

source/custom/visa_service.custom.cpp
bkeryan: On further thought, casting away const here would be a bug: refcounted implementations of `basic_string` don't unshare the string when you call a const method. Using `&buffer[0]` or `&buffer.front()` is correct because these methods have non-const overloads which unshare the string.

bkeryan: BTW, calling `&buffer[0]` or `&buffer.front()` on an empty `vector` is undefined behavior, but with `basic_string` it returns a reference to a NUL terminator. C++11 even allows overwriting the NUL terminator (with another NUL terminator): https://en.cppreference.com/w/cpp/string/basic_string/operator_at

dmondrik: "Interesting" note about the empty container (shudder). Not an issue here since we first reserve the size.

source/custom/visa_service.custom.cpp
dmondrik: I hadn't looked at the other ones. The custom code generally looks good to me. There are copies of course for cases where the structures are too different. But I did find in generated code a couple of possibilities for improvement. I created a "tech debt" issue #1011

source/custom/visa_service.custom.cpp
```diff
@@ -447,32 +448,34 @@ ::grpc::Status VisaService::UsbControlIn(::grpc::ServerContext* context, const U
...
     ViUInt16 w_value = request->w_value();
     ViUInt16 w_index = request->w_index();
     ViUInt16 w_length = request->w_length();
-    std::string buffer(return_count, '\0');
+    std::string* buffer = response->mutable_buffer();
+    buffer->resize(w_length);
     ViUInt16 return_count{};
-    auto status = library_->UsbControlIn(vi, bm_request_type, b_request, w_value, w_index, w_length, (ViByte*)buffer.data(), &return_count);
-    if (!status_ok(status)) {
+    auto status = library_->UsbControlIn(vi, bm_request_type, b_request, w_value, w_index, w_length, reinterpret_cast<ViByte*>(&buffer[0]), &return_count);
```
dmondrik, line 464: Hang on, I don't think this is doing what you want. I think it needs to be `(&buffer->front())`

I think what it might be doing right now is treating buffer as an array of strings, taking the first string object, and casting that pointer to the desired type.

But if that's true, your calls to Read right now should be failing on the prebuild. The only thing that's clear to me right now is this kind of C++ code gets confusing quickly.

dmondrik: And yes, that's exactly what happened. It crashed as soon as it called `WriteAndRead_Matches`.

danielhuani: `&buffer->front()` still throws exception when it tries to read 0 byte, and extra logic will need to handle the 0 byte case. I will revert back to the C style type convert.

source/custom/visa_service.custom.cpp
```diff
@@ -344,7 +351,7 @@ ::grpc::Status VisaService::Read(::grpc::ServerContext* context, const ReadReque
     std::string* buffer = response->mutable_buffer();
     buffer->resize(count);
     ViUInt32 return_count{};
-    auto status = library_->Read(vi, reinterpret_cast<ViByte*>(&buffer[0]), count, &return_count);
+    auto status = library_->Read(vi, get_buffer_data_pointer<std::string>(buffer, count), count, &return_count);
```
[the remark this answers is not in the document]
danielhuani: Right, I thought it would need to make the type explicitly to instantiate the template function. But it turns out that removing the type compiles just fine.

source/custom/visa_service.custom.cpp
danielhuani: Commit 4

source/custom/visa_service.custom.cpp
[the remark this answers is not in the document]
danielhuani: commit 5

PR #1010, opened 2023-10-06
Add register based tests for in/out/move-in/move-out

dmondrik:

Adds system auto tests for high level register based VISA operations.

- Increasing API coverage in parallel with client side development.
- Clean up some build warnings while I'm at it.
- Add some message based tests for methods that weren't being covered before.

Added `VisaRegisterBasedLoopbackTest` which passes.

### DIFF
--- a/source/tests/system/visa_driver_api_tests.cpp
+++ b/source/tests/system/visa_driver_api_tests.cpp
@@ -177,6 +177,8 @@ class VisaMessageBasedLoopbackTest : public VisaDriverApiTest {
     echoserver_.stop();
   }
 
+  const uint32_t kFastTimeoutMsec = 500;
+
   void write(const std::string& data)
   {
     auto response = client::write(GetStub(), GetNamedSession(), data);
@@ -227,6 +229,7 @@ TEST_F(VisaMessageBasedLoopbackTest, WriteTwice_ReadOnce_Matches)
 TEST_F(VisaMessageBasedLoopbackTest, ReadMoreThanWritten_ReturnTimeoutErrorWithDataInResponsePacket)
 {
   std::string writeData = "Visa gRPC read/write test";
+  set_uint32_attribute(visa::VisaAttribute::VISA_ATTRIBUTE_TMO_VALUE, kFastTimeoutMsec);
   write(writeData);
   read(writeData.size() + 1, writeData, VI_ERROR_TMO);
 }
@@ -247,6 +250,7 @@ TEST_F(VisaMessageBasedLoopbackTest, ReadZeroBytes_ReturnSuccess)
 
 TEST_F(VisaMessageBasedLoopbackTest, ReadWithoutWrite_ThrowsError)
 {
+  set_uint32_attribute(visa::VisaAttribute::VISA_ATTRIBUTE_TMO_VALUE, kFastTimeoutMsec);
   EXPECT_THROW_DRIVER_ERROR(
     read(10, "", -1),
     VI_ERROR_TMO);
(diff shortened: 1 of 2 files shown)


source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -258,6 +270,284 @@ TEST_F(VisaDriverLoopbackTest, WriteWithTermChar_ReadMatches)
...
+  set_bool_attribute(visa::VisaAttribute::VISA_ATTRIBUTE_TERMCHAR_EN, true);
+  set_uint32_attribute(visa::VisaAttribute::VISA_ATTRIBUTE_IO_PROT, 4); // VI_PROT_4882_STRS 
+  assert_trigger(visa::TRIGGER_PROTOCOL_PROT_DEFAULT);
+  read(255, "*TRG\n", VI_SUCCESS_TERM_CHAR);
+}
+
+TEST_F(VisaMessageBasedLoopbackTest, Flush_ClearsBuffer)
+{
+  write("Some data in the pipe\n");
+  set_bool_attribute(visa::VisaAttribute::VISA_ATTRIBUTE_TERMCHAR_EN, true);
```
dmondrik, line 284: This attribute is irrelevant for this test. I'll remove it in the next commit.

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -258,6 +270,284 @@ TEST_F(VisaDriverLoopbackTest, WriteWithTermChar_ReadMatches)
...
+  ViUInt64 buffer[testLength] = { 0x01234567DEADBEEF, 0x02340234, 0x00, 0x867543210ABCDEF, 0xC0FFEE };
+  for (int i = 0; i < testLength; ++i) {
+    out64(startingOffset + i * 8, buffer[i]);
+  }
+  auto readBuffer = moveIn64(startingOffset, testLength);
+  for (int i = 0; i < testLength; ++i) {
+    EXPECT_EQ(readBuffer[i], buffer[i]);
+  }
+}
+
```
danielhuani, line 550: Looks good!
Some ideas for test: Maybe add some tests to do Out64 to write a 64bit data while using In8/In16/In32 to read part of it to check the bit order is aligned with the methods.
Does it allow to read memory whose address is not aligned with byte size? It could be an error case test to read a 64bit data from address 0x1.

danielhuani (approved):

source/tests/system/visa_driver_api_tests.cpp
dmondrik: A few thoughts on why I don't plan to add those here:
- Checking alignment of reads and writes of different sizes is dependent on architecture-specific behavior, specifically big-endian vs. little-endian. Yes we know what architecture our tests run on, but it's not a good idea to depend on that.
- Querying on a misaligned address for a given size will return an error. That's why you see my 8-bit operations start at an odd address but not the others.
- I don't see the value of error cases for the register based operations at all, really. That's for testing the VISA driver. These tests are ensuring that we correctly pass the input parameters from the request packet to the VISA driver and copy the output parameters to the response packet. The error case testing is important in scenarios where our code has to do something special, like handle partial buffer content in the message based operations.

danielhuani: `These tests are ensuring that we correctly pass the input parameters from the request packet to the VISA driver and copy the output parameters to the response packet.` 
Good point!

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -258,6 +270,284 @@ TEST_F(VisaDriverLoopbackTest, WriteWithTermChar_ReadMatches)
...
+  const int testLength = 5;
+  const int startingOffset = 16;
+  ViUInt64 buffer[testLength] = { 0x01234567DEADBEEF, 0x02340234, 0x00, 0x867543210ABCDEF, 0xC0FFEE };
+  for (int i = 0; i < testLength; ++i) {
+    out64(startingOffset + i * 8, buffer[i]);
+  }
+  auto readBuffer = moveIn64(startingOffset, testLength);
+  for (int i = 0; i < testLength; ++i) {
+    EXPECT_EQ(readBuffer[i], buffer[i]);
+  }
```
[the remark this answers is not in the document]
danielhuani: Need to be `google::protobuf::RepeatedField<ViUInt64> buffer = { 0x01234567DEADBEEF, 0x02340234, 0x00, 0x867543210ABCDEF, 0xC0FFEE };`
The types needs to be aligned for `EXPECT_THAT`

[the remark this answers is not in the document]
dmondrik: It's more work to me than it's worth. I'm still getting too much compiler spewage.

reckenro (approved):

astarche (approved):

PR #1012, opened 2023-10-09
Add support for queued asynchronous read/write operations

dmondrik:

Add support for `ReadAsync` and `WriteAsync`, along with the ability to query and close the I/O completion event when done.

[User Story 2522258](https://dev.azure.com/ni/DevCentral/_workitems/edit/2522258): [grpc-device server] Implement custom functions for visa asynchronous I/O in grpc-device

Added duplicates for many existing read/write tests in `VisaMessageBasedLoopbackTest`.


source/codegen/metadata/visa/custom_header.mako
```diff
@@ -16,13 +16,15 @@ bytes_attrs = []
...
-    if ("MEM_BASE" in attr['name']
+    if ("_BASE" in attr['name']
+      or "_COUNT" in attr['name']
       or "MEM_SIZE" in attr['name']
-      or "WIN_BASE" in attr['name']
       or "WIN_SIZE" in attr['name']
     ):
         u32_attrs.append(f"{attr_name}_32")
         u64_attrs.append(f"{attr_name}_64")
+    elif "VI_ATTR_BUFFER" == attr['name']:
```
dmondrik, line 26: This is necessary to prevent it from being put in with 'string' attributes which have a defined 256-character limit.

source/custom/nixnet_converters.h
```diff
@@ -106,7 +106,7 @@ struct FrameHolder {
         throw std::invalid_argument("All FrameBufferRequest instances in repeated field should have same oneof set for the frame");
       }
 
-      u32 frame_size = get_frame_buffer_size(1, grpc_frame.enet().frame_data().length(),Protocol::PROTOCOL_ENET);
+      u32 frame_size = get_frame_buffer_size(1, static_cast<u32>(grpc_frame.enet().frame_data().length()), Protocol::PROTOCOL_ENET);
```
dmondrik, line 109: This just cleans up some build warnings I had noticed. Not directly related to this PR.

source/custom/visa_service.custom.cpp
```diff
@@ -34,9 +45,18 @@ class DriverErrorException : public std::runtime_error {
   }
 };
 
+static void ServerObjectCleanup(visa_grpc::VisaService::LibrarySharedPtr library)
+{
+  std::unique_lock<std::shared_mutex> lock(s_server_object_lock);
+  library->Close(s_defaultRM);
+  s_defaultRM = VI_NULL;
+}
```
dmondrik, line 53: I also need to clear `s_async_operations` from here. I'll do that on the next commit. While these operations wouldn't leak under normal circumstances, it could happen if the client crashes between starting and finishing an operation.

source/custom/visa_service.custom.cpp
```diff
@@ -131,6 +185,30 @@ ::grpc::Status VisaService::ConvertApiErrorStatusForViObject(::grpc::ServerConte
...
+  if (context->IsCancelled()) {
+    return ::grpc::Status::CANCELLED;
+  }
+  try {
+    ViEvent event_handle = request->event_handle();
+    auto status = library_->CloseEvent(event_handle);
+    if (!status_ok(status)) {
+      return ConvertApiErrorStatusForViSession(context, status, 0);
+    }
+    auto matching_event_lambda = [&](const VisaAsyncOperation& v) { return v.event == event_handle; };
```
astarche, line 201: Consider: could events be implemented as a grpc-device session? 2 possible benefits:

1. Allows for "client allocated IDs" via "named sessions". This enables replay APIs because you can have the same named event across sessions.
1. If it works, the session repository could be helpful reusable code.

I think xnet does something like this with the dbrefs.

I realize this is not a minor suggestion. And I think the way the VISA API works, there's a lot of low level back and forth (like the memory API) that wouldn't work for replay testing anyway. Just an idea; in case it's useful.

source/tests/system/visa_driver_api_tests.cpp
```diff
@@ -186,14 +189,54 @@ class VisaMessageBasedLoopbackTest : public VisaDriverApiTest {
     EXPECT_EQ(data.size(), response.return_count());
   }
 
+  void write_async(const std::string& data)
```
astarche, line 192: Would it make sense for all of this to be in a single method (possibly a streaming one?) for `grpc-device`? Is there anything similar in a LV API (or other higher-level language)?

Matching the low-level driver API (like this) is a good default for `grpc-device`. We can always add higher-level/async variants later. So, I don't have a problem with it as-is.

astarche (approved):

reckenro (approved):

danielhuani (approved):

source/custom/visa_service.custom.cpp
dmondrik: This was brought up earlier in the development and I declined for these reasons:
- Events happen when the driver detects them. 
- You cannot just access one when you want.
- They are extremely temporal / short-lived.

source/tests/system/visa_driver_api_tests.cpp
dmondrik: I agree this could make sense as a higher-level abstraction to add later. We need the 1:1 mapping for VISA to work properly in LabVIEW. I'm not going to mark this as an action item though, just pointing out that it's a reasonable feature someday.

source/custom/visa_service.custom.cpp
bkeryan: @astarche, here's the previous discussion thread: https://github.com/ni/grpc-device/pull/992#discussion_r1313515516

PR #1013, opened 2023-10-12
Add low-level register based tests for VISA.

dmondrik:

- Add peek/poke system tests for VISA.
- Modify the test setup to use `ASSERT` macros so the tests are skipped if the setup fails.
- Modify the gRPC error message to include optional text such as the VISA resource string.
  - In addition to being helpful in general, we really need it to get a better idea of how the CI is failing now.

Final testing for [User Story 2540593](https://dev.azure.com/ni/DevCentral/_workitems/edit/2540593): [grpc_device_server] implement the register based methods

Ran the system tests locally with `--gtest_filter=*Visa*`:
```
[==========] 41 tests from 4 test suites ran. (3279 ms total)
[  PASSED  ] 41 tests.
```

### DIFF
--- a/source/tests/system/visa_session_tests.cpp
+++ b/source/tests/system/visa_session_tests.cpp
@@ -88,20 +88,24 @@ TEST_F(VisaSessionTest, OpenSession_CloseSession_ClosesDriverSession)
 
 TEST_F(VisaSessionTest, OpenWithErrorFromDriver_ReturnsDriverErrorWithUserErrorMessage)
 {
-  EXPECT_THROW_DRIVER_ERROR_WITH_SUBSTR({
-    visa::OpenResponse init_response;
-    call_open(kVisaTestInvalidInstrumentDescriptor, visa::LOCK_STATE_NO_LOCK, "", 0, &init_response);
-  },
-                                        kInvalidRsrc, kVisaErrorInstrumentDescriptorNotFoundMessage);
+  EXPECT_THROW_DRIVER_ERROR_WITH_SUBSTR(
+    {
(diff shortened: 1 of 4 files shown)


danielhuani (approved):

source/tests/system/visa_session_tests.cpp
```diff
@@ -95,6 +95,15 @@ TEST_F(VisaSessionTest, OpenWithErrorFromDriver_ReturnsDriverErrorWithUserErrorM
                                         kInvalidRsrc, kVisaErrorInstrumentDescriptorNotFoundMessage);
 }
 
+TEST_F(VisaSessionTest, OpenWithErrorFromDriver_ReturnsDriverErrorWithResourceDescriptor)
+{
+  EXPECT_THROW_DRIVER_ERROR_WITH_SUBSTR({
```
reckenro, line 100: Nit: I think this formatting is correct with our clang-format file, but an alternative I personally think reads better would be
```cpp
EXPECT_THROW_DRIVER_ERROR_WITH_SUBSTR(
      {
        visa::OpenResponse init_response;
        call_open(kVisaTestInvalidInstrumentDescriptor, visa::LOCK_STATE_NO_LOCK, "", 0, &init_response);
      },
      kInvalidRsrc,
      kVisaTestInvalidInstrumentDescriptor);
```
Same in the test directly above but this is minor and preference so up to you.

reckenro (approved):

source/tests/system/visa_session_tests.cpp
dmondrik: Agreed.

PR #1015, opened 2023-10-16
Add PXI failure trigger index to trailing metadata

dmondrik:

- The failure index from `PxiReserveTriggers` should be returned as part of the trailing metadata. That means the method needs to be handled in custom code.
- Add `ni-failure-index` to the context in all cases.

On failure, the response object does not get returned.

Built locally.

### DIFF
--- a/source/codegen/metadata/visa/functions.py
+++ b/source/codegen/metadata/visa/functions.py
@@ -1367,7 +1367,6 @@ functions = {
         'status_expression': '0'
     },
     'PxiReserveTriggers': {
-        'codegen_method': 'CustomCode',
         'parameters': [
             {
                 'direction': 'in',
@@ -1401,6 +1400,7 @@ functions = {
                 'direction': 'out',
                 'include_in_proto': False,
                 'name': 'failureIndex',
+                'return_on_error_key': 'ni-failure-index',
                 'type': 'ViInt16'
             }
         ],
--- a/source/codegen/metadata/visa/functions.py
+++ b/source/codegen/metadata/visa/functions.py
@@ -1367,6 +1367,7 @@ functions = {
         'status_expression': '0'
     },
     'PxiReserveTriggers': {
+        'codegen_method': 'CustomCode',
         'parameters': [
             {
                 'direction': 'in',
@@ -1398,6 +1399,7 @@ functions = {
             },
             {
                 'direction': 'out',
+                'include_in_proto': False,
                 'name': 'failureIndex',
                 'type': 'ViInt16'
             }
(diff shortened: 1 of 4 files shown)


source/codegen/metadata/visa/functions.py
```diff
@@ -1398,6 +1399,7 @@
             },
             {
                 'direction': 'out',
+                'include_in_proto': False,
```
reckenro, line 1402: Optional: Seeing it in this context with the custom implementation reminded me of the tag `'return_on_error_key'`. What you have in the custom code looks good to me but alternatively you might be able to remove the `CustomCode` for this function and then have both `include_in_proto': False,` like you have here and also `'return_on_error_key': 'ni-failing-index'`.

It'd be interesting to see if that generates what you need. Though looking at the DAQmx usage of the key it seems to add the trailing metadata if there's an error or not instead of only when there's an error like in your custom implementation so that might be problematic: https://github.com/ni/grpc-device/blob/1e69024ec764188e78ff35d430618f918ffb0723/generated/nidaqmx/nidaqmx_service.cpp#L13070-13073

reckenro (approved):

source/custom/visa_service.custom.cpp
```diff
@@ -467,6 +469,56 @@ ::grpc::Status VisaService::ParseRsrc(::grpc::ServerContext* context, const Pars
...
+    std::vector<ViInt16> trig_lines;
+    std::transform(
+      trig_lines_request.begin(),
+      trig_lines_request.end(),
+      std::back_inserter(trig_lines),
+      [](auto x) { return (ViInt16)x; }); 
+    ViInt16 failure_index {};
+    auto status = library_->PxiReserveTriggers(vi, cnt, trig_buses.data(), trig_lines.data(), &failure_index);
+    if (!status_ok(status)) {
+      context->AddTrailingMetadata("ni-failing-index", std::to_string(failure_index));
```
astarche, line 511: The doc you called out in your email said:

> if this function returns an error code directly related to reserving triggers. [...]. For any other status code returned, the value of this output parameter is undefined and should not be used.

So it seems this code would have Undefined Behavior for any non-trigger-related errors. Is it practical to trap those specific errors? Or maybe by code inspection decide that it's not really UB? Also, if those non-trigger-error cases return `-1`, would you prefer to have `ni-failing-index` be unset in that case to be more consistent with success?

Related note: should `failure_index` be initialized to `-1`?

astarche (approved):

source/codegen/metadata/visa/functions.py
dmondrik: That works well. Thanks!

source/custom/visa_service.custom.cpp
dmondrik: It's not practical to look at specific error codes because vendor-specific error codes can always extend it. This is called so rarely that it's not a performance concern. Returning it via the metadata in all cases seems reasonable to me.

danielhuani (approved):

PR #1017, opened 2023-10-18
Removed the Initialize Api with channel parameter

alenkani:

we have decided to  remove the the API "RFmxInstr_InitializeWithChannel "
Channel information from separate parameter is not expected now. hence we need to remove the added Initialize API from both C and .NET
The RFSA/G Driver has decided not to expose this API, contrary to the previous plan of making it public. In light of this decision, RFmx will also refrain from exposing it.

https://dev.azure.com/ni/DevCentral/_workitems/edit/2548517/

Locally built grpc-device.


astarche:
Marking binary-breaking. Removing an API is binary breaking because existing compiled clients that use the API will break when targeting new servers. "source breaking" changes are something like "renaming fields" where existing compiled clients continue working (the binary format on the wire is the same) but will need to update when they re-run `protoc`.

generated/nirfmxinstr/nirfmxinstr.proto
```diff
@@ -74,7 +74,6 @@ service NiRFmxInstr {
   rpc GetSelfCalibrateLastTemperature(GetSelfCalibrateLastTemperatureRequest) returns (GetSelfCalibrateLastTemperatureResponse);
   rpc GetSignalConfigurationNames(GetSignalConfigurationNamesRequest) returns (GetSignalConfigurationNamesResponse);
   rpc Initialize(InitializeRequest) returns (InitializeResponse);
-  rpc InitializeWithChannel(InitializeWithChannelRequest) returns (InitializeWithChannelResponse);
```
astarche, line 77: Please provide a better description because we need to include it in the release notes. My understanding is that you think this is OK to remove because:

1. It was added very recently (2 months ago).
2. It is also being deleted from the C API (so there's nothing for grpc-device call and you've decided to break at the driver API level).
3. Even though it is binary breaking, the failure mode is clear and clean (failure to initialize).

I don't see a justification for *why* the method is being removed that I understand. You linked a task that said "Channel information from separate parameter is not expected now", which doesn't explain it to me. Please come up with a message that you would share with customers explaining why this API is removed.

astarche (changes requested):
Please update the description with a better message to clients explaining why the method is being removed.

generated/nirfmxinstr/nirfmxinstr.proto
alenkani: update the description

reckenro (approved):

PR #1019, opened 2023-10-27
Adding 6 python examples to rfsg, rfmxwlan, and rfmxnr

JonP-NI:
### What does this Pull Request accomplish?

This PR adds 6 new python examples to RFSG, RFmx WLAN, and RFmx NR.

### Why should this Pull Request be merged?

These examples help broaden the base of examples for RFSG, RFmx WLAN, and RFmx NR. The examples are modeled on the corresponding examples found in the shipped RFSG/RFmx examples in .NET and LabVIEW.

### What testing has been done?

These were all tested on a Ubuntu VM, communicating with real NI hardware (5830 VST) over a network.

PR #1027, opened 2023-12-12
Remove unnecessary folder and fix black validation.

reckenro:

After we approved a [recent PR](https://github.com/ni/grpc-device/pull/1019), an extra commit added this .idea folder. We don't want it in the repo.

Also, fixing up black version to avoid format validation issues.

Clean up repo and get build passing.

Note on [the first run](https://github.com/ni/grpc-device/actions/runs/7184972414/job/19567246822) that python validation fails after installing a new `black` alpha version [24.1a1](https://github.com/psf/black/releases/tag/24.1a1) that's intended to change formatting rules slightly for 2024.

CI


python_build_requirements.txt
```diff
@@ -2,7 +2,7 @@ contextlib2==21.6.0
 Mako==1.2.2
 MarkupSafe==2.0.1
 schema==0.7.4
-black>=21.10b0
+black==23.3.0
```
reckenro, line 5: This is to avoid getting an alpha version installed that includes proposed formatting changes for 2024. Setting it to the same version used in the [ni-python-styleguide](https://github.com/ni/python-styleguide/blob/cc47dab3ee0b56be594f956834c5e29d5124e4b9/poetry.lock#L3).

maxxboehme (approved):

PR #1043, opened 2024-02-20
Fix runtime environment test failing in some situations.

reckenro:

Fixes a unit test to be more reliable.

Brought up by @amstewart who saw the test failing on a generic debian 12 container.
Failed locally for me in the same way.

Test passes locally now. Build will verify it continues to pass there too.

### DIFF
--- a/source/tests/unit/library_set_runtime_environment_tests.cpp
+++ b/source/tests/unit/library_set_runtime_environment_tests.cpp
@@ -2,6 +2,7 @@
 // Use NI-SWITCH instead of NI-FAKE because nifake.h is missing a bunch of functions specified in metadata
 // and they're difficult to properly add in source.
 #include <niswitch/niswitch_library.h>
+
 #include "mock_shared_library.h"
 #include "version.h"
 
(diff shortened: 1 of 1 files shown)


source/tests/unit/library_set_runtime_environment_tests.cpp
```diff
@@ -20,15 +21,15 @@ TEST(LibrarySetRuntimeEnvironmentTests, Library_SetRuntimeEnvironmentFound_Calls
...
 
   const char* func_name = "niSwitch_SetRuntimeEnvironment";
 
   EXPECT_CALL(*p_shared_library, get_function_pointer(_))
-    .WillRepeatedly(Return(nullptr));
+      .WillRepeatedly(Return(nullptr));
   // We must return an address for SetRuntimeEnvironment or we won't call it.
-  EXPECT_CALL(*p_shared_library, get_function_pointer(func_name))
-    .WillOnce(Return(const_cast<const void*>(reinterpret_cast<void*>(&FunctionToPointAt))));
+  EXPECT_CALL(*p_shared_library, get_function_pointer(::testing::StrEq(func_name)))
```
reckenro, line 31: Note: This was the fix. Without it, locally and on the generic debian 12 container where someone else was running it, it wouldn't match here and we'd just get a nullptr for the "niSwitch_SetRuntimeEnvironment".

amstewart (approved):

astarche (approved):

PR #1045, opened 2024-03-05
CMakeLists: refactor to optionally use native libs

amstewart:
The current CMakeLists logic assumes that you want to build against the libraries distributed within the submodules, except in rare CROSSCOMPILING cases that are not fully functional.

Refactor the CMakeLists logic so that linking against native build libraries is a first-class feature.

* Adds options to configure the modular sections of the build.
  * `grpc_device_USE_LEGACY_NILRT_TOOLCHAIN` should be enabled when you want to use the NILRT cross-compiling toolchain from NI.com.
  * `grpc_device_USE_SUBMODULE_DEPS` should be enabled when you want to build and use the dependencies from the project submodules.
  * `grpc_device_USE_PYTHON_VENV` should be enabled when you want to build and use the optional python virtual environment distributed with this project.
  * The submodule_deps and python venv are enabled by default, to continue behavior from the previous CMakeLists logic, and to ease builds on windows machines - where the distro is less likely to provide those facilities.
  * All options should be disabled on native linux builds (including OE), where it is normally desirable for the built grpc-device binaries to match the remainder of the distribution.
* Fixes the library links and includes for the various test targets, when not using the submodule deps. This was previously broken.
* Enables testing (`enable_testing()`) for all build configurations, since the testing targets should now always build.
  * This may be a mistake for the legacy toolchain build case. I'll rework this change if it breaks the PR/CI.

This patchset includes necessary changes to enable building on NILRT, within OpenEmbedded.

* [x] Setup a debian 12 build environment with the necessary distribution dependencies to build this project.
* [x] Built the project with/without the submodule dependencies and verified that all targets work in both cases. The Unit, Integration, and Compilation tests all pass. The System tests do not fully pass - but that is due to configuration differences I expect between the debian build system and a proper grpc device.
* [x] With this patchset, NILRT 11's OE build environment can now compile the server binary. It *could* also compile the Test targets, but they are blocked by an incomplete grpc package in upstream OE.
* [ ] All PR stages pass with this change.


christag-ni:

christag-ni:

amstewart:
Closed. Functionality implemented by merged PR #1082.

PR #1047, opened 2024-03-13
Updated gRPC Device code for all 24C1 RFmx APIs

RahulBenaka:
gRPC Supported personalities - BT, LTE, NR, SpecAn and WLAN 

What does this Pull Request accomplish?
Updates niRFmxSpecAn in grpc-device-scrapigen to 24.3.0
Updates niRFmxLTE in grpc-device-scrapigen to 24.3.0
Updates niRFmxNR in grpc-device-scrapigen to 24.3.0
Updates niRFmxWLAN in grpc-device-scrapigen to 24.3.0
Updates the gRPC Scrapigen Device Code for SpecAn Marker APIs.
Updates the gRPC Scrapigen Device Code for LTE List Mode APIs, attributes and enums.
Updates the gRPC Scrapigen Device Code for new BT ModSpectrum API and attributes.

Why should this Pull Request be merged?
Updated to match the upcoming release of niRFmxSpecAn 24.3.0
Updated to match the upcoming release of niRFmxLTE 24.3.0
Updated to match the upcoming release of niRFmxNR 24.3.0
Updated to match the upcoming release of niRFmxWLAN 24.3.0

What testing has been done?
Manually inspected generated files.
Copied files from grpc-device-scrapigen/out/rfmxspecan/export/ to grpc-device/source/codegen/metadata/nirfmxspecan/ and built grpc-device successfully.
Manually inspected nirfmxspecan.proto file.
Copied files from grpc-device-scrapigen/out/rfmxlte/export/ to grpc-device/source/codegen/metadata/nirfmxlte/ and built grpc-device successfully.
Manually inspected nirfmxlte.proto file.
Copied files from grpc-device-scrapigen\out\rfmxbluetooth\export to grpc-device/source/codegen/metadata/ nirfmxbluetooth / and built grpc-device successfully.
Manually inspected nirfmxbluetooth.proto file.
Copied files from grpc-device-scrapigen\out\rfmxnr\export to grpc-device/source/codegen/metadata/ nirfmxnr/ and built grpc-device successfully.
Manually inspected nirfmxnr.proto file.
Copied files from grpc-device-scrapigen\out\rfmxwlan\export to grpc-device/source/codegen/metadata/ nirfmxwlan/ and built grpc-device successfully.
Manually inspected nirfmxwlan.proto file.
Manually inspected nirfmxbluetooth.proto file.

PR #1048, opened 2024-03-20
Avoid binary-breaking change in RFmx LTE

reckenro:

Avoids changing a field number in the RFmx LTE Proto from the previous release.

Reverses the binary breaking change introduced in #1047 

The build.

### DIFF
--- a/generated/nirfmxlte/nirfmxlte.proto
+++ b/generated/nirfmxlte/nirfmxlte.proto
@@ -2582,8 +2582,8 @@ message CfgPSSCHModulationTypeRequest {
   nidevice_grpc.Session instrument = 1;
   string selector_string = 2;
   oneof modulation_type_enum {
-    PsschModulationType modulation_type = 3;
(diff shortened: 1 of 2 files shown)


generated/nirfmxlte/nirfmxlte.proto
```diff
@@ -2582,8 +2582,8 @@ message CfgPSSCHModulationTypeRequest {
   nidevice_grpc.Session instrument = 1;
   string selector_string = 2;
   oneof modulation_type_enum {
-    PsschModulationType modulation_type = 3;
-    int32 modulation_type_raw = 4;
+    int32 modulation_type_raw = 3;
```
reckenro, line 2585: We want to keep the modulation_type_raw field number as `3` so old clients will still work passing forward their ints for field `3` and it won't be interpreted as an enum value.

astarche (approved):

maxxboehme (approved):

PR #1029, opened 2023-12-14
RFmxSpecAn Update to 24.0

alenkani:
### What does this Pull Request accomplish?
Updates niRFmxSpecAn in grpc-device to 24.0
Updated manually niRFmxSpecAn_exclude.h file to skip property specific Set/Get APIs as per the design.

### Why should this Pull Request be merged?
Updated to match the upcoming release of niRFmxSpecAn 24.0.0
There are Lot of apis rearranged because of Code clean Up in header file 

### What testing has been done?
Manually inspected generated files.
Copied files from grpc-device-scrapigen/out/rfmxSpecAn/export/ to grpc-device/source/codegen/metadata/nirfmxSpecAn/ and built grpc-device successfully.
Manually inspected nirfmxspecan.proto file.

PR #1049, opened 2024-03-21
Correcting enum values and maintaining backward compatability

RahulBenaka:
What does this Pull Request accomplish?
Avoids changing a field number in the RFmx LTE Proto from the previous release.
Avoids changing a field number in the RFmxSpecAn Proto from the previous release

Reverses the binary breaking change introduced in https://github.com/ni/grpc-device/pull/1047
Adding the special implementation in the functions_addon.py files of SpecAn and LTE to make sure correct code is generated in future.

What testing has been done?
The build.

### DIFF
--- a/generated/nirfmxspecan/nirfmxspecan.proto
+++ b/generated/nirfmxspecan/nirfmxspecan.proto
@@ -5350,12 +5350,12 @@ message IDPDCfgReferenceWaveformRequest {
   double dx = 4;
   repeated nidevice_grpc.NIComplexNumberF32 reference_waveform = 5;
   oneof idle_duration_present_enum {
-    IdpdReferenceWaveformIdleDurationPresent idle_duration_present = 6;
-    int32 idle_duration_present_raw = 7;
+    int32 idle_duration_present_raw = 6;
+    IdpdReferenceWaveformIdleDurationPresent idle_duration_present = 8;
   }
   oneof signal_type_enum {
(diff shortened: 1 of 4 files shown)


generated/nirfmxspecan/nirfmxspecan.proto
```diff
@@ -5350,12 +5350,12 @@ message IDPDCfgReferenceWaveformRequest {
   double dx = 4;
   repeated nidevice_grpc.NIComplexNumberF32 reference_waveform = 5;
   oneof idle_duration_present_enum {
-    IdpdReferenceWaveformIdleDurationPresent idle_duration_present = 6;
-    int32 idle_duration_present_raw = 7;
+    int32 idle_duration_present_raw = 6;
```
reckenro, line 5353: For clarity on these two, this is fixing the breaking change that we missed in https://github.com/ni/grpc-device/pull/1029/files#diff-698499203913617af06118a7f6c4ff2607d89dc18eeacdd7ae3ac538f09e264b

That change hadn't gone into the previous release so this is fixing it before it gets into this release.

Before (in the previous release), `idle_duration_present` was field 6 and `signal_type` was field 7. So this change is avoiding a binary breaking change getting into this release.

reckenro (approved):

astarche (approved):

maxxboehme (approved):

PR #1050, opened 2024-03-26
Update CMakeLists.txt to 2.6.0

christag-ni:

Bumps the version in CMake file

To reflect that the next grpc-device release will be 2.6.0

None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.12.0)
 
 project(ni_grpc_device_server
   LANGUAGES C CXX
-  VERSION 2.5.0)
+  VERSION 2.6.0)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(CreateVirtualEnvironment)


reckenro (changes requested):
We should bump [the version in AzDo](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/grpc_device/pipeline/templates/version.yml&version=GBmain&line=9&lineEnd=10&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) first, then we can bump here. Otherwise, if this goes into main as is it'll create another 2.5 adhoc export.

christag-ni:
> We should bump [the version in AzDo](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/grpc_device/pipeline/templates/version.yml&version=GBmain&line=9&lineEnd=10&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) first, then we can bump here. Otherwise, if this goes into main as is it'll create another 2.5 adhoc export.

Wouldn't doing it in this order mean that, for a small moment, ni-central will be pulling in a version of grpc-device that doesn't yet have builds?

astarche (approved):

reckenro:
> > We should bump [the version in AzDo](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/grpc_device/pipeline/templates/version.yml&version=GBmain&line=9&lineEnd=10&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) first, then we can bump here. Otherwise, if this goes into main as is it'll create another 2.5 adhoc export.
> 
> Wouldn't doing it in this order mean that, for a small moment, ni-central will be pulling in a version of grpc-device that doesn't yet have builds?

I don't think so, ni-central is only notified of builds from grpc-device. Changes on the ni-central side won't trigger a new adhoc build (see [here in the pipeline file](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/src/grpc_device/pipeline/pipeline-ci.yml&version=GBmain&line=11&lineEnd=12&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)).

christag-ni:
AzDO changes have been merged
https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/681880

PR #1055, opened 2024-05-06
Fix sync tests to what is expected

maxxboehme:

Some of the sync system tests have been failing for a while. Specifically the ones that check for errors. This is probably because they require real hardware and are not ran as part of the CI.

I updated the tests to work as expected with the current implementation. It appears that #692 did some changes where the error and message is reported through the grpc status instead of the response status. Updated the tests work with those changes.

Fixes sync tests so we can rely on their output to identify issues with any changes.

built and ran tests locally.

### DIFF
--- a/source/tests/system/nisync_driver_api_tests.cpp
+++ b/source/tests/system/nisync_driver_api_tests.cpp
@@ -1,4 +1,5 @@
 #include <gtest/gtest.h>
+#include <nlohmann/json.hpp>
 
 #include "device_server.h"
 #include "enumerate_devices.h"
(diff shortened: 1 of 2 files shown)


maxxboehme:
xnet test ConvertFramesToFromSignalsFromExample_CompareFrames_ValuesMatch fails on Ubuntu repeatedly. Does not appear to be caused by this change. @reckenro @astarche

reckenro (approved):

reckenro:
> xnet test ConvertFramesToFromSignalsFromExample_CompareFrames_ValuesMatch fails on Ubuntu repeatedly

Yeah, I can't see your change affecting this. Didn't see a Bug/Issue either so created one: https://github.com/ni/grpc-device/issues/1058

astarche (approved):

PR #1057, opened 2024-05-07
Updated gRPC Device code for BT, LTE and SpecAn 24C2 Cycle RFmx APIs

RahulBenaka:
Updates niRFmxSpecAn in grpc-device-scrapigen to 24.5.0
Updates niRFmxLTE in grpc-device-scrapigen to 24.5.0
Updates niRFmxBT in grpc-device-scrapigen to 24.5.0
Updates the gRPC Scrapigen Device Code for SpecAn APIs which are created for the specific feature request from the customer: Addition of Sequential FFT to Spectrum measurement with configurable overlap type property.
Updates the gRPC Scrapigen Device Code for LTE TXP measurement APIs, attributes and enums.
Updates the gRPC Scrapigen Device Code for new BT enum value for an attribute.

Updated to match the upcoming release of niRFmxSpecAn 24.5.0
Updated to match the upcoming release of niRFmxLTE 24.5.0
Updated to match the upcoming release of niRFmxBT 24.5.0

What testing has been done?
Manually inspected generated files.
Copied files from grpc-device-scrapigen/out/rfmxspecan/export/ to grpc-device/source/codegen/metadata/nirfmxspecan/ and built grpc-device successfully.
Manually inspected nirfmxspecan.proto file.
Copied files from grpc-device-scrapigen/out/rfmxlte/export/ to grpc-device/source/codegen/metadata/nirfmxlte/ and built grpc-device successfully.
Manually inspected nirfmxlte.proto file.
Copied files from grpc-device-scrapigen\out\rfmxbluetooth\export to grpc-device/source/codegen/metadata/ nirfmxbluetooth / and built grpc-device successfully.
Manually inspected nirfmxbluetooth.proto file.

### Breaking Changes
RFmx SpecAn: removing `timeout` field from `MarkerFetchFunctionValueRequest` to match Driver API. Technically breaking for grpc-device but this method was busted in grpc-device in the previous release because the C API never released with the timeout parameter, and those changes weren’t carried over to grpc-device.

### DIFF
--- a/generated/nirfmxbluetooth/nirfmxbluetooth.proto
+++ b/generated/nirfmxbluetooth/nirfmxbluetooth.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-RFMXBLUETOOTH API metadata version 24.3.0
+// This file is generated from NI-RFMXBLUETOOTH API metadata version 24.5.0
 //---------------------------------------------------------------------
 // Proto file for the NI-RFMXBLUETOOTH Metadata
 //---------------------------------------------------------------------
@@ -432,6 +432,7 @@ enum PacketType {
   PACKET_TYPE_3_EV5 = 15;
   PACKET_TYPE_LE = 16;
   PACKET_TYPE_LE_CS = 17;
+  PACKET_TYPE_LE_HDT = 18;
 }
 
 enum PayloadBitPattern {
@@ -562,6 +563,7 @@ enum NiRFmxBluetoothInt32AttributeValues {
   NIRFMXBLUETOOTH_INT32_PACKET_TYPE_3_EV5 = 15;
   NIRFMXBLUETOOTH_INT32_PACKET_TYPE_LE = 16;
   NIRFMXBLUETOOTH_INT32_PACKET_TYPE_LE_CS = 17;
+  NIRFMXBLUETOOTH_INT32_PACKET_TYPE_LE_HDT = 18;
   NIRFMXBLUETOOTH_INT32_PAYLOAD_BIT_PATTERN_STANDARD_DEFINED = 0;
   NIRFMXBLUETOOTH_INT32_PAYLOAD_BIT_PATTERN_11110000 = 1;
   NIRFMXBLUETOOTH_INT32_PAYLOAD_BIT_PATTERN_10101010 = 2;
--- a/generated/nirfmxlte/nirfmxlte.proto
+++ b/generated/nirfmxlte/nirfmxlte.proto
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-RFMXLTE API metadata version 24.3.0
+// This file is generated from NI-RFMXLTE API metadata version 24.5.0
 //---------------------------------------------------------------------
 // Proto file for the NI-RFMXLTE Metadata
 //---------------------------------------------------------------------
@@ -304,6 +304,10 @@ service NiRFmxLTE {
   rpc SlotPhaseFetchSamplePhaseErrorLinearFitTrace(SlotPhaseFetchSamplePhaseErrorLinearFitTraceRequest) returns (SlotPhaseFetchSamplePhaseErrorLinearFitTraceResponse);
   rpc SlotPowerCfgMeasurementInterval(SlotPowerCfgMeasurementIntervalRequest) returns (SlotPowerCfgMeasurementIntervalResponse);
   rpc SlotPowerFetchPowers(SlotPowerFetchPowersRequest) returns (SlotPowerFetchPowersResponse);
+  rpc TXPCfgAveraging(TXPCfgAveragingRequest) returns (TXPCfgAveragingResponse);
+  rpc TXPCfgMeasurementOffsetAndInterval(TXPCfgMeasurementOffsetAndIntervalRequest) returns (TXPCfgMeasurementOffsetAndIntervalResponse);
+  rpc TXPFetchMeasurement(TXPFetchMeasurementRequest) returns (TXPFetchMeasurementResponse);
+  rpc TXPFetchPowerTrace(TXPFetchPowerTraceRequest) returns (TXPFetchPowerTraceResponse);
   rpc WaitForAcquisitionComplete(WaitForAcquisitionCompleteRequest) returns (WaitForAcquisitionCompleteResponse);
   rpc WaitForMeasurementComplete(WaitForMeasurementCompleteRequest) returns (WaitForMeasurementCompleteResponse);
 }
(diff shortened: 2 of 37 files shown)


generated/nirfmxspecan/nirfmxspecan.proto
```diff
@@ -5350,12 +5368,12 @@ message IDPDCfgReferenceWaveformRequest {
   double dx = 4;
   repeated nidevice_grpc.NIComplexNumberF32 reference_waveform = 5;
   oneof idle_duration_present_enum {
-    int32 idle_duration_present_raw = 6;
-    IdpdReferenceWaveformIdleDurationPresent idle_duration_present = 8;
+    IdpdReferenceWaveformIdleDurationPresent idle_duration_present = 6;
```
ThangamV-NI, line 5371: Looks like these are also source breaking changes!!!

generated/nirfmxlte/nirfmxlte.proto
```diff
@@ -2582,8 +2603,8 @@ message CfgPSSCHModulationTypeRequest {
   nidevice_grpc.Session instrument = 1;
   string selector_string = 2;
   oneof modulation_type_enum {
-    int32 modulation_type_raw = 3;
-    PsschModulationType modulation_type = 4;
+    PsschModulationType modulation_type = 3;
```
ThangamV-NI, line 2606: Isn't this a source breaking change?

source/codegen/metadata/nirfmxlte/functions.py
```diff
@@ -2511,8 +2511,6 @@
             {
                 'direction': 'in',
                 'enum': 'PsschModulationType',
-                'grpc_field_number': '4',
```
astarche, line 2514: This is what is causing the source breaking changes. Maybe there was a hand-edit that was not applied to scrapigen?

reckenro: I think this the versions of these functions coming from scrapigen were intended to be overridden with the ones specified in `functions_addon.py` for both files in [this PR](https://github.com/ni/grpc-device/pull/1049/files).

I think the problem is the [`__init__.py` for these](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/nirfmxlte/__init__.py) don't use the `functions_addon.py` file. 

@RahulBenaka , I think the fix is to update the __init__.py for SpecAn and LTE to be like the [ni_scope one](https://github.com/ni/grpc-device/blob/main/source/codegen/metadata/niscope/__init__.py) for instance where it takes in the `functions_addon` file and updates the functions with `metadata['functions'].update(functions_override_metadata)`.

generated/nirfmxspecan/nirfmxspecan.proto
```diff
@@ -5889,7 +5907,6 @@ message MarkerCfgYLocationResponse {
 message MarkerFetchFunctionValueRequest {
   nidevice_grpc.Session instrument = 1;
   string selector_string = 2;
-  double timeout = 3;
```
reckenro, line 5892: This is breaking too, and not related to the other ones. I added [a comment in the scrapigen review ](https://github.com/ni/grpc-device-scrapigen/pull/230/files#r1592921949) about what is causing this but I'm not sure if it's intended or not on that end.

reckenro (changes requested):
Want to see another round after the breaking changes are addressed.

generated/nirfmxbluetooth/nirfmxbluetooth.proto
```diff
@@ -1,6 +1,6 @@
 
 //---------------------------------------------------------------------
-// This file is generated from NI-RFMXBLUETOOTH API metadata version 24.3.0
+// This file is generated from NI-RFMXBLUETOOTH API metadata version 24.5.0
```
reckenro, line 3: Make sure to update [the README](https://github.com/ni/grpc-device/blob/main/README.md#supported-ni-drivers) as well for all these version bumps

generated/nirfmxspecan/nirfmxspecan.proto
RahulBenaka: The issue what we are seeing now was fixed in the PR: #1049. This was a permanent fix as we had added the (hard-coded) desried function values in functions_addons.py of the respective personalities. 

But when I used scrapigen to generate the code this time, It is not able to pick up the data from the functions_addons.py file and instead giving the same issue which was reported previously. 

Was there any logical change done on the scrapigen tool? If it is not clear, then we might have to file a bug.

generated/nirfmxspecan/nirfmxspecan.proto
RahulBenaka: The issue what we are seeing now was fixed in the PR: #1049. This was a permanent fix as we had added the (hard-coded) desired function values in functions_addons.py of the respective personalities. 

But when I used scrapigen to generate the code this time, It is not able to pick up the data from the functions_addons.py file and instead giving the same issue which was reported previously. 

Was there any logical change done on the scrapigen tool? If it is not clear, then we might have to file a bug.
@reckenro

generated/nirfmxspecan/nirfmxspecan.proto
RahulBenaka: Same comment as above i.e. Even if I correct this source breaking change by adding appropriate code in functions_addons.py file, I won't be able to generate the desired code.

generated/nirfmxbluetooth/nirfmxbluetooth.proto
RahulBenaka: Updated the Readme.MD file

generated/nirfmxlte/nirfmxlte.proto
RahulBenaka: Yes. Same comment as above

generated/nirfmxspecan/nirfmxspecan.proto
reckenro: @ThangamV-NI , Rahul mentioned in the scrapigen changes that there was [this work item](https://dev.azure.com/ni/DevCentral/_workitems/edit/2679274/) for removing this parameter: 

So, is this just an expected breaking change for the API / proto file for SpecAn?

source/codegen/metadata/nirfmxlte/functions.py
RahulBenaka: This fix worked. I have updated the code for bluetooth and specan personality

generated/nirfmxspecan/nirfmxspecan.proto
ThangamV-NI: @reckenro Yes, It is.

RahulBenaka: @reckenro I have updated the PR with source-breaking and binary-breaking tags

reckenro: Alright, sounds good then. FYI @maxxboehme that we'll have a breaking change this release for SpecAn.

generated/nirfmxspecan/nirfmxspecan.proto
reckenro: Worked through these with the functions_addons.py and __init__.py changes.

reckenro (approved):

generated/nirfmxspecan/nirfmxspecan.proto
maxxboehme: I guess since for this API grpc-device has never worked having this change is fine.

maxxboehme (approved):

PR #1064, opened 2024-06-17
Stop running NiXnetLINDriverApiTests.ConvertFramesToFromSignalsFromEx…

maxxboehme:
Stop running NiXnetLINDriverApiTests.ConvertFramesToFromSignalsFromExample_CompareFrames_ValuesMatch on Ubuntu

Stop running NiXnetLINDriverApiTests.ConvertFramesToFromSignalsFromExample_CompareFrames_ValuesMatch on Ubuntu. Testing locally and internal NI ATS is not able to reproduce the issue seen when running the github actions. More debugging is needed on why this is happening there. So this does not because noise for PR reviews will will keep watching the test runs on the internal ATS while debugging more into why this is failing in the system used for running the system tests.

Stop the Ubuntu tests from reporting the same test failure that appears to only be failing with the configuration used in the actions pipeline.

Local testing and internal ATS testing. Discussed in internal BUG.

### DIFF
--- a/.github/workflows/run_ubuntu_system_tests.yml
+++ b/.github/workflows/run_ubuntu_system_tests.yml
@@ -26,5 +26,5 @@ jobs:
         tar -xvzf ni-grpc-device-tests-linux-glibc2_31-x64.tar.gz
 
     - name: Run System Tests
-      run: ./SystemTestsRunner --gtest_filter=-NiRFSA*:NiRFSG*:NiDCPower*CalSelfCalibrate*
+      run: ./SystemTestsRunner --gtest_filter=-NiRFSA*:NiRFSG*:NiDCPower*CalSelfCalibrate*:NiXnetLINDriverApiTests.ConvertFramesToFromSignalsFromExample_CompareFrames_ValuesMatch
       timeout-minutes: 27


.github/workflows/run_ubuntu_system_tests.yml
```diff
@@ -26,5 +26,5 @@ jobs:
         tar -xvzf ni-grpc-device-tests-linux-glibc2_31-x64.tar.gz
 
     - name: Run System Tests
-      run: ./SystemTestsRunner --gtest_filter=-NiRFSA*:NiRFSG*:NiDCPower*CalSelfCalibrate*
+      run: ./SystemTestsRunner --gtest_filter=-NiRFSA*:NiRFSG*:NiDCPower*CalSelfCalibrate*:NiXnetLINDriverApiTests.ConvertFramesToFromSignalsFromExample_CompareFrames_ValuesMatch
```
astarche, line 29: I'm happy with this solution. But why is this used instead of a conditional GTEST_SKIP ([example](https://github.com/ni/grpc-device/blob/a47a73ebfd58fa1f621a1c7207586efc359a92e9/source/tests/system/nidigital_driver_api_tests.cpp#L51))?

astarche (approved):

reckenro (approved):

maxxboehme:
Windows System tests fail with just an error code. This does not appear to be related to this change.

PR #1071, opened 2024-07-01
Build using Ninja on Windows and Linux Desktop

maxxboehme:

Update the build step to use the ninja build system for both Windows and Linux. Also simplifies the build steps and removes things that were not being used.

Also a side note is that on Windows using ninja provides better developer efficiency as it appears to track dependencies better and reduces the amount of things being rebuilt after a first full build unlike Visual Studio which I have experience rebuilding much of the project over repeatedly even though there were no changes to those files. This doesn't help with the build but might help with developers wanting to build grpc-device this way.

- Build times are reduced using ninja
  - Windows Average build times
    - Before Change: 1 hr 30 min
    - After Change: 50 min (about a 45% build time reduction)
  - Linux Average build times
    - Before Change:  1 hr
    - After Change: 50 min (about a 17% build time reduction)
- The steps have been simplified to just cmake command calls making it easier to understand. 
- Removed things like CMAKE_VERSION which didn't appear to be doing anything as the build before were using the version that was provided by the github provided runner images.

- PR Build
  - Windows still specifies using MSVC 19.40.33811.0 for the CC and CXX compilers
  - Linux still specifies using GNU 9.4.0 for the CC and CXX compilers

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -53,7 +53,7 @@ jobs:
       run: git submodule update --init --recursive --depth=1
 
     - name: Setup Ninja
-      uses: ashutoshvarma/setup-ninja@v1
+      uses: ashutoshvarma/setup-ninja@v1.1
       with:
           version: 1.12.1
 
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -63,7 +63,7 @@ jobs:
       env:
           CC: ${{ matrix.config.cc }}
           CXX: ${{ matrix.config.cxx }}
-      run: |
+      run: >
         cmake
         -S .
         -B build
@@ -71,7 +71,7 @@ jobs:
         ${{ matrix.config.cmake_generator }}
 
     - name: Build
-      run: |
+      run: >
         cmake
         --build build
         --config ${{ matrix.config.build_type }}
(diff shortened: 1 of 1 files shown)


reckenro (approved):

reckenro:
This one's good to go whenever you're ready since Alex is out.

.github/workflows/build_cmake.yml
```diff
@@ -50,44 +44,38 @@ jobs:
     - name: Update Submodules
       run: git submodule update --init --recursive --depth=1
 
+    - name: Setup Ninja
+      uses: ashutoshvarma/setup-ninja@v1.1
```
bkeryan, line 48: @maxxboehme This hasn't released since 2021. Does it install an up-to-date version of Ninja? It uses an outdated version of Node.js and outdated npm dependencies.

maxxboehme: There is a version [argument](https://github.com/ashutoshvarma/setup-ninja/blob/master/action.yml) that can be past to specify the ninja version. I will look into an alternative way to install ninja onto the build machines.

bkeryan: Oops, I didn't notice the `version:` field.

https://github.com/ni/grpc-device/pull/1072/ looks good.

PR #1079, opened 2024-08-08
Move to using GITHUB_OUTPUT from set-output

### LINKED ISSUE #908: Update actions to avoid set-output
We use set-output in a few of our actions: [build_nilrt.yml](https://github.com/ni/grpc-device/blob/main/.github/workflows/build_nilrt.yml), [create_release.yml](https://github.com/ni/grpc-device/blob/main/.github/workflows/create_release.yml), and [validate_linux_rt.yml](https://github.com/ni/grpc-device/blob/main/.github/workflows/validate_linux_rt.yml).

It is planned to be disabled on May 31, 2023: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

We should update them to use the GITHUB_OUTPUT environment file that they recommend as an alternative.

[AB#2359405](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2359405)

maxxboehme:
Fixes #908

set-output is deprecated and there are warnings being logged in our CI build specifying so. Here is a github blog article about it, [GitHub Actions: Deprecating save-state and set-output commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). I followed that article along with  this github doc [Workflow commands for GitHub Actions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#environment-files) on how to use the new environment file correctly. 

Move away from deprecated set-output to github's new suggested workflow using `$GITHUB_OUTPUT` environment file.

PR build

### DIFF
--- a/.github/workflows/build_nilrt.yml
+++ b/.github/workflows/build_nilrt.yml
@@ -35,7 +35,7 @@ jobs:
       id: submodules
       run: |
         git submodule update --init --recursive
-        echo "::set-output name=grpc-commit::$(git rev-parse --short @:./third_party/grpc)"
+        echo "grpc-commit=$(git rev-parse --short @:./third_party/grpc)" >> $GITHUB_OUTPUT
 
     - name: Cache Host OS gRPC Support
       uses: actions/cache@v4
--- a/.github/workflows/validate_linux_rt.yml
+++ b/.github/workflows/validate_linux_rt.yml
@@ -22,23 +22,23 @@ jobs:
     - name: Get the latest tag
       id: latesttag
       run: >-
-        echo "::set-output name=tag::$(
+        echo "tag=$(
           git tag --list --sort=-version:refname | grep '^\(v[0-9]\+.[0-9]\+.[0-9]\+\)$' | head -1
-        )"
+        )" >> $GITHUB_OUTPUT
 
     - name: Get latest tag without leading v
       id: latesttagwithoutv
       run: >-
-        echo "::set-output name=tag::$(
+        echo ":tag=$(
           git tag --list --sort=-version:refname | grep '^\(v[0-9]\+.[0-9]\+.[0-9]\+\)$' | head -1 | cut -c2-
-        )"
+        )" >> $GITHUB_OUTPUT
 
     - name: Get the second latest tag
       id: secondlatesttag
       run: >-
-        echo "::set-output name=tag::$(
+        echo "tag=$(
           git tag --list --sort=-version:refname | grep '^\(v[0-9]\+.[0-9]\+.[0-9]\+\)$' | sed '1d' | head -1
-        )"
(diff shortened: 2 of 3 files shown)


.github/workflows/build_nilrt.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       id: submodules
       run: |
         git submodule update --init --recursive
-        echo "::set-output name=grpc-commit::$(git rev-parse --short @:./third_party/grpc)"
```
astarche, line 38: I think @reckenro is more familiar with this code than I am. Are all of these validated by a PR build? What should we want for after submit to make sure everything is bubbling through?

astarche (approved):

.github/workflows/validate_linux_rt.yml
```diff
@@ -22,23 +22,23 @@ jobs:
...
+        echo "tag=$(
           git tag --list --sort=-version:refname | grep '^\(v[0-9]\+.[0-9]\+.[0-9]\+\)$' | head -1
-        )"
+        )" >> $GITHUB_OUTPUT
 
     - name: Get latest tag without leading v
       id: latesttagwithoutv
       run: >-
-        echo "::set-output name=tag::$(
+        echo ":tag=$(
```
reckenro, line 32: Left a leading `:` in front of the `tag=` here.

.github/workflows/build_nilrt.yml
reckenro: This one is validated in the PR build since it uses the grpc-commit as part of the key in the "Cache Host OS gRPC Support" step. And in this build it's still using the grpc-commit (0a82c02) correctly for the key:
```
Run actions/cache@v4
  with:
    path: third_party/grpc/build/
    key: ni/grpc-device-Linux-hostgrpccache-0a82c02
    enableCrossOsArchive: false
    fail-on-cache-miss: false
    lookup-only: false
    save-always: false
```

For the other two, one's used to create the release when a release branch is published and the validate_linux_rt one is run after the release itself is published. I think you could potentially run / trigger those on a fork of the repo but that might be tedious and more effort than it's worth. Otherwise I'd just be prepared to validate at the time of the next release.

reckenro (approved):

.github/workflows/validate_linux_rt.yml
maxxboehme: Fixed in latest commit

.github/workflows/build_nilrt.yml
maxxboehme: The changes are small and follow a simple pattern as the change that was tested in the PR. I don't believe it will be much to debug when releasing if there is an issue.

PR #1080, opened 2024-08-14
FPGA Get/Set APIs

amehra-ni:

This PR contains code generation for Get/Set FPGA functions as well as functions with enum as params.

- Update config.py files so the new data type can map correctly.
- Updated metadata.

- No compile or codegen issues.
- Manually verified server handlers for all APIs and compared them with [documentation for the APIs mentioned](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=%2Fsrc%2Frio%2Friodriverd%2Ffpgainterfacecapi%2Fsource%2Fcvi%2Fbin%2FNiFpga_fp.h&version=GBmain&_a=contents).
- Manually tested few of the APIs.


astarche (approved):

reckenro (approved):

source/codegen/metadata/nifpga/config.py
```diff
@@ -27,7 +27,8 @@
         'uint16_t': 'uint32',
         'uint32_t': 'uint32',
         'uint64_t': 'uint64',  
-        'NiFpga_Bool': 'bool'     
+        'NiFpga_Bool': 'bool',
+        'NiFpga_FifoProperty': 'uint32',
```
doshirohan, line 31: Should grpc_type be `FifoProperty` enum here?

doshirohan (approved):

source/codegen/metadata/nifpga/config.py
amehra-ni: I believe this should be uint32 only because, in the proto, it will be represented as shown below. We need one parameter to be of the _FifoProperty_ type and the other to be _raw_. If we set _FifoProperty_ here, the _raw_ parameter would also be of the _FifoProperty_ type, which I think is incorrect.

oneof property_enum {
    FifoProperty property = 3;
    uint32 property_raw = 4;
  }

PR #1089, opened 2024-09-08
Adding LTE  and (correcting) SpecAn 24C3 APIs

RahulBenaka:

Updated the APIs added as part of niRFmxSpecAn 24C3 APIs in grpc-device-scrapigen
Updates the gRPC Scrapigen Device Code for hiding SpecAn DVD DPR attributes.
Updates niRFmxLTE in grpc-device-scrapigen to 24.8.0
Updates the gRPC Scrapigen Device Code for LTE Evm APIs.

Updated to match the upcoming release of niRFmxLTE 24.8.0

Manually inspected generated files.
Copied files from grpc-device-scrapigen/out/rfmxspecan/export/ to grpc-device/source/codegen/metadata/nirfmxspecan/ and built grpc-device successfully.
Manually inspected nirfmxspecan.proto file.
Copied files from grpc-device-scrapigen/out/rfmxlte/export/ to grpc-device/source/codegen/metadata/nirfmxlte/ and built grpc-device successfully.
Manually inspected nirfmxlte.proto file.


generated/nirfmxspecan/nirfmxspecan.proto
```diff
@@ -986,9 +986,7 @@ enum NiRFmxSpecAnAttribute {
   NIRFMXSPECAN_ATTRIBUTE_DPD_PRE_DPD_CARRIER_BANDWIDTH = 1110116;
   NIRFMXSPECAN_ATTRIBUTE_DPD_IQ_ORIGIN_OFFSET_CORRECTION_ENABLED = 1110117;
   NIRFMXSPECAN_ATTRIBUTE_DPD_DVR_NUMBER_OF_SEGMENTS = 1110119;
-  NIRFMXSPECAN_ATTRIBUTE_DPD_DVR_LINEAR_MEMORY_DEPTH = 1110120;
```
maxxboehme, line 989: @reckenro @amstewart Correct if wrong but I believe the removing of enums is fine as clients building before would still just pass the value to our sever and we would just pass along. The issue would be if the meaning of the enum changed in a future update.

amstewart: @astarche

RahulBenaka:
Can you please approve my PR? I need to submit 2 more PR's which are dependent on this submission.

@maxxboehme @reckenro @astarche @amstewart

generated/nirfmxspecan/nirfmxspecan.proto
astarche: Yeah, removing enum values is "source breaking" so any existing clients that use the enum value will be broken when they uptake new protos. Usually when enums like these are removed, they were not working in the first place. So the runtime behavior will change from one type of error to another, which is usually OK. OR we are intentionally breaking  behavior to remove a feature.

If you ever use the enum value to mean something else ... yeah that would be more of a concern for sure. Ideally these are reserved somehow in your metadata.

I've tagged this as source breaking. Can you add to the description a line that some enum values are removed, why they are removed, and what the change in behavior would be (i.e., "new error"). If there is any chance that someone was doing something useful with the values, add a binary-breaking tag as well.

astarche (approved):

reckenro (approved):

PR #1101, opened 2024-10-08
Cherrypick Adding LTE List APIs (#1100)

maxxboehme:

This cherry-picks #1100 into the releases/2.7 branch to create a 2.7.1 release version of grpc-device

This change was too late to be added to the 2.7 release of grpc-device. Doing an off cycle release so that clients can use the changes easier. 

PR system tests


maxxboehme:
After discussion it is probably easier to just create a 2.8 release.

PR #1104, opened 2024-10-09
Use native concurrency functionality and separate CI and PR actions

maxxboehme:

Migrate to using the native [concurrency ](https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/)functionality provided by github actions. This migration was suggested by the [styfle/cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) we were using in their README.

Also created a separate ci.yml and pr.yml for easier sorting when looking at the actions tab. This also mimics what other projects like [nidaqmx-python](https://github.com/ni/nidaqmx-python/tree/master/.github/workflows) and [measurement-plugin-labview](https://github.com/ni/measurement-plugin-labview/tree/main/.github/workflows) follow. This also makes it clear on the rules that for PR builds if another action is kicked off it would cancel the previous while the same is not true for CI builds. 

- Clear division of CI and PR builds
- Migrate to using native github action functionality for canceling multiple PR jobs running at once for the same PR.

PR build


reckenro (approved):
Looks good to me. When you're ready to submit you'll need to change main branch's protection policies to switch over the required checks to the new ones.

astarche (approved):

maxxboehme:
Updated required checks for both "main" and "releases" branches.

PR #1114, opened 2024-10-28
Add initial auto-generation support for streaming APIs

amehra-ni:
This PR adds initial support for auto-generation of streaming APIs. This might not be complete, but we will be iteratively covering cases in follow-ups.

- **Metadata Update**: Enhanced the metadata for FPGA functions to include additional information regarding streaming. This updated metadata is utilized in the **metadata_mutation.py** to create streaming functions and integrate them into the original function list. list.
- **Condition Check for Streaming API**: Added condition checks in service.cpp.mako and other relevant files to generate streaming APIs. These checks invoke helper Mako files to generate the streaming function body.
**Struct Generation**: First, generate the necessary structs for the streaming API.
**Moniker Functions**: Next, generate the Moniker functions required for streaming.
**Function Generation**: Finally, generate the streaming functions themselves.
- **Service Helpers**: Added logic in **service_helpers.mako** to auto-generate the streaming function bodies. The generated functions are designed to handle various parameter and streaming types, similar to the example provided [here](https://github.com/vvacharya/grpc-device/blob/users/vacharya/nifpga-stream-prototype/source/custom/nifpga_service.custom.cpp)
- **Proto Mako**: Added support for including additional proto files in the proto.mako file.
- In a follow-up PR, plan to extract common helper methods that can be used by both the new and existing code. This iterative approach will enhance the auto-generation process to handle additional cases more effectively.

Both grpc-server and sideband server runs fine. I'm able to use the client code and call the server with all the new streaming APIs and talk with the VI.


source/codegen/metadata/nifpga/config.py
```diff
@@ -13,6 +13,7 @@
     'code_readiness': 'Release',
     'driver_name': 'NI-FPGA',
     'status_ok': 'status >= 0',
+    'additional_protos': ['data_moniker.proto'],
```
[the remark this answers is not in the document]
doshirohan: If not, can this and 3 other configurations you have added below be replaced with single config `supports_streaming`? I am not sure if for below 3 configurations if there will be scenario where drivers supports streaming APIs but not through moniker, arenas et al.

source/codegen/common_helpers.py
```diff
@@ -1194,3 +1202,17 @@ def get_params_needing_initialization(parameters: List[dict]) -> List[dict]:
     * Outputs that are calculated/populated after the API call.
     """
     return [p for p in parameters if not (is_return_value(p) or is_get_last_error_output_param(p))]
+
+
+def filter_data_moniker_functions(functions):
```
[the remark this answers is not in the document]
amehra-ni: Updated

source/codegen/common_helpers.py
```diff
@@ -1194,3 +1202,17 @@ def get_params_needing_initialization(parameters: List[dict]) -> List[dict]:
...
+    """Return function metadata only for functions that use the data moniker service."""
+    return [
+        name for name, function in functions.items() if function.get("data_moniker_support", False)
+    ]
+
+
+def get_data_moniker_function_name(function_name, function_data):
+    """Return the corresponding moniker function name for the given C API function."""
+    if function_data.get("moniker_cname", None):
```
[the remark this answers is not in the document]
amehra-ni: This is not needed. It was needed before when I had some other changes in _functions.py_ but I have removed those, hence these are not needed anymore and removed it.

source/codegen/metadata_mutation.py
```diff
@@ -139,6 +151,37 @@ def mutate(metadata: dict):
...
+            "direction": "out",
+            "grpc_type": "ni.data_monikers.Moniker",
+            "name": "moniker",
+            "type": "ni::data_monikers::Moniker",
+        }
+    )
+
+    # Construct the streaming API metadata
+    streaming_metadata = {
+        "data_moniker_support": True,
```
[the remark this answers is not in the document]
amehra-ni: No, both of them would be True, replace it with **is_streaming_api**

source/codegen/metadata_mutation.py
```diff
@@ -117,15 +118,26 @@ def mutate(metadata: dict):
     if config["driver_name"] in _ALREADY_MUTATED:
         return
 
+    # Generate new streaming APIs for functions that support it.
+    all_functions = {}
+
+    for function_name, function in metadata["functions"].items():
+        all_functions[function_name] = function
+
+        if function.get("moniker_supports_streaming") is True:
```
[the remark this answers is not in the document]
amehra-ni: Updated

source/codegen/metadata/nifpga/config.py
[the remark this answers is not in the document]
amehra-ni: Currently, we have three separate configurations that may or may not be needed independently. While it's possible that these configurations could be consolidated into a single configuration at the moment, I'm not entirely sure if there might be future use cases that would require these configurations to be separated.

source/codegen/templates/register_all_services.cpp.mako
```diff
@@ -38,6 +42,9 @@ std::shared_ptr<std::vector<std::shared_ptr<void>>> register_all_services(
   auto session_repository = std::make_shared<nidevice_grpc::SessionRepository>();
   service_vector->push_back(session_repository);
   nidevice_grpc::register_core_services(service_vector, server_builder, session_repository, feature_toggles);
+  auto moniker_service = std::make_shared<ni::data_monikers::DataMonikerService>();
```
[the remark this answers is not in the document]
amehra-ni: Moved it behind the feature toggle.

source/codegen/metadata/nifpga/config.py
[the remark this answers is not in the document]
amehra-ni: As discussed offline, there is no specific usage of other tags, I have merged them into one now. We can update in the future if any requirement comes up.

source/codegen/templates/service.cpp.mako
```diff
@@ -13,6 +13,7 @@ custom_types = common_helpers.get_custom_types(config)
 has_async_functions = any(service_helpers.get_async_functions(functions))
 has_two_dimension_functions = any(service_helpers.get_functions_with_two_dimension_param(functions))
 function_names = service_helpers.filter_proto_rpc_functions_to_generate(functions)
```
doshirohan, line 15: Rename `function_names` to `functions_to_generate` to indicate what this list really contains. After that, other two places below where you are calling `filter_proto_rpc_functions_to_generate`, you can just iterate over `functions_to_generate`. One of these is admittedly existing usage, but seems unnecessary.

source/codegen/templates/service.cpp.mako
```diff
@@ -13,6 +13,7 @@ custom_types = common_helpers.get_custom_types(config)
...
+streaming_functions = common_helpers.filter_streaming_functions(functions)
```
doshirohan, line 16: Rename this to `streaming_functions_to_generate`. Then pass `functions_to_generate` and `functions` to `filter_streaming_functions` API so that it returns only those functions which need to be generated.

Places where you check `% if streaming_functions` below will return true even if all functions in `streaming_functions` are marked false for generation and you will end up generating code when there are no streaming functions generated in cpp file.

source/codegen/templates/service.cpp.mako
```diff
@@ -110,6 +125,9 @@ namespace ${config["namespace_component"]}_grpc {
     response_type = service_helpers.get_response_type(method_name)
     is_async_streaming = common_helpers.has_async_streaming_response(function_data)
 %>\
+% if function_name in streaming_functions:
```
doshirohan, line 128: This really should be an `elif function_name in streaming_functions_to_generate` before the last `else` which calls `define_simple_method_body` and then change your helper from `define_streaming_api` to `define_streaming_api_body`. Most of your implementation in `define_streaming_api` after you remove `struct` and moniker methods from there is same as non-async function definition in `else` block starting at line 122/140 below.

source/codegen/templates/service.h.mako
```diff
@@ -59,6 +60,20 @@ struct ${service_class_prefix}FeatureToggles
   bool ${service_helpers.get_toggle_member_name(toggle)};
 % endfor
 };
+% if config.get("use_moniker_service", False):
+void RegisterMonikers();
```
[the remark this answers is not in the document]
amehra-ni: Yes, it was a typo. I have updated it to _RegisterMonikerEndpoints_

source/codegen/templates/service.h.mako
```diff
@@ -59,6 +60,20 @@ struct ${service_class_prefix}FeatureToggles
   bool ${service_helpers.get_toggle_member_name(toggle)};
 % endfor
 };
+% if config.get("use_moniker_service", False):
```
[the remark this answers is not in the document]
amehra-ni: Updated

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,198 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+    catch (std::exception& ex) {
+        return ::grpc::Status(::grpc::UNKNOWN, ex.what());
+    }
+}
+<%
+  arg_string= service_helpers.create_args(parameters)
+  arg_string = arg_string.replace(", &moniker", "").strip()
+  data_type = streaming_type.replace("[]", "")
+%>\
+::grpc::Status ${moniker_function_name}(void* data, google::protobuf::Arena& arena, google::protobuf::Any& packedData)
```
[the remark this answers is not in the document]
amehra-ni: Moved them after RegisterMonikerEndPoints

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,198 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+  struct_name = f"Moniker{function_name.replace('Begin', '')}Data" if function_name.startswith('Begin') else f"Moniker{function_name}Data"
+  moniker_function_name = service_helpers.create_moniker_function_name(function_name)
+  c_api_name = service_helpers.get_c_api_name(function_name)
+  streaming_param = common_helpers.get_streaming_parameter(parameters)
+  streaming_type = streaming_param['type']
+  coerced_type, is_coerced_type_present = service_helpers.get_coerced_type_and_presence(streaming_type)
+  grpc_streaming_type = streaming_param['grpc_streaming_type']
+  is_array = common_helpers.is_array(streaming_type)
+%>\
+struct ${struct_name}
```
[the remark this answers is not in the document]
amehra-ni: Moved them at the desired location mentioned.

source/codegen/templates/service.cpp.mako
amehra-ni: Correct. Updated

source/codegen/templates/service.cpp.mako
amehra-ni: Updated

source/codegen/templates/service.cpp.mako
amehra-ni: Updated.

source/codegen/templates/service.h.mako
```diff
@@ -17,6 +17,7 @@ custom_types = common_helpers.get_custom_types(config)
 (input_custom_types, output_custom_types) = common_helpers.get_input_and_output_custom_types(config, functions)
 resource_repository_deps = service_helpers.get_driver_shared_resource_repository_ptr_deps(config, functions)
 resource_handle_types = service_helpers.get_resource_handle_types(config)
+data_moniker_functions = common_helpers.filter_streaming_functions(functions)
```
doshirohan, line 20: Similar to `service.cpp.mako`, this also needs to generate entry for only those functions which are not marked for no-codegen. This does have equivalent logic to get functions to generate on line 82/98 below. Move it here `functions_to_generate = common_helpers.filter_proto_rpc_functions(functions)` and then pass that to `filter_streaming_functions` API so that you generate only those moniker APIs that need to be generated.

source/codegen/templates/service.h.mako
```diff
@@ -59,6 +60,21 @@ struct ${service_class_prefix}FeatureToggles
...
 };
+% if data_moniker_functions:
+
+void RegisterMonikerEndpoints();
+
+% for function in data_moniker_functions:
+<%
+  method_name = common_helpers.get_data_moniker_function_name(function)
+%>\
+% if config.get("has_streaming_api", False):
```
doshirohan, line 71: You don't need this `if` block here. Only generate the signature with `arena`.

source/codegen/templates/service.cpp.mako
```diff
@@ -100,6 +111,20 @@ namespace ${config["namespace_component"]}_grpc {
   }
 
 % endif
+% if streaming_functions_to_generate:
+void RegisterMonikerEndpoints()
```
doshirohan, line 115: This needs to be indented one more level (same as `CopyBytesToEnums` function above). And the method body will then indent one more level.

doshirohan: Same for `define_moniker_function_body` as well.

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,205 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+    ${struct_name}* function_data = (${struct_name}*)data;
+    auto library = function_data->library;
+    ${initialize_non_array_parameters(input_params, streaming_param)}\
+    % if streaming_param and streaming_param['direction'] == 'out':
+        ${handle_out_direction(c_api_name, arg_string, data_type, streaming_type, is_coerced_type_present, streaming_param)}\
+    % elif streaming_param and streaming_param['direction'] == 'in':
+        ${handle_in_direction(c_api_name, arg_string, data_type, grpc_streaming_type, streaming_type, coerced_type, is_coerced_type_present, is_array)}\
+    % endif
+    if (status < 0) {
+      std::cout << "${moniker_function_name} error: " << status << std::endl;
```
doshirohan, line 327: Even in case of error you return `Status::OK` below. Replace this `cout` with `ConvertApiErrorStatusForNiFpga_Session` call similar to how it is being done for other FPGA APIs. Probably calling `populate_error_check` should do that for you.

doshirohan: Actually, I think you need to call `populate_response`. The "response" object has `status` field which is supposed to contain `status` returned from driver call. Right now following highlighted status is never returned to caller.

image

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,205 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+  streaming_param = common_helpers.get_streaming_parameter(parameters)
+%>\
+${initialize_streaming_input_param(function_name, input_params, parameters, streaming_param)}
+      ${struct_name}* data = new ${struct_name}();
+      ${initialize_begin_input_param(input_params, streaming_param)}\
+      data->library = std::shared_ptr<${service_class_prefix}LibraryInterface>(library_);
+      ${initialize_service_output_params(output_params)}\
+      ni::data_monikers::Moniker* moniker = new ni::data_monikers::Moniker();
+      ni::data_monikers::DataMonikerService::RegisterMonikerInstance("${moniker_function_name}", data, *moniker);
+      response->set_allocated_moniker(moniker);
```
doshirohan, line 352: Also `response->set_status(0)` to indicate success?

source/codegen/metadata/nifpga/config.py
```diff
@@ -30,6 +31,7 @@
         'NiFpga_Bool': 'bool',
         'NiFpga_FifoProperty': 'uint32',
     },
+    'has_streaming_api': True,
```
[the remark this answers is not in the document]
doshirohan: or maybe `has_moniker_streaming_apis` if that sound more verbose... but is essential to differentiate this from grpc streaming APIs.

source/codegen/templates/service.h.mako
amehra-ni: updated

source/codegen/templates/service_helpers.mako
amehra-ni: This is a Moniker function, hence there is no response object available in this.
Also, for ConvertApiErrorStatusForNiFpga_Session we need a context object as well but it's not available in the Moniker function.

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,205 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+</%def>
+
+<%def name="define_moniker_streaming_structs(function_name, function_data, parameters)">\
+<%
+  config = data['config']
+  input_params = [p for p in function_data['parameters'] if common_helpers.is_input_parameter(p)]
+  streaming_param = common_helpers.get_streaming_parameter(function_data['parameters'])
+  service_class_prefix = config["service_class_prefix"]
+  grpc_streaming_type = streaming_param['grpc_streaming_type']
+  struct_name = f"Moniker{function_name.replace('Begin', '')}Data" if function_name.startswith('Begin') else f"Moniker{function_name}Data"
```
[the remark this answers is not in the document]
amehra-ni: updated

source/codegen/templates/service.cpp.mako
```diff
@@ -99,6 +104,16 @@ namespace ${config["namespace_component"]}_grpc {
     }
   }
 
+% endif
+% if streaming_functions:
+void RegisterMoniker()
+{
+% for function_name in service_helpers.filter_proto_rpc_functions_to_generate(functions):
+% if function_name in streaming_functions:
```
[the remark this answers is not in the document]
amehra-ni: resolved

maxxboehme (changes requested):
>Both grpc and sideband servers runs fine.

Can you add more details about the testing done since no automated testing was added in the PR.

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,227 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+  service_class_prefix = config["service_class_prefix"]
+  request_param = service_helpers.get_request_param(function_name)
+  response_param = service_helpers.get_response_param(function_name)
+  struct_name = common_helpers.get_data_moniker_struct_name(function_name)
+  moniker_function_name = common_helpers.get_data_moniker_function_name(function_name)
+  streaming_param = common_helpers.get_streaming_parameter(parameters)
+  streaming_param_to_include = common_helpers.get_input_streaming_param(parameters)
+%>\
+${initialize_streaming_input_param(function_name, streaming_param_to_include, parameters, streaming_param)}
+      ${struct_name}* data = new ${struct_name}();
```
astarche, line 344: More raw `new`s here. Can some of these be allocated on the stack? Or can they be a `std::unique` pointers?

There is one case where we may "need" raw pointers: if we are passing the data to asynchronous code that is not in our control. The data needs to be kept around after this function returns (cannot be a stack variable) and the other code may not support managing data with `unique_ptr`s (or other smart pointers).

 But we need to be clear when that is happening and what the cleanup strategy is.

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,227 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+        std::back_inserter(array),
+        [](auto x) {
+          if (x < std::numeric_limits<${data_type}>::min() || x > std::numeric_limits<${data_type}>::max()) {
+            std::string message("value " + std::to_string(x) + " doesn't fit in datatype ${data_type}");
+            throw nidevice_grpc::ValueOutOfRangeException(message);
+          }
+          return static_cast<${data_type}>(x);
+        });
+% else:
+    auto data_array = ${grpc_streaming_type.lower()}_message.value();
```
astarche, line 460: Consider if any of your `auto`s should be `const auto x&` or some other qualifiers to avoid copy on assignment and ensure that it is not modified. Several questions to ask:

1. Is the data simple and pointer size or smaller? If so: use just `auto`, a copy is fine.
2. Are you doing a `move` or `initialization` where you want the data to be owned by the stack variable? just `auto` is fine.
3. Do you need a copy of the data to modify it? Or snapshot it at this moment in time? Just `auto`.
4. Do you need read-only access larger-than-pointer-size data? Use `const auto x&`?
5. Do you need to modify the data in place? Consider `auto x&` or use a pointer (smart pointer?) to modify the data.

generated/nifpga/nifpga_service.cpp
```diff
@@ -41,6 +306,817 @@ namespace nifpga_grpc {
...
+        std::transform(
+            array,
+            array + size,
+            function_data->data.mutable_value()->begin(),
+            [&](auto x) {
+                return x;
+            });
+        packedData.PackFrom(function_data->data);
+    }
+
```
maxxboehme, line 394: I see a new but never see a delete in this function. Seems like a memory leak. To add to @astarche points. We should use a std::vector or std::unique_ptr. Until there is a reason shouldn't ever see a call to new being assigned to a raw pointer.

generated/nifpga/nifpga_service.cpp
```diff
@@ -2852,6 +4828,34 @@ namespace nifpga_grpc {
...
+      MonikerWriteU8Data* data = new MonikerWriteU8Data();
+            data->session = session;
+      data->control = control;
+      data->library = std::shared_ptr<NiFpgaLibraryInterface>(library_);
+      
+      ni::data_monikers::Moniker* moniker = new ni::data_monikers::Moniker();
+      ni::data_monikers::DataMonikerService::RegisterMonikerInstance("MonikerWriteU8", data, *moniker);
+      response->set_allocated_moniker(moniker);
+      response->set_status(0);
```
maxxboehme, line 4851: Is ownership of this moniker pointer being transferred to RegisterMonikerInstance? If so using a std::unique_ptr and then releasing it to the RegisterMonikerInstance would help show that else it again seems like a memory leak.

maxxboehme (changes requested):

source/codegen/templates/service_helpers.mako
amehra-ni: As for these things you mentioned I don't need the data in the auto to modify data in place or any other reason.
I just need the data for the moment and the data is pretty simple. I'd stay with `auto`.

Thanks for detailed explanation.

source/codegen/templates/service_helpers.mako
amehra-ni: Updated it to use unique_ptr.

generated/nifpga/nifpga_service.cpp
amehra-ni: I have cleared up the memory by using delete. I have mentioned reasoning in Alex comments above.

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,227 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
+
+<%def name="streaming_handle_out_direction_array(c_api_name, arg_string, data_type, streaming_param)">\
+<%
+   is_coerced = service_helpers.is_output_array_that_needs_coercion(streaming_param)
+   size = service_helpers.get_size_param_name(streaming_param)
+%>\
+% if is_coerced:
+std::vector<${data_type}> array(${size});
+% else:
+${data_type}* array = new ${data_type}[${size}];
```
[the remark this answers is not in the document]
amehra-ni: I have updated to clean up the memory using delete.

We would need to change how create_args works if I need to use vector here as I would then need to pass the pointer of the vector which create_args doesn't do right now(It only creates them for coercion type). I feel we don't have enough information while creating the arg for streaming such that I can twist the logic.(Maybe I need to add more info in metadata?) I'm currently looking into it.

generated/nifpga/nifpga_service.cpp
amehra-ni: I have transferred the ownership of the moniker to the set_allocated_moniker, which now takes the responsibility and ensure its memory management.

generated/nifpga/nifpga_service.cpp
```diff
@@ -41,6 +306,821 @@ namespace nifpga_grpc {
...
+        std::transform(
+            array,
+            array + size,
+            function_data->data.mutable_value()->begin(),
+            [&](auto x) {
+                return x;
+            });
+        packedData.PackFrom(function_data->data);
+    }
+    delete[] array;
```
maxxboehme, line 394: Would one of the following changes not work for array in cases like this?
std::vector
```C++
std::vector<int32_t> array(size);
auto status = library->ReadArrayI32(session, indicator, array.data(), size);
```
std::unique_ptr
```C++
auto array = std::make_unique<int32_t[]>(size);
auto status = library->ReadArrayI32(session, indicator, array, size);
```

source/codegen/templates/service_helpers.mako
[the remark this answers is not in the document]
astarche: OK, I'm not sure I fully understand how this fits together. But I appreciate that you're working through it. Most likely anywhere using a `new` array should be replaced with a `std::vector` and use `array.data()` to get a pointer to the underlying array data when needed (i.e., to pass to a driver). So I don't think that *needs* to be a special condition, but maybe existing code would need to change to use `std::vector` consistently. But there could be other reasons to need more info in metadata.

generated/nifpga/nifpga_service.cpp
amehra-ni: Updated, resolving the comment.

source/codegen/templates/service_helpers.mako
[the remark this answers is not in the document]
amehra-ni: Updated as discussed offline. Using vector now.

source/server/data_moniker_service.cpp
```diff
@@ -168,7 +168,6 @@ Status DataMonikerService::BeginSidebandStream(ServerContext* context, const Beg
 
   char identifier[32] = {};
   InitOwnerSidebandData(strategy, bufferSize, identifier);
-  
   response->set_strategy(request->strategy());
   response->set_sideband_identifier(identifier);
   response->set_connection_url(GetConnectionAddress(strategy));
```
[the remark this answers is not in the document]
amehra-ni: I have created a task and tagged you under the task in which we will be fixing this.

generated/nifpga/nifpga_service.cpp
```diff
@@ -2017,6 +3661,34 @@ namespace nifpga_grpc {
...
+    }
+    try {
+      auto session_grpc_session = request->session();
+      NiFpga_Session session = session_repository_->access_session(session_grpc_session.name());
+      uint32_t control = request->control();
+
+      auto data = std::make_unique<MonikerWriteArrayI8Data>();
+            data->session = session;
+      data->control = control;
+      data->library = std::shared_ptr<NiFpgaLibraryInterface>(library_);
```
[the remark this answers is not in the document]
amehra-ni: Fixed the indentation at various places.

source/codegen/templates/register_all_services.cpp.mako
```diff
@@ -60,9 +67,20 @@ std::shared_ptr<std::vector<std::shared_ptr<void>>> register_all_services(
...
 </%block>\
+% endfor
+% for driver in drivers:
+<%
+  config = driver["config"]
+  namespace = f"{config['namespace_component']}_grpc"
+%>\
+% if config.get("use_moniker_service", False):
+<%block filter="common_helpers.os_conditional_compile_block(config)">\
+  ${namespace}::RegisterMonikers();
```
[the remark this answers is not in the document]
amehra-ni: Updated

maxxboehme (approved):

source/codegen/templates/register_all_services.cpp.mako
```diff
@@ -24,6 +24,9 @@ repository_type_to_config = service_helpers.list_session_repository_handle_types
 %>\
 <%block filter="common_helpers.os_conditional_compile_block(config)">\
 #include "${module_name}/${module_name}_service_registrar.h"
+% if config.get("has_moniker_streaming_apis", False):
```
doshirohan, line 27: You don't need this now that you are not calling `nifpga_grpc::RegisterMonikerEndpoints();` from `resgister_all_services.cpp`. And then restore/remove blank line at the end of the file so that this file doesn't appear in the review.

reckenro (approved):

source/codegen/templates/register_all_services.cpp.mako
amehra-ni: Updated

source/codegen/templates/service_helpers.mako
amehra-ni: I'll talk to you offline and will raise follow-up based on the discussion.

source/codegen/templates/service_helpers.mako
```diff
@@ -271,6 +271,198 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
       return ::grpc::Status::OK;\
 </%def>
 
+<%def name="register_moniker_functions(function_name)">
+<%  moniker_function_name = service_helpers.create_moniker_function_name(function_name)
+%>  ::ni::data_monikers::DataMonikerService::RegisterMonikerEndpoint("${moniker_function_name}", ${moniker_function_name});
```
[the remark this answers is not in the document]
amehra-ni: will raise follow-up for this and one other comment after having the discussion with you.

PR #1116, opened 2024-11-06
Fix Accept addrlen initialization

jonathanou:

- Properly initialize the address length variable being passed to the `nxaccept` call.
- Fix mirrored to https://github.com/ni/grpc-device-scrapigen/pull/252

Similar to #1107 , the address is also not being populated in the `Accept` call.

Manually tested that by setting `addrlen` to `sizeof(addr.storage)` the grpc call is now getting the client address.

### DIFF
--- a/generated/nixnetsocket/nixnetsocket_service.cpp
+++ b/generated/nixnetsocket/nixnetsocket_service.cpp
@@ -58,7 +58,7 @@ namespace nixnetsocket_grpc {
       auto initialization_behavior = request->initialization_behavior();
 
       auto addr = allocate_output_storage<nxsockaddr, SockAddr>();
-      nxsocklen_t addrlen {};
+      auto addrlen = static_cast<nxsocklen_t>(sizeof(addr.storage));
       bool new_session_initialized {};
       auto init_lambda = [&] () {
         auto socket_out = library_->Accept(socket, &addr, &addrlen);
--- a/source/codegen/metadata/nixnetsocket/functions.py
+++ b/source/codegen/metadata/nixnetsocket/functions.py
@@ -21,7 +21,8 @@ functions = {
                 'direction': 'out',
                 'name': 'addrlen',
                 'include_in_proto': False,
-                'type': 'nxsocklen_t'
+                'type': 'nxsocklen_t',
+                'hardcoded_value': 'static_cast<nxsocklen_t>(sizeof(addr.storage))'
             },
             {
                 'direction': 'out',


generated/nixnetsocket/nixnetsocket_service.cpp
```diff
@@ -58,7 +58,7 @@ namespace nixnetsocket_grpc {
       auto initialization_behavior = request->initialization_behavior();
 
       auto addr = allocate_output_storage<nxsockaddr, SockAddr>();
-      nxsocklen_t addrlen {};
+      auto addrlen = static_cast<nxsocklen_t>(sizeof(addr.storage));
```
maxxboehme, line 61: Is this something that can't be easily tested using an integration or system test to make sure it wouldn't be broken in the future.

jonathanou: This requires xnet hardware, so I'm guessing it's not easy?  If there is an existing test class that can test with hardware, please let me know.

astarche: The hardware tests I think are the "loopback" tests in this class: https://github.com/ni/grpc-device/blob/main/source/tests/system/nixnetsocket_driver_api_tests.cpp

I don't remember it too clearly but it looks like it requires some "discovered_device" that provides the "required_interfaces" ENET1-4: https://github.com/ni/grpc-device/blob/bd1ba154bf5a29fe0975e196b5a902389eb1d645/source/tests/system/nixnetsocket_driver_api_tests.cpp#L202

Definitely we could discuss testing gaps or options for automation. I'm not sure who an "active" driver owner would be for xnet socket.

astarche (approved):

generated/nixnetsocket/nixnetsocket_service.cpp
jonathanou: @maxxboehme let me know how we should proceed here.  Who is the xnet driver owner that we can find to add an auto test (also need the a test for #1107 too) ?  Can we track this through an issue while this fix goes in first?

maxxboehme: For this size of change I wouldn't expect setting up a test suite with hardware. Please add more details about the setup you used when doing your manual testing like all the hardware used with model numbers along with software versions and that would be fine to close this conversation.

maxxboehme (approved):

reckenro (approved):

generated/nixnetsocket/nixnetsocket_service.cpp
jonathanou: Manual test was done using PXIe-8523/8623, running LV 2023 RT image with fix applied to grpc-device 2.6.

PR #1118, opened 2024-11-13
Add streaming API for remaining FPGA functions.

amehra-ni:

This PR adds support for the remaining FPGA APIs.

- Added necessary message in custom_proto.
- Updated functions.py list with the leftover functions.
- Updated to **${streaming_type} value {}**; (as done for normal functions )such that it fits all datatype initialization.
- Initialize vector properly as when we call array_storage.data(), it returns a pointer to the underlying array, but since the vector is empty, this pointer is not valid for writing. Hence, initialize it properly now with the size.
- Updated the ownership transfer of the pointer using release() now for the **data** variable.

Streaming functions get called correctly, and the data is read/written correctly. This is one example where I read/write data in an indicator in the loop. (I have tested this for normal i32 type, but it will still work for other types as implementation is same internally) This is for scaler.
image
This is for array.
image


source/codegen/templates/service_helpers.mako
```diff
@@ -341,12 +341,12 @@ ${populate_response(function_data=function_data, parameters=parameters)}\
...
 %>\
 ${initialize_streaming_input_param(function_name, streaming_param_to_include, parameters, streaming_param)}
-      auto data = std::make_unique<${struct_name}>();\
+      ${struct_name}* data = new ${struct_name}();\
       ${initialize_begin_input_param(streaming_param_to_include, streaming_param)}\
       data->library = std::shared_ptr<${service_class_prefix}LibraryInterface>(library_);
       ${initialize_service_output_params(output_params)}\
       auto moniker = std::make_unique<ni::data_monikers::Moniker>();
-      ni::data_monikers::DataMonikerService::RegisterMonikerInstance("${moniker_function_name}", data.get(), *moniker);
+      ni::data_monikers::DataMonikerService::RegisterMonikerInstance("${moniker_function_name}", data, *moniker);
```
maxxboehme, line 349: If any of those other calls between calling new and what I assume is the RegisterMonikerInstance taking some ownership of the pointer throws an exception. It isn't caught here at least but is higher in the stack we then have a memory leak.

Continue to use a unique_ptr but to show ownership transfer of the pointer use release() instead of get().

maxxboehme (changes requested):

source/codegen/templates/service_helpers.mako
amehra-ni: Yes, tested it out. this works. Resolving this comment.

astarche (approved):

reckenro (approved):

PR #1122, opened 2024-11-18
Moniker streaming functions return standard reponse message instead of custom data type

doshirohan:
- Values returned from streaming APIs are now in `<function-name>StreamingResponse` message.
- Most of the custom codegen written for Moniker APIs was needed because output was not in standard "response" message. With this change, I was able to delete most of the custom codegen written for streaming functionality.

Moniker functions used for streaming data to/from clients used to return custom data types defined in `custom_proto.mako` file for DAQ and FPGA. This has multiple limitations.
- The custom types had only one field `value` for streaming. Read APIs with multiple out parameters need multiple values in response.
- Response was not provided in standard `<function-name>Response` message that other grpc APIs use.

- Existing integration tests for FPGA streaming APIs pass.
- Executed sample client tests with StreamRead/StreamWrite and validated results.

### DIFF
--- a/source/codegen/common_helpers.py
+++ b/source/codegen/common_helpers.py
@@ -1204,24 +1204,24 @@ def get_params_needing_initialization(parameters: List[dict]) -> List[dict]:
     return [p for p in parameters if not (is_return_value(p) or is_get_last_error_output_param(p))]
 
 
-def filter_moniker_streaming_functions(functions, functions_to_generate):
+def filter_moniker_streaming_functions(functions: dict, functions_to_generate: List[str]) -> List[str]:
     """Return streaming functions that need to be generated."""
     return [
         name for name in functions_to_generate if is_moniker_streaming_function(functions[name])
     ]
 
 
-def is_moniker_streaming_function(function):
+def is_moniker_streaming_function(function: dict) -> bool:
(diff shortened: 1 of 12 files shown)


Raghav-NI (approved):

source/codegen/common_helpers.py
```diff
@@ -1207,18 +1207,58 @@ def get_params_needing_initialization(parameters: List[dict]) -> List[dict]:
...
+
 def get_data_moniker_function_name(function_name):
     """Return the corresponding moniker function name for the given C API function."""
     return function_name.replace("Begin", "Moniker")
 
 
-def get_data_moniker_struct_name(function_name):
-    """Return the corresponding moniker function name for the given C API function."""
-    return f"{function_name.replace('Begin', 'Moniker')}Data"
+def get_data_moniker_struct_name(begin_function_name):
```
reckenro, line 1224: Can you go through the new/updated methods and add type hints for the inputs and output? It helps see at a glance what's expected of the methods.

reckenro (approved):

source/codegen/common_helpers.py
doshirohan: Added typehints.

PR #1126, opened 2024-11-21
Make sideband port configurable via json file.

amehra-ni:
There were some unresolved comments in the [previous PR](https://github.com/ni/grpc-device/pull/1113). Two of them are being addressed in this PR.

- The sideband port should not be hardcoded and instead should be configurable via service_cofig.json file.
- Exposed a new member sideband_port and used that for initialization.
- Created out a helper which extracts port. Used it at two places for extracting port.

- No functional behavior change for grpc-server.
- We can now configure the sideband port.

### DIFF
--- a/source/server/server_configuration_parser.cpp
+++ b/source/server/server_configuration_parser.cpp
@@ -17,6 +17,7 @@ namespace nidevice_grpc {
 static const char* kDefaultFilename = "server_config.json";
 static const char* kAddressJsonKey = "address";
 static const char* kPortJsonKey = "port";
+static const char* kSidebandAddressJsonKey = "sideband_address";
 static const char* kSidebandPortJsonKey = "sideband_port";
 static const char* kServerCertJsonKey = "server_cert";
 static const char* kServerKeyJsonKey = "server_key";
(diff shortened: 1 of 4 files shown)


asumit (approved):

source/server/server_configuration_parser.cpp
```diff
@@ -141,6 +142,21 @@ int ServerConfigurationParser::parse_max_message_size() const
...
 
+int ServerConfigurationParser::parse_sideband_port() const
+{
+  auto sideband_port = config_file_.find(kSidebandPortJsonKey);
+  if (sideband_port != config_file_.end()) {
+    try {
+      return sideband_port->get<int>();
+    }
+    catch (const nlohmann::json::type_error& ex) {
+      throw InvalidMaxMessageSizeException(ex.what());
```
reckenro, line 153: This should probably throw a `WrongPortTypeException(ex.what());` instead.

Also, consider checking the port range like `parse_port` on line 271 below, and if it's not in a valid range throwing an `InvalidPortException`.

reckenro (approved):

source/server/server_configuration_parser.cpp
amehra-ni: I've created a helper which does the port extraction.
I'm now using that at two places to get the port.

PR #1129, opened 2024-11-25
Add codegen support for Daqmx Read apis which have multiple out params.

Raghav-NI:

This PR adds codegen support for remaining Daqmx Read apis which have multiple out params. The generation logic for multiple out params in response has been done as part of [Rohan's PR](https://github.com/ni/grpc-device/pull/1122) 

- Generated streaming counterpart for all the remaining Daqmx read apis needed for streaming.

I am able to build and ran server without any errors. We will test the apis with the integration test we are working on.


source/codegen/metadata/nidaqmx/functions.py
```diff
@@ -17612,9 +17612,10 @@
             {
                 'ctypes_data_type': 'ctypes.c_int',
                 'direction': 'out',
+                'grpc_streaming_type': 'I32Data',
```
reckenro, line 17615: Should these changes be coming from [scrapigen](https://github.com/ni/grpc-device-scrapigen)? I didn't see these changes on that side.

Raghav-NI: Yes, this metadata should be coming from scrapigen, the PR on scrapigen with this change is not published yet as we are making changes for some more apis. It will be out tomorrow, for now I have copied the metadata from there itself.

reckenro (approved):

PR #1134, opened 2024-11-29
Add test for coerced type (read/write API for FPGA)

amehra-ni:
The int32 and int64 implementations are similar as neither of them are coerced. Therefore, I have updated the _int32_ test to an _int16_ test to ensure we can test both behaviors.

- For the read API, I updated the read type to **int16_t** to match what is read from the real API.
- For the write API, I updated type for data written to **pb::int16**, even though proto expects i32, as that's how user applications will have data they want to write.

- All tests pass successfully.


christag-ni:

doshirohan (approved):

reckenro (approved):

christag-ni:

PR #1136, opened 2024-12-03
Make CPU affinity used for CPU pinning configurable via json file

Raghav-NI:

We want to configure cpu affinity used for cpu pinning for different process configurable via json file.

1. User-Configurable CPU affinity:

    -  Users can explicitly assign processor for:

          -   Sideband stream read/write thread
          -   Stream write threads

    - Why?

        - Some systems have more CPUs available, while others have fewer CPUs with shared workloads.
        - Configurability allows users to optimize CPU usage based on their system constraints or dedicated tasks.
        - It gives users control to explicitly decide how resources are distributed rather than relying on the scheduler.

2. Default Behavior: "No Affinity" (-1):

     - By default, threads are not pinned to any specific CPU, allowing the OS scheduler to decide the CPU assignment 
     dynamically.

3. Separate CPU Cores for Critical Threads:

    - Users can assign separate CPU cores for non- sideband threads, sideband-read/write, and server threads.

    - Why?
    
        -   To isolate performance-critical sideband operations (e.g., low-latency tasks) from server threads to minimize contention.
        -   Separating cores ensures better performance for threads that are sensitive to cache invalidation, refreshes, or CPU contention.

- Best Practices for configuration:

    - Use Default (-1): For general systems where CPU pinning is unnecessary, default "no affinity" provides broad compatibility.
    - Assign Dedicated Cores: On systems with more CPUs, pin performance-critical threads (e.g., sideband tasks) to isolated cores.
    - Avoid Contention: Separate sideband and server threads onto different cores to reduce interference, particularly for real-time or low-latency workloads.

I am able to build grpc-server and run it successfully with and without cpu pinning configuration in json file.


source/config/localhost_config.json
```diff
@@ -7,5 +7,10 @@
        "server_cert": "",
        "server_key": "",
        "root_cert": ""
+    },
+    "core_configuration": {
```
asumit, line 11: the default values should be 0

source/config/localhost_config.json
```diff
@@ -7,5 +7,10 @@
...
+       "sideband_read_write_core": 8,
```
asumit, line 12: I think CpuCoreConfiguration or StreamingCoreConfiguration would be better

Raghav-NI: renamed to `StreamingCoreConfiguration`

source/config/localhost_config.json
Raghav-NI: core indexing starts from 0 so we can't set default value as 0, hence setting it to -1 which indicates that any available core can be used and cpu pinning is not needed.

source/server/server_configuration_parser.cpp
```diff
@@ -287,6 +292,47 @@ int ServerConfigurationParser::parse_port_with_key(const std::string& key) const
...
+        cpu_affinity.sideband_read_write = parse_cpu_affinity_with_key(kSidebandReadWriteKey);
+        cpu_affinity.stream_write = parse_cpu_affinity_with_key(kStreamWriteKey);
+        cpu_affinity.server = parse_cpu_affinity_with_key(kServerKey);
+    }
+    else{
+      // -1 is set as the default value for the cores which indicates that any available core can be used.
+      cpu_affinity.sideband_read_write = -1;
+      cpu_affinity.stream_write = -1;
+      cpu_affinity.server = -1;
+    }
```
maxxboehme, line 310: I feel like the struct should be set to the default values on creation. That way it is in a valid state even if one of the parse function throws an exception.

maxxboehme (changes requested):

source/server/server_configuration_parser.cpp
Raghav-NI: That makes sense, set it to default value in while creating in `cpu_affinity_configuration.h` and removed the else block highlighted above.

source/server/core_server.cpp
```diff
@@ -272,10 +276,12 @@ int main(int argc, char** argv)
     schedParam.sched_priority = 95;
     sched_setscheduler(0, SCHED_FIFO, &schedParam);
 
-    cpu_set_t cpuSet;
-    CPU_ZERO(&cpuSet);
-    CPU_SET(6, &cpuSet);
-    sched_setaffinity(0, sizeof(cpu_set_t), &cpuSet);
+    if (config.cpu_affinity.server >= 0) {
```
astarche, line 279: The `grpc-device` grpc server is multi-threaded, right? Should we be setting the affinity of grpc worker threads?

source/server/data_moniker_service.cpp
```diff
@@ -240,10 +248,12 @@ Status DataMonikerService::StreamRead(ServerContext* context, const MonikerList*
 Status DataMonikerService::StreamWrite(ServerContext* context, ServerReaderWriter<StreamWriteResponse, MonikerWriteRequest>* stream)
 {
 #ifndef _WIN32
-  cpu_set_t cpuSet;
-  CPU_ZERO(&cpuSet);
-  CPU_SET(1, &cpuSet);
-  sched_setaffinity(0, sizeof(cpu_set_t), &cpuSet);
+  if(c_CpuAffinityConfig.stream_write >= 0) {
+    cpu_set_t cpuSet;
```
astarche, line 252: Nit: this code is duplicated a few times. Consider a helper method or something like that so it's easier to tweak.

source/config/localhost_config.json
```diff
@@ -7,5 +7,10 @@
        "server_cert": "",
        "server_key": "",
        "root_cert": ""
+    },
+    "cpu_affinity_configuration": {
```
astarche, line 11: The PR doesn't describe "why" we are doing this very well.

The pre-existing code already assigns the sideband loop to a single CPU so that it can be isolated and cache invalidations/refreshes are minimized in performance critical code.

 There are a few different features here:

1. User can decide *which* CPU to assign. Why? Because they have more/fewer CPUs? They have more dedicated tasks? We want them to think about it?
2. The default changes to "no affinity"
3. User can assign the CPU of the server and the non-sideband read/write threads. Why? For performance of those threads? Or to keep them away from the core that the sideband thread uses?

source/config/localhost_config.json
```diff
@@ -7,5 +7,10 @@
...
+       "sideband_read_write": -1,
+       "stream_write": -1,
```
astarche, line 13: This wording might be confusing because we will likely use grpc streaming in other contexts. We could have a normal "streaming read" or "streaming write" API that just calls the driver in a loop rather than using the special "moniker" system.

Also, isn't the stream-based implementation also read/write?

I think the key term here is `moniker`. Monikers are a special system for streaming that has a specific dedicated processing thread. It's less important for `sideband` because `sideband` is a moniker-specific concept. So this could be `moniker_stream_processor`, `moniker_stream`, `moniker_stream_read_write`.

source/server/core_server.cpp
astarche: OVERALL I don't think this feature is correct in how it handles the "server". If most users don't need/use the "server" affinity feature, maybe it is OK.

The reason someone would use "server" affinity, as I understand it, is to make sure that the server *DOES NOT* use the CPUS that are assigned to the moniker processing loop threads. The "server" should not be single-CPU, so it doesn't make sense to assign it a single CPU affinity.

I think what *WOULD* make sense: IF sideband and/or moniker-stream affinity are set, THEN set the server to use all CPUS EXCEPT the ones that are assigned to the processing threads. That could actually be the logic for `-1` for the processing threads as well (IF sideband affinity is set AND stream affinity is NOT SET, THEN stream is assigned to all other CPUs).

astarche (changes requested):
Please respond to questions about `server` affinity and the high level question about clarifying the purpose/priorities for this change. Also consider that this PR is the first-pass documentation for HOW to use the feature, so include details about best practices for config.

source/server/core_server.cpp
asumit: @astarche I think what you are saying makes sense. If we set the cpu affinity of the server main thread to a specific thread, it just makes the server startup pinned on a specific thread. Subsequent server calls( whatever is served from 31763) from the client would anyways be  in specific threads. So this affinity would not do much except probably make server startup time better.

Raghav-NI:
> We are planning to create a release branch on the 18th, and your PR will need to be completed before then for the changes to be in effect. Do you want these changes included in this release?

Yes, I was on leave, this PR should be completed today (17th Dec).

Raghav-NI:
> Please respond to questions about `server` affinity and the high level question about clarifying the purpose/priorities for this change. Also consider that this PR is the first-pass documentation for HOW to use the feature, so include details about best practices for config.

Updated PR description.

source/config/localhost_config.json
Raghav-NI: Renamed as suggested. 
kept the first one as `moniker_sideband_stream_read_write` to separate it from the newly added configuration for `moniker_stream_read_write`.

source/config/localhost_config.json
Raghav-NI: Updated the PR description.

source/server/data_moniker_service.cpp
Raghav-NI: Created a helper as suggested.

source/server/core_server.cpp
Raghav-NI: Removed the configuration for server.

source/server/data_moniker_service.cpp
```diff
@@ -45,6 +46,13 @@ static void SysFsWrite(const std::string& fileName, const std::string& value)
 
 namespace ni::data_monikers {
 
+static MonikerStreamProcessor c_StreamProcessor;
+
+void configure_moniker_stream_processor(const MonikerStreamProcessor& stream_processor)
+{
+  c_StreamProcessor = stream_processor;
```
asumit, line 53: change it to s_StreamProcessor

source/server/data_moniker_service.cpp
```diff
@@ -103,20 +111,31 @@ void DataMonikerService::InitiateMonikerList(const MonikerList& monikers, Endpoi
   }
 }
 
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+#ifndef _WIN32
+void set_moniker_stream_processor(int stream_processor)
+{
+  if(stream_processor >= 0) {
```
asumit, line 119: nit: space after if

asumit (changes requested):

source/server/data_moniker_service.cpp
```diff
@@ -103,20 +111,31 @@ void DataMonikerService::InitiateMonikerList(const MonikerList& monikers, Endpoi
...
+  if ((strategy == ::SidebandStrategy::RDMA_LOW_LATENCY ||
+      strategy == ::SidebandStrategy::SOCKETS_LOW_LATENCY) && c_StreamProcessor.moniker_sideband_stream_read_write >= 0) {
     pid_t threadId = syscall(SYS_gettid);
     ::SysFsWrite("/dev/cgroup/cpuset/LabVIEW_tl_set/tasks", std::to_string(threadId));
 
-    cpu_set_t cpuSet;
-    CPU_ZERO(&cpuSet);
-    CPU_SET(8, &cpuSet);
-    sched_setaffinity(threadId, sizeof(cpu_set_t), &cpuSet);
+    set_moniker_stream_processor(c_StreamProcessor.moniker_sideband_stream_read_write);
```
asumit, line 138: This code should set the cpu affinity on a specific thread id. the setter function is setting it on thread 0 which I think is for the whole process.

source/server/data_moniker_service.cpp
```diff
@@ -103,20 +111,31 @@ void DataMonikerService::InitiateMonikerList(const MonikerList& monikers, Endpoi
   }
 }
 
+//---------------------------------------------------------------------
+//---------------------------------------------------------------------
+#ifndef _WIN32
+void set_moniker_stream_processor(int stream_processor)
```
asumit, line 117: In the context of this method we are just setting the cpu affinity, irrespective of who would be calling be it moniker or any other code. We could name is set_cpu_affinity and take in the thread_id as well

asumit (changes requested):

source/server/data_moniker_service.cpp
Raghav-NI: As discussed in chat, passing thread id as 0 to `sched_setaffinity `means it will set the affinity for the calling thread, which is what we need and hence we don't need to get the calling threadId explicitely and pass it here.

source/config/localhost_config.json
```diff
@@ -7,5 +7,12 @@
        "server_cert": "",
        "server_key": "",
        "root_cert": ""
+    },
+    "moniker_stream_processor": {
```
doshirohan, line 11: Since these are not applicable for most users, I would rather leave this out of default config.json file. Thoughts? @asumit

doshirohan, line 11: This can just be `processor_affinity`. Individual flags below call out that these are for "moniker streaming" APIs. This also gives us flexibility to add more configurations related to cpu pinning in future if needed which might not related to streaming.

source/server/moniker_stream_processor.h
```diff
@@ -0,0 +1,11 @@
+#ifndef MONIKER_STREAM_PROCESSOR_H
```
doshirohan, line 1: If you change the configuration name from `moniker_stream_processor` to `processor_affinity`, everything here should including file name should change to `processor_affinity` or even better `processor_affinity_configuration`

source/server/data_moniker_service.cpp
doshirohan: >the main purpose of this helper is thus to assign a specific processor to a thread for moniker streaming apis.
Hence the name, set_moniker_stream_processor sounds correct to me.

The API (or rather the private helper) doesn't have anything to do with moniker. Update name to `set_cpu_affinity` and change parameter name to `cpu`. The parameter as well has nothing to do with "stream".

source/server/server_configuration_parser.cpp
```diff
@@ -287,6 +293,41 @@ int ServerConfigurationParser::parse_port_with_key(const std::string& key) const
...
+        stream_processor.moniker_stream_write = parse_moniker_stream_processor_with_key(kMonikerStreamWriteKey);
+        stream_processor.moniker_stream_read = parse_moniker_stream_processor_with_key(kMonikerStreamReadKey);
+        stream_processor.moniker_stream_read_write = parse_moniker_stream_processor_with_key(kMonikerStreamReadWriteKey);
+    }
+    return stream_processor;
+}
+
+int ServerConfigurationParser::parse_moniker_stream_processor_with_key(const std::string& key) const
+{
+    int parsed_core = -1;
```
doshirohan, line 312: I think just `value` would've been sufficient for this, but I see all other methods have `parsed_<name>` variables. `parsed_cpu` will be better name.

Raghav-NI: Updated

source/server/moniker_stream_processor.h
Raghav-NI: Keeping it same as discussed in chat, refer to the Alex's comment above.

source/config/localhost_config.json
Raghav-NI: Removed

source/server/data_moniker_service.cpp
Raghav-NI: Updated

christag-ni:

asumit (approved):

reckenro (approved):

PR #1137, opened 2024-12-04
Adding server version always in the command display

yardov:
Is quite important to know what version of the server is running and not only when the --version is passed.

The server when executed will always display the version running.

All software needs to have a way to know the version running irrespective if a tag was passed. This ensure in our customers that they know the correct version is running.

Ideally, this should be queryable by a client but that's a task for another day.

None.

### DIFF
--- a/source/server/core_server.cpp
+++ b/source/server/core_server.cpp
@@ -88,6 +88,15 @@ static void RunServer(const ServerConfiguration& config)
         config.config_file_path.c_str());
   }
 
+  std::string string_kNiDeviceGrpcBranchName(nidevice_grpc::kNiDeviceGrpcBranchName);
+  if (string_kNiDeviceGrpcBranchName.rfind("releases", 0) == 0) {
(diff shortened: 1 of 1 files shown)


source/server/core_server.cpp
```diff
@@ -88,6 +88,15 @@ static void RunServer(const ServerConfiguration& config)
         config.config_file_path.c_str());
   }
 
+  std::string string_kNiDeviceGrpcBranchName(nidevice_grpc::kNiDeviceGrpcBranchName);
```
maxxboehme, line 91: This is the same logic as in the --version case. Please create a static function in this file. that is then called in both places so the logic for both of these will always stay consistent. 
https://github.com/ni/grpc-device/blob/d2f2862128e30d758aff738db6964d6c6dcf36d6/source/server/core_server.cpp#L206

maxxboehme (changes requested):

maxxboehme:
@yardov planning to create a release branch of grpc-device 2.9.0 on the 18th. If want this change in the 2.9.0 release please make sure to continue progress on the PR.

christag-ni:

christag-ni:

PR #1144, opened 2024-12-26
Separate out feature toggles for moniker streaming and sideband streaming

doshirohan:

We had single feature toggle `sideband-streaming` for moniker based streaming as well as sideband streaming support for moniker based streaming. This change separates out both of these into different feature toggles.

Moniker based streaming that uses sideband streaming mechanism opens up unsecured raw socket for communication. To make grpc-device more secure, we want sideband streaming to be OFF by default so that additional port is not opened up for users that don't need sideband based streaming mechanism. Since moniker based streaming is behind feature toggle, I have added sideband streaming with moniker as a feature toggle as well. In future, sideband streaming using moniker could be a configuration instead of feature toggle.

Validated with local client examples that
- Moniker streaming APIs are enabled only when `moniker_streaming` feature toggle is set to `true`.
- Sideband streaming is enabled only when both `moniker_streaming` and `moniker_streaming_sideband_support` feature toggles are set to `true`.

### DIFF
--- a/source/server/core_server.cpp
+++ b/source/server/core_server.cpp
@@ -114,7 +114,7 @@ static void RunServer(const ServerConfiguration& config)
       return;
     }
     server = builder.BuildAndStart();
-    if (ni::data_monikers::is_sideband_streaming_enabled(config.feature_toggles)) {
+    if (ni::data_monikers::is_moniker_streaming_enabled(config.feature_toggles)) {
       auto sideband_socket_thread = new std::thread(RunSidebandSocketsAccept, config.sideband_address.c_str(), config.sideband_port);
       ni::data_monikers::configure_moniker_stream_processor(config.stream_processor);
       // auto sideband_rdma_send_thread = new std::thread(AcceptSidebandRdmaSendRequests);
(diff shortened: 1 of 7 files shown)


Raghav-NI (approved):

source/server/core_server.cpp
```diff
@@ -266,7 +268,7 @@ int main(int argc, char** argv)
 #if defined(_WIN32)
   nidevice_grpc::set_console_ctrl_handler(&StopServer);
 #else
-  if (ni::data_monikers::is_sideband_streaming_enabled(config.feature_toggles)) {
+  if (ni::data_monikers::is_moniker_streaming_enabled(config.feature_toggles)) {
     SysFsWrite("/dev/cgroup/cpuset/system_set/cpus", "0-5");
```
asumit, line 272: Should we remove this hardcoded values  assigning cpus?

asumit (approved):

source/server/core_server.cpp
doshirohan: Yes. I think they got left out when Raghav removed configuration support for core server cpu affinity. I will create separate PR for better tracking and merge that as well in some time.

PR #1147, opened 2025-01-02
Add example for moniker based streaming

doshirohan:

Add example for moniker based streaming for DAQmx APIs

The example (alongwith wiki) will be useful for users trying to write moniker based streaming applications for grpc-device.

Example builds and runs successfully.

### DIFF
--- a/examples/nidaqmx/moniker-streaming/CMakeLists.txt
+++ b/examples/nidaqmx/moniker-streaming/CMakeLists.txt
@@ -1,0 +1,95 @@
+# cmake build file for C++ MonikerStreamingClient example.
+# Assumes protobuf and gRPC have been installed using cmake.
+
+cmake_minimum_required(VERSION 3.16)
+
+project(MonikerStreamingClient C CXX)
+
+find_package(Protobuf CONFIG REQUIRED)
+message(STATUS "Using protobuf ${Protobuf_VERSION}")
+
+set(_PROTOBUF_LIBPROTOBUF protobuf::libprotobuf)
+find_program(_PROTOBUF_PROTOC protoc)
+find_package(gRPC CONFIG REQUIRED)
+message(STATUS "Using gRPC ${gRPC_VERSION}")
+set(_GRPC_GRPCPP gRPC::grpc++)
+find_program(_GRPC_CPP_PLUGIN_EXECUTABLE grpc_cpp_plugin)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+
+# Proto files
+get_filename_component(daqmx_proto "../../../generated/nidaqmx/nidaqmx.proto" ABSOLUTE)
+get_filename_component(daqmx_proto_path "${daqmx_proto}" PATH)
+get_filename_component(session_proto "../../../imports/protobuf/session.proto" ABSOLUTE)
+get_filename_component(session_proto_path "${session_proto}" PATH)
+get_filename_component(data_moniker_proto "../../../imports/protobuf/data_moniker.proto" ABSOLUTE)
+get_filename_component(data_moniker_proto_path "${data_moniker_proto}" PATH)
+
+# Generated sources
+set(daqmx_proto_srcs "${CMAKE_CURRENT_BINARY_DIR}/nidaqmx.pb.cc")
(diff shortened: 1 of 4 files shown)


examples/nidaqmx/moniker-streaming/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+find_program(_PROTOBUF_PROTOC protoc)
+find_package(gRPC CONFIG REQUIRED)
+message(STATUS "Using gRPC ${gRPC_VERSION}")
+set(_GRPC_GRPCPP gRPC::grpc++)
+find_program(_GRPC_CPP_PLUGIN_EXECUTABLE grpc_cpp_plugin)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+
+# Proto files
+get_filename_component(daqmx_proto "../../../generated/nidaqmx/nidaqmx.proto" ABSOLUTE)
```
asumit, line 21: Can we use path relative to {CMAKE_SOURCE_DIR}

examples/nidaqmx/moniker-streaming/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+        --plugin=protoc-gen-grpc="${_GRPC_CPP_PLUGIN_EXECUTABLE}"
+        "${session_proto}"
+    DEPENDS "${session_proto}")
+
+# Generated sources
+set(data_moniker_proto_srcs "${CMAKE_CURRENT_BINARY_DIR}/data_moniker.pb.cc")
+set(data_moniker_proto_hdrs "${CMAKE_CURRENT_BINARY_DIR}/data_moniker.pb.h")
+set(data_moniker_grpc_srcs "${CMAKE_CURRENT_BINARY_DIR}/data_moniker.grpc.pb.cc")
+set(data_moniker_grpc_hdrs "${CMAKE_CURRENT_BINARY_DIR}/data_moniker.grpc.pb.h")
+add_custom_command(
```
asumit, line 65: We could create a helper function to generate grpc plumbing code that can be used for the 3 calls here. Similar to how we do it in the main CMakeLists.txt.

Since this is an example client, I do not have a strong recommendation for it, just something nice to have.

examples/nidaqmx/moniker-streaming/stream-read-analog-input-using-moniker-streaming.cpp
```diff
@@ -0,0 +1,245 @@
...
+      begin_read_context);
+    auto daqmx_read_moniker = new ni::data_monikers::Moniker(begin_read_response.moniker());
+
+    ::grpc::ClientContext stream_read_context;
+    ni::data_monikers::MonikerList read_request;
+    read_request.mutable_read_monikers()->AddAllocated(daqmx_read_moniker);
+    auto read_stream = moniker_service.StreamRead(&stream_read_context, read_request);
+
+    // Read data
+    for(int i=0; i<NUM_ITERATIONS; i++) {
```
asumit, line 192: nit : space after "for" , space around the operators = and <

asumit (approved):

examples/nidaqmx/moniker-streaming/CMakeLists.txt
doshirohan: There are two ways customers can potentially consume this CMakeLists.txt
1. Directly from this folder after cloning the grpc-device repo:
- In this case `{CMAKE_SOURCE_DIR}` will point to this directory, not root of the grpc-device repo. So I will still need to de-reference relative path using "../../../". I am not sure specific advantage of using the variable in this case, but I see it being done elsewhere in other repos. So I will follow the pattern and update to `{CMAKE_SOURCE_DIR}/../../../generated/nidaqmx/nidaqmx.proto`.
2. From the `ni-grpc-device-client` zip files we attach to the _Release_:
- In that case this relative path doesn't make any sense as protos are stored in `proto` directory in the root of the archive. The users will get error when trying to build this and will need to update relative path of protos appropriately.

Does this make sense to you? Or do you think we should rather make it work for archive in the Release and someone building from grpc-device clone should get error and need to fix? I guess either way doesn't really make much difference really.

examples/nidaqmx/moniker-streaming/CMakeLists.txt
doshirohan: Added `GenerateGrpcSources` helper.

PR #1161, opened 2025-04-01
Update grpc from 1.51.1 to 1.69.0

ckoellin:

Update the version of grpc from 1.51.1 to 1.69.0. There were a number of classes of changes that needed to happen that I will outline below.

- A subset of the nixnetsocket API was getting replaced by Windows #defines from winsock2. This occurred in a number of places including the protobuf codegen. #undef problematic defines.
- The keyword "shutdown" was colliding with Windows header defines so I renamed to shutdown_server.
- The keyword "GetCurrentTime" was colliding with Windows header defines so push/undef/pop define.
- API deprecation of "CreateMessage" is now replaced with "Create".
- The removal of grpc::internal::GrpcLibraryInitializer from grpc.
- Collision of pb = ::google::protobuf. Updated a few headers and tests to use fully qualified names instead.
- Update the grpc submodule to point to 1.69.0. The newest version is 1.71.0 but there are currently issues if you try and use it as a git submodule. See https://github.com/grpc/grpc/issues/38606
- Update AIFiniteAcquisition_ReadWithTimeoutTooSmall_SamplesNotYetAvailableStatusReturned since the test was failing the asserts after checking for SamplesNotYetAvailableStatus. To me the test was trying to validate too much with the last asserts possibly not always being true. I updated the test to only validate SamplesNotYetAvailableStatus.

This addresses known CVEs in grpc 1.51.1.

Builds for Release/Debug on Windows (VS 2022), Linux Desktop(g++ 9.4.0), and Linux RT(g++ 13.3.0). All Unit and Integration tests passed. I could only run a subset of the System tests given I do no have all the necessary HW required but the ones I could run passed. I ran a sample application that acquired voltage data from a simulated NI-DAQmx PCIe-6363.


maxxboehme:
Looking at the length of your latest PR vs the CI it looks like updating added around 30 minutes to the Windows and Linux build. And it appears to be mostly in the build step and not the added step of clearing up space.

maxxboehme (changes requested):

astarche (approved):

ckoellin:
> Looking at the length of your latest PR vs the CI it looks like updating added around 30 minutes to the Windows and Linux build. And it appears to be mostly in the build step and not the added step of clearing up space.

Unfortunately I don't think there is much we can do about this. Even if we were to pull grpc from a conan package, by default this would still compile it as a source dependency given how much a client build environment can very (compiler version, compiler flags, static vs dynamic, etc.)

astarche:
> > Looking at the length of your latest PR vs the CI it looks like updating added around 30 minutes to the Windows and Linux build. And it appears to be mostly in the build step and not the added step of clearing up space.
> 
> Unfortunately I don't think there is much we can do about this. Even if we were to pull grpc from a conan package, by default this would still compile it as a source dependency given how much a client build environment can very (compiler version, compiler flags, static vs dynamic, etc.)

@ckoellin @maxxboehme I never got this to work, but supposedly you could use something like [`ccache` in a GH action](https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/) to speed up builds of code that rarely changes.

Is it possible that there are grpc/proto build options that would reduce what is built? Is it the actual grpc build time that increases? Or perhaps the proto codegen/templates for our generated services has gotten more complex?

We could also use custom build agents with more power. GH action agents are borderline for a large C++ build. When I was working on grpc-device, we didn't really have a way to use custom agents. But I think there are more options now (the machines we use for system tests?).

I don't need this to be resolved with this PR. Just offering ideas in case anything is useful.

maxxboehme:
> > > Looking at the length of your latest PR vs the CI it looks like updating added around 30 minutes to the Windows and Linux build. And it appears to be mostly in the build step and not the added step of clearing up space.
> > 
> > 
> > Unfortunately I don't think there is much we can do about this. Even if we were to pull grpc from a conan package, by default this would still compile it as a source dependency given how much a client build environment can very (compiler version, compiler flags, static vs dynamic, etc.)
> 
> @ckoellin @maxxboehme I never got this to work, but supposedly you could use something like [`ccache` in a GH action](https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/) to speed up builds of code that rarely changes.
> 
> Is it possible that there are grpc/proto build options that would reduce what is built? Is it the actual grpc build time that increases? Or perhaps the proto codegen/templates for our generated services has gotten more complex?
> 
> We could also use custom build agents with more power. GH action agents are borderline for a large C++ build. When I was working on grpc-device, we didn't really have a way to use custom agents. But I think there are more options now (the machines we use for system tests?).
> 
> I don't need this to be resolved with this PR. Just offering ideas in case anything is useful.

Experience so far is that the custom agents have not been the easiest to maintain. It took several weeks to get both the Ubuntu and Windows system tests in a workable state the last time there were issues.  Maintaining them has been a very one off process vs the other build processes we have.

maxxboehme (approved):
Approving but don't merge until ready to look into the internal build pipeline updates needed as well. Don't want to have that broken for too long a period of time.

maxxboehme:
> Approving but don't merge until ready to look into the internal build pipeline updates needed as well. Don't want to have that broken for too long a period of time.

Never mind I see you kept the "Multi Config" and commented why.

bkeryan:
@astarche wrote:
> @ckoellin @maxxboehme I never got this to work, but supposedly you could use something like [`ccache` in a GH action](https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/) to speed up builds of code that rarely changes.

Even without ccache, you might be able to use actions/cache to cache the build directory. For example, PR builds could start with the most recent successful CI build and do an incremental build against it.

How big is the build directory? Each repo is limited to 10 GB of total cache:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

PR #1165, opened 2025-04-07
add component.yml for sbom_info

christag-ni:
Introduces component.yml file that resolves [AB#3017003](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3017003)

We should be including component.yml files in our ad-hoc exports as well 

None

### DIFF
--- a/component.yml
+++ b/component.yml
@@ -1,8 +1,8 @@
 component:
   sbom_info:
-    - name: grpc
-      version: 1.69.0
+    - name: grpc-device
+      version: 2.11.0
       identifiers:
-        cpe: cpe:2.3:a:grpc:grpc:1.69.0:*:*:*:*:*:*:*
+        cpe: cpe:2.3:a:grpc-device:grpc-device:2.11.0:*:*:*:*:*:*:*
       licenses:
-        - legal_package: gRPC157
+        - legal_package: grpc-device151
--- a/component.yml
+++ b/component.yml
@@ -1,6 +1,6 @@
 component:
   sbom_info:
-    - name:grpc
+    - name: grpc
       version: 1.69.0
       identifiers:
         cpe: cpe:2.3:a:grpc:grpc:1.69.0:*:*:*:*:-:*:*
--- a/component.yml
+++ b/component.yml
@@ -5,4 +5,4 @@ component:
       identifiers:
         cpe: cpe:2.3:a:grpc:grpc:1.69.0:*:*:*:*:-:*:*
       licenses:
-        - legal_package: gRPC151
+        - legal_package: gRPC157
--- a/component.yml
+++ b/component.yml
@@ -3,6 +3,6 @@ component:
     - name: grpc
       version: 1.69.0
       identifiers:
-        cpe: cpe:2.3:a:grpc:grpc:1.69.0:*:*:*:*:-:*:*
+        cpe: cpe:2.3:a:grpc:grpc:1.69.0:*:*:*:*:*:*:*
       licenses:
         - legal_package: gRPC157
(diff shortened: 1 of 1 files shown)


christag-ni:
@reckenro, do you know for which components I'm supposed to include sbom info? Here's an example with some grpc components
https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2FSource%2FLicenseManagement%2FnilmCompatibilityDistribution%2Fnilm_compatibility_server%2Fcomponent.yml&version=GBmain&_a=contents

reckenro:
> @reckenro, do you know for which components I'm supposed to include sbom info? Here's an example with some grpc components https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2FSource%2FLicenseManagement%2FnilmCompatibilityDistribution%2Fnilm_compatibility_server%2Fcomponent.yml&version=GBmain&_a=contents

I'm not sure. I honestly don't have too much experience with SBOM. I would guess since we're including gRPC we'd include the other third-party submodules we pull in in [https://github.com/ni/grpc-device/tree/main/third_party](https://github.com/ni/grpc-device/tree/main/third_party).

Do we also include items for the things in [https://github.com/ni/grpc-device/blob/main/python_build_requirements.txt](https://github.com/ni/grpc-device/blob/main/python_build_requirements.txt)? I'm not sure.

component.yml
```diff
@@ -0,0 +1,8 @@
+component:
+  sbom_info:
+    - name: grpc
+      version: 1.69.0
+      identifiers:
+        cpe: cpe:2.3:a:grpc:grpc:1.69.0:*:*:*:*:-:*:*
+      licenses:
+        - legal_package: gRPC151
```
dmondrik, line 8: In our installers I see some `gRPC157` instances but that's the highest I see. You should ask Fred if that is sufficient or if a newer one is warranted.

dmondrik (approved):
The file looks generally good to me.

dmondrik:
> I'm not sure. I honestly don't have too much experience with SBOM. I would guess since we're including gRPC we'd include the other third-party submodules we pull in in https://github.com/ni/grpc-device/tree/main/third_party.
> 
> Do we also include items for the things in https://github.com/ni/grpc-device/blob/main/python_build_requirements.txt? I'm not sure.

Regarding sub-modules, to me it would depend. If they are build-time dependencies only, it's probably not necessary. If it's included in the final production code, that sounds important yes.

bkeryan:
> Introduces component.yml file that resolves bug

FYI, if you write "AB#xxxxxx", a bot should automatically linkify it after you save the PR.

reckenro (approved):

reckenro:
> > I'm not sure. I honestly don't have too much experience with SBOM. I would guess since we're including gRPC we'd include the other third-party submodules we pull in in https://github.com/ni/grpc-device/tree/main/third_party.
> > Do we also include items for the things in https://github.com/ni/grpc-device/blob/main/python_build_requirements.txt? I'm not sure.
> 
> Regarding sub-modules, to me it would depend. If they are build-time dependencies only, it's probably not necessary. If it's included in the final production code, that sounds important yes.

That makes sense, in that case I would also include JSON and utfcpp submodules since those are used in production code. Also, [grpc-sideband](https://github.com/ni/grpc-sideband/tree/e351b75f2df9d932fb7993520429d7c680031864) but that's ours so guessing it doesn't need to be included.

astarche:
@reckenro @christag-ni  Does this file need to include transitive dependencies? Or does tooling handle that? `grpc` has a lot of dependencies, for example.

astarche (approved):

PR #1177, opened 2025-06-27
fix nifgen grpc-device tests

danielhuani:

(https://dev.azure.com/ni/DevCentral/_workitems/edit/3176369)

- Fix an expected error code for nifgen test, which was updated in nifgen driver in 2023 https://ni.visualstudio.com/DevCentral/_workitems/edit/2434002
- Fix the simulated device type

image

### DIFF
--- a/source/tests/system/nifgen_session_tests.cpp
+++ b/source/tests/system/nifgen_session_tests.cpp
@@ -12,12 +12,12 @@ namespace system {
 
 namespace fgen = nifgen_grpc;
 
-const int kInvalidFgenRsrc = -1074134944;
+const int kInvalidFgenRsrc = -1074134964;
 const int kInvalidFgenSession = -1074130544;
-const char* kViErrorFgenResourceNotFoundMessage = "Insufficient location information or resource not present in the system.\n\nInvalid Identifier: ";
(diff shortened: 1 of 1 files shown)


reckenro (approved):

danielhuani:
It seems out CI build is still using the old version of nifgen driver. I think we will need to upgrade it to the latest version.

reckenro:
> It seems the [PR / Run CI / run_win_system_tests](https://github.com/ni/grpc-device/actions/runs/15929262364/job/44940804419?pr=1177) is still using the old version of nifgen driver. I think we will need to upgrade it to the latest version.

I believe the feed for the Windows VM these tests run against is defined here: https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/eng/iac/cfg_mgmt/salt/srv/salt/grpcbot/win/resources/driver_feeds_win.yml&version=GBmain

I believe we'd been updating that as needed but not on any regular cadence. @maxxboehme can correct me if I'm wrong.

Regardless, after that's updated you could run this pipeline to regenerate the VM in AWS for these tests to run against: https://dev.azure.com/ni/DevCentral/_build?definitionId=9309

danielhuani:
Some other NiRFmxWLANDriverApiTests still failed even with the niRFmx driver 23.0. It looks like an intermittent error for the RF test by looking at the CI history failures. This shall not block this PR gets in @reckenro  @maxxboehme

PR #1198, opened 2025-09-22
update OS on windows build

christag-ni:
windows-latest is now windows server 2025, we want to use 2022

If we don't do this we'll keep seeing failures

None


astarche (approved):

maxxboehme (approved):
I think in a later change we can look into using windows-latest after updating the "sdk: 10.0.22621.0" argument to msvc-dev-cmd. Since this hit us right when releasing though we can move forward with this change.

PR #1203, opened 2025-10-16
Add ReadAnalogWaveforms to NiDAQmxService

mikeprosserni:

- Added ReadAnalogWaveforms method to the NiDAQmxService.
- Updated code generation with support for CustomCodeNoLibrary. This is needed for ReadAnalogWaveforms because it doesn't have a matching library function. (Instead, it calls `library_->InternalReadAnalogWaveformPerChan()`, which has very different parameters)
- Introduced new waveform attributes and functions in metadata.
- Updated CMake build process to include new waveform protobuf files.
- Added Ninja build instructions to CONTRIBUTING.md.

[AB#3424627](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3424627)

```
PS C:\dev\fireserp\grpc-device> .\build\Debug\SystemTestsRunner.exe --gtest_filter="*ReadAnalogWaveforms*"                                                             
Note: Google Test filter = *ReadAnalogWaveforms*
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from NiDAQmxDriverApiTests
[ RUN      ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithNoAttributeMode_ReturnsWaveformData
[       OK ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithNoAttributeMode_ReturnsWaveformData (4493 ms)
[ RUN      ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithTimingMode_ReturnsWaveformDataWithTimingInfo
[       OK ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithTimingMode_ReturnsWaveformDataWithTimingInfo (128 ms)
[ RUN      ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithExtendedPropertiesMode_ReturnsWaveformDataWithAttributes
[       OK ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithExtendedPropertiesMode_ReturnsWaveformDataWithAttributes (15 ms)
[ RUN      ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithTimingAndExtendedPropertiesMode_ReturnsWaveformDataWithTimingAndAttributes
[       OK ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_WithTimingAndExtendedPropertiesMode_ReturnsWaveformDataWithTimingAndAttributes (81 ms)
[----------] 4 tests from NiDAQmxDriverApiTests (4746 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (4754 ms total)
[  PASSED  ] 4 tests.
```

### DIFF
--- a/source/tests/unit/precision_timestamp_converter_tests.cpp
+++ b/source/tests/unit/precision_timestamp_converter_tests.cpp
@@ -48,7 +48,6 @@ TEST(PrecisionTimestampConverterTests, ConvertOneSecondOfTicks_ReturnsOneSecond)
 struct SecondsAndFractionalTestParam {
   int64_t ticks_offset;
   int64_t seconds;
-  double fraction;
   uint64_t btf_fractional_ticks;
 };
 
@@ -68,21 +67,21 @@ INSTANTIATE_TEST_SUITE_P(
   SecondsAndFractionalTests,
   PrecisionTimestampConverterSecondsAndFractionalTests,
   ::testing::Values(
-    SecondsAndFractionalTestParam{  0000000,  0, 0.00,      0x0000000000000000ULL},
-    SecondsAndFractionalTestParam{  2500000,  0, 0.25,      0x4000000000000000ULL},
-    SecondsAndFractionalTestParam{  5000000,  0, 0.50,      0x8000000000000000ULL},
-    SecondsAndFractionalTestParam{  7500000,  0, 0.75,      0xC000000000000000ULL},
-    SecondsAndFractionalTestParam{  9999999,  0, 0.9999999, 0xFFFFFE5280D65800ULL},
-    SecondsAndFractionalTestParam{ 12500000,  1, 0.25,      0x4000000000000000ULL},
-    SecondsAndFractionalTestParam{ 25000000,  2, 0.50,      0x8000000000000000ULL},
-    SecondsAndFractionalTestParam{ 37500000,  3, 0.75,      0xC000000000000000ULL},
-    SecondsAndFractionalTestParam{ -2500000, -0, 0.25,      0x4000000000000000ULL},
-    SecondsAndFractionalTestParam{ -5000000, -0, 0.50,      0x8000000000000000ULL},
-    SecondsAndFractionalTestParam{ -7500000, -0, 0.75,      0xC000000000000000ULL},
-    SecondsAndFractionalTestParam{ -9999999, -0, 0.9999999, 0xFFFFFE5280D65800ULL},
-    SecondsAndFractionalTestParam{-12500000, -1, 0.25,      0x4000000000000000ULL},
-    SecondsAndFractionalTestParam{-25000000, -2, 0.25,      0x8000000000000000ULL},
-    SecondsAndFractionalTestParam{-37500000, -3, 0.75,      0xC000000000000000ULL}
+    SecondsAndFractionalTestParam{  0000000,  0, 0x0000000000000000ULL},
+    SecondsAndFractionalTestParam{  2500000,  0, 0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{  5000000,  0, 0x8000000000000000ULL},
+    SecondsAndFractionalTestParam{  7500000,  0, 0xC000000000000000ULL},
+    SecondsAndFractionalTestParam{  9999999,  0, 0xFFFFFE5280D65800ULL},
+    SecondsAndFractionalTestParam{ 12500000,  1, 0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{ 25000000,  2, 0x8000000000000000ULL},
+    SecondsAndFractionalTestParam{ 37500000,  3, 0xC000000000000000ULL},
+    SecondsAndFractionalTestParam{ -2500000, -0, 0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{ -5000000, -0, 0x8000000000000000ULL},
+    SecondsAndFractionalTestParam{ -7500000, -0, 0xC000000000000000ULL},
+    SecondsAndFractionalTestParam{ -9999999, -0, 0xFFFFFE5280D65800ULL},
+    SecondsAndFractionalTestParam{-12500000, -1, 0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{-25000000, -2, 0x8000000000000000ULL},
+    SecondsAndFractionalTestParam{-37500000, -3, 0xC000000000000000ULL}
   )
 );
 


.github/workflows/build_cmake.yml
```diff
@@ -25,7 +25,8 @@ jobs:
             cc: "gcc-9", cxx: "g++-9",
             glibc_version: "2_31",
             cmake_generator: '-G "Ninja"',
-            build_type: RelWithDebInfo
+            build_type: RelWithDebInfo,
```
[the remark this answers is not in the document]
bkeryan: Also, how long does it take to build locally with -O2? Does it eventually succeed?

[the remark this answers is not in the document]
bkeryan: VTA is a potential cause of resource exhaustion. This feature tracks variable locations for each line of code in order to produce debug info that accurately shows the variable values when you step through the function, even if it's inlined. However, when your code has hundreds of variable assignments in the same function, this feature can take a lot of CPU and memory. I don't know exactly why.

VTA only affects the debugging experience, not execution speed, so if turning it off solves your problem, then it's a better solution than using -O1 and turning off a bunch of optimizations.

Please try updating https://github.com/ni/grpc-device/blob/main/source/codegen/templates/service.cpp.mako to use `#pragma GCC optimize` to scope the `-fno-var-tracking-assignments` or `-O1` to `*_library.cpp`. You will need to guard this pragma with `#ifdef __GNUC__` or similar.

source/codegen/templates/service.cpp.mako
```diff
@@ -51,6 +51,9 @@ resource_repository_deps = service_helpers.get_driver_shared_resource_repository
 % endif
 % if any_non_mockable_functions:
 #include "${module_name}_library.h"
+#ifdef __GNUC__
```
bkeryan: Move this below the last `#include` so it only affects this file, not the `#include`.

source/codegen/templates/service.cpp.mako
bkeryan: Wait, which file is causing the problem? service.cpp or library.cpp? I thought you said it was library.cpp.

.github/workflows/build_cmake.yml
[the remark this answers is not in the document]
bkeryan: I checked out your branch on my laptop, commented out the `-fno-var-tracking-assignments`, and ran `time cmake --build . -j 16 --config RelWithDebInfo` on WSL:

```
real    21m28.891s
user    298m4.721s
sys     30m30.097s
```

Running `.ninja_log` through https://github.com/nico/ninjatracing shows

image

It seems like all of the DAQmx files are problem files, taking >3 minutes to compile.

source/codegen/metadata/nidaqmx/enums_addon.py
```diff
@@ -1,7 +1,33 @@
 # These dictionaries are applied to the generated enums dictionary at build time
 # Any changes to the API should be made here. enums.py is code generated
 
-enums_override_metadata = {}
+enums_override_metadata = {
+    'WaveformAttributeMode': {
```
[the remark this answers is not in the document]
bkeryan: @mikeprosserni If scrapigen needs CVI annotations, copy the version from exported nicai/tools/daqmxInternalAPI.h

reckenro (approved):

source/codegen/metadata/nidaqmx/enums_addon.py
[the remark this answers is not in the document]
bkeryan: Oops, wrong enum, this one doesn't exist in C

bkeryan (commented):
I'll review the actual code today. So far I've only looked at the build problems.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -10,4 +101,139 @@ ::grpc::Status NiDAQmxService::ConvertApiErrorStatusForTaskHandle(::grpc::Server
...
+    }
+
+    uInt32 num_channels = 0;
+    auto status = library_->GetReadAttributeUInt32(task, ReadUInt32Attribute::READ_ATTRIBUTE_NUM_CHANS, &num_channels);
+    if (!status_ok(status)) {
+      return ConvertApiErrorStatusForTaskHandle(context, status, task);
+    }
+
+    if (num_channels == 0) {
+      return ::grpc::Status(::grpc::INVALID_ARGUMENT, "No channels found in task");
```
bkeryan: READ_ATTRIBUTE_NUM_CHANS isn't the number of channels in the task, it's the number of channels specified by DAQmx_Read_ChannelsToRead.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,7 +1,98 @@
...
+// Returns true if it's safe to use outputs of a method with the given status.
+inline bool status_ok(int32 status)
+{
+  return status >= 0;
+}
+
+// Callback data structure to pass waveform pointers to the callback
+struct WaveformCallbackData {
+  std::vector<::ni::protobuf::types::DoubleAnalogWaveform*> waveforms;
+};
```
bkeryan: I think you can replace this struct with `RepeatedPtrField<DoubleAnalogWaveform>`. It's already an STL-style container.

Also, when you add DigitalWaveform support, I think you can template this function on the waveform type.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,7 +1,98 @@
 #include <nidaqmx/nidaqmx_service.h>
+#include <vector>
+#include <memory>
+#include <cstdint>
+#include "NIDAQmxInternalWaveform.h"
 
 namespace nidaqmx_grpc {
 
+constexpr int64 TICKS_PER_SECOND = 10000000; // 100ns ticks per second
+constexpr double TICKS_TO_SECONDS = 1e-7; // Convert 100ns ticks to seconds
```
bkeryan: SECONDS_PER_TICK?

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -10,4 +101,139 @@ ::grpc::Status NiDAQmxService::ConvertApiErrorStatusForTaskHandle(::grpc::Server
...
+        }
+
+        auto* t0 = waveform->mutable_t0();
+        // Convert from 100ns ticks (DAQmx format) to PrecisionTimestamp
+        // t0_array[i] contains 100ns ticks since Jan 1, 0001 (.NET DateTime epoch)
+        const int64_t seconds = t0_array[i] / TICKS_PER_SECOND;
+        const int64_t fractional_ticks = t0_array[i] % TICKS_PER_SECOND;
+        
+        t0->set_seconds(seconds);
+        t0->set_fractional_seconds(fractional_ticks);
```
bkeryan: This isn't converting fractional_ticks from 100 ns ticks to NI-BTF ticks. It's just taking the remainder, in 100 ns ticks, and treating them as NI-BTF ticks.

Look at how TimeSpanToCVITime does it in the .NET API. It calculates the fractional seconds as a double, then multiplies the double by ulong.MaxValue to convert it to NI-BTF ticks.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1479,6 +1511,198 @@ TEST_F(NiDAQmxDriverApiTests, SetPolynomialForwardCoefficients_GetPolynomialForw
...
+    // Get current time in seconds since year 1 AD (Jan 1, 0001) - .NET DateTime epoch
+    // This matches the format used by DAQmxInternalReadAnalogWaveformPerChan
+    // From year 1 AD to 1970-01-01 is 719162 days * 24 * 3600 = 62135596800 seconds
+    const auto epoch_offset_year1_to_1970 = 62135596800LL;
+    auto now = std::chrono::duration_cast<std::chrono::seconds>(
+        std::chrono::system_clock::now().time_since_epoch()).count();
+    auto now_since_year1 = now + epoch_offset_year1_to_1970;
+    const auto& timestamp = waveform.t0();
+    EXPECT_NEAR(timestamp.seconds(), now_since_year1, 1);
+    EXPECT_NE(timestamp.fractional_seconds(), 0);
```
bkeryan: Random timestamps have fractional seconds of 0, sometimes.

source/custom/nidaqmx_service.custom.cpp
mikeprosserni: this now uses `google::protobuf::RepeatedPtrField<::ni::protobuf::types::DoubleAnalogWaveform>*`

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -10,4 +101,139 @@ ::grpc::Status NiDAQmxService::ConvertApiErrorStatusForTaskHandle(::grpc::Server
...
+    }
+
+    for (uInt32 i = 0; i < num_channels; ++i) {
+      auto* waveform = response->mutable_waveforms(i);
+      
+      auto* y_data = waveform->mutable_y_data();
+      y_data->Reserve(samples_per_chan_read);
+      for (int32 j = 0; j < samples_per_chan_read; ++j) {
+        y_data->Add(read_arrays[i][j]);
+      }
```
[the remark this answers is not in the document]
mikeprosserni: This now uses `y_data->Add(read_arrays[i].data(), read_arrays[i].data() + samples_per_chan_read);`. MSVC did not complain.

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -1,7 +1,98 @@
...
+        } else {
+          return -1;
+        }
+        break;
+        
+      case DAQmx_Val_WfmAttrType_String:
+        if (value_size_in_bytes > 0) {
+          const char* str_val = static_cast<const char*>(value);
+          const std::string string_val(str_val, value_size_in_bytes - 1);
+          attr_value.set_string_value(string_val);
```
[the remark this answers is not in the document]
mikeprosserni: This now uses `convert_to_grpc(std::string(str_val), &string_val);`

source/custom/nidaqmx_service.custom.cpp
mikeprosserni: This is now in converters.h:
```
// Convert ticks (100ns since Jan 1, 0001) to PrecisionTimestamp
inline void convert_ticks_to_precision_timestamp(int64 ticks, ::ni::protobuf::types::PrecisionTimestamp* timestamp)
{
  const double seconds = static_cast<double>(ticks) * SecondsPerTick;
  const int64 seconds_int = static_cast<int64>(std::floor(seconds));
  timestamp->set_seconds(seconds_int);
  const double fractional_seconds = std::abs(seconds - static_cast<double>(seconds_int));
  const uint64_t fractional_seconds_uint = static_cast<uint64_t>(fractional_seconds * UINT64_MAX);
  timestamp->set_fractional_seconds(fractional_seconds_uint);
}
```

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -178,6 +180,19 @@ class NiDAQmxDriverApiTests : public Test {
     return create_ai_voltage_chan(request, response);
   }
 
+  ::grpc::Status create_two_ai_voltage_chans(double min_val, double max_val, CreateAIVoltageChanResponse& response = ThrowawayResponse<CreateAIVoltageChanResponse>::response())
```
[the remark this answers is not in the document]
mikeprosserni: Now that you mention it, I think I prefer the idea of create_ai_voltage_chan() taking in the name. I've removed create_two_ai_voltage_chans() and upgraded create_ai_voltage_chan() instead.

source/tests/system/nidaqmx_driver_api_tests.cpp
mikeprosserni: I've removed this assertion. (during development, I was having a problem where the fractional seconds were always zero. And now we have unit tests for the conversion function.)

source/custom/nidaqmx_service.custom.cpp
```diff
@@ -203,27 +197,12 @@ ::grpc::Status NiDAQmxService::ReadAnalogWaveforms(::grpc::ServerContext* contex
       
       auto* y_data = waveform->mutable_y_data();
       y_data->Reserve(samples_per_chan_read);
-      for (int32 j = 0; j < samples_per_chan_read; ++j) {
-        y_data->Add(read_arrays[i][j]);
-      }
+      y_data->Add(read_arrays[i].data(), read_arrays[i].data() + samples_per_chan_read);
```
bkeryan: Add() with an iterator range should grow the array up front, so you don't need Reserve() anymore.

source/tests/unit/precision_timestamp_converter_tests.cpp
bkeryan: Most of these can be value-parameterized on a tuple of (ticks, expected)
https://google.github.io/googletest/advanced.html#value-parameterized-tests

source/server/converters.h
```diff
@@ -341,10 +342,10 @@ inline void convert_to_grpc(const SmtSpectrumInfoType& input, nidevice_grpc::Smt
 }
 
 const int64 SecondsFromCVI1904EpochTo1970Epoch = 2082844800LL;
+const int64 SecondsFromDAQmx0001EpochToCVI1904Epoch = -((static_cast<int64>(0xfffffff2) << 32) | 0x0493b980); // extracted from NITYPES_ABSOLUTETIME_EPOCH_BIAS_FROM_0001 in ni-central/src/platform_services/abstractions/niatomicd/nitypes/source/nitypes/time/AbsoluteTime.h
```
bkeryan: Y0001 is not a DAQmx concept.

```suggestion
const int64 SecondsFromDotNet0001EpochToCVI1904Epoch = -((static_cast<int64>(0xfffffff2) << 32) | 0x0493b980); // extracted from NITYPES_ABSOLUTETIME_EPOCH_BIAS_FROM_0001 in ni-central/src/platform_services/abstractions/niatomicd/nitypes/source/nitypes/time/AbsoluteTime.h
```

source/server/converters.h
```diff
@@ -341,10 +342,10 @@ inline void convert_to_grpc(const SmtSpectrumInfoType& input, nidevice_grpc::Smt
...
 const double TwoToSixtyFour = (double)(1 << 31) * (double)(1 << 31) * (double)(1 << 2);
 const double NanosecondsPerSecond = 1000000000.0;
-const int64 TicksPerSecond = 1e7; // each tick is 100ns
-const double SecondsPerTick = 1e-7;
+const int64 DotNetTicksPerSecond = 1e7; // each tick is 100ns
```
bkeryan: 1e7 is a floating-point literal

source/server/converters.h
bkeryan: 0xFFFFFFF2_0493b980 is negative, so you negate to get the positive number of seconds from Y0001 to Y1904.

Then when you use it:

> const int64 dot_net_ticks_since_1904 = dot_net_ticks - SecondsFromDAQmx0001EpochToCVI1904Epoch * DotNetTicksPerSecond;

dot_net_ticks is the ticks since Y0001, which includes the ticks between Y0001 and Y1904, so you subtract the number of ticks between Y0001 and Y1904. Ok, I think this makes sense.

source/server/converters.h
```diff
@@ -371,15 +372,21 @@ inline CVIAbsoluteTime convert_from_grpc(const google::protobuf::Timestamp& valu
...
-  timestamp->set_fractional_seconds(fractional_seconds_uint);
+  const int64 dot_net_ticks_since_1904 = dot_net_ticks - SecondsFromDAQmx0001EpochToCVI1904Epoch * DotNetTicksPerSecond;
+  const double total_seconds = static_cast<double>(dot_net_ticks_since_1904) / DotNetTicksPerSecond;
+  double integer_part;
+  double fractional_part = std::modf(total_seconds, &integer_part);
+  
+  if (fractional_part < 0) {
+    integer_part -= 1;
+    fractional_part += 1;
+  }
```
bkeryan: This is incorrect logic for NI-BTF. 

```
    double msb;
    double lsb = modf(-1.25, &msb);
    printf("%f %f\n", msb, lsb);
```

msb=-1.0 lsb=-0.25

You add/subtract to get
msb=-2.0 lsb=0.75

NI-BTF is fixed point, so the fractional part is an extension of the integer part and shares its sign. You are treating the fractional part as if it's always added, regardless of the sign of the integer part. I think struct timeval might work this way, but NI-BTF doesn't.

A fixed point number of msb=-2.0 lsb=0.75 has an absolute value of 2.75. The correct value would be msb=-1.0 lsb=0.25.

source/tests/unit/precision_timestamp_converter_tests.cpp
```diff
@@ -8,84 +8,139 @@ namespace tests {
...
+  EXPECT_EQ(timestamp.seconds(), param.seconds) << "Failed for " << param.description;
+  EXPECT_EQ(timestamp.fractional_seconds(), static_cast<uint64_t>(param.fraction * TwoToSixtyFour)) << "Failed for " << param.description;
 }
 
-TEST(PrecisionTimestampConverterTests, ConvertSingleTick_ReturnsCorrectFractionalSeconds)
+INSTANTIATE_TEST_SUITE_P(
+  SecondsAndFractionalTests,
+  PrecisionTimestampConverterSecondsAndFractionalTests,
+  ::testing::Values(
+    SecondsAndFractionalTestParam{0, 0.25, DotNetTicksPerSecond / 4, "quarter second"},
```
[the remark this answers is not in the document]
mikeprosserni: I've added lots of negative offset test cases.

source/tests/unit/precision_timestamp_converter_tests.cpp
```diff
@@ -58,53 +58,31 @@ TEST_P(PrecisionTimestampConverterSecondsAndFractionalTests, ConvertSecondsAndFr
...
+    SecondsAndFractionalTestParam{  9999999,  0, 0.9999999, 0xFFFFFE5280D65800ULL},
+    SecondsAndFractionalTestParam{ 12500000,  1, 0.25,      0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{ 25000000,  2, 0.50,      0x8000000000000000ULL},
+    SecondsAndFractionalTestParam{ 37500000,  3, 0.75,      0xC000000000000000ULL},
+    SecondsAndFractionalTestParam{ -2500000, -0, 0.25,      0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{ -5000000, -0, 0.50,      0x8000000000000000ULL},
+    SecondsAndFractionalTestParam{ -7500000, -0, 0.75,      0xC000000000000000ULL},
+    SecondsAndFractionalTestParam{ -9999999, -0, 0.9999999, 0xFFFFFE5280D65800ULL},
+    SecondsAndFractionalTestParam{-12500000, -1, 0.25,      0x4000000000000000ULL},
+    SecondsAndFractionalTestParam{-25000000, -2, 0.25,      0x8000000000000000ULL},
```
bkeryan: param.fraction is wrong for -2.5, which highlights that this field is no longer used.

mikeprosserni: removed the unused fraction field

PR #1211, opened 2025-10-27
Copilot Build Instructions

mikeprosserni:

I've been experimenting with using Copilot's Agent Mode with the grpc-device repo, and the biggest challenge is that copilot is not able to do a build correctly when cmake is configured with ninja. Here's what happens when you try:

image

To fix this, I've added a [copilot-instructions.md](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions) file that tells copilot to use a `copilot-build.ps1` script that I've created. This build script always calls  `vcvars64.bat` to set the environment variables before calling cmake.

This allows Copilot Agent Mode to do complete build/test/fix loops on its own.

image

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -11,7 +11,7 @@ This repository requires specific environment setup before building. **Always** 
 
 ### Critical: Preferred Build Method
 
-**Always use the copilot-build.ps1 script for all builds (recommended for Copilot):**
+**Always use the copilot-build.ps1 script for all builds:**
 
 ```powershell
 # Default Debug build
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,83 @@
...
+This repository requires specific environment setup before building. **Always** follow these guidelines when running build commands.
+
+### Prerequisites
+
+1. **Visual Studio Build Tools**: Ensure Visual Studio Build Tools are installed with C++ development components
+2. **CMake**: Make sure CMake is available in the system PATH
+
+### Critical: Preferred Build Method
+
+**Always use the copilot-build.ps1 script for all builds (recommended for Copilot):**
```
reckenro: I would take off this (recommended for Copilot) part. I'm assuming these instructions are only for copilot and therefore the Always directive is correct.

maxxboehme (approved):

PR #1223, opened 2025-11-06
Add support for READ_ALL_AVAILABLE in ReadAnalogWaveforms and ReadDigitalWaveforms

mikeprosserni:

This adds support for `READ_ALL_AVAILABLE` in `ReadAnalogWaveforms()` and `ReadDigitalWaveforms()`, using `DAQmx_DefaultNumberOfSamplesToRead` if `number_of_samples_per_channel` is -1.

[AB#3460923](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3460923)

Added these tests:
```
[       OK ] NiDAQmxDriverApiTests.ReadAnalogWaveforms_ReadAllAvailable_ReturnsAvailableSamples (13 ms)
[       OK ] NiDAQmxDriverApiTests.ReadDigitalWaveforms_ReadAllAvailable_ReturnsAvailableSamples (11 ms)
```

### DIFF
--- a/source/tests/system/nidaqmx_driver_api_tests.cpp
+++ b/source/tests/system/nidaqmx_driver_api_tests.cpp
@@ -1959,12 +1959,18 @@ TEST_F(NiDAQmxDriverApiTests, ReadAnalogWaveforms_WithTimingAndExtendedPropertie
 
 TEST_F(NiDAQmxDriverApiTests, ReadAnalogWaveforms_ReadAllAvailable_ReturnsAvailableSamples)
 {
+  const auto FINITE_SAMPLES = 100;
   const auto TIMEOUT = 10.0;
   CreateAIVoltageChanResponse create_channel_response;
   auto create_channel_status = create_ai_voltage_chan("gRPCSystemTestDAQ/ai0", "ai0", -1.0, 1.0, create_channel_response);
   EXPECT_SUCCESS(create_channel_status, create_channel_response);
   create_channel_status = create_ai_voltage_chan("gRPCSystemTestDAQ/ai1", "ai1", -1.0, 1.0, create_channel_response);
   EXPECT_SUCCESS(create_channel_status, create_channel_response);
+
+  auto timing_request = create_cfg_samp_clk_timing_request(1000.0, Edge1::EDGE1_RISING, AcquisitionType::ACQUISITION_TYPE_FINITE_SAMPS, FINITE_SAMPLES);
+  CfgSampClkTimingResponse timing_response;
+  auto timing_status = cfg_samp_clk_timing(timing_request, timing_response);
+  EXPECT_SUCCESS(timing_status, timing_response);
 
   start_task();
   ::grpc::ClientContext context;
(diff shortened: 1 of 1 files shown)


generated/nidaqmx/nidaqmx.proto
bkeryan: I suppose you want nidaqmx.proto to support GetDefaultNumberOfSamplesToRead so you can call it from Task._calculate_num_samps_per_chan in Python? That makes sense, but now that I am reviewing this, I see it's more complicated than I thought.

The C API's ReadAnalogF64 does not allocate the read buffer for you. It requires you to allocate a buffer up front and tell it how big the buffer is. If you specify num_samps_per_chan=-1 in order to read all available samples, then it queries the available samples and caps the value based on the read buffer size.

The .NET API doesn't have the same constraint. You can call reader.ReadMultiSample(-1) and it allocates as much memory as it needs to, capping only at the .NET array size limit. (Note: the available samples per channel are limited by the task's input buffer size anyway.)

gRPC and Python can also allocate memory on your behalf, so it would make sense for them to handle -1 and return an appropriately sized buffer.

However, grpc-device is a translation of the C API, so its ReadAnalogF64 has an array_size_in_samps parameter so you can tell it the read buffer size. This still enables read all available samples to work, the same way it does in C, but it means that you have to pass in a buffer size that is not required in .NET or Python.

Adding GetDefaultNumberOfSamplesToRead to nidaqmx.proto would allow the Python API to query the buffer size it needs to allocate, so it could then call ReadAnalogF64 with that buffer size. This can work, but it makes 2 RPCs when it should only really need one.

I think the best long-term solution would be to change ReadAnalogF64's array_size_in_samps parameter to be optional.
- If you specify array_size_in_samps, then it passes that to the C API and everything works like today.
- If you don't specify array_size_in_samps, then it calculates the required array size based on num_samps_per_chan and GetReadNumChans. If num_samps_per_chan == -1, then it also uses GetDefaultNumberOfSamplesToRead.

Caveats:
- GetDefaultNumberOfSamplesToRead requires NI-DAQmx 24.5 or 24.8, I forget which. This means that if you use it without a version check or fallback, then you may break read all available use cases that worked before.
- Unscaled and raw reads also support read all available, but translating from num_samps_per_chan to a buffer size is a little more complicated for these. I think raw reads require another internal property.

bkeryan (changes requested):

generated/nidaqmx/nidaqmx.proto
bkeryan: More issues with adding GetDefaultNumberOfSamplesToRead to nidaqmx.proto:

- We generally don't add single-attribute get/set functions to nidaqmx.proto. Instead, we use generic get/set -functions like GetReadAttributeUInt32.
- GetDefaultNumberOfSamplesToRead is an internal API, so it belongs in nidaqmx_restricted.proto.
 
I think adding it to a restricted proto also avoids setting a precedent of adding single-attribute get/set functions to nidaqmx.proto.

mikeprosserni: I've changed everything to use GetReadAttributeUInt32(DAQmx_DefaultNumberOfSamplesToRead) instead of GetDefaultNumberOfSamplesToRead(). So, I don't need to update any proto files anymore.

bkeryan: Ok, I filed https://github.com/ni/grpc-device/issues/1224 about the non-waveform read functions.

source/tests/system/nidaqmx_driver_api_tests.cpp
```diff
@@ -1956,6 +1957,41 @@ TEST_F(NiDAQmxDriverApiTests, ReadAnalogWaveforms_WithTimingAndExtendedPropertie
...
+  start_task();
+  ::grpc::ClientContext context;
+  ReadAnalogWaveformsResponse read_response;
+  auto read_status = read_analog_waveforms(context, READ_ALL_AVAILABLE, TIMEOUT, WaveformAttributeMode::WAVEFORM_ATTRIBUTE_MODE_NONE, read_response);
+  stop_task();
+
+  EXPECT_SUCCESS(read_status, read_response);
+  EXPECT_EQ(read_response.status(), DAQMX_SUCCESS);
+  EXPECT_EQ(read_response.waveforms_size(), 2);
+  EXPECT_EQ(read_response.samps_per_chan_read(), 1);
```
bkeryan: Configure finite sample clock timing and validate that the sample count matches the number you passed to the timing function.

This also applies to the digital test.

reckenro (approved):

PR #1238, opened 2026-02-24
Update example package versions

astarche:

Update `examples/pyproject.toml` versions to latest.

Update `python` version constraints to match actively supported python versions.

Update action to use the oldest non-EoL python: 3.10.

Sample commands for similar update:

```
poetry add  grpcio@latest
poetry add --group dev grpcio-tools@latest mypy@latest mypy-protobuf@latest types-protobuf@latest types-grpcio@latest
poetry update
poetry lock
```

Discussion: https://github.com/ni/grpc-device/pull/1237#issuecomment-3953570275

Mix of old and new versions in example pyproject is causing build errors.

Users testing examples will use `pip` to install latest version.

Unbreak builds.

PR build.

### DIFF
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -35,4 +35,4 @@ jobs:
 
     - name: Validate examples (except aio)
       run: |
-        python source/codegen/validate_examples.py --exclude=".*-aio\.py$"
+        python source/codegen/validate_examples.py --exclude="-aio\.py$"


examples/pyproject.toml
```diff
@@ -6,18 +6,18 @@ authors = ["Christian Aguilar <christian.aguilar@ni.com>"]
...
-mypy-protobuf = "3.6.0"
-types-protobuf = ">=5.28.3.20241203"
-types-grpcio = ">=1.0"
-black = ">=23.3.0"
+grpcio-tools = "^1.78.1"
+mypy = "^1.19.1"
+mypy-protobuf = "^5.0.0"
+types-protobuf = "^6.32.1.20260221"
+types-grpcio = "^1.0.0.20251009"
+black = "^23.3.0"
```
astarche: I think there were some behavior changes in the last 2 major versions of black that would reformat a few files. So as a quick fix I'm sticking it on `23`.

Ideally we should get in sync with a specific ni-python-styleguide that we're also using in the code gen modules. But `black` is just used for internal style/consistency, not needed for correctness or a "contract" with clients. So improvements there are orthogonal to this change. At least until it starts breaking builds.

If there's a better quick fix I can try that too.

examples/pyproject.toml
```diff
@@ -6,18 +6,18 @@ authors = ["Christian Aguilar <christian.aguilar@ni.com>"]
...
-mypy = ">=1.0"
-mypy-protobuf = "3.6.0"
-types-protobuf = ">=5.28.3.20241203"
-types-grpcio = ">=1.0"
-black = ">=23.3.0"
+grpcio-tools = "^1.78.1"
+mypy = "^1.19.1"
+mypy-protobuf = "^5.0.0"
+types-protobuf = "^6.32.1.20260221"
+types-grpcio = "^1.0.0.20251009"
```
bkeryan: Node-style ^ versions have an implicit upper bound based on semver, so they require regularly updating `pyproject.toml` to keep dependencies up to date. For example, for Node applications, Renovate Bot can automatically change "^1.0.0" to "^1.0.0 || ^2.0.0" when version 2.x is released: https://docs.renovatebot.com/configuration-options/#rangestrategy

The current guidance for NI Python projects is to use `>=` by default and only use `^` or a bare version when you need an upper bound. That way, we can regularly update `poetry.lock` without having to also update `pyproject.toml`.

bkeryan: > Also, the Pythonic syntax for an upper bound is `black = ">=23.3,<24"`. AFAIK, Poetry is the only Python tool that also supports the Node-style syntax.

BTW switching to the standard pyproject.toml dependency schema (which is optional after upgrading to Poetry 2.x) would require using the Python version format, not the Node version format.

PR #1259, opened 2026-04-24
Add client defaults config to artifacts

texasaggie97:

Adds the gRPC Device client defaults TLS configuration file to the upstream repo and includes it in released client artifacts.

This change:
- adds the client defaults YAML as a source-owned config file
- updates client artifact staging so the file is exported in the client bundle under `config/`
- keeps the release workflow unchanged by using the existing client staging path that already feeds the published zip and tarball artifacts

The client defaults config needs to be sourced from the `grpc-device` repo instead of being owned only downstream in `ni-central`.

Merging this change makes the GitHub repo the source of truth for the client config payload and ensures published client artifacts contain the file needed by downstream consumers and installers.

- Ran the client staging script locally and verified the staged output contains `config/ni-grpc-device.defaults.yml`
- Built the repo locally on Windows in the Visual Studio 2022 developer environment with:
  - `cmake --build build --config RelWithDebInfo --target ni_grpc_device_server`

The local build completed successfully.

### DIFF
--- a/source/config/ni-grpc-device.defaults.yml
+++ b/source/config/ni-grpc-device.defaults.yml
@@ -1,0 +1,9 @@
+file_version: 1
+# TLS is disabled by default. Change certificate_mode to Unmanaged to enable
+# client authentication and provide paths to your certificate chain and private
+# key.
+certificate_mode: Disabled
+certificate_chain: File://$(UserServicesDir)/$(Service)/cert.pem
+certificate_key: File://$(UserServicesDir)/$(Service)/key.pem
+server_mode: Disabled
+trusted_certificates: SystemDefault
--- a/source/codegen/stage_client_files.py
+++ b/source/codegen/stage_client_files.py
@@ -100,6 +100,10 @@ class _ArtifactLocations:
     def metadata_dir(self) -> Path:
         return self.repo_root / "source" / "codegen" / "metadata"
 
+    @property
+    def config_dir(self) -> Path:
+        return self.repo_root / "source" / "config"
+
 
 def _get_release_proto_files(
     artifact_locations: _ArtifactLocations, readiness: _ArtifactReadiness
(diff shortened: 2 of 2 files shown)


source/config/ni-grpc-device.defaults.yml
```diff
@@ -0,0 +1,9 @@
+file_version: 1
```
shastriUF, line 1: @astarche @bkeryan We had talked about making TLS-config secure by default earlier today... Does this meet your expectations?

texasaggie97: FWIW, the values I have set for these is for mTLS to be _disabled_. There will be workflows in HWCU releasing soon to configure it.

bkeryan: I was suggesting that the server should "fail safe" if you delete `server_config.json`.

reckenro (approved):

source/config/ni-grpc-device.defaults.yml
bkeryan: Anyway, this PR is about the client config so I think our server config discussion is outside the scope of this PR.

astarche: The (relatively) secure default config that we've used in the past is "localhost only, TLS disabled" listener in the server. Yeah, as Brad said, we should (at least) make that the default in code, not just the config file.

Being able to default to a default/ni-managed certificate config would be great. I'm not sure if that's in the scope of Mark's work. Defaulting to "broken / prompt for cert or config" could also be a workable option. Especially if many of our distribution mechanisms (i.e., ni installers) are able to configure a deployment-specific default on top of that.

I think overall this change makes sense. This work gives us more/better options for default server config. But changing default server config is separate from this change.

astarche (approved):

PR #1260, opened 2026-05-01
Support ni-tls-config Directory Scheme and Add TLS Configuration Templates

texasaggie97:
## What does this Pull Request accomplish?

This PR enables mutual TLS support in grpc-device through integration with NI managed certificate configuration via ni-tls-config. The work includes:

1. **TLS loader enhancements**  
   Refactors `TlsConfigLoader` to use ni-tls-config contents APIs instead of direct file-path parsing, enabling `directory://` support for trusted certificate bundles.

2. **TLS configuration templates with explicit role naming**  
   Introduces role-explicit template source names so usage is clear at the source level:
   - `ni-grpc-device.server.conf.yml`
   - `ni-grpc-device.server.caps.yml`
   - `ni-grpc-device.client.defaults.yml`
   - `ni-grpc-device.client.caps.yml`

3. **Canonical runtime filename mapping during staging**  
   Preserves canonical ni-tls-config runtime filenames during staging so runtime behavior stays compatible:
   - server source templates stage/copy to `ni-grpc-device.conf.yml` and `ni-grpc-device.caps.yml`
   - client source templates stage to `ni-grpc-device.defaults.yml` and `ni-grpc-device.caps.yml`

4. **Client artifact staging cleanup**  
   Updates client staging to use `client.d` as the staging folder (instead of `config`) for ni-tls client artifacts, improving clarity and resolving reviewer feedback around naming intent.

5. **Example configuration and feature gating**  
   Adds `server_nitlsconfig.json` to demonstrate secure configuration with:
   - `"security": "ni-tls-config"`
   - `"feature_toggles": { "ni-tls-config": true }`

6. **CI/artifact updates**  
   Updates workflow artifact packaging and staging logic so the new template naming model and canonical output mapping are included consistently.

7. **URI scheme normalization**  
   Standardizes URI schemes to lowercase `file://` and `directory://` for consistency.

## Why should this Pull Request be merged?

This change improves TLS security support while reducing naming ambiguity for maintainers. It keeps runtime compatibility by preserving canonical ni-tls-config filenames, but makes source ownership and intent explicit with server/client template naming. It also aligns artifact layout with clearer `client.d` staging semantics and addresses reviewer concern about confusing export naming behavior.

## What testing has been done?

- TLS loader builds and unit tests pass with the contents API integration.
- End-to-end mutual TLS handshake validated on Windows host and NI Linux RT target.
- Server correctly loads trusted certificates via `directory://`.
- Client and server certificate exchange flow validated using automation.
- Workflow and staging logic validated for template inclusion and canonical filename outputs.
- URI normalization validated with lowercase `file://` and `directory://` schemes.
- YAML template syntax validated.


source/codegen/stage_client_files.py
```diff
@@ -157,6 +157,10 @@ def stage_client_files(output_path: Path, ignore_release_readiness: bool):
     config_path = output_path / "config"
     config_path.mkdir(parents=True)
     copy2(artifact_locations.config_dir / "ni-grpc-device.defaults.yml", config_path)
+    copy2(
+        artifact_locations.config_dir / "ni-grpc-device.client.caps.yml",
```
dmondrik, line 164: This seems  a little odd to drop the word 'client' from the export. Why 2 different names? And why is it treated differently from the line above yet exported to the same directory?

dmondrik (approved):

source/codegen/stage_client_files.py
texasaggie97: This is good catch. The idea is to differentiate it from from the server conf file. But this made it inconsistent with the other files. I have added client/server to all the ni-tls-config files as appropriate but then stage them using their canonical name that `ni-tls-config` expects. Also I changed the folder the client files are put in to `client.d` since that is what will be used by the installer.

source/config/server_nitlsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "port": 31763,
+    "sideband_address": "[::1]",
+    "sideband_port": 50055,
+    "security" : "ni-tls-config",
+    "code_readiness": "RestrictedRelease",
```
astarche: What is the reason for setting `RestrictedRelease` readiness here? That is probably common for NI-internal clients of `grpc-device` to use internal driver features. So, I can guess why there's an affinity between the `ni-tls-config` features and `RestrictedRelease`. But I don't think they're directly related, and we usually don't have non-default `code_readiness` in examples.

I'm OK with this if it's helpful to you, though.

source/config/server_nitlsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
astarche, line 1: If you're using/expecting to listen on all NICs, include `"address": "[::]"` here explicitly (similar to the localhost example).

astarche (approved):

reckenro (approved):

source/config/server_nitlsconfig.json
texasaggie97: I started with this guide: https://www.ni.com/docs/en-US/bundle/instrumentstudio/page/grpc-remote-panel-config.html

This is the InstrumentStudio guide for using gRPC with RFSA, but NI-DMM uses the same infrastructure. I found that it was needed here also. I did not look in to _why_ it was needed though. I also found that that guide was missing the sideband entries so I added them this way and it worked. Again I did not investigate _why_ it worked or was required.

Given all that, I will leave it here.

source/config/server_nitlsconfig.json
texasaggie97: Based on https://www.ni.com/docs/en-US/bundle/instrumentstudio/page/grpc-remote-panel-config.html I removed the "address" entry. I will leave it as it is.

source/config/server_nitlsconfig.json
reckenro: This was a historical thing. A while back, to have access to methods in [nidmm_restricted.proto file](https://github.com/ni/grpc-device/blob/2e43c2412b0fd3a4fb7ff4ee42270df60fced10f/generated/nidmm_restricted/nidmm_restricted.proto#L18), for instance, you'd have to set the code_readiness to one of the "Restricted..." ones.

But since https://github.com/ni/grpc-device/pull/842 that's not needed. The server will translate RestrictedRelease to Release code_readiness for the purposes of exposing the RPCs on the server. We kept around "Restricted...." code_readiness support for backwards compatibility.

So, essentially, `"code_readiness": "RestrictedRelease"` should behave the same as  `"code_readiness": "Release"`. And furthermore, the config parser defaults to Release if the code_readiness isn't specified in the config.

source/config/server_nitlsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+    "port": 31763,
+    "sideband_address": "[::1]",
+    "sideband_port": 50055,
```
maxxboehme: I think for the default nitls configuration we should not have sideband enabled.

maxxboehme (commented):
With these new client and server configuration files. I think there should be a readme update clearing up their uses.

source/config/server_nitlsconfig.json
texasaggie97: Based on internal discussion, simplified this file to show only ni-tls-config specific items.

source/config/server_nitlsconfig.json
texasaggie97: Removed

maxxboehme (approved):

PR #1261, opened 2026-05-22
CMakeLists: honor USE_SUBMODULE_LIBS when including

amstewart:

A section of the CMakeLists attempts to include protobuf headers from the third_party submodules unconditionally, even if the submodules are not supposed to be used for the build.

Branch these includes behind a check for if USE_SUBMODULE_LIBS is enabled.

**Also**. Upgrade the python runtime version used in the ``validate_python`` step of the PR build to 3.11.0. Python 3.9.0 does not support one of the functions needed by poetry 1.8.2.

NILRT builds the grpc-sideband library as a dynamic lib and should import its headers from ``/usr/includes`` rather than the submodules.

* [x] Built these changes in the NILRT OE environment and confirmed that ni-grpc-device compiles without the submodules.

### DIFF
--- a/.github/workflows/validate_python.yml
+++ b/.github/workflows/validate_python.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Setup python3
       uses: actions/setup-python@v5
       with:
-        python-version: '3.9.x'
+        python-version: '3.11.x'
 
     - name: Install build dependencies
       run: python -m pip install -r python_build_requirements.txt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,11 +319,12 @@ function(GenerateGrpcSources)
   set(output_files "${GENERATE_ARGS_OUTPUT}")
   set(proto_file "${GENERATE_ARGS_PROTO}")
   if(USE_SUBMODULE_LIBS)
-    set(protobuf_includes_arg 
-      -I ${CMAKE_SOURCE_DIR}/third_party/grpc/third_party/protobuf/src/
-      -I ${CMAKE_SOURCE_DIR}/third_party/ni-apis/ni/grpcdevice/v1/ # for session.proto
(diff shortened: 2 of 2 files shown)


amstewart:
I've just discovered some other issues with the CMakeLists that I'll have to patch for NILRT. So I'll append those changes to this PR as well, before taking it final.

amstewart:
* Nevermind, I don't need to make any other core changes.

# Patch V2
* Fixed the validate_python GH workflow.

amstewart:
# Patch V3
* I *did* have to patch one more thing. Unconditionally include the ni-apis headers, since they are part of a source repo that we (NILRT) and other distros would not make into its own package.

.github/workflows/validate_python.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Setup python3
       uses: actions/setup-python@v5
       with:
-        python-version: '3.9.x'
+        python-version: '3.11.x'
```
reckenro, line 22: This version bump seems fine, but can you briefly mention it and the brief reasoning for it in the PR description?

reckenro (approved):

astarche (approved):

maxxboehme (approved):

.github/workflows/validate_python.yml
amstewart: It's described in [the commit](https://github.com/ni/grpc-device/pull/1261/changes/3958ba4c81f2f2a69c608bf792844f791e3a222b). But sure, I can copy it to the PR.

reckenro: Oh, gotcha. A little redundant then. Appreciate it though.

I asked because most people in this repo do a squash merge with PR descripton as commit message. I see you didn't do that, though, so that commit you referenced is in main's history.

PR #1263, opened 2026-06-02
Copy server TLS config into output directory

texasaggie97:

Adds a post-build CMake copy step for `server_nitlsconfig.json` so the TLS configuration file is copied into the `ni_grpc_device_server` binary output directory.

The server expects its runtime configuration files to be available in the output directory. This change ensures the TLS config is staged alongside the server binary the same way other server config artifacts already are, which makes local runs and deployments more reliable.

No additional tests were run for this change.

The change is limited to build-time staging logic in `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -676,6 +676,15 @@ add_custom_command(
             $<TARGET_FILE_DIR:ni_grpc_device_server>/server_config.json)
 
 #----------------------------------------------------------------------
+# Copy server_nitlsconfig.json to binary output directory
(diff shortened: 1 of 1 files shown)


amstewart (approved):

amstewart:
Should probably mention that this is primarily to fix the GH workflow CI runs.

reckenro (approved):

astarche (approved):

maxxboehme (approved):

PR #1280, opened 2026-07-28
Audit log in successful connection cases

ryanwixon-emerson:

This PR adds simple audit logging capabilities to `grpc-device`. We're using `spdlog` to log to the Windows Event Log and Syslog.

gRPC includes `GlobalCallbacks` functionality which allows us to define functions invoked at every RPC call. This allows us to easily log the existence of successfully connected clients.

The current logging functionality in `grpc-device` provides an API to log to the terminal and a corresponding API to override this function with your own custom option. This applies for all logging. Since overriding with a `spdlog` API would affect all callers, I've instead opted to add a new `log_to_audit_source` function which will always use the `spdlog` sync. 
 
Our goal is to add audit logging for the CRA. The server should log a message when clients succeed/fail when authenticating. This PR adds the required framework to make this possible and implements the success scenario.

A future PR to add functionality to log when a client fails to authenticate will follow. 

I tested this in a Windows -> Windows workflow and confirmed that the audit events appear in Windows Event Log as expected. Due to the difficulty of testing other workflows, I'll wait until this is merged to check Linux workflows.


source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+    nidevice_grpc::logging::log_to_audit_source(nidevice_grpc::logging::Level_Info, "Remote client successfully connected from %s", peer.c_str());
+}
+
+void ClientConnectionLogger::PostSynchronousRequest(grpc::ServerContext*)
+{
+}
+
+void register_client_connection_logger()
+{
+  // gRPC does not take ownership of the callbacks object, so it will be a static object.
```
ckoellin: It seems like this was a bug that has been fixed in newer versions of grpc. Is grpc-device using a version of grpc that has this fix? https://github.com/grpc/grpc/issues/37607. From what I am seeing it looks like grpc-device is still on a version that has the bug.

CMakeLists.txt
ckoellin, line 101: Looks like we are missing `set(_SPDLOG spdlog::spdlog)` in this case. You might also have to add "add_subdirectory(third_party/spdlog ... EXCLUDE_FROM_ALL). Similar to the above "add_subdirectory(third_party/grpc ...)

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+}
+
+void ClientConnectionLogger::PreSynchronousRequest(grpc::ServerContext* context)
+{
+  const auto peer = context->peer();
+  std::string ip, port;
+
+  // Try to parse the IP + Port out of the peer uri so that the log message can be formatted nicely. If it fails, just log the raw string.
```
ckoellin, line 84: I believe PreSynchronousRequest fires for every sync RPC, including health-check and reflection calls. A DAQmx read loop could emit thousands of identical "successfully connected" events per second, flooding the Event Log, rolling other audit records out of retention, and putting a synchronous Event Log write on the RPC hot path.

Should be easy to test if this is the case. If so we should reintroduce some form of bounded dedup (per-peer/per-session cache) or rate limiting.

.gitmodules
```diff
@@ -13,7 +13,9 @@
 [submodule "third_party/googletest"]
 	path = third_party/googletest
 	url = https://github.com/google/googletest.git
+[submodule "third_party/spdlog"]
```
ckoellin, line 16: We should update https://github.com/ni/grpc-device/blob/main/ThirdPartyNotices.txt with spdlog

source/server/logging.cpp
```diff
@@ -37,5 +46,59 @@ void log(Level level, const char* fmt, ...)
...
 
+namespace {
+
+std::shared_ptr<spdlog::logger> get_audit_logger()
+{
+  static std::shared_ptr<spdlog::logger> audit_logger = []() {
+#if defined(_WIN32)
+    auto sink = std::make_shared<spdlog::sinks::win_eventlog_sink_mt>("ni-grpc-device-server");
+#else
+    auto sink = std::make_shared<spdlog::sinks::syslog_sink_mt>("ni-grpc-device-server", LOG_PID, LOG_USER, /*enable_formatting=*/true);
```
ckoellin: It looks like the ctor of syslog_sink_mt [openlog](https://github.com/gabime/spdlog/blob/989d28dd768adef702538634857942db24e1058a/include/spdlog/sinks/syslog_sink.h#L33) and the syslog_logging already in grpc-device also calls [openlog](https://github.com/ni/grpc-device/blob/fbedc85c2c207e1276efc80347ca64c9ab92be32/source/server/linux/syslog_logging.cpp#L20) which my understanding is that these are process wide which could clobber syslog_options. We might need to reconcile [syslog_logging.cpp](https://github.com/ni/grpc-device/blob/main/source/server/linux/syslog_logging.cpp) to go through our new spdlog facilities instead of syslog directly.

source/server/logging.cpp
```diff
@@ -37,5 +46,59 @@ void log(Level level, const char* fmt, ...)
...
+}
+
+void log_to_audit_source(Level level, const char* fmt, ...)
+{
+  va_list args;
+  va_start(args, fmt);
+  const std::string message = format_message(fmt, args);
+  va_end(args);
+
+  auto audit_logger = get_audit_logger();
```
ckoellin: I believe get_audit_logger can throw a spdlog_ex which could result in an escaped exception terminating the process. We should add a try/catch here and probably fall back to log(...) in the catch.

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+  // Strip ipv6 brackets, e.g. "[::1]" -> "::1".
+  if (ip.size() >= 2 && ip.front() == '[' && ip.back() == ']')
+    ip = ip.substr(1, ip.size() - 2);
+
+  return !ip.empty() && !port.empty();
+}
+
+}
+
+void ClientConnectionLogger::PreSynchronousRequest(grpc::ServerContext* context)
```
ckoellin, line 61: Should we also include the server context->auth_context()? You might be able to include the transport security type in the message.

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return false;
+
+  ip = peer.substr(scheme_end + 1, port_pos - scheme_end - 1);
+  port = peer.substr(port_pos + 1);
+
+  // Strip ipv6 brackets, e.g. "[::1]" -> "::1".
+  if (ip.size() >= 2 && ip.front() == '[' && ip.back() == ']')
+    ip = ip.substr(1, ip.size() - 2);
+
+  return !ip.empty() && !port.empty();
```
ckoellin, line 58: Turns ipv6:[::1]:5000 into ::1:5000.
keep the brackets. Once you do, the function reduces to stripping the ipv4:/ipv6: scheme prefix, since the remainder is already host:port — much less code and no failure path.

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,54 @@
+#include "client_connection_logger.h"
+
+#include "logging.h"
+
+namespace nidevice_grpc {
+
+namespace {
+
+bool parse_peer(const std::string& peer, std::string& ip, std::string& port)
```
ckoellin, line 29: Probably worth adding some unit tests for.

ckoellin (commented):
A few things to fix but overall looks good.

source/server/client_connection_logger.cpp
ryanwixon-emerson: Yes, this definitely fires for every RPC. I did have functionality to only log a certain IP + Port combination once, but I got rid of it. I can easily add it back in, although I think it would be best to filter based on IP alone since the call picks a different port each time. So during the server's lifetime, there will only be a log on the first RPC call from a given IP.

README.md
```diff
@@ -191,3 +191,15 @@ With this setting the server reads TLS configuration at startup from the per-ser
 If `"security": "ni-tls-config"` is configured but the `ni-tls-config` library is not installed, the server logs an error and exits instead of starting insecurely.
 
 Once `ni-tls-config` is enabled, use NI Hardware Configuration Utility on each client machine to configure the desired security settings.
+
+### Audit Logging
+
+The server logs audit messages to the Windows Event Log and Linux Syslog. On Windows, events are logged from the `ni-grpc-device-server` source. These events will be automatically placed under the default Application log (seen under the Windows folder in the Event Viewer app). It is recommended to add the following registry key to properly register it as a source:
```
bkeryan, line 197: Would it make sense for the readme to say that running NI's official installer for grpc-device (included with InstrumentStudio, etc.) does this automatically?

source/server/client_connection_logger.cpp
ryanwixon-emerson: It looks like this does let me include the type and the cert's common name as well. I'll update the log to add them both.

README.md
```diff
@@ -191,3 +191,17 @@ With this setting the server reads TLS configuration at startup from the per-ser
...
+
+`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\EventLog\National Instruments\ni-grpc-device-server`
+
+Then add the following values:
+  - `EventMessageFile (REG_EXPAND_SZ): %systemroot%\System32\mscoree.dll`
+  - `TypesSupported (REG_DWORD): 7`
+
+This recommendation is taken directly from [spdlog's documentation](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/win_eventlog_sink.h).
+
+NI's official installer for `grpc-device` automatically sets this registry key. This installer is included with installations of Instument Studio, several MI drivers, and more.
```
bkeryan:
```suggestion
NI's official installer for `grpc-device` automatically sets this registry key. This installer is included with installations of InstrumentStudio, several MI drivers, and more.
```

source/server/logging.cpp
ryanwixon-emerson: I don't think that function can throw anything. Was there a specific part of it that you had in mind?

ckoellin: I believe the constructors of win_eventlog and syslog_sink can throw in get_audit_logger.

source/server/client_connection_logger.cpp
ckoellin: To circle back on this, if we are using a version with the bug I think an acceptable solution could be.

```
  // gRPC stores this in an owning shared_ptr (see Server::SetGlobalCallbacks in
  // server_cc.cc) and deletes it at static destruction. Even if that changes to
  // non-owning, intentionally leaking one process-wide object is correct and
  // avoids a static-destruction-order hazard against grpc::Server.
  grpc::Server::SetGlobalCallbacks(new ClientConnectionLogger());
```

ryanwixon-emerson: Yes, it appears we're on gRPC v1.69, which was released after the bug was reported but before it was addressed. I implemented your suggestion.

source/server/logging.cpp
ryanwixon-emerson: I didn't realize there were already syslog capabilities already built into this. I modeled a couple of different solutions to this problem but couldn't find one that was satisfying while still preserving the `spdlog` usage for Linux. Because of this I'm opting to use `spdlog` only for the Windows logging and using the existing implementation for the Linux end.

Audit logging will still go through `log_to_audit_source` (since I still want to enforce adding the prefix that the Windows logs get automatically), but it now redirects into the direct syslog implementation.

This has the added benefit of removing `spdlog` as a build dependency for Linux, as we can now gate the Windows implementation + helpers behind preprocessor directives.

source/server/logging.cpp
ryanwixon-emerson: Maybe I'm missing something? I don't think this is the case:

https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/sinks/win_eventlog_sink.h
```
    win_eventlog_sink(std::string const &source,
                      DWORD event_id = 1000 /* according to mscoree.dll */)
        : source_(source),
          event_id_(event_id) {
        try {
            current_user_sid_ = internal::sid_t::get_current_user_sid();
        } catch (...) {
            // get_current_user_sid() is unlikely to fail and if it does, we can still proceed
            // without current_user_sid but in the event log the record will have no user name
        }
    }
```

The only thing the constructor does is set some private variables. Everything else is contained in a try/catch that will swallow all exceptions.

(`syslog_sink` is no longer used due to refactors)

astarche (approved):

reckenro (approved):

maxxboehme (approved):

source/server/logging.cpp
ckoellin: Yeah, probing copilot a bit more on this there should not be an issue.

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+}
+
+bool parse_peer(const std::string& peer, std::string& ip, std::string& port)
+{
+  if (peer.rfind("ipv4:", 0) != 0 && peer.rfind("ipv6:", 0) != 0)
+    return false;
+
+  const auto scheme_end = peer.find(':');
+  const auto port_pos = peer.rfind(':');
```
ckoellin: I believe for ipv6 if we have something like "ipv6:[::1]" or "ipv6:[2001:db8::1]" rfind(':') finds a colon inside the IPv6 address when no port is present. When the address starts with [, locate ] and require the separator to immediately follow it. I believe it would look something like

```
const auto addr_start = scheme_end + 1;
size_t port_pos;
if (peer[addr_start] == '[') {
  const auto bracket_end = peer.find(']', addr_start);
  if (bracket_end == std::string::npos || bracket_end + 1 >= peer.size() || peer[bracket_end + 1] != ':')
    return false;
  port_pos = bracket_end + 1;
}
else {
  port_pos = peer.rfind(':');
  if (port_pos <= scheme_end) return false;
}
```

So also be easy to add a unit test for this.

source/server/client_connection_logger.h
```diff
@@ -0,0 +1,32 @@
...
+
+// Implementation of callbacks that the grpc::Server will invoke for individual client RPC calls.
+class ClientConnectionLogger : public grpc::Server::GlobalCallbacks {
+ public:
+  void PreSynchronousRequest(grpc::ServerContext* context) override;
+  void PostSynchronousRequest(grpc::ServerContext* context) override;
+
+ private:
+  std::mutex seen_ips_mutex_;
+  std::unordered_set<std::string> seen_ips_;
```
ckoellin, line 29: As of now seen_ips_ will grow unbounded for the lifetime of the server. The can be especially problematic for very long lifetimes. We should use a fixed-capacity cache with eviction (e.g. cap at a few thousand entries), or time-windowed entries that expire.

Related:

The auth description is captured on the first RPC from an IP and never revisited. Consequences:
- Reconnects are never recorded.
- A client that first connects unauthenticated and later presents a certificate (or connects with a different certificate) is never logged again — the audit record actively misstates the current state.
- Multiple distinct clients behind NAT, or several on one host, collapse to one record.

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,79 @@
...
+  if (port_pos <= scheme_end)
+    return false;
+
+  ip = peer.substr(scheme_end + 1, port_pos - scheme_end - 1);
+  port = peer.substr(port_pos + 1);
+
+  return !ip.empty() && !port.empty();
+}
+
+void ClientConnectionLogger::PreSynchronousRequest(grpc::ServerContext* context)
```
ckoellin, line 61: Should probably add a `grpc::ServerContext* context` null check.

CMakeLists.txt
```diff
@@ -71,6 +71,10 @@ if(USE_SUBMODULE_LIBS)
   add_subdirectory(third_party/json ${CMAKE_CURRENT_BINARY_DIR}/json)
   add_subdirectory(third_party/utfcpp ${CMAKE_CURRENT_BINARY_DIR}/utfcpp)
   add_subdirectory(third_party/grpc-sideband ${CMAKE_CURRENT_BINARY_DIR}/grpc-sideband)
+  if(WIN32)
+    add_subdirectory(third_party/spdlog ${CMAKE_CURRENT_BINARY_DIR}/spdlog EXCLUDE_FROM_ALL)
+    set(_SPDLOG spdlog::spdlog)
+  endif()
```
ckoellin, line 76: I believe you will need `find_package(spdlog REQUIRED)` in the else case when USE_SUBMODULE_LIBS is off.

ryanwixon-emerson: I think turning off `USE_SUBMODULE_LIBS` on Windows builds is technically (although not explicitly) unsupported. You can see that if you do that you are unconditionally sent down a path for building Linux binaries:

```
  if(USE_NILRT_LEGACY_TOOLCHAIN)
    # The archetypical NILRT SDK toolchain build case
    add_subdirectory(third_party/grpc ${CMAKE_CURRENT_BINARY_DIR}/grpc EXCLUDE_FROM_ALL)

    set(_REFLECTION grpc++_reflection)
    set(_GRPC_GRPCPP grpc++)
    set(_PROTOBUF_LIBPROTOBUF libprotobuf)
  else()
    # The archetypical linux build case (including OpenEmbedded)
    find_library(_GRPC_GRPCPP grpc++)
    find_library(_PROTOBUF_LIBPROTOBUF protobuf)
    find_library(_REFLECTION grpc++_reflection)
    find_library(_GRPC_SIDEBAND ni_grpc_sideband REQUIRED)

    find_package(gRPC REQUIRED)
    find_package(GTest REQUIRED)
    find_package(nlohmann_json REQUIRED)
    find_package(utf8cpp REQUIRED)

    set(_GRPC gRPC::grpc)
    set(_UTF8CPP utf8cpp::utf8cpp)
  endif()
```

Building the LinuxRT binaries is a valid use case if you're cross-compiling on Windows. But this doesn't need `spdlog` anymore since the Linux paths have been refactored to use the preexisting syslog mechanisms.

Building desktop Linux binaries on Windows is an invalid use case and will fail. So if you disable `USE_SUBMODULE_LIB`, you also need to enable `USE_NILRT_LEGACY_TOOLCHAIN`.

It seems to me that there is no supported method to build Windows binaries with `USE_SUBMODULE_LIBS=off`, so we shouldn't add any references to `spdlog` in this case.

ckoellin: Ok makes sense.

source/server/client_connection_logger.h
ryanwixon-emerson: I've implemented an eviction mechanism which will drop the first recorded client after the client exceeds 3000 entries. This could potentially cause some confusion for someone looking through the logs as they will only see their connections being logged sometimes. Unfortunately, the reality is that we will probably have to live with that for now since the alternative is either logging every connection (spamming the log) or continue to cache everything with undefined size.

If gRPC stabilizes the experimental options that could improve the logging process this is something that could definitely be revisited to provide a more robust solution, but I think this is about the best we can do with what we have today.

ckoellin (approved):

PR #1281, opened 2026-07-31
Audit log in failed connection cases

ryanwixon-emerson:

This PR is a follow up to #1280 and is based on it. The relevant changes are mostly in `ClientConnectionLogger.cpp`, and once #1280 is merged it will shrink considerably.

This PR adds handshake failure logging to `grpc-device`. Due to the limitations of gRPC, the mechanism that it uses to log differs slightly from the success logging, which fires once per RPC call. This new mechanism instead creates a sink using `absl`, which will catch all of gRPC's internal logging and bubble it up into a callback function.

Since this naturally surfaces a very large number of logs (everything gRPC ever sends), I'm opting to filter for the word "handshake" when determining which logs to actually display to the user. This results in messages being logged on mTLS failures (such as when the client does not provide a certificate) and, from what I can tell, nowhere else.

Our goal is to add audit logging for the CRA. The server should log a message when clients succeed/fail when authenticating. This PR completes the implementation by adding an implementation for the failure scenario.

#1280 is the PR which laid the groundwork for this change and implemented the success scenario.

I tested this in a Windows -> Windows workflow and confirmed that the audit events appear in Windows Event Log as expected. Due to the difficulty of testing other workflows, I'll wait until this is merged to check Linux workflows.


source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,123 @@
...
+      return logging::Level_Info;
+  }
+}
+
+// This sink captures gRPC's own internal logs and calls Send for each of them. We only use it to log handshake failures.
+class AuditLogSink : public absl::LogSink {
+ public:
+  void Send(const absl::LogEntry& entry) override
+  {
+    if (!absl::StrContainsIgnoreCase(entry.text_message(), "handshake"))
```
astarche: I have 2 "thinking out loud" comments on this:

1. Is it possible that grpc has improved or will improve support for connection logging as part of its own CRA compliance efforts? It's quite cool that we've managed to make this work in a pretty reasonable way in a difficult environment. But it's hard to imagine that every C++ grpc server will need to do all of this to comply with CRA. 
2. What are some interesting failure modes of this text match? Is it possible for a "failed connection" to be missing the `handshake` token? Is it possible for a "normal message" to include the `handshake` token? Is it possible for a client (unauthenticated client?) to influence either failure? Would using an additional/different filter (like "severity"?) improve it?

Most likely what you're doing is about right. Just checking.

source/server/client_connection_logger.cpp
```diff
@@ -0,0 +1,123 @@
...
+  grpc::Server::SetGlobalCallbacks(new ClientConnectionLogger());
+}
+
+void register_grpc_log_sink()
+{
+  absl::InitializeLog();
+
+  // Abseil does not take ownership of the sink, so it will be a static object.
+  static AuditLogSink sink;
+  absl::AddLogSink(&sink);
```
astarche: Do we need to RemoveLogSink?

astarche:
Looks reasonable. Waiting for this to be rebased on the merged base commit (the diff is showing both for me)

source/server/client_connection_logger.cpp
ckoellin: I chatted with Ryan about this a bit and I believe the preferred mechanism for authentication logging moving forward will be to use https://grpc.io/docs/guides/interceptors/. Currently these are under the experimental namespace so we did not want to use these in their current state. Mainly to avoid any API friction that could be incurred when upgrading to newer versions of gRPC in grpc-device. Once these have stabilized and have been moved out of experimental it's definitely worth looking into.

ryanwixon-emerson: Yes, gRPC has several experimental options that would likely streamline the logging process. They've existed for a while but have not been stabilized yet. I'm not sure if there are plans on their end to do this at the moment, but maybe the CRA deadline will provide some motivation. In particular, there's a [CertificateVerifier](https://github.com/grpc/grpc/blob/master/src/cpp/common/tls_certificate_verifier.cc) which seems to have the capability to provide function hooks that will fire during a handshake itself.

From what I can tell throughout my testing the "handshake" token seems to be sufficient as a filtering method. I saw the correct message in all failure cases and did not notice any extraneous error messages appearing at any other point either. Exploring the gRPC source with Copilot seems to align with this behavior; I found only 7 error messages that include it, including the one I saw in testing. It _is_ possible for a failed connection to not include the token, but the scope of this addition is only to log authentication failures, not any kind of failure.

I think that the current state is probably the best solution we have for today; the aforementioned `CertificateVerifier` is probably the best way to do it in a less "hacky" way and would definitely be something to consider if gRPC stabilizes it.

source/server/client_connection_logger.cpp
ryanwixon-emerson: I don't think this would have caused a problem in practice since the sink itself is not destroyed until the end of the process, but to make it a bit cleaner I refactored this a bit. We're now using a RAII wrapper class that will own the sink and call `AddLogSink`/`RemoveLogSink` on construction and deconstruction.

source/server/client_connection_logger.cpp
```diff
@@ -100,4 +107,62 @@ void register_client_connection_logger()
...
+    AuditLogSinkWrapper& operator=(const AuditLogSinkWrapper&) = delete;
+    AuditLogSinkWrapper(AuditLogSinkWrapper&&) = delete;
+    AuditLogSinkWrapper& operator=(AuditLogSinkWrapper&&) = delete;
+
+  private:
+    class AuditLogSink : public absl::LogSink {
+    public:
+      void Send(const absl::LogEntry& entry) override
+      {
+        if (!absl::StrContainsIgnoreCase(entry.text_message(), "handshake"))
```
ckoellin: I believe this might also match against some of the handshaker logging that can happen since handshake is a substring of that. ex. https://github.com/grpc/grpc/blob/838ece08491e32cdbbd83360e0637f2543eede99/src/core/credentials/transport/ssl/ssl_credentials.cc#L193. We should probably make this making a bit more strict so that we only match against the things we really want. If it's easy enough it's probably worth adding some unit tests for this.

source/server/client_connection_logger.cpp
```diff
@@ -100,4 +107,62 @@ void register_client_connection_logger()
...
+  }
+}
+
+// This sink captures gRPC's own internal logs and calls Send for each of them. We only use it to log handshake failures.
+class AuditLogSinkWrapper {
+  public:
+    AuditLogSinkWrapper() {
+      absl::InitializeLog();
+      absl::AddLogSink(&sink_);
+    }
```
ckoellin, line 140: We may need to add something like the following here.

`
// gpr_log_verbosity_init() may raise the global floor above INFO and silence audit records.
if (absl::MinLogLevel() > absl::LogSeverityAtLeast::kInfo) {
  logging::log(logging::Level_Error, "GRPC_VERBOSITY suppresses connection-failure audit logging; forcing INFO.");
  absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfo);
}
`

ckoellin (approved):

maxxboehme (approved):

source/server/client_connection_logger.cpp
ryanwixon-emerson: I'm now matching against a regex string that looks for "handshake" and "fail" in the same message. After searching through the various logging strings in gRPC, this looks like the best way to differentiate between actual handshake errors and internal gRPC errors related to handshakes.

Unit testing this seems like it would be very involved, so I'll probably opt to just manually verify that I'm still seeing the messages I expect.

reckenro (approved):

source/server/logging.cpp
```diff
@@ -53,9 +53,9 @@ namespace {
 std::shared_ptr<spdlog::logger> get_event_log_logger()
 {
   static std::shared_ptr<spdlog::logger> audit_logger = []() {
-    auto sink = std::make_shared<spdlog::sinks::win_eventlog_sink_mt>("ni-grpc-device-server");
+    auto sink = std::make_shared<spdlog::sinks::win_eventlog_sink_mt>("ni-grpc-device");
```
ryanwixon-emerson, line 56: Adding in a quick unrelated change to this PR as I'm going to be changing the Windows Event Log source name to prevent the logs from being redundant. Today they are prefixed with "[ni-grpc-device-server][Server]".

astarche (approved):

PR #1287, opened 2026-09-07
Adding daqmx metadata for Analog Path Delay feature

karhin1990:

Pull in daqmx metadata that contain Analog Path Delay attributes then regenerate nidaqmx.proto to pick up those attributes.

- Feature spec: https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/docs/specs/daqmx/459x/459x_phase_correction/459x_phase_correction.md&_a=preview
- PR that generates daqmx metadata: https://github.com/ni/grpc-device-scrapigen/pull/288

To support Analog Path Delay attribute access through grpc interpreter in python API

Local build succeeded and generated nidaqmx.proto that contain Analog Path Delay attributes.


source/codegen/metadata/nidaqmx/attributes.py
```diff
@@ -18905,6 +18905,100 @@
             'python_description': 'Specifies the current minimum value you expect to measure. This value is in the units you specify with a units property. When you query this property, it returns the coerced current minimum value that the device can measure with the current settings.',
             'resettable': True,
             'type': 'float64'
+        },
+        12790: {
```
reckenro, line 18909: Please evaluate if these changes should also bump the api_version in the [config.py file](https://github.com/ni/grpc-device/blob/c1ddf51a6bcc9d28f1b4116b986f0ae9d23dc2f8/source/codegen/metadata/nidaqmx/config.py#L3) (that file generates a line at the top of the proto mentioning the API version).

And if so, update there and in the [README's driver table versions](https://github.com/ni/grpc-device#supported-ni-drivers)

reckenro (approved):

source/codegen/metadata/nidaqmx/attributes.py
karhin1990: Yes I think we should bump the version since these attributes are newly introduced in 26.8.0, thanks for reminding.
Fixed.

PR #1291, opened 2026-09-17
Make code flexible and ready for alternative builds

texasaggie97-ni:

- Add configurable driver-service selection to CMake, generated server capabilities, and service registration
- Use host protobuf and gRPC code generators during cross-compilation
- Improve NI Linux RT portability across server, driver-service, and test code
- Add compatibility patches for older NI Linux RT C++ standard library headers
- Document configurable driver builds

Prepare code for additional compile targets

PR Build


cmake/patches/grpc-remove-redundant-cxx17-includes.patch
```diff
@@ -0,0 +1,58 @@
+diff --git a/src/core/ext/transport/chaotic_good/server_transport.h b/src/core/ext/transport/chaotic_good/server_transport.h
```
maxxboehme, line 1: I'm a little nervous about having to deal with patches. Did you try seeing if this is maybe fixed in a newer version of grpc? Also when are these patches applied?

